PR #2, opened 2024-01-11
chore: add theme colours and text styles

alexanderleegs:


src/theme/foundations/colors.ts
```diff
@@ -0,0 +1,25 @@
+export const colors = {
```
harishv7, line 1: Are these meant to be used with design system?

src/theme/textStyles.ts
```diff
@@ -0,0 +1,80 @@
...
+  "button-text": {
+    textAlign: "center",
+    fontSize: "1rem",
+    fontWeight: 600,
+    lineHeight: "2rem" /* 200% */,
+    textTransform: "uppercase",
+  },
+
+  link: {
+    fontSize: "1.125rem",
```
harishv7, line 74: Where did we get these styles from?

src/theme/foundations/colors.ts
alexanderleegs: Hmm currently these are copied over from our existing template! Not sure if we're planning to revamp isomer template colours as well

src/theme/textStyles.ts
alexanderleegs: Also copied over from our existing template, might need to change depending on the sgds revamp

src/theme/foundations/colors.ts
harishv7: Got it, yea we should maintain current colors. Was just checking if this is compatible with design system or it expects in a different type structure but guess we can figure that later

Maybe we should put into a folder structure. This can be called as Isomer classic

In case we want to "load" a different theme in future we can create a new folder with "modern/theme.ts for example

harishv7 (approved):

src/theme/foundations/colors.ts
alexanderleegs: sure, that makes sense! moved in https://github.com/isomerpages/isomer-components/pull/2/commits/bbb8f1c2e1d8425dcd27d9116873df7f5f18d692

PR #4, opened 2024-02-27
Update Infobar

jacobkwan:
https://linear.app/ogp/issue/ISOM-734/create-infobar-component-in-isomer-next
1. use flex-col with gap instead of setting paddings between children (I think its cleaner?)
2. add missing `md:w-1/2` to replicate our current Infobar's maxWidth media queries
3. use short circuit rendering so that missing subtitle / description etc does not spawn a ghost `gap` in the flex-cols
4. add `items-center` to button so that the arrow is still centralised if the label is multiline

didn't really touch font styles because I'm not sure if those will change/if we really want to follow the current blueprint.css to a T

Also fix a bug where if `sectionIdx = 0`, it would show bg gray instead of white

Recording showing layout for convenience:

https://github.com/isomerpages/isomer-components/assets/77665182/f60ccb6b-ee36-4023-bd78-90cefe5c03c6


jacobkwan:
Also not sure if you noticed this while developing but seems like some of the tailwind utility classes are overridden by something else? I don't think its good that its being built as `!important` right...

Screenshot 2024-02-27 at 9 21 58 PM

dcshzj:
@jacobkwan is it possible to add some stories with different permutations and length of text?

jacobkwan:
> @jacobkwan is it possible to add some stories with different permutations and length of text?

added here
[f3fcb10](https://github.com/isomerpages/isomer-components/pull/4/commits/f3fcb10eb093e30cbdc6c1c1ecfbddd42e4d46b2)

dcshzj:
!run chromatic

dcshzj (approved):
LGTM! I think the padding on mobile view might be a bit too much but we can fix that later when we get designer to review.

PR #5, opened 2024-02-28
feat: update infopic

jacobkwan:
Simplify `Infopic` slightly and match current responsive behaviour a little closer

1. change most `lg` to `md` as `md` screen should see two col layout based on current behaviour
2. Use `my-auto` to fix the similar issue in infobar where button arrow was not center aligned if label is multiline
3. use `flex-col` with `gap` instead of padding and margins everywhere in `TextComponent`
4. tweak paddings here and there to better match current Infopic

again, didn't really touch the fonts


dcshzj:
!run chromatic

src/components/Infopic/Infopic.tsx
```diff
@@ -36,32 +36,32 @@ const TextComponent = ({
...
-      <section aria-labelledby="information-heading" className="mt-4">
-        <div className="mt-4 space-y-6">
+    <div className="flex flex-col gap-4 lg:px-8">
+      {subtitle && (
+        <p className="text-subtitle uppercase tracking-widest">{subtitle}</p>
+      )}
+      {title && (
+        <h1 className="text-secondary text-5xl font-semibold">{title}</h1>
+      )}
+      {title && (
```
dcshzj, line 46: question(blocking): should this be description instead?

src/components/Infopic/Infopic.tsx
```diff
@@ -98,29 +98,31 @@ const InfoPic = ({
...
-            title={title}
-            subtitle={subtitle}
-            description={description}
-            buttonLabel={button}
-            buttonUrl={url}
-          />
-        </InfopicContentWrapper>
-      )}
+      <InfopicContentWrapper shouldShowMobile={false}>
+        {(sectionIndex ?? 0) % 2 === 0 ? (
```
dcshzj, line 102: thought: this binds the left right appearance of the picture with the grey background. I was thinking perhaps we can have a separate prop to define the left and right alignment of the picture?

If this is the current behaviour then let's park this for now and we can follow up with design.

src/components/Infopic/Infopic.stories.tsx
```diff
@@ -11,8 +11,20 @@ export default {
...
-export const Default = Template.bind({})
-Default.args = {
+export const DefaultLeft = Template.bind({})
+DefaultLeft.args = {
+  sectionIndex: 0,
+  title: "Thank you for attending the roadshows!",
+  subtitle: "Coming soon to your hood",
+  description: "Catch the highlights from the roadshows here.",
+  alt: "alt",
+  imageUrl: "https://picsum.photos/200/200",
```
dcshzj, line 21: nit: can use a permanent placeholder image instead

src/components/Infopic/Infopic.tsx
```diff
@@ -36,32 +36,32 @@ const TextComponent = ({
...
-        >
-          {buttonLabel}
+        </section>
+      )}
+      {buttonLabel && buttonUrl && (
+        <div className="text-lg font-semibold uppercase tracking-wid">
+          <a
+            className="flex gap-2 text-secondary font-semibold underline uppercase"
+            href={buttonUrl}
+          >
```
dcshzj, line 56: question(non-blocking): should this be using the LinkComponent instead? Although this might have to handle both internal and external links.

src/components/Infopic/Infopic.tsx
```diff
@@ -36,32 +36,32 @@ const TextComponent = ({
...
-        <div className="mt-4 space-y-6">
+    <div className="flex flex-col gap-4 lg:px-8">
+      {subtitle && (
+        <p className="text-subtitle uppercase tracking-widest">{subtitle}</p>
+      )}
+      {title && (
+        <h1 className="text-secondary text-5xl font-semibold">{title}</h1>
+      )}
+      {title && (
+        <section aria-labelledby="information-heading">
```
dcshzj, line 47: nit: we should drop this `aria-labelledby` prop here, since the description has text itself and can have different content from the title.

src/components/Infopic/Infopic.tsx
jacobkwan: [5edfc06](https://github.com/isomerpages/isomer-components/pull/5/commits/5edfc0634cdb7ca8e9d1d4c25f89a9a489c14be6)

src/components/Infopic/Infopic.tsx
jacobkwan: [52f5b1f](https://github.com/isomerpages/isomer-components/pull/5/commits/52f5b1ffb114c5ab30ca7c18affd10b08ab3b81c)

oh yea you're right!
current behaviour is just that it alternates from left to right, whereas gray or white depends on section id

src/components/Infopic/Infopic.stories.tsx
jacobkwan: [cae3ee4](https://github.com/isomerpages/isomer-components/pull/5/commits/cae3ee46da5caeabd2587ccb55bb88efc60660c6)

src/components/Infopic/Infopic.tsx
jacobkwan: [679b0be](https://github.com/isomerpages/isomer-components/pull/5/commits/679b0be8a33bc98b98cfc6ed6b547348d3f3fdbd)

src/components/Infopic/Infopic.tsx
jacobkwan: oh, does that mean I should add Nextjs as a dependency, cus currently it seems like we're just using vanilla React

jacobkwan:
!run chromatic

src/components/Infopic/Infopic.tsx
jacobkwan: lol a little easter egg for you because I thought I understood the homepage.html template correctly but something was different on CMS


https://github.com/isomerpages/isomer-components/assets/77665182/39122d85-96d1-4fa4-9910-8447af835ce7

dcshzj: LOL damn

src/components/Infopic/Infopic.tsx
```diff
@@ -36,32 +37,28 @@ const TextComponent = ({
...
-        </div>
-        <h1 className="text-secondary text-5xl font-semibold pb-4">{title}</h1>
-      </div>
+    <div className="flex flex-col gap-4 lg:px-8">
+      {subtitle && (
+        <p className="text-subtitle uppercase tracking-widest">{subtitle}</p>
+      )}
+      {title && (
+        <h1 className="text-secondary text-5xl font-semibold">{title}</h1>
+      )}
```
dcshzj, line 46: question(blocking): sorry I missed out this part earlier. I think for infopic, minimally the `title` and the `imageUrl` should be required? Otherwise it is possible for the user to supply nothing and get nothing.

jacobkwan: yep agreed, for some reason I thought the props were purposely made optional by harish hahaha, will change 
in the spirit of making good components, I think it makes sense to make alt text compulsory too then?

[dc64df4](https://github.com/isomerpages/isomer-components/pull/5/commits/dc64df4d9a6ad9b1c4848c372e5b5df4b1b9e701)

jacobkwan:
!run chromatic

dcshzj (approved):

PR #8, opened 2024-03-01
feat(hero): add hero layouts with key highlights

dcshzj:
References:
- Infobox (mobile): https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/homepage/hero/components/hero-infobox-mobile.html
- Infobox (desktop): https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/homepage/hero/components/hero-infobox-desktop.html
- Dropdown: https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/homepage/hero/components/hero-dropdown.html
- Side layout: https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/homepage/hero/hero-side-layout.html
- Hero component: https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/homepage/hero.html

### DIFF
--- a/src/classic/components/Hero/Hero.stories.tsx
+++ b/src/classic/components/Hero/Hero.stories.tsx
@@ -20,6 +20,28 @@ Default.args = {
   subtitle: "Free, fast, easy",
   buttonLabel: "Find out if I'm a good fit",
   buttonUrl: "/contact",
+  keyHighlights: [
+    {
+      title: "Key highlight 1",
+      description: "This is a key highlight",
+      url: "/key-highlight-1",
+    },
+    {
+      title: "Key highlight 2",
+      description: "This is another key highlight 1",
+      url: "/key-highlight-2",
+    },
+    {
+      title: "Key highlight 3",
+      description: "This is another key highlight 2",
+      url: "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
+    },
+    {
+      title: "Key highlight 4",
+      description: "This is another key highlight 3",
+      url: "/key-highlight-4",
+    },
+  ],
 }
 
 // Side layout
--- a/src/classic/components/Hero/Hero.stories.tsx
+++ b/src/classic/components/Hero/Hero.stories.tsx
@@ -13,7 +13,14 @@ const Template: StoryFn<HeroProps> = (args) => <Hero {...args} />
 
 // Default scenario
 export const Default = Template.bind({})
-Default.args = {}
+Default.args = {
+  variant: "center",
+  backgroundUrl: "https://ohno.isomer.gov.sg/images/hero-banner.png",
+  title: "Easily set up good government websites",
+  subtitle: "Free, fast, easy",
+  buttonLabel: "Find out if I'm a good fit",
+  buttonUrl: "/contact",
+}
 
 // Side layout
 export const SideButton = Template.bind({})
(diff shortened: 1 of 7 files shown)


dcshzj:
!run chromatic

src/components/Hero/HeroInfobox.tsx
```diff
@@ -0,0 +1,219 @@
...
+              </div>
+            )}
+          </div>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+export const HeroInfoboxMobile = ({
```
jacobkwan, line 177: I think the mobile view might be a little off, not sure if there's a simpler way to think about the permutations? 

Screenshot 2024-03-01 at 6 03 50 PM

src/components/Hero/HeroInfobox.tsx
```diff
@@ -0,0 +1,219 @@
...
+import { HeroDropdown, type HeroDropdownProps } from "./HeroDropdown"
+
+type HeroInfoboxVariants =
+  | HeroSideProps["variant"]
+  | HeroFloatingProps["variant"]
+
+export interface HeroInfoboxProps extends HeroHeadingProps, HeroButtonProps {
+  alignment?: "left" | "right"
+  backgroundColor?: "black" | "white" | "gray"
+  size?: "sm" | "md"
```
jacobkwan, line 16: a bit of a controversial opinion but I feel like the `size` prop doesn't really make sense.
it feels like it should be dependent on the content as well as the screen size instead

it will also allow us to reason about our css and responsive behaviour easier as compared to the jankness below atm

Screenshot 2024-03-01 at 6 14 55 PM

src/components/Hero/HeroInfobox.tsx
```diff
@@ -0,0 +1,219 @@
...
+    sm: "w-min w-1/3 xl:w-1/2",
+    md: "w-1/2",
+  },
+  floating: {
+    sm: "w-1/3",
+    md: "w-1/2",
+  },
+}
+
+export const HeroInfoboxDesktop = ({
```
jacobkwan, line 54: is it possible to simplify the code structure by operating under some invariants? (whether enforced or not)
e.g
`default -> HeroInfoBoxMobile`
`md -> HeroInfoBoxTablet`
`lg onwards -> HeroInfoBoxDesktop`

asking because I feel like there seems to be quite a lot of responsive css-ing going on in those 3 components, which affects readability and ability to reason about our UI
Plus its confusing why `HeroInfoBoxTablet` and `HeroInfoBarMobile` would have `lg:xxx` classes

I think would be neater to stick to those invariants, then we just need to make sure we use the usual `md:hidden` pattern when consuming the 3 components like you've done in `HeroSide` and `HeroFloating`



edit: I just realised that `HeroSide` only uses Mobile and Desktop, but I think this still makes sense, just that maybe need to make Desktop compatible with both `md` and `lg` 🤔

src/components/Hero/Hero.tsx
```diff
@@ -1,57 +1,154 @@
...
+  )
+}
+
+const HeroFloating = (
+  props: Omit<HeroFloatingProps, keyof HeroCommonProps>,
+) => {
+  return (
+    <div className="md:p-12">
+      {/* Desktop view 1024 above */}
+      <div className="invisible lg:visible hidden lg:block max-w-screen-xl mx-auto my-0">
```
jacobkwan, line 100: first time seeing `invisible`, went to look it up
can I ask whats the rationale for using it here instead of hidden?
https://tailwindcss.com/docs/visibility#making-elements-invisible

I would have thought we wouldn't want it to appear in the DOM (not sure why it isn't affecting layout currently tbh)

jacobkwan (commented):
from playing around on storybook, some of the components look a bit off on smaller screens, but I'm also not sure if its just a responsive font-size problem?

src/components/Hero/Hero.tsx
dcshzj: Hmm for some reason I learnt before that we should specify `visibility: hidden` together with `display: none` to support screen readers, but doing a search on this topic did not yield anything and in fact, screen readers are able to detect if something is hidden using the `display: none` property. Adjusted in 214340f70037ea61e777450e9572320a4aebf48c!

src/components/Hero/HeroInfobox.tsx
dcshzj: I think the main issue is with the button not being responsive. Deprioritising this for now as it is intended to take from the Button component directly.

dcshzj:
Going to merge this first, the design side we can adjust later since this is a classic component

PR #9, opened 2024-03-01
feat: add basic button component

jacobkwan:
Update classic button and add next button


jacobkwan:
!run chromatic

src/next/components/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
+import { ButtonProps } from "~/common"
+import { SUPPORTED_ICONS_MAP } from "~/common/Button"
+
+// Theme specific config
+const NEXT_BUTTON_COLOR_VARIANTS = ["black", "white"] as const
+export type NextButtonColorVariant = (typeof NEXT_BUTTON_COLOR_VARIANTS)[number]
```
jacobkwan, line 6: wasnt sure if I should define here or in /common because it is theme specific but at the same time I thought would have circular imports. but in the end it worked so I left it here

src/next/components/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
...
+// https://tailwindcss.com/docs/content-configuration#dynamic-class-names
+const colorVariantToClassMap: Record<NextButtonColorVariant, string> = {
+  black: "bg-[#333333] text-white",
+  white: "bg-white text-[#333333]",
+}
```
jacobkwan, line 11: design used `#333333` which is not within the predefined tailwind colours, closest would be `gray-700` or `gray-800`
not sure if there's a neater way to use this colour?

jacobkwan:
also realised that design is using Inter as the font for the prototype whereas classic uses Lato
looking at how to best override default fonts for different themes, not sure if you have any ideas?

src/classic/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ButtonProps } from "~/common"
+
+export default {
+  title: "Classic/Components/Button",
+  component: Button,
+  argTypes: {},
+} as Meta
+
+// Template for stories
+const Template: StoryFn<ButtonProps> = (args) => <Button {...args} />
```
dcshzj, line 12: nit: would be best to pick only the props that are used here so the docs don't show the other unused props

src/next/components/Button/Button.tsx
dcshzj: Hmm I think this might be better to put inside `~/common`? We might want to use primary/secondary/inverse in this case and let themes define what primary/secondary/inverse means.

src/next/components/Button/Button.tsx
dcshzj: Yeah we can use this for now, I think the comment can be removed as well

src/next/components/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <a
+      href={href}
+      target={href.startsWith("http") ? "_blank" : undefined}
+      rel={href.startsWith("http") ? "noopener noreferrer nofollow" : undefined}
+      type="button"
+      className={`${colorVariantClass} inline-flex items-center gap-1 px-5 py-4`}
+    >
+      <Label />
```
dcshzj, line 42: nit: how about left icon?

dcshzj (approved):

src/next/components/Button/Button.tsx
jacobkwan: oh, yea I didn't include it for now since I didn't see it in design's buttons for the MTI prototype

jacobkwan:
merging first as a first pass, can improve as we improve the theming ideas

PR #10, opened 2024-03-04
feat: add infocols component

jacobkwan:
Adds Classic InfoCols component which follows current behaviour closely
Main difference: use grid instead of janky toggling between flex-row + flex-wrap and flex-col
As such, responsive behaviour is exactly the same but implementation is simpler

Adds Next InfoCols component which follows MTI Prototype on figma (mostly)
1. Didn't focus too much on responsiveness as couldn't find design, but its still responsive
2. when comparing storybook and figma, there is still a visible difference. This should be because of the font family difference. Everything else should be correct i.e `font-weight`, `font-size`, `text-color`
3. and of course `secondary` colour is different from figma (blue vs orange)
4. not sure whats the best way to handle the random text colours being used in figma, they are just hardcoded atm, `#2C2E34` and `#333333`

References:
`isomerpages-template`: `infocols.html`


jacobkwan:
!run chromatic

src/classic/components/InfoCols/InfoCols.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+      description: "Infobox description",
+    },
+    {
+      title: "Infobox title",
+      description: "Infobox description",
+    },
+  ],
+}
+
+export const ThreeInfoBoxesLongText = Template.bind({})
```
harishv7, line 116: just curious, why are we developing both classic + next component simultaneously?

src/classic/components/InfoCols/InfoCols.tsx
```diff
@@ -0,0 +1,101 @@
...
+  // Follows current behaviour: if there are 4 info boxes, xl screen should stay at 2 columns instead of going up to 3
+  const xlColsClass =
+    infoBoxes.length === 1
+      ? "xl:grid-cols-1"
+      : infoBoxes.length % 2 === 0
+        ? "xl:grid-cols-2"
+        : "xl:grid-cols-3"
+
+  return (
+    <div>
```
harishv7, line 32: whats this top level div for?

src/classic/components/InfoCols/InfoCols.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <div
+        className={`grid grid-cols-1 gap-4 justify-between ${mdColsClass} ${xlColsClass}`}
+      >
+        {infoBoxes.map((infoBox, idx) => (
+          <div
+            key={idx}
```
harishv7, line 38: it's usually advised not to use the idx as the key. can we use any other property of infobox as the key?

src/classic/components/InfoCols/InfoCols.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const InfoColsFooter = ({
+  buttonLabel,
+  buttonUrl,
+}: Pick<InfoColsProps, "buttonLabel" | "buttonUrl">) => {
+  return (
+    buttonLabel &&
+    buttonUrl && (
+      <div className="text-lg font-semibold uppercase">
+        <a
```
harishv7, line 60: depending on the type of link (internal or extenal), this might need to accept NextLink as well
https://nextjs.org/docs/pages/api-reference/components/link

src/classic/components/InfoCols/InfoCols.tsx
```diff
@@ -0,0 +1,101 @@
...
+const InfoCols = ({
+  sectionIdx,
+  title,
+  subtitle,
+  buttonLabel,
+  buttonUrl,
+  infoBoxes,
+}: InfoColsProps) => {
+  return (
+    <HomepageSectionWrapper sectionIndex={sectionIdx}>
```
harishv7, line 89: Is it still right to call this HomepageSectionWrapper? This component should be agonstic of that homepage or not right?

src/classic/components/InfoCols/InfoCols.tsx
jacobkwan: I think there are 2 ways we can fix this
1. uuid on FE
2. add `id` field to backend

src/classic/components/InfoCols/InfoCols.tsx
jacobkwan: ah good point
I think it's fine to keep it as this for Classic InfoCols since this is the assumption made by the current Classic template on CMS, aka InfoCols / InfoPic etc are all Homepage components

I'll decouple this for Next

src/next/components/InfoCols/InfoCols.tsx
```diff
@@ -0,0 +1,77 @@
...
+  )
+}
+
+const InfoBoxes = ({ infoBoxes }: Pick<InfoColsProps, "infoBoxes">) => {
+  return (
+    <div className="grid grid-cols-1 gap-x-8 gap-y-12 md:grid-cols-2 xl:grid-cols-3">
+      {infoBoxes.map((infoBox, idx) => (
+        <div key={idx} className="flex flex-col gap-4 items-start text-left">
+          <InfoBoxIcon icon={infoBox.icon} />
+          <div className="flex flex-col gap-1 items-start text-[#2C2E34]">
```
harishv7, line 33: why is the color hardcoded here?

src/next/components/InfoCols/InfoCols.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <h3 className="font-bold">{infoBox.title}</h3>
+            <p>{infoBox.description}</p>
+          </div>
+          {infoBox.buttonLabel && infoBox.buttonUrl && (
+            <a
```
harishv7, line 38: Might need to use the LinkComponent instead that I merged recently

src/classic/components/InfoCols/InfoCols.tsx
harishv7: can we just use the title as the key?

jacobkwan: i don't think that's a good idea, from my understanding:
1. react keys should be static/independent of the fields itself; it kinda defeats the point of having a key if when we change the title the entire component has to re-render because the key has changed
2. we also have no guarantees that titles are unique, for e.g current CMS behaviour uses some hardcoded string as title when adding new infoboxes

src/next/components/InfoCols/InfoCols.tsx
jacobkwan: originally hardcoded it as wasn't sure if this was just a special universal "black" for next theme or was intended to be site primary or secondary
will check with design and add the colours to config as required!

src/next/components/InfoCols/InfoCols.tsx
jacobkwan: [4da0053](https://github.com/isomerpages/isomer-components/pull/10/commits/4da005345aae3ed41794a3a59388d1259f9887e8)

src/next/components/InfoCols/InfoCols.tsx
jacobkwan: checked with design and added
[b920866](https://github.com/isomerpages/isomer-components/pull/10/commits/b920866d3535d3306898c5e2794bcd7d6a33a0b5)

jacobkwan:
merging first, can improve the `key` next time

PR #11, opened 2024-03-04
chore: split classic and next

jacobkwan:
moves currently built components to /classic, update storybook hierarchical `title` accordingly

moves component props to /common

have not added /next, will do in separate PR, starting with InfoPic and Button

also didn't really change render engine since its not priority now and requires more thought

**TODO: after merge, merge main to open PRs and make necessary changes**


dcshzj (approved):
nit: can we name the files in `~/common` to have `.ts` extension instead of `.tsx`?

PR #13, opened 2024-03-06
ISOM-778: Feat/allow link components

harishv7:
Resolves ISOM-778

- Allows custom link components from Next.js
- Usage is demonstrated in Navbar
- Minor fix to build script


src/engine/render.tsx
```diff
@@ -194,11 +212,12 @@ const RenderEngine = ({
...
   sitemap,
   components,
+  LinkComponent,
 }: IsomerBaseSchema): React.ReactNode => {
   if (components && components.length > 0) {
     const collatedComponents = components.map(
       (component: IsomerComponent, idx: number) => {
-        return <div key={idx}>{getComponent(component)}</div>
+        return <div key={idx}>{getComponent(component, LinkComponent)}</div>
       },
```
jacobkwan, line 221: might be tripping but it seems like we have some circular logic between layouts and render engine
seemed to be a loop when I was tracing to understand who supplies the `LinkComponent` prop

src/engine/render.tsx
```diff
@@ -40,14 +40,18 @@ export interface IsomerBaseSchema {
   components: IsomerComponent[]
   sitemap?: Sitemap
   indexable?: string[] // specifies which keys are indexable
+  LinkComponent?: any // Next.js Link
```
jacobkwan, line 43: is there a reason why we don't want to type the `LinkComponent`?
like `<a>` or `NextLink`
it would allow us to have better typed props on the components like in `Navbar`

src/engine/render.tsx
```diff
@@ -194,11 +212,12 @@ const RenderEngine = ({
   config,
   sitemap,
   components,
+  LinkComponent,
 }: IsomerBaseSchema): React.ReactNode => {
   if (components && components.length > 0) {
     const collatedComponents = components.map(
       (component: IsomerComponent, idx: number) => {
-        return <div key={idx}>{getComponent(component)}</div>
+        return <div key={idx}>{getComponent(component, LinkComponent)}</div>
```
jacobkwan, line 220: from what I gather, the design here is basically dependency injection where users of our components can decide what is the actual element used for links

just thinking out loud, perhaps in the future we could configure this once instead of passing it through a few functions, all the way to the component itself? for e.g I don't see why a layout would need to know about `LinkComponent`

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,178 @@
...
+    searchUrl?: string
+  }
+  LinkComponent: any
+}
+
+export const IsomerNav = ({
+  logo,
+  links,
+  search = { isEnabled: false },
+  LinkComponent = "a",
```
jacobkwan, line 38: should this be `NextLink` since this is the next themed navbar?

jacobkwan:
non blocking comments, looks good component wise
anyway I think the whole `render` side of things will probably change quite a bit so no point spending too much effort on it now

src/engine/render.tsx
harishv7: yea, there is circular or rather recursive logic - can update this in future PRs

src/engine/render.tsx
harishv7: fair point, this PR demonstrates the use of NextLink injected but does not optimise the render logic as such. again, refactoring the render logic will be done in a separate PR

src/engine/render.tsx
harishv7: Hmm yea, the only issue was to type as NextLink I need the dependency of nextjs, but this was supposed to be framework agnostic so I left it as any for now. Need a better way to type this

PR #18, opened 2024-03-12
feat: add Next Infobar component

jacobkwan:
## Problem
Closes https://linear.app/ogp/issue/ISOM-765/create-infobar-component-in-isomer-next

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


dcshzj:
!run chromatic

src/common/Infobar.ts
```diff
@@ -5,6 +5,8 @@ export interface InfobarProps {
   description?: string
   buttonLabel?: string
   buttonUrl?: string
+  secondaryButtonLabel?: string
+  secondaryButtonUrl?: string
```
seaerchin, line 9: how are we going to avoid prop drilling/build nested components in the future? this way of doing things seems to be limited to _just_ the infobar - note, for example, that this wants a nested button but is accepting props for it rather than accepting a button as a child. 

this might complicate things in the future, especially as we build more complex components. worst case is we end up with like a 30 props component (smth that happens in the wild)

src/templates/next/components/Infobar/Infobar.tsx
```diff
@@ -0,0 +1,41 @@
...
+          )}
+          {secondaryButtonLabel && secondaryButtonUrl && (
+            <Button
+              label={secondaryButtonLabel}
+              href={secondaryButtonUrl}
+              colorVariant="white"
+            />
+          )}
+        </div>
+      </div>
```
seaerchin, line 36: checking to see if these are mobile responsive

src/common/Infobar.ts
jacobkwan: yup agreed on the points raised, to be honest I think we need to spend some time to think about this after the prototype is done, when we start shifting towards designing the render engine and CMS portion of things

my initial guess is that we could design the render engine such that schemas would still contain nested props but the render engine would be able to parse the button related fields, construct the sub component (`Button` in this case) and pass them to the `Infobar`

src/templates/next/components/Infobar/Infobar.tsx
jacobkwan: good catch, forgot about it as it wasn't in design, opting to make it cols on smaller screens for now
[5c1ce43](https://github.com/isomerpages/isomer-components/pull/18/commits/5c1ce430cf949e0d9db82e25071823ed4fcc7a13)

seaerchin: Just an aside, how much do we care about a11y? I don't know what's the stance here but just pointing it out

src/common/Infobar.ts
seaerchin: If this is something to be decided and this is a temporary component that'll be refactored then sure.

dcshzj (approved):

PR #19, opened 2024-03-13
feat: add theme support to storybook

dcshzj:
## Problem

We currently default to using Lato for everything, even though Isomer Next uses Inter.

## Solution

**Breaking Changes**

**Features**:

- Use Storybook's native support for switching themes to allow us to specify the theme to use.

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -91,20 +91,12 @@ const preview: Preview = {
 }
 
 export const decorators: any = [
-  withThemeByClassName({
-    themes: {
-      light: "light",
-      dark: "dark",
-    },
-    defaultTheme: "light",
-  }),
   withThemeByDataAttribute({
     themes: {
-      light: "light",
-      dark: "dark",
+      "Isomer Classic": "isomer-classic",
+      "Isomer Next": "isomer-next",
     },
-    defaultTheme: "light",
-    attributeName: "data-mode",
+    defaultTheme: "Isomer Next",
   }),
 ]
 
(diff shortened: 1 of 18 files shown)


.storybook/preview.ts
```diff
@@ -91,20 +91,12 @@ const preview: Preview = {
...
-  }),
   withThemeByDataAttribute({
     themes: {
-      light: "light",
-      dark: "dark",
+      "Isomer Classic": "isomer-classic",
+      "Isomer Next": "isomer-next",
     },
-    defaultTheme: "light",
-    attributeName: "data-mode",
```
jacobkwan, line 107: do we not need to specify `attributeName: "data-theme"`? or is that the default attributeName

src/index.css
```diff
@@ -1,4 +1,15 @@
...
+
+@layer base {
+  html[data-theme="isomer-next"] {
+    font-family: "Inter", system-ui, sans-serif;
+  }
+
+  html[data-theme="isomer-classic"] {
+    font-family: "Lato", system-ui, sans-serif;
+  }
+}
```
jacobkwan, line 15: since we have this, should we remove the `fontFamily` setting from tailwindconfig?

.storybook/preview.ts
dcshzj: Yup that's [the default](https://github.com/storybookjs/storybook/blob/next/code/addons/themes/docs/api.md#withthemebydataattribute).

src/index.css
dcshzj: Yeah good point, removed in efe45a1b1c5f3b6c071bcc7a3a870be4cb85309e.

jacobkwan (approved):

PR #22, opened 2024-03-13
feat: add clear button variants and make button more flexible

jacobkwan:
## Problem
add button variants and colorScheme

hover colors are abit jank but can fix when design reviews

Closes https://linear.app/ogp/issue/ISOM-780/add-clear-button-variant

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


jacobkwan:
!run chromatic

dcshzj:
Hmm @jacobkwan for this one I'm a bit surprised at your approach. Is it possible to do something similar to [how Chakra/OGPDS does it](https://chakra-ui.com/docs/components/button/usage)? The `variant` prop can take in the outline/clear options, then the colorScheme can determine how the button colour looks like.

jacobkwan:
> Hmm @jacobkwan for this one I'm a bit surprised at your approach. Is it possible to do something similar to [how Chakra/OGPDS does it](https://chakra-ui.com/docs/components/button/usage)? The `variant` prop can take in the outline/clear options, then the colorScheme can determine how the button colour looks like.
> 
> I'm thinking using the variant approach would be more extensible in the future.

its ok I'm also surprised after seeing how chakra did it, clown moment

src/templates/next/components/Button/Button.stories.tsx
```diff
@@ -23,38 +23,88 @@ Default.args = {
   href: "/faq",
 }
 
-export const WithRightIcon = Template.bind({})
-WithRightIcon.args = {
+export const White = Template.bind({})
```
seaerchin, line 26: can we just do a 2 level map over `colorScheme` + `varaint`

src/templates/next/components/Button/Button.tsx
```diff
@@ -1,45 +1,120 @@
...
-
+const BaseButton = ({
+  label,
+  href,
+  rightIcon,
+  className = "",
+  isLinkVariant = false,
+}: ButtonProps & { className?: string; isLinkVariant?: boolean }) => {
   return (
     <a
```
seaerchin, line 31: not part of this pr but the base button is an `a`?

src/templates/next/components/Button/Button.tsx
```diff
@@ -1,45 +1,120 @@
...
-        <Icon className="size-6" />
-      </div>
-    )
+const RightIcon = ({ rightIcon }: Pick<ButtonProps, "rightIcon">) => {
+  if (!rightIcon) {
+    return null
   }
+  const Icon = SUPPORTED_ICONS_MAP[rightIcon]
+  return (
+    <div>
```
seaerchin, line 17: this isn't an issue but just curious why we need a `div` here

seaerchin (commented):
trying to get familiar with the codebase, some comments

src/templates/next/components/Button/Button.tsx
jacobkwan: yep, because semantically they are all links

src/templates/next/components/Button/Button.tsx
jacobkwan: originally added because the icon was shrinking on smaller screens, but fixed

[1f67649](https://github.com/isomerpages/isomer-components/pull/22/commits/1f6764961d958f58a61e6e1d00c1f6f67d572bf8)

src/templates/next/components/Button/Button.stories.tsx
jacobkwan: don't understand this comment, can elaborate?

seaerchin: The current way of doing seems like we'd need to create different variations manually when all that's changing is `colorScheme` and `variant`. 

Maybe we should do a 2 level map like
```ts
variants.map(() => colourSchemes.map)
```

src/templates/next/components/Button/Button.tsx
seaerchin: Kek

seaerchin: This effectively blocks us from having actual buttons - if that's possible on next, then this decision needs to be well thought through.

jacobkwan: fair, but I think it'll be quite easy to just differentiate them in the future if needed
this decision now doesn't really lock us into anything long term

src/templates/next/components/Button/Button.stories.tsx
jacobkwan: [2437d66](https://github.com/isomerpages/isomer-components/pull/22/commits/2437d66a722c85b851a2d420ad60d35b7b796158)

https://65dd949d255950353c94d5cc-efgijxxgyf.chromatic.com/?path=/story/next-components-button--colors-and-variants

PR #25, opened 2024-03-14
feat: use separate theme presets for each template

dcshzj:
## Problem

Our theme presets is a little messy right now.

## Solution

**Breaking Changes**

**Features**:

- Separate the theme presets for each template. This also allows us to publish the presets as a separate NPM package in the future, and downstream sites can just import the package, layer the site-specific theme colours on top, and build the final site.

### DIFF
--- a/src/presets/classic.js
+++ b/src/presets/classic.js
@@ -3,6 +3,21 @@ export default {
   theme: {
     extend: {
       colors: {
+        border: {
+          light: "#d6d6d6",
+        },
+        canvas: {
+          base: "#ffffff",
+          inverse: "#000000",
+          translucentGrey: "#00000080",
+        },
+        header: "#2164da",
+        headings: "#6d58bb",
+        navItems: "#323232",
+        paragraph: "#344054",
+        prose: "#484848",
+        subtitle: "#344054",
+        subtleLink: "#767676",
         site: {
           primary: {
             DEFAULT: "#6031b6",
@@ -13,6 +28,16 @@ export default {
           },
         },
       },
+      typography: ({ theme }) => ({
+        isomer: {
+          css: {
+            "--tw-prose-body": theme("colors.prose"),
+            "--tw-prose-headings": theme("colors.headings"),
+            "--tw-prose-bullets": theme("colors.prose"),
+            "--tw-prose-links": theme("colors.site.secondary"),
+          },
+        },
+      }),
     },
   },
 }
--- a/src/presets/classic.js
+++ b/src/presets/classic.js
@@ -1,0 +1,18 @@
+/** @type {import('tailwindcss').Config} */
+export default {
+  theme: {
+    extend: {
+      colors: {
+        site: {
+          primary: {
+            DEFAULT: "#6031b6",
+            hover: "#4b268e",
+          },
+          secondary: {
+            DEFAULT: "#4372d6",
+          },
+        },
+      },
+    },
+  },
+}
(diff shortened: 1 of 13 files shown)


src/presets/classic.js
```diff
@@ -0,0 +1,43 @@
...
+        site: {
+          primary: {
+            DEFAULT: "#6031b6",
+            hover: "#4b268e",
+          },
+          secondary: {
+            DEFAULT: "#4372d6",
+          },
+        },
+      },
```
jacobkwan, line 30: just checking why for classic we put the site colors into the preset but next we dont

tailwind.config.js
```diff
@@ -1,61 +1,27 @@
 /** @type {import('tailwindcss').Config} */
 export default {
   content: ["./index.html", "./src/**/*.{js,ts,jsx,tsx}"],
+  presets: [
+    require("./src/presets/next.js"),
+    // Note: This is here temporarily until we can figure out how to load the
+    // presets dynamically depending on the template being used.
+    require("./src/presets/classic.js"),
+  ],
   theme: {
```
jacobkwan, line 10: how does it work if `next.js` and `classic.js` have the same color keys but different values? does it override each other or

src/templates/next/components/InfoCols/InfoCols.tsx
```diff
@@ -19,7 +19,7 @@ const InfoBoxIcon = ({ icon }: { icon?: SupportedIconName }) => {
   const Icon = SUPPORTED_ICONS_MAP[icon]
   return (
     <div>
-      <Icon className="w-10 h-auto text-secondary" />
+      <Icon className="w-10 h-auto text-site-secondary" />
```
jacobkwan, line 22: yeeesh can help me change this to primary 😆  it's supposed to be the orange according to design haha, brown icons mad ugly

jacobkwan:
thanks for adding primary.100! it will be used in `ContentPageHeader` in case you're wondering

tailwind.config.js
dcshzj: Yeah, the [last configuration will win](https://tailwindcss.com/docs/presets#extending-multiple-presets). In this case we will take the classic one, but if it is configured inside `tailwind.config.js` then it will override that.

So it's like this:

next.js < classic.js < tailwind.config.js

src/presets/classic.js
dcshzj: This is a set of defaults, honestly not too sure about next yet cos we haven't really finalised the set of site-specific colours, but I can add it inside next just in case.

dcshzj: Added in 0eef8dc0c1dc9a95fa438b8f72335b322ced07f8.

src/templates/next/components/InfoCols/InfoCols.tsx
dcshzj: Updated in 0eef8dc0c1dc9a95fa438b8f72335b322ced07f8.

dcshzj:
> I will fix the other `next` components once this is merged (Button for e.g)

Yeah can sounds good

tailwind.config.js
jacobkwan: oh yikes, yea that's not great since its quite likely that presets use the same keys given that its a common design language and all

for now it seems that somehow we dont have any overlaps but I think quite scary in long term lol

jacobkwan (approved):
should we create a task to further "decouple" different theme presets so that we don't have to worry about them overriding each other?

dcshzj:
> should we create a task to further "decouple" different theme presets so that we don't have to worry about them overriding each other?

Yeah can, let me put that in but will deprioritise for now

PR #27, opened 2024-03-14
feat: add ordered and unordered list components for next

dcshzj:
Adds the ordered and unordered lists as components for the next template.

Also fixed paragraph component to add an icon for external links.


dcshzj:
!run chromatic

src/templates/next/components/OrderedList/OrderedList.tsx
```diff
@@ -0,0 +1,25 @@
...
+import UnorderedList from "../UnorderedList"
+import Paragraph from "../Paragraph"
+
+const OrderedList = ({ start, items }: OrderedListProps) => {
+  return (
+    <ol className="list-decimal ps-8 my-1" start={start}>
+      {items.map((item) => {
+        if (typeof item === "string") {
+          return (
+            <li key={Math.random()} className="[&_p]:inline pl-2 my-1">
```
jacobkwan, line 11: should we add a TODO here for when we think about rendering the components in the CMS? question applies for unordered list too
For the static site itself I guess it doesnt matter since the key will never change after build

src/templates/next/components/UnorderedList/UnorderedList.tsx
```diff
@@ -0,0 +1,25 @@
+import type { UnorderedListProps } from "~/common"
+import OrderedList from "../OrderedList"
+import Paragraph from "../Paragraph"
+
+const UnorderedList = ({ items }: UnorderedListProps) => {
+  return (
+    <ul className="list-disc ps-8 my-1">
```
jacobkwan, line 7: from eye balling it seems like padding in design is more generous
https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=1516-13636&mode=design&t=cDs1tIIi9QBLaFaE-0

jacobkwan (commented):
actually I have a question regarding Content in Isomer Next
does this mean we are not going with the tiptap editor route anymore but with complete block editing instead?

also this means our lists cannot nest other types of blocks (e.g images) but only other lists or strings right?

src/templates/next/components/OrderedList/OrderedList.tsx
dcshzj: Hmm I'm a bit more inclined to leave it out first, cos I feel the decisions made here should not be influenced by how the CMS would handle it. I think we can look at everything again when we start building the CMS side of things.

src/templates/next/components/UnorderedList/UnorderedList.tsx
dcshzj: Hmm I feel might be a bit too generous but sure, updated in 7bc1d1a154830958de76e9cb89f531c5ddd2f78f.

jacobkwan (approved):

PR #34, opened 2024-03-18
feat: add toc component

jacobkwan:
## Problem
https://linear.app/ogp/issue/ISOM-783/create-toc-component-in-isomer-next

https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=1597-5191&mode=design&t=kbBJVaROCZKq7Kep-0

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


jacobkwan:
!run chromatic

src/common/TableOfContents.ts
```diff
@@ -0,0 +1,12 @@
+export interface HeadingLink {
+  content: string
+  anchorLink: string
+  level: number
+}
+
+export interface TableOfContentsProps {
+  type: "tableofcontents"
+  headings: HeadingLink[]
```
dcshzj, line 9: nit: maybe we can name this as `items` instead?

src/templates/next/components/TableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,25 @@
...
+const TableOfContents = ({ headings }: TableOfContentsProps) => {
+  return (
+    <div className="flex flex-col gap-3 pl-5 border-l-4 border-content w-[18.75rem]">
+      <p className="text-lg text-content-strong font-semibold">On this page</p>
+      <div className="flex flex-col gap-3">
+        {headings.map((heading) => (
+          <div className="flex flex-row gap-2 items-start">
+            <BiRightArrowAlt className="size-6 shrink-0 text-interaction-main" />
+            <a
+              href={heading.anchorLink}
```
dcshzj, line 13: question: this reminds me, it doesn't seem like we are auto-generating the anchor tags for the heading component. Should we do that as well?

This component is likely going to be auto-generated, so users will only say they want a TOC block. The contents will be handled by us.

src/templates/next/components/TableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,25 @@
+import { BiRightArrowAlt } from "react-icons/bi"
+import { TableOfContentsProps } from "~/common"
+
+const TableOfContents = ({ headings }: TableOfContentsProps) => {
+  return (
+    <div className="flex flex-col gap-3 pl-5 border-l-4 border-content w-[18.75rem]">
+      <p className="text-lg text-content-strong font-semibold">On this page</p>
+      <div className="flex flex-col gap-3">
+        {headings.map((heading) => (
```
dcshzj, line 9: nit: can do prop spreading instead

src/templates/next/components/TableOfContents/TableOfContents.tsx
jacobkwan: yea, I thought of that but decided not to because I didn't want to couple the anchor link creation with this component

I imagine the anchor link creation should be whenever we create a heading then we add `id=foo` to the heading
Then when the user adds ToC, we get the anchor links from the headings and pass it into this ToC component

src/templates/next/components/TableOfContents/TableOfContents.tsx
jacobkwan: [bff3d13](https://github.com/isomerpages/isomer-components/pull/34/commits/bff3d13e9d5f1f30f3e4bd132777a7626b5313a1)

src/common/TableOfContents.ts
jacobkwan: [bff3d13](https://github.com/isomerpages/isomer-components/pull/34/commits/bff3d13e9d5f1f30f3e4bd132777a7626b5313a1)

src/templates/next/components/TableOfContents/TableOfContents.tsx
dcshzj: Yeah oops I mean like creating the anchor tags for the Heading component, so that the anchor links would work here.

src/common/TableOfContents.ts
```diff
@@ -0,0 +1,12 @@
+export interface HeadingLink {
+  content: string
+  anchorLink: string
+  level: number
```
harishv7, line 4: can this be a optional prop?

src/common/TableOfContents.ts
```diff
@@ -0,0 +1,12 @@
...
+}
+
+export interface TableOfContentsProps {
+  type: "tableofcontents"
```
harishv7, line 8: why dont we use camelcase for naming convention?

jacobkwan: sure, @dcshzj was there any reason why you went with this naming convention originally? If not I can change them all in another PR

src/templates/next/components/TableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,22 @@
+import { BiRightArrowAlt } from "react-icons/bi"
+import { TableOfContentsProps } from "~/common"
+
+const TableOfContents = ({ items }: TableOfContentsProps) => {
+  return (
+    <div className="flex flex-col gap-3 pl-5 border-l-4 border-content w-[18.75rem]">
+      <p className="text-lg text-content-strong font-semibold">On this page</p>
```
seaerchin, line 7: not too fussed on this but do check with design if this should be customisable

src/templates/next/components/TableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,22 @@
+import { BiRightArrowAlt } from "react-icons/bi"
+import { TableOfContentsProps } from "~/common"
+
+const TableOfContents = ({ items }: TableOfContentsProps) => {
+  return (
+    <div className="flex flex-col gap-3 pl-5 border-l-4 border-content w-[18.75rem]">
```
seaerchin, line 6: should we auto collapse on mobile views? might be that the toc takes up a huge amount of space prior to content. (can check w/ design, also ok if we don't)

src/common/TableOfContents.ts
seaerchin, line 4: this doesn't seem to be used oddly enough. we might want to remove it

seaerchin (commented):
again, trying to get familiar with the codebase, so pinch of salt.

this toc implementation is fine, just a question for the future - do we want to support nested levels and also autogeneration of toc? asking because eg: h1 should be above h2, but if we pass in manually it's possible to circumvent this

src/common/TableOfContents.ts
dcshzj: No preference here, originally went for lowercase alphabet cos I saw it more like an ID. Can change if needed.

src/templates/next/components/TableOfContents/TableOfContents.tsx
jacobkwan: [2e70a26](https://github.com/isomerpages/isomer-components/pull/34/commits/2e70a26ddf45267e07829c4ba5c376d6b2328bf2)

added `id` to `HeadingProps` for now, not auto populating `id` from `content` as there is no guarantee of headings being unique

I envision it eventually as 
1. user adds Heading, CMS populates `id` for user, whether it's UUID or content with suffix
2. user adds toc, CMS grabs the ids from the Heading (only h1s for now) and generates TOC automatically

src/templates/next/components/TableOfContents/TableOfContents.tsx
jacobkwan: yup it's not customisable

src/templates/next/components/TableOfContents/TableOfContents.tsx
jacobkwan: yea, we are still showing on mobile for now

jacobkwan:
> this toc implementation is fine, just a question for the future - do we want to support nested levels and also autogeneration of toc? asking because eg: h1 should be above h2, but if we pass in manually it's possible to circumvent this

yup toc should be autogenerated based on the content, this is just the component we will use to render it
and for now we will only pull in the H2s to the ToC (H1 is the title)
@seaerchin

seaerchin (commented):
lgtm but cos i lack context hence not approving

src/templates/next/components/TableOfContents/TableOfContents.tsx
dcshzj: Yeah I think this solution should be ok (to add the `id` field) but not quite ok with the CMS being the one that adds it. I think eventually it would be the render engine that does this (because it would have context on which headings there are), and we make it completely transparent to the user editing the page.

src/common/TableOfContents.ts
dcshzj: Eh sorry I disagree here. The prop may not be used but we want to make sure minimally the props that we need are correct. I think the level is definitely needed because we need to perform indentation on the TOC.

@jacobkwan in the interest of time I think yes it's ok if we don't implement the logic to indent the items in the TOC, but I think it is still useful to have this right now.

dcshzj (approved):
Clarified offline, missed the comment about only supporting H2s

PR #36, opened 2024-03-20
feat(img): add img component

kishore03109:
## Problem

add images

### DIFF
--- a/src/templates/next/components/Image/Image.tsx
+++ b/src/templates/next/components/Image/Image.tsx
@@ -1,7 +1,7 @@
-import type { ImageProps } from "~/common"
+import { ImageProps } from "~/common"
 
 const Image = ({ src, alt, width }: Omit<ImageProps, "type">) => (
-  <img src={src} alt={alt} width={`${width ?? 100}%`} height={"auto"} />
+  <img src={src} alt={alt} width={`${width ?? 100}%`} height="auto" />
 )
 
 const ImageComponent = ({ src, alt, width, href }: ImageProps) => {
--- a/src/templates/next/components/Image/Image.stories.tsx
+++ b/src/templates/next/components/Image/Image.stories.tsx
@@ -1,0 +1,45 @@
+import type { Meta, StoryFn } from "@storybook/react"
+import Image from "./Image"
+import type { ImageProps } from "~/common"
+
+export default {
+  title: "Next/Components/Image",
+  component: Image,
+  argTypes: {},
+  parameters: {
+    themes: {
+      themeOverride: "Isomer Next",
+    },
+  },
+} as Meta
+
+// Template for stories
+const Template: StoryFn<ImageProps> = (args) => <Image {...args} />
+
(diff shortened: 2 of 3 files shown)


kishore03109:
* **#36** <a href="https://app.graphite.dev/github/pr/isomerpages/isomer-components/36?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @kishore03109 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

kishore03109:
!run-e2e

src/templates/next/components/Image/Image.tsx
```diff
@@ -0,0 +1,27 @@
+import { ImageProps } from "~/common"
+
+const Image = ({ src, alt, width }: ImageProps) => (
+  <img src={src} alt={alt} width={`${width ?? 100}%`} />
```
jacobkwan, line 4: curious why we added `height=auto` for classic but not here?

src/templates/classic/components/Image/Image.tsx
```diff
@@ -1,12 +1,10 @@
...
 import { ImageProps } from "~/common"
 
 const Image = ({ src, alt, width }: ImageProps) => (
-  <img src={src} alt={alt} width={`${width ?? 100}%`} />
+  <img src={src} alt={alt} width={`${width ?? 100}%`} height="auto" />
 )
 
-
-const ImageComponent = ({ src, alt, width, href }: ImageProps) => {
+const ImageComponent = ({ src, alt, width, href, type }: ImageProps) => {
```
jacobkwan, line 7: I don't think we need to destructure `type`, it's only used by the render engine when deciding what component to render

kishore03109: hmm its just a typing thing at the moment, since the props for Image are `ImageProps`, in which `type` is required. 

type error noticed in main line today:
Screenshot 2024-03-21 at 8 18 05 AM

src/templates/next/components/Image/Image.tsx
kishore03109: tiptap put as auto, so i just added it in. 

dont have strong opinions about this at the moment, will just add this to the classic as well

src/templates/classic/components/Image/Image.tsx
kishore03109: wah actually got quite a number of errors, opening #40 to make this more obvious to devs. after fix lets make it a requirement to passing ci before merge to main line!

jacobkwan: ah ok, yea should probably use`Omit<ComponentProps, "type">` for the actual component props
the `type` is only needed by render engine

can see Next `Paragraph` for reference

kishore03109: ya thats better, changed

PR #37, opened 2024-03-20
feat: introduce navbar component

dcshzj:
Figma: https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=1811-9913&mode=dev

Note:
- There is no animation currently except for hovering, so the navbar items will appear immediately instead of easing in. Deprioritising building that animation first.
- Also adjusted the orderedlist prop such that the `start` prop is optional.

### DIFF
--- a/src/common/Navbar.ts
+++ b/src/common/Navbar.ts
@@ -1,21 +1,26 @@
-export interface NavbarLink {
-  type: "single" | "dropdown"
+interface NavbarLocalSearchProps {
+  type: "localSearch"
+  searchUrl: string
+}
+
+interface NavbarSearchSGProps {
+  type: "searchSG"
+}
+
+export interface NavbarItem {
   name: string
-  eventKey?: string
-  url?: string
-  links?: NavbarLink[]
+  url: string
+  description?: string
+  items?: Omit<NavbarItem, "items">[]
 }
 
-export interface IsomerNavProps {
+export interface NavbarProps {
   type: "navbar"
-  id?: string
-  logo: { url: string; alt: string }
-
-  links: NavbarLink[]
-  search?: {
-    isEnabled: boolean
-    searchUrl?: string
-  }
+  logoUrl: string
+  logoAlt: string
+  search?: NavbarLocalSearchProps | NavbarSearchSGProps
+  items: NavbarItem[]
+  LinkComponent?: any
 }
 
-export default IsomerNavProps
+export default NavbarProps
--- a/src/presets/next.js
+++ b/src/presets/next.js
@@ -47,6 +47,9 @@ export default {
             hover: "#333333",
             active: "#333333",
           },
+          sub: {
+            DEFAULT: "#f3f3f3",
+          },
         },
         site: {
           primary: {
--- a/src/templates/next/components/Navbar/Navbar.stories.tsx
+++ b/src/templates/next/components/Navbar/Navbar.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta, StoryFn } from "@storybook/react"
-import Navbar, { IsomerNavProps, NavbarLink } from "./Navbar"
-import { Navbar as NavbarConfig } from "~/config/navbar"
+import type { NavbarProps } from "~/common"
+import Navbar from "./Navbar"
 
 export default {
   title: "Next/Components/Navbar",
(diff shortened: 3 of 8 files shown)


src/common/Navbar.ts
```diff
@@ -1,21 +1,26 @@
...
-  links: NavbarLink[]
-  search?: {
-    isEnabled: boolean
-    searchUrl?: string
-  }
+  logoUrl: string
+  logoAlt: string
+  search?: NavbarLocalSearchProps | NavbarSearchSGProps
+  items: NavbarItem[]
+  LinkComponent?: any
```
seaerchin, line 23: not expecting this; seems dangerous to do, especially in a TS codebase

src/presets/next.js
```diff
@@ -47,6 +47,9 @@ export default {
             hover: "#333333",
             active: "#333333",
           },
+          sub: {
+            DEFAULT: "#f3f3f3",
```
seaerchin, line 51: odd ot see capitalised properties when the rest is lower case

src/templates/next/components/Navbar/Navbar.stories.tsx
```diff
@@ -14,13 +14,104 @@ export default {
...
+  },
+  items: [
+    {
+      name: "Expandable nav item",
+      url: "/item-one",
+      items: [
+        {
+          name: "PA's network one",
+          url: "/item-one/pa-network-one",
+          description: "Click here and brace yourself for mild disappointment.",
```
seaerchin, line 37: kekw

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
-  return classes.filter(Boolean).join(" ")
-}
+const Navbar = ({
+  logoUrl,
+  logoAlt,
+  search,
+  items,
+  LinkComponent = "a",
+}: NavbarProps) => {
+  const [navItemOpen, setNavItemOpen] = useState(-1)
```
seaerchin, line 21: could we rename to `currentlyOpenNavItem` or `openNavItemIdx`? reads abit oddly now

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+  const [isHamburgerOpen, setIsHamburgerOpen] = useState(false)
+  const [isSearchOpen, setIsSearchOpen] = useState(false)
 
-export interface IsomerNavProps {
-  id?: string
-  logo: { url: string; alt: string }
+  const currentPathName =
+    typeof window !== "undefined" ? window.location.pathname : ""
```
seaerchin, line 26: nit: flip teh condition so easier to read.

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+              type="submit"
+              aria-label="Search this site"
+            >
+              <BiSearch className="text-2xl mt-0.5" />
+            </button>
+          </form>
+        )}
+
+        {/* Hamburger menu for small screens */}
+        <button className="block xl:hidden h-10">
```
seaerchin, line 76: seems to be a disjunct between actual buttons and what we have as a button in our codebase (actually a link) 

not sure if important enuf to fix?

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+          {isHamburgerOpen ? (
+            <button
```
seaerchin, line 78: we seem to be nesting a `button` within another `button`?

i understand that the above button seems to be hidden on small screens but this nevertheless bad behaviour. could we change it?

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+              onClick={() => {
+                setIsHamburgerOpen(!isHamburgerOpen)
```
seaerchin, line 80: supernit: prefer setting to `false` since we're not toggling

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+                setNavItemOpen(-1)
+              }}
+              aria-label="Close navigation menu"
+            >
+              Close
+              <BiX className="inline ml-1 -mt-0.5 text-2xl" />
+            </button>
+          ) : (
+            <button
+              onClick={() => setIsHamburgerOpen(!isHamburgerOpen)}
```
seaerchin, line 90: supernit: prefer setting to `true` since we're not toggling

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+        </button>
+      </div>
+
+      {/* Search bar (for mobile/tablet) */}
+      <div className="block lg:hidden">
+        {isSearchOpen && search && search.type === "localSearch" && (
+          <form action={search.searchUrl} method="get">
+            <div className="flex flex-row gap-4 h-10 px-14">
+              <input
+                type="search"
```
seaerchin, line 105: technically `input`s are supposed to have a corresponding `label`. (see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)) 

i only mention this because it seems liek we care abuot a11y

src/templates/next/components/Navbar/Navbar.tsx
```diff
@@ -1,178 +1,285 @@
...
+                          Close
+                          <BiX className="inline ml-1 -mt-0.5 text-2xl" />
+                        </button>
+                      </div>
+
+                      <ul className="flex flex-row flex-wrap pt-12 gap-x-36 gap-y-8">
+                        {items.map((subItem) => (
+                          <li key={subItem.name} className="w-2/5">
+                            <div className="flex flex-col gap-1">
+                              <LinkComponent
```
seaerchin, line 191: can we put the props given to `linkComponent` here as the type

seaerchin (commented):
some comments, due to lack of familiarity w codebase atm, feel free to ignore. 

again, this reads more like we're trying to use props to dictate the behaviour of certain elements, whereas we should be looking to pass in a component that does it for us. 

eg: `Navbar` uses a custom search component that's baked in here whereas perhaps what we want is to have it dictate only layout?
```ts
<Navbar items = {items}>
<Searchbar search = {search}>
<SearchContent /> 
</Navbar>
```

src/common/Navbar.ts
dcshzj: Deprioritising this

src/presets/next.js
dcshzj: Tailwind feature: https://tailwindcss.com/docs/theme#core-plugins

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: Using `<button>` here would be semantically correct, because it is not a link like our usual buttons.

dcshzj: Using `<button>` here would be semantically correct, because it is not a link like our usual buttons.

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: Adjusted in 33a5297969cb617a3746f06d60fcfb9ee361f6c1 to use `<div>` instead on the parent, missed this out when changing.

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: 33a5297969cb617a3746f06d60fcfb9ee361f6c1

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: 33a5297969cb617a3746f06d60fcfb9ee361f6c1

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: Added `title` and `aria-label` instead, recommended in https://stackoverflow.com/a/23151986

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: Not sure what you mean here, the `LinkComponent` is effectively the `<a>` tag. The actual typings of `LinkComponent` will be fixed later.

src/templates/next/components/Navbar/Navbar.tsx
dcshzj: 68770a2b87545dbe121d276aa45e6b29c7788612

src/templates/next/components/Navbar/Navbar.tsx
seaerchin: Yea i meant that we should relook at our buttons. 

Understand that it's out of scope so ok with this

src/templates/next/components/Navbar/Navbar.tsx
seaerchin: Ok sgtm

seaerchin:
Lgtm but maybe wait on @jacobkwan or @alexanderleegs ?

jacobkwan:
!run chromatic

tsconfig.json
```diff
@@ -23,7 +23,7 @@
 
     /* Additional Checks */
     "noUnusedLocals": false /* Report errors on unused locals. */,
-    "noUnusedParameters": true /* Report errors on unused parameters. */,
+    "noUnusedParameters": false /* Report errors on unused parameters. */,
```
jacobkwan, line 26: would prefer keeping this for easy visibility, since we have a couple of unused props in components as their functionalities aren't implemented yet

jacobkwan (commented):
mostly design related things, can go through during the review
1. responsiveness (padding right before leaving desktop variant, not sure where the search should go based on design
2. text color for expanded menu items should be a mixture of `text-content` and `text-content-medium`
3. underline offset for links should be larger

tsconfig.json
dcshzj: Re-enabled in c8c9e11f7ec67ceb5e79785977ce53690b23b6f4. It spilled over cos I was trying to build the repo (enabling this will block builds).

PR #45, opened 2024-03-22
feat: add Next ArticleCard

jacobkwan:
## Problem
https://linear.app/ogp/issue/ISOM-727/create-articlecards-component-in-isomer-next

https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=2098-12269&mode=design&t=34nfAOpBPgpKlyQl-0

Width of image on large screen not quite right but couldn't get it to work so deprio for now


jacobkwan:
!run chromatic

src/templates/next/components/ArticleCard/ArticleCard.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+Default.args = {
+  lastUpdated: "December 2, 2023",
+  category: "Research",
+  title:
+    "A journal on microscopic plastic and their correlation to the number of staycations enjoyed per millennials between the ages of 30-42, substantiated by research from IDK university",
+  url: "/",
+  description:
+    "We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year. We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year.",
+  image: {
+    src: "https://picsum.photos/500/500",
```
dcshzj, line 29: nit: I think we shouldn't use a randomly generated image cos otherwise Chromatic will keep saying that there are changes to be reviewed, which is quite annoying.

src/common/ArticleCard.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface ArticleCardProps {
+  type: "articleCard"
+  lastUpdated: string
+  category: string
+  title: string
+  url: string
+  description: string
+  image?: Image
+  variant: "file" | "article"
+  fileDetails?: FileDetails
```
dcshzj, line 24: supernit: can we make this prop only available if the variant is set to `file`? That way it can be set as required.

src/templates/next/components/ArticleCard/ArticleCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+          className={`${Paragraph[2]} text-content-medium line-clamp-3 sm:line-clamp-2`}
+        >
+          {description}
+        </p>
+      </div>
+      {variant === "file" && fileDetails && (
+        <div
+          className={`${Paragraph[1]} text-hyperlink underline underline-offset-2`}
+        >{`Download (${fileDetails.type.toUpperCase()}, ${
+          fileDetails.size
```
dcshzj, line 59: question: wanted to check, would this be automatically populated somewhere or will the site editor be the one responsible?

src/templates/next/components/ArticleCard/ArticleCard.stories.tsx
jacobkwan: oh ya I just wanted to see how the image would resize responsively, will change back to placeholder

src/templates/next/components/ArticleCard/ArticleCard.tsx
jacobkwan: should be automatically populated when the site editor uploads a file

src/common/ArticleCard.ts
jacobkwan: [79659e4](https://github.com/isomerpages/isomer-components/pull/45/commits/79659e486148c1f71e324b29c01eb7896623372d)

src/templates/next/components/ArticleCard/ArticleCard.stories.tsx
jacobkwan: [519b473](https://github.com/isomerpages/isomer-components/pull/45/commits/519b4732b365199d9c1ce386a3e458bc0d508908)

src/templates/next/components/ArticleCard/ArticleCard.tsx
dcshzj: Hmm okay can, I think we might have to think about how to implement this on the render engine level, not a blocker for this PR.

Some thoughts:
1. Files can only be uploaded into the `public/files` folder and should not be present in any other place.
2. This means that the article card contains the URL to the file itself, and it should contain the information inside the `_collection.json` file
3. At build time we get information about the file from the link, then compute the file size to feed into the render engine.
4. Render engine builds the page with all the props available, providing this information into the CollectionCard component.

dcshzj (approved):

PR #46, opened 2024-03-25
Feat/collection layout v1

jacobkwan:
## Problem
First pass of Collections layout
Includes some clean ups to render engine

1. Refactor to use discriminated union for schema of different layouts
3. putting the Collection items as `page.items`, actually I was hoping don't need to include it in the json file at all, and instead construct the props from file system at build time in `getStaticProps`. But realised with files it's different because they don't follow the file system hierarchy, unlike a normal nested article page. @dcshzj can discuss this also when you're free

also as an aside I think some things are abit confusing / may get confusing over time if not on same page
like our concept of `Props` / `Schema` abit unclear,
Props sometimes need `type` (when talking to the render engine, in json files), sometimes dont need (when talking to the Component directly)
some props exposed to user, some not
etc etc

Wondering if there's a way to neaten up this front 😭 

### TODO
- [ ] have to change the json schema files in next-template repo and mti-corp

### DIFF
--- a/src/common/MetaHead.ts
+++ b/src/common/MetaHead.ts
@@ -1,6 +1,5 @@
 export interface MetaHeadProps {
   type: "metahead"
-  language?: "en"
   title: string
   description?: string
   noIndex?: boolean
--- a/src/common/MetaHead.ts
+++ b/src/common/MetaHead.ts
@@ -1,5 +1,6 @@
 export interface MetaHeadProps {
   type: "metahead"
+  language?: "en"
   title: string
   description?: string
   noIndex?: boolean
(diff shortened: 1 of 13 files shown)


jacobkwan:
!run chromatic

src/common/MetaHead.ts
```diff
@@ -1,5 +1,6 @@
 export interface MetaHeadProps {
   type: "metahead"
+  language?: "en"
```
dcshzj, line 3: nit: can remove this, the head tag would not be able to set the html language sadly

src/engine/types.ts
```diff
@@ -59,7 +60,7 @@ export type IsomerComponent = IsomerComponentProps & {
   indexable?: string[]
 }
 
-export type IsomerLayout = "homepage" | "content"
+export type IsomerLayout = "homepage" | "content" | "collection"
```
dcshzj, line 63: nit: this is no longer used I think

dcshzj (approved):

PR #47, opened 2024-03-26
feat: introduce next theme filter component

dcshzj:
Filter component for collection layout

Figma (for checkbox): https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=2190-15364&mode=design&t=GXqdgVHLaK0jqZMS-4

Figma (for the component itself): https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=2092-9617&mode=design&t=GXqdgVHLaK0jqZMS-4


src/templates/next/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { BiChevronDown } from "react-icons/bi"
+import { FilterProps } from "~/common"
+import { Heading } from "../../typography/Heading"
+
+const Filter = ({
+  filters,
+  appliedFilters,
+  setAppliedFilters,
+}: Omit<FilterProps, "type">) => {
+  const [showFilter, setShowFilter] = useState<Record<string, boolean>>({})
```
seaerchin, line 11: does `useState` still work with server sided react?

jacobkwan:
!run chromatic

src/templates/next/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const filterIndex = appliedFilters.findIndex(
+      (filter) => filter.id === filterId,
+    )
+    if (filterIndex > -1) {
+      const itemIndex = appliedFilters[filterIndex].items.findIndex(
+        (item) => item.id === itemId,
+      )
+      if (itemIndex > -1) {
+        const newAppliedFilters = [...appliedFilters]
+        newAppliedFilters[filterIndex].items.splice(itemIndex, 1)
```
seaerchin, line 23: what does this do? `itemIndex` could be greater than 1 and this seems to work iff `itemIndex <= 1`

seaerchin (commented):
low-value review but quite hard for me to discern how this actually works in practice + how we can guarantee that the filtering is correct + extensible next time if required. 

maybe we can huddle for this or have a chromatic?

src/templates/next/components/Filter/Filter.tsx
jacobkwan: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#deletecount

believe it's just removing the specific item from the applied filter items, seems correct to me

src/templates/next/layouts/render.tsx
```diff
@@ -43,6 +44,8 @@ export const renderComponent = ({
       return <CollectionCard {...component} />
     case "contentpageheader":
       return <ContentPageHeader {...component} />
+    case "filter":
+      return <Filter {...component} />
```
jacobkwan, line 48: similar comment left on the CollectionSearch PR if we really consider `Filter` a "component"

jacobkwan (commented):
Some design nits, can go through during review but roughly
rounded borders for hover/active effects, checkbox also rounded etc
font colour of `Filter by`

src/common/Filter.ts
```diff
@@ -0,0 +1,29 @@
+interface FilterItem {
+  id: string
+  label: string
+  count?: number
```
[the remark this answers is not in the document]
dcshzj: Yeah good point, originally was following design but don't make sense to have some counts missing. Updated in 5b2eb609005fc2460029f66d2f757dc103c64f35.

src/templates/next/components/Filter/Filter.tsx
dcshzj: Yup

src/templates/next/layouts/render.tsx
dcshzj: Let me reply there

src/templates/next/components/Filter/Filter.tsx
jacobkwan: actually im abit confused on this front
https://nextjs.org/learn/react-foundations/server-and-client-components

I thought it wouldn't work since next uses server components by default, but I also don't see anywhere in our `isomer-next-base-template` that we use the `use client` directive

can fill me in on what im missing here @dcshzj

dcshzj: Yup, it is mainly because we are using the pages router, which [renders components on the client side by default](https://www.wearedevelopers.com/magazine/next-js-app-router-explained). In contrast, the app router renders components on the server side by default, hence the need for `use client` in that scenario. Because the default is already client-side, so we don't need to add this directive at all in our components.

So when we are using SSG, we are building the pages on the server and [rendering as much as we can on the server](https://stackoverflow.com/a/75550697). Then, the HTML and CSS, along with the related JS content needed for client-side functionality to work, will be sent to the user. Hence, you will see the site with the proper HTML sent correctly to the user, but there will still be some client-side actions that are performed using the JS executed on the user's browser. The `useState` hook is one of them.

The server-side of things comes in under the `getStaticProps` function, so it knows what props to pass into the components to render as much as possible on the server. The `getStaticPaths` is the other function to know which routes the server needs to "access" to render for. This just leaves places where React hooks (which are client-side) are used, and that is then offloaded to the user's browser to execute.

So my "yup" earlier was referring to the fact that `useState` will still work regardless, but this is still done on the client side and not rendered on the server.

src/templates/next/layouts/render.tsx
dcshzj: Converted in 87c41a1a8d9db43baa97121307638aca908fde7c.

jacobkwan (approved):
LGTM can merge after resolving conflicts

PR #48, opened 2024-03-26
feat: introduce CollectionSearch component

dcshzj:
Figma: https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=2190-15656&mode=design&t=GXqdgVHLaK0jqZMS-4


jacobkwan:
!run chromatic

src/common/CollectionSearch.ts
```diff
@@ -0,0 +1,8 @@
+export interface CollectionSearchProps {
+  type: "collectionSearch"
```
jacobkwan, line 2: not sure if we want to remove `type` here since this component shouldn't be rendered by `renderComponent` directly?
was thinking of separating these two types of components

src/templates/next/components/CollectionSearch/CollectionSearch.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const CollectionSearch = ({
+  placeholder,
+  search,
+  setSearch,
+}: Omit<CollectionSearchProps, "type">) => {
+  return (
+    <label className="relative block">
+      <span className="sr-only">{placeholder}</span>
+      <span className="absolute inset-y-0 left-0 flex items-center pl-4">
```
jacobkwan, line 12: just curious why need to use relative and absolute for positioning the search icon

src/templates/next/layouts/render.tsx
```diff
@@ -41,6 +42,8 @@ export const renderComponent = ({
       return <Callout {...component} />
     case "collectionCard":
       return <CollectionCard {...component} />
+    case "collectionSearch":
+      return <CollectionSearch {...component} />
```
jacobkwan, line 46: related to earlier comment, actually I'm not sure if we should expose such "useless standalone" components to `renderComponent` since it should only ever be used in the context of say the `CollectionLayout`

src/templates/next/components/CollectionSearch/CollectionSearch.tsx
dcshzj: The `relative` here is for the `absolute` positioning to know which parent element to position relative to. Otherwise it will position based on the document itself.

This part references https://tailwindcss.com/docs/hover-focus-and-other-states#placeholder-text

jacobkwan: oh i meant like doesn't padding solve that already, but nvm non-issue

src/common/CollectionSearch.ts
dcshzj: Wah been thinking hard about this question. Fundamentally this is asking where do we draw the line on whether a component is for "internal" consumption (i.e. called directly from a layout) or for "external" consumption (i.e. added and edited directly by the site editor).

Right now this line is drawn at the schema level, as it tracks the enum of the components that it can use. The render engine is "dumb" in the sense that it will just render the appropriate component based on the type, but it does not care who is using the component.

If we remove the `type` here and call the component directly from the layout, means we shift the line to the layout level, and the schema is just a representation of what is codified inside the layout and component props. Not sure if this would be confusing because the schema contains information that cannot be done in TypeScript (e.g. prop input must match a certain pattern), so it will have to be some kind of hybrid.

I'm a bit more inclined towards keeping the current way, which means `renderComponent` doesn't care who is using it, and leave it entirely to the schema to determine which components are meant for internal or external consumption. Thoughts?

jacobkwan: don't quite understand why the line shifts, isn't the schema still telling what is allowed in the layout? In this case this entire concept of Search would be absent from the schema no?

dcshzj: > wait in this case, regardless you're going to call the Search component directly from the layout no? Or you mean you still want to wrap it in the renderComponent call?

Yeah originally intended to have it wrapped inside the `renderComponent` call which would mean that there is only a single entry point to use the components. The way I originally saw it was that components and layouts are like two different libraries, and layouts are making use of components using a single API endpoint.

> don't quite understand why the line shifts, isn't the schema still telling what is allowed in the layout? In this case this entire concept of Search would be absent from the schema no?

The line shifted because the schema is not the single source of truth on what kind of components a layout can take. The `renderComponent` function also has some responsibility in limiting the kind of components that are available. For instance:

1. If the `renderComponent` function contains all the components in the component library, then only the schema knows that there is a subset of components that can be used in a particular layout.
2. However, if the `renderComponent` function contains only some of the components in the component library, then it is possible for the schema to allow using components that are not supported by `renderComponent` and the schema will not know that it is not supported. The schema itself will also have its own list, which means that only the intersect (components in both lists) are supported, when the schema (i.e. the CMS) would expect that all the components in the schema (which is a subset of all available components) are supported.

There is a slight distinction here but I think as we discussed, there is a technical limitation that prevents the schema from being the single source of truth. The universe of components will need to be split into "internal" and "public" components, and the `renderComponent` function will have all "public" components available. So we will need to clearly distinguish the "internal" components such that it is not possible for anyone to consume them unless through another component or a layout.

src/templates/next/components/shared/CollectionSearch/CollectionSearch.tsx
```diff
@@ -0,0 +1,29 @@
+import { BiSearch } from "react-icons/bi"
+import type { CollectionSearchProps } from "../../../types/CollectionSearch"
+
+const CollectionSearch = ({
+  placeholder,
+  search,
+  setSearch,
+}: Omit<CollectionSearchProps, "type">) => {
+  return (
```
seaerchin, line 9: not entirely sure if it's possible but we might benefit from trying to debounce here? 

something we guard against is users typing in rapidly and we trigger researches on each intermediate query (eg: a-> ab -> abb -> abba)

src/templates/next/components/shared/CollectionSearch/CollectionSearch.tsx
```diff
@@ -0,0 +1,29 @@
+import { BiSearch } from "react-icons/bi"
+import type { CollectionSearchProps } from "../../../types/CollectionSearch"
```
seaerchin, line 2: an aside but maybe can do path aliasing here

src/common/CollectionSearch.ts
seaerchin: i think there's a slight disjoint here - the `type` should be marked on the component itself as that's a distinguishing property of what the component is (yes, it's internal and it won't be seen by others but still) 

a quick and dirty solution here might just be to have another property marking whether this is a top-level component seen by `renderComponent` or it's to be consumed by another component.

separately, we might want to flesh out our parsing strategy for the json (the ambiguity seems to be the reason for this question here)

jacobkwan: I think we should have the mental model that `type` is only a concept for the `RenderEngine` 
and the `RenderEngine` is only for components exposed on the CMS e.g Infobar, Infopic
the jsonschema and the page jsons should only care about these exposed components

on that note, I think we should use different terminology when talking about external vs internal components.
maybe `blocks` (external) vs `components` (internal)
will need to clean up our components and be very deliberate about what is a block and what is a component
for e.g `ContentPageHeader`. Should it be a `block`? Users shouldn't be able to add multiple of these in their Content Pages right? So do we make it a component instead and use it directly in the `ContentPageLayout`?

anyway yea I think the last commits you pushed are ok @dcshzj , so I'll approve first

jacobkwan (approved):
good point raised about debouncing, can fix in v2 (?)

src/templates/next/components/shared/CollectionSearch/CollectionSearch.tsx
dcshzj: Yeah I think it's a good point, but not going to optimise things for now until we integrate everything together

PR #50, opened 2024-03-27
feat: introduce pagination component for next

dcshzj:
Figma: https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=2190-15437&mode=design&t=BbAXYmG7rNWLCepZ-4


dcshzj:
!run chromatic

src/templates/next/components/shared/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
+import { BiLeftArrowAlt, BiRightArrowAlt } from "react-icons/bi"
+import type { PaginationProps } from "../../../types/Pagination"
+
+export const Pagination = ({
+  totalItems,
+  itemsPerPage,
```
seaerchin, line 6: note that we need some level of check for `itemsPerPage` - doesn't have to be here as it'll be used both on FE + end user site. 

maybe we can just default this to 10?

src/templates/next/components/shared/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+}: Omit<PaginationProps, "type">) => {
+  const totalPages = Math.ceil(totalItems / itemsPerPage)
+
+  return (
+    <nav className="flex flex-row w-full" aria-label="Pagination">
+      {/* Previous button */}
+      <button
+        className="flex flex-row p-1 gap-1 align-middle cursor-pointer disabled:cursor-not-allowed disabled:text-interaction-sub disabled:hover:bg-transparent hover:bg-interaction-main-subtle-hover"
+        aria-label="Previous page"
+        disabled={currPage === 1}
```
seaerchin, line 18: i think the assumption that the starting index is 1 is pretty important to make clear! 

not sure where the caller is but yea

src/templates/next/components/shared/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+        onClick={() => {
+          if (currPage !== 1) {
+            setCurrPage(currPage - 1)
+          }
```
seaerchin, line 22: we might want to strengthen this check, especially if the `currPage` is taken from page params (eg: `www.something.com/test?p=1` - users can manually access `p=0` and bug this)

src/templates/next/components/shared/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+        }}
+      >
+        <BiLeftArrowAlt className="text-2xl my-auto" />
+        <p className="hidden xs:inline underline text-lg">Previous</p>
```
seaerchin, line 26: i assume this is for a11y purposes but why not jjust attach a `label` to the button

src/templates/next/components/shared/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </button>
+
+      {/* Spacing */}
+      <div className="flex-grow" />
+
+      {/* Page number */}
+      <p className="my-auto xs:text-xl xs:leading-8">
+        Page {currPage} of {totalPages}
```
seaerchin, line 34: glancing at this seems wrong - maybe we need to close this in ``?

src/templates/next/components/shared/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+      {/* Next button */}
+      <button
+        className="flex flex-row p-1 gap-1 align-middle cursor-pointer disabled:cursor-not-allowed disabled:text-interaction-sub disabled:hover:bg-transparent hover:bg-interaction-main-subtle-hover"
+        aria-label="Next page"
+        disabled={currPage === totalPages}
+        onClick={() => {
+          if (currPage !== totalPages) {
+            setCurrPage(currPage + 1)
+          }
```
seaerchin, line 48: same comment here to strengthen said ceck

src/templates/next/components/shared/Pagination/Pagination.tsx
dcshzj: Done in 74a3c42f3bbc2aaed2c3acf63453b9e2b6a05e00.

src/templates/next/components/shared/Pagination/Pagination.tsx
dcshzj: This is correct, can verify using chromatic.

src/templates/next/components/shared/Pagination/Pagination.tsx
dcshzj: Done in 74a3c42f3bbc2aaed2c3acf63453b9e2b6a05e00.

src/templates/next/components/shared/Pagination/Pagination.tsx
dcshzj: The label is added at the top in the button using `aria-label`.

src/templates/next/components/shared/Pagination/Pagination.tsx
dcshzj: Hmm not doing this. I feel that the `itemsPerPage` is a required prop that should be provided by the caller. Providing defaults, understandably sane ones, may add a layer of complexity as it makes the caller provide one thing and the component providing another, which is then combined to implement the functionality of the component.

Overall I think forcing the caller to control everything would be safer in the longer term. Just in case, I added a minimum of 1 to the `itemsPerPage` prop below.

src/templates/next/components/shared/Pagination/Pagination.tsx
dcshzj: Yeah I was being explicit here with using page as the terminology to indicate that it should start with 1.

jacobkwan (approved):

PR #51, opened 2024-03-27
Fix/component review round 1

jacobkwan:
https://www.notion.so/opengov/Next-Component-Design-Reviews-302206df3b7442aa8e242baf01ce5b69

Adding layout level gutters in a separate PR


jacobkwan:
!run chromatic

src/templates/next/components/Siderail/Siderail.tsx
```diff
@@ -30,36 +30,46 @@ const SiderailMobile = ({
...
-                    >
-                      {title}
-                    </LinkComponent>
-                  )
-                })}
-            </div>
-          )
-        })}
+                {isCurrent ? (
+                  <p className="font-bold block py-4 px-5">{title}</p>
```
dcshzj, line 41: nit: can add an `aria-current="page"` in the `<li>` to indicate that the item represents the current page.

I think got a few other accessibility stuff to fix but going to ignore for now since this is mostly design fixes + we should use a proper accessibility checker later for all our components.

src/templates/next/components/Siderail/Siderail.tsx
```diff
@@ -82,37 +92,46 @@ const SiderailDesktop = ({
...
+                </LinkComponent>
+              )}
+              {isCurrent && (
+                <ul>
+                  {childPages?.map(({ url, title }, index) => {
+                    return (
+                      <li>
+                        <LinkComponent
+                          href={url}
+                          className={`block py-2.5 pl-10 hover:bg-interaction-sub active:underline active:underline-offset-2 ${
```
dcshzj, line 120: nit: should we have a padding-right here also similar to mobile?

dcshzj (approved):
No blockers

PR #52, opened 2024-03-27
fix: add gutters for homepage layout

jacobkwan:
Use custom css class "component-content" and target it at the layout level to enforce max width and padding-x

Going with this approach so that components can be layout agnostic (as much as possible)
There are still sensible defaults on the component - you will see some `px`, but these will then be overridden by the styles specified by the layout.

Also restructured the edited components slightly, such that top level div for each component does not contain padding/width related styles. Imagine that one day if needed, the bg-color could just be easily added to that div.

Only made the changes to the components that are in Homepage layout, but still nearly died trying to check that nothing broke.

Didn't fix Hero Split variant as it has two background images which seems incompatible with this implementation @dcshzj 💀 

Not sure if components like Paragraph, Heading, Un/OrderedList are available to Homepage? @dcshzj

Also added all the components that I understand to be available to Homepage layout, to Homepage story

Will do a separate one for ContentPage layout 💀


jacobkwan:
!run chromatic

src/templates/next/layouts/Homepage/Homepage.tsx
```diff
@@ -10,9 +11,13 @@ const HomepageLayout = ({
 }: HomePageSchema) => {
   return (
     <Skeleton site={site} page={page}>
-      {content.map((component) =>
-        renderComponent({ component, LinkComponent }),
-      )}
+      <div
+        className={`[&_.${ComponentContent}]:max-w-[1240px] [&_.${ComponentContent}]:px-6 [&_.${ComponentContent}]:md:px-10 [&_.${ComponentContent}]:mx-auto`}
```
dcshzj, line 15: question: might be trippin' but [on Chromatic](https://65dd949d255950353c94d5cc-tqzldlvqhy.chromatic.com/iframe.html?args=&id=next-layouts-homepage--default&viewMode=story) it doesn't seem like the max width is applied?

Screenshot 2024-03-28 at 08 22 24

src/templates/next/components/shared/CustomCssClass.tsx
```diff
@@ -0,0 +1 @@
+export const ComponentContent = "component-content"
```
dcshzj, line 1: nit: can rename this to a `.ts` file + don't need caps at the start since this is not a React component

src/templates/next/layouts/Homepage/Homepage.tsx
jacobkwan: 😭 works locally need to debug

src/templates/next/components/Infopic/Infopic.tsx
```diff
@@ -59,39 +60,47 @@ const SideBySideInfoPic = ({
...
-            buttonLabel={button}
-            buttonUrl={url}
-            className="w-2/3"
-          />
-          <ImageComponent src={src} alt={alt} className="w-1/3" />
+      </div>
+      {/* Desktop */}
+      <div className="hidden lg:block">
+        <div className={`${ComponentContent} flex flex-row gap-14 px-10 py-24`}>
+          {isTextOnRight ? (
```
dcshzj, line 79: nit: maybe can consider using `flex-row-reverse` here?

src/templates/next/components/Infopic/Infopic.tsx
```diff
@@ -132,19 +129,37 @@ const SidePartInfoPic = ({
...
-              buttonLabel={button}
-              buttonUrl={url}
-            />
-          </div>
-          <ImageComponent src={src} alt={alt} className="w-1/2" />
+      </div>
+      {/* Desktop */}
+      <div className="hidden lg:block">
+        <div className={`${ComponentContent} flex flex-row gap-10 px-10`}>
+          {isTextOnRight ? (
```
dcshzj, line 136: nit: same thoughts here

src/templates/next/layouts/Homepage/Homepage.tsx
```diff
@@ -10,9 +10,13 @@ const HomepageLayout = ({
 }: HomePageSchema) => {
   return (
     <Skeleton site={site} page={page}>
-      {content.map((component) =>
-        renderComponent({ component, LinkComponent }),
-      )}
+      <div
+        className={`[&_.component-content]:max-w-[1240px] [&_.component-content]:px-6 [&_.component-content]:md:px-10 [&_.component-content]:mx-auto`}
```
dcshzj, line 14: nit: can add a comment here about its relationship with the `ComponentContent` constant and the reason why we are not using it directly? Just in case we forget in the future.

dcshzj (approved):
Non-blocking comments

src/templates/next/layouts/Homepage/Homepage.tsx
jacobkwan: [319bacd](https://github.com/isomerpages/isomer-components/pull/52/commits/319bacd6a2cc9e1440fc098548305161343f8ddf)

src/templates/next/components/Infopic/Infopic.tsx
jacobkwan: great suggestion 👍 
[07f73eb](https://github.com/isomerpages/isomer-components/pull/52/commits/07f73eb430d3f38a59c452e491d9725cf5ed3fb0)

PR #55, opened 2024-04-01
feat: integrate collection layout components

dcshzj:
Integrate various components into the collection layout.

Todo:

- [x] Adjust mobile friendliness of page
- [x] Tie various states together with the original items list

### DIFF
--- a/src/templates/next/layouts/Collection/Collection.tsx
+++ b/src/templates/next/layouts/Collection/Collection.tsx
@@ -35,17 +35,62 @@ const getFilters = (categories: Record<string, number>): FilterType[] => {
   ]
 }
 
+const getFilteredResults = (
+  items: CollectionPageSchema["page"]["items"],
+  appliedFilters: AppliedFilter[],
+  search: string,
+  itemsPerPage: number,
+  currPage: number,
+): CollectionPageSchema["page"]["items"] => {
+  const filteredItems = items.filter((item) => {
+    if (
+      search !== "" &&
+      !item.title.toLowerCase().includes(search.toLowerCase()) &&
+      !item.description.toLowerCase().includes(search.toLowerCase())
+    ) {
+      return false
+    }
+
+    for (const filter of appliedFilters) {
+      if (filter.id === "category") {
+        if (
+          !filter.items.some(
+            (appliedFilterItem) =>
+              appliedFilterItem.id === item.category.toLowerCase(),
+          )
+        ) {
+          return false
+        }
+      }
+    }
+    return true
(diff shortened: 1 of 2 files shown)


dcshzj:
!run chromatic

src/templates/next/layouts/Collection/Collection.tsx
```diff
@@ -1,15 +1,97 @@
...
+    if (
+      search !== "" &&
+      !item.title.toLowerCase().includes(search.toLowerCase()) &&
+      !item.description.toLowerCase().includes(search.toLowerCase())
+    ) {
+      return false
+    }
+
+    for (const filter of appliedFilters) {
+      if (filter.id === "category") {
```
seaerchin, line 55: abit unclear why this magic string works - i know that this is via what we specify here but maybe we should make it a const in this file?

src/templates/next/layouts/Collection/Collection.tsx
```diff
@@ -1,15 +1,97 @@
...
+          !filter.items.some(
+            (appliedFilterItem) =>
+              appliedFilterItem.id === item.category.toLowerCase(),
+          )
+        ) {
+          return false
+        }
+      }
+    }
+    return true
```
seaerchin, line 66: nit: can return directly. also, it might be clearer if we flip the condition and do `filter.items.every`

src/templates/next/layouts/Collection/Collection.tsx
```diff
@@ -1,15 +1,97 @@
...
+  })
+
+  return filteredItems.slice(
+    (currPage - 1) * itemsPerPage,
```
seaerchin, line 70: just wondering if it's possible for `currPage` to be 0 here (i know we do 1 indexing, but maybe we accept user input as-is?)

src/templates/next/layouts/Collection/Collection.tsx
```diff
@@ -1,15 +1,97 @@
...
+          )
+        ) {
+          return false
+        }
+      }
+    }
+    return true
+  })
+
+  return filteredItems.slice(
```
jacobkwan, line 69: would prefer separating out the pagination logic out of this method, especially given its name `getFilteredResults`

think would be neater to separate into separate functions and chain them

items -> filtered -> sorted -> paginated items

can be a v2 but wondering if makes sense to `useMemo` on 
`items -> filtered -> sorted` to not have to re-render the entire component when the only thing changing is the `currPage` (which i expect to be quite a common user pattern)

PR #60, opened 2024-04-02
feat: add SearchSG input box

dcshzj:
Followed [the onboarding guide](https://docs.developer.tech.gov.sg/docs/searchsg-onboarding-guide/getting-started?id=step-21-load-the-searchconfigjs-script).

Also adjusted the local search input box to standardise the location of the search box between both types of search.

[Context on](https://redirect.github.com/storybookjs/storybook/issues/24429#issuecomment-1936395039) why I did the `useEffect` to load the SearchSG script inside the storybooks.


dcshzj:
!run chromatic

src/templates/next/layouts/Content/Content.stories.tsx
```diff
@@ -32,6 +32,10 @@ Default.args = {
       siteNavItems: [],
     },
     lastUpdated: "1 Jan 2021",
+    search: {
```
jacobkwan, line 35: confused why these are localSearch for Collection and Content layout instead of searchSG?
I thought it's a site wide thing that is used for the navbar on the entire site

src/templates/next/layouts/Homepage/Homepage.stories.tsx
```diff
@@ -32,6 +44,10 @@ Default.args = {
       siteNavItems: [],
     },
     lastUpdated: "1 Jan 2021",
+    search: {
+      type: "searchSG",
+      clientId: "7946e346-993e-41c7-bd81-26a3999dc3f4",
```
jacobkwan, line 49: should this be in env vars?

src/templates/next/types/LocalSearchInputBox.ts
```diff
@@ -0,0 +1,5 @@
+export interface LocalSearchInputBoxProps {
```
jacobkwan, line 1: confused why there's `NavbarLocalSearchProps` and there's this also (same for searchSG)
is it better to move the `type` here, get rid of the NavbarXXX types and then do the discriminated union on these two types instead?

Will have to move these types out of /next also

src/templates/next/components/shared/SearchSGInputBox/SearchSGInputBox.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+  parameters: {
+    themes: {
+      themeOverride: "Isomer Next",
+    },
+  },
+} as Meta
+
+// Template for stories
+const Template: StoryFn<Omit<SearchSGInputBoxProps, "clientId">> = (args) => {
+  const TEST_CLIENT_ID = "7946e346-993e-41c7-bd81-26a3999dc3f4"
```
jacobkwan, line 19: same comment about env var here

src/templates/next/layouts/Skeleton/Skeleton.tsx
```diff
@@ -35,13 +39,11 @@ export const Skeleton = ({
           type: "navbar",
           logoUrl: site.logoUrl,
           logoAlt: site.siteName,
-          search: {
-            type: "localSearch",
-            searchUrl: "/search",
-          },
+          search: site.search,
```
jacobkwan, line 42: don't have to be in this PR but I guess we would eventually want to move these Skeleton components to `/shared` and stop using `renderComponent` for them

jacobkwan:
also the stuff about CSP in their docs, we need to do anything?

src/templates/next/types/LocalSearchInputBox.ts
dcshzj: Yeah good point, adjusted in 6c387b237453581a1443fe6f589bcc337429a394!

src/templates/next/layouts/Content/Content.stories.tsx
dcshzj: This is inside the context of a story so I kept it to local search so I could avoid duplicating the `useEffect` logic + have different layouts to visualise how the local search and SearchSG looks like within a page.

src/templates/next/layouts/Homepage/Homepage.stories.tsx
dcshzj: Hmm this one is public info (took it from MFA's website), so I think can hardcode it here?

PR #66, opened 2024-04-09
feat(storybook): add a11y

kishore03109:
## Problem
Storybook comes with a11y support out of thet box. might be benefical to add this in doing development to ensure our components are up to standard.

also, changing the viewports addon as a dev dep rather than a direct dep.


jacobkwan (approved):
nice, thanks!

PR #67, opened 2024-04-09
feat(notification): add notification

kishore03109:
## Problem

add notification [here](https://linear.app/ogp/issue/ISOM-845/create-site-notification-component-in-isomer-next)

## Note to reviewers
Not too sure what our stance is on the colours (see that we are defining the colours at quite a number of places without tokens) is this intentional ah? else adding tokens should be a separate pr ya 

https://github.com/isomerpages/isomer-components/assets/42832651/e9314822-6afe-43bb-b06c-c3b1594abdee


kishore03109:
!run chromatic

src/templates/next/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import { Meta, StoryFn } from "@storybook/react"
+import Notification from "./Notification"
+import { NotificationProps } from "~/common/Notification"
+
+export default {
+  title: "Next/Components/Notification",
```
dcshzj, line 6: issue: I think the site notification should be an internal component, since the user would not use this component directly in the page.

suggestion: move the component into the `shared` folder and use the component directly within the various layouts in the next theme.

src/templates/next/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,17 @@
+import { NotificationProps } from "~/common/Notification"
+import BaseParagraph from "../shared/Paragraph"
+import { RxCross2 } from "react-icons/rx"
+
+const Notification = ({ title }: NotificationProps) => {
+  return (
+    <div className="flex flex-col sm:flex-row justify-between gap-6 bg-[#FEF4E8] py-4 px-5 sm:px-10">
```
dcshzj, line 7: nit: I think the `#fef4e8` here represents `site.primary.100`, so we can use that colour token instead (can refer to `next.ts` inside the `presets` folder for the colour tokens specific to the next theme).

src/templates/next/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <BaseParagraph className="text-base " content={title}></BaseParagraph>
+      <button className="flex flex-row ml-auto items-center gap-1">
+        <text className="text-lg">Dismiss</text>
+        <RxCross2 className="h-full" />
```
dcshzj, line 11: nit: we haven't standardise across the codebase yet sadly but we should only be using box icons, so this should be `BiX` instead.

src/templates/next/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,17 @@
+import { NotificationProps } from "~/common/Notification"
+import BaseParagraph from "../shared/Paragraph"
+import { RxCross2 } from "react-icons/rx"
+
+const Notification = ({ title }: NotificationProps) => {
+  return (
+    <div className="flex flex-col sm:flex-row justify-between gap-6 bg-[#FEF4E8] py-4 px-5 sm:px-10">
+      <BaseParagraph className="text-base " content={title}></BaseParagraph>
+      <button className="flex flex-row ml-auto items-center gap-1">
```
dcshzj, line 9: suggestion: the button should ideally work with the `useState` hook in React to hide the entire notification

src/templates/next/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,17 @@
...
+        <text className="text-lg">Dismiss</text>
```
dcshzj, line 10: nit: should be `<p>` tag instead? Also the colour should be `text-content`.

src/common/Notification.ts
```diff
@@ -0,0 +1,3 @@
+export interface NotificationProps {
+  title: string
```
dcshzj, line 2: nit: maybe can rename it as `content` instead?

jacobkwan:
> Not too sure what our stance is on the colours (see that we are defining the colours at quite a number of places without tokens) is this intentional ah? else adding tokens should be a separate pr ya

Most of our codebase should be using colour tokens already (maybe not /classic) - see `src/presets/next.ts`!
If design has added new tokens for your component then can just define the token here and use it
If you see any hardcoded colours in the codebase can help to replace it with the token / define the token if missing, thanks 😄

src/templates/next/components/shared/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import { Meta, StoryFn } from "@storybook/react"
+import Notification from "./Notification"
+import { NotificationProps } from "~/common/Notification"
+
+export default {
+  title: "Next/Components/Notification",
```
dcshzj, line 6:
```suggestion
  title: "Next/Internal Components/Notification",
```

src/templates/next/components/shared/Notification/Notification.tsx
```diff
@@ -0,0 +1,30 @@
+import { NotificationProps } from "~/common/Notification"
+
+import { BiX } from "react-icons/bi"
+import { useState } from "react"
+import BaseParagraph from "../Paragraph"
+
+const Notification = ({ content: title }: NotificationProps) => {
```
dcshzj, line 7: nit: I think we don't need to rename the prop here?

src/templates/next/components/shared/Notification/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./Notification"
```
dcshzj, line 1: question: this component isn't being used in any of the layouts, should it be added into the Skeleton layout?

src/templates/next/components/shared/Notification/Notification.tsx
```diff
@@ -0,0 +1,30 @@
+import { NotificationProps } from "~/common/Notification"
```
dcshzj, line 1: nit: can export the props in `~/common/index.ts` so the import here is simplified

kishore03109: [88000ab](https://github.com/isomerpages/isomer-components/pull/67/commits/88000ab71881e74a26f77b82c12b38498b6edca1)

src/templates/next/components/shared/Notification/index.ts
kishore03109: [88000ab](https://github.com/isomerpages/isomer-components/pull/67/commits/88000ab71881e74a26f77b82c12b38498b6edca1)

src/templates/next/components/shared/Notification/Notification.tsx
kishore03109: [88000ab](https://github.com/isomerpages/isomer-components/pull/67/commits/88000ab71881e74a26f77b82c12b38498b6edca1)

src/templates/next/components/shared/Notification/Notification.stories.tsx
kishore03109: [88000ab](https://github.com/isomerpages/isomer-components/pull/67/commits/88000ab71881e74a26f77b82c12b38498b6edca1)

dcshzj (approved):

PR #68, opened 2024-04-09
feat: introduce article layout for next theme

dcshzj:
Figma: https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=1903-8895&mode=design&t=MljLE7P0W7is1gj7-4

Not done yet, will do in separate PR:
- Related articles column (still need to understand what exactly is related)
- Footnotes (not sure what this entails either)
- About the author (Figma says this is not needed but might change)


dcshzj:
!run chromatic

src/common/ArticlePageHeader.ts
```diff
@@ -0,0 +1,12 @@
+import type BreadcrumbProps from "./Breadcrumb"
+
+export interface ArticlePageHeaderProps {
+  breadcrumb: BreadcrumbProps
+  category: string
+  title: string
+  date: string
+  summary: string[]
+  LinkComponent?: any
```
kishore03109, line 9: hmm why any? :( 
minimally should be a jsx.element/string?

src/templates/next/components/shared/ArticlePageHeader/ArticlePageHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <div className="px-4 lg:px-5">
+      <div className="max-w-[960px] mx-auto">
+        <div className="my-8 lg:my-16">
+          <Breadcrumb links={breadcrumb.links} LinkComponent={LinkComponent} />
+        </div>
+
+        <div className="mb-3 text-lg text-content-medium">{category}</div>
+
+        <div className="flex flex-col gap-9 max-w-[660px]">
```
kishore03109, line 23: whats our stance here ah? should we not be defining these `max-w-[660px]` in presets?

src/utils/getBreadcrumbFromSiteMap.ts
```diff
@@ -0,0 +1,33 @@
...
+  let currentPath = ""
+
+  for (const pathSegment of permalink) {
+    currentPath += "/" + pathSegment
+    const nextNode = node.children?.find(
+      (node) => node.permalink === currentPath,
+    )
+
+    if (!nextNode) {
+      // TODO: handle this unexpected case where cannot traverse to permalink in the sitemap
```
kishore03109, line 21: hmm why not just throw / log for now so no silent failures ah?

kishore03109 (commented):
hmm the related columns doesnt shift down in smaller screens tho. understand the content inside the related content is not done, but the shift down behaviour is confirmed already right?

Screenshot 2024-04-09 at 3 40 02 PM

dcshzj:
> hmm the related columns doesnt shift down in smaller screens tho. understand the content inside the related content is not done, but the shift down behaviour is confirmed already right?

Right yes, forgot about that part, adjusted in b79d429a2b14dac038131ad07b55bf8d8636a1f0.

src/common/ArticlePageHeader.ts
dcshzj: Haven't got the time to figure out the appropriate typing for this, so kept it as any for now.

In practice this `LinkComponent` is basically the `Link` from the `next/link` package, but because this library is framework-agnostic, we don't want to import/include this dependency. For the typing to be strict here, we must make sure that it is compatible with at least what `Link` is without including the actual Next.js dependency.

src/templates/next/components/shared/ArticlePageHeader/ArticlePageHeader.tsx
dcshzj: For the layouts-level stuff I don't think we have fully finalised the list of breakpoints/container widths yet, so this is hard-coded for now pending confirmation from design.

src/utils/getBreadcrumbFromSiteMap.ts
dcshzj: Both the permalink and the sitemap are autogenerated from the same data source, which is the folder structure of the site repository stored on the file system, so this case shouldn't happen. Opted to deprioritise this until we minimally achieve feature parity.

src/common/ArticlePageHeader.ts
kishore03109: hmm i dont understand this. if at the end of the day, the render engine expects certain props (that the next package defines for the component link) and renders the next/link as a component form the next package, arent we already not dependent? 
we could always just use the same shape tho right? even if another framework is used, as long as the shape can be converted it is ok?

dcshzj: Hmm no, we are not dependent on the next package, because it is valid for a user to provide the `a` string as the LinkComponent and the component will still render correctly. If we add typings here, we run the risk that the site cannot build when the `next/link` package is actually used. We will need to spend time investigating and experimenting to know what is the appropriate type to use here.

src/utils/getBreadcrumbFromSiteMap.ts
dcshzj: We currently don't have a logging mechanism setup yet, so we might not be able to do this right now. If we do a `console.log` here it will print out to the browser's console which would be seen by site visitors.

kishore03109 (approved):

src/common/ArticlePageHeader.ts
kishore03109: discussed offline, we are doing this so that we remain type agnostic

PR #72, opened 2024-04-19
feat: introduce divider component for next theme

dcshzj:
Figma: https://www.figma.com/file/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?type=design&node-id=2213-18339&mode=design&t=dug8hwIYBU3DtzOn-4

No examples currently on how it is potentially going to be used, but keeping the margins to be the same as paragraph first.


src/templates/next/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,7 @@
+import type { DividerProps } from "~/common"
+
+const Divider = ({}: DividerProps) => {
+  return <hr className="bg-divider-medium my-6" />
```
harishv7, line 4: do we need the `w-full` or similar width control to take the full width of the parent container?

dcshzj: Hmm, the `<hr />` tag [automatically has](https://stackoverflow.com/a/21921785) a `width: 100%` set, so we don't need to specify further here.

dcshzj:
Merging first since no blocking comments

PR #73, opened 2024-04-23
feat: introduce iframe component for next theme

dcshzj:
Support having iframe embeds in Isomer Next.

Notes:
- The `title` attribute is required for accessibility, so it is a required prop of the component
- The padding-top hack is a setup suggested here: https://www.w3schools.com/howto/howto_css_responsive_iframes.asp
- Instagram embeds are NOT supported yet, will be creating a separate component instead (I might be doing that for FormSG as well)


src/utils/getSanitizedIframeEmbed.ts
```diff
@@ -0,0 +1,38 @@
+import DOMPurify from "isomorphic-dompurify"
+
+// Sanitize iframe embeds to remove any potentially harmful attributes
+// and to insert the iframe title for accessibility
+export const getSanitizedIframeWithTitle = (content: string, title: string) => {
+  DOMPurify.addHook("beforeSanitizeElements", (curr) => {
+    // Add title attribute to iframe elements for accessibility
```
kishore03109, line 7: nit: misplaced comment

kishore03109 (approved):

PR #74, opened 2024-04-25
fix: design fixes for keystatistics component

dcshzj:
Closes ISOM-966.

Design to confirm:
- [ ] How to handle large numbers as one of the key statistics item

### DIFF
--- a/src/templates/next/components/complex/KeyStatistics/KeyStatistics.stories.tsx
+++ b/src/templates/next/components/complex/KeyStatistics/KeyStatistics.stories.tsx
@@ -32,6 +32,24 @@ Side.args = {
   ],
 }
 
+export const SideLong = Template.bind({})
+SideLong.args = {
+  variant: "side",
+  title: "Key economic indicators",
+  statistics: [
+    {
+      label: "Advance GDP Estimates, 4Q 2023 (YoY)",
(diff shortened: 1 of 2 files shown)


dcshzj:
!run chromatic

src/templates/next/components/KeyStatistics/KeyStatistics.tsx
```diff
@@ -1,7 +1,26 @@
...
+  Record<number, string>
+> = {
+  side: {
+    1: "basis-full",
+    2: "basis-[calc((100%-2.5rem)/2)]",
+    3: "basis-[calc((100%-5rem)/3)]",
+  },
+  top: {
+    1: "basis-full",
+    2: "basis-[calc((100%-2.5rem)/2)]",
```
harishv7, line 15: do we need to do these calc ourselves or can consider using presets?

harishv7 (approved):
lgtm, minor comment

src/templates/next/components/KeyStatistics/KeyStatistics.tsx
dcshzj: So far this is something that is quite specific for key statistics component, so opted to not use presets for this.

PR #81, opened 2024-04-29
feat: add tiptap compatibility to components

dcshzj:
> [!IMPORTANT]
> This PR is **NOT** intended to be merged, it is just a proof of concept to help us make a decision on recursive vs 1D structure.

Refactor the components to make them Tiptap compatible.

### Things that are not available yet

- Many of the components (commented out inside `src/tiptap/components.ts`), was not done originally as this is just a POC
- Table component - need to write our own node view to bypass the ReactNodeViewRenderer provided by Tiptap, this is a similar implementation on how they do it inside `@tiptap/extension-table`
    - Reason: The ReactNodeViewRenderer and NodeViewContent adds additional `<div>` tags on top of the tags that were originally defined as part of the table cells/rows, which breaks HTML rendering because now `<tr>` is not a direct child of `<tbody>`.
- Converting JSON schema into the Prosemirror schema (currently defined in `src/tiptap/components.ts`)
    - In theory it should be possible to just convert so just need to create the parser to churn out the Prosemirror schema
- Converting Tiptap content schema to the actual page schema
    - In theory this should be quite easy to do now that the props now more closely resemble the content schema provided by Tiptap

### How to review?

This PR is a proof-of-concept on making the components compatible with Tiptap, so that we can have page editing that looks something like this:
Screenshot 2024-04-29 at 11 15 10

Clicking outside of the editor and the user can immediately preview the page:
Screenshot 2024-04-29 at 11 15 26

So, a couple of pointers/feedback that would be ideal:

1. Introduction of NodeViewWrapper, NodeViewContent and ContentHole
    1. The NodeViewWrapper is a Tiptap concept where we "take control of the rendering" of the component within the Tiptap editor, basically telling Tiptap that the rendering of any child elements inside the NodeViewWrapper will be done by us.
    2. The NodeViewContent is also another Tiptap concept that "returns back the control of the rendering" back to Tiptap. This is to indicate to Tiptap that this is the part that we should allow the site editor to be able to modify content within the component.
    3. The ContentHole is the same as the NodeViewContent but adds support for when the Tiptap editor is not in play (i.e. we are rendering for the actual site rather than rendering in the context of the Tiptap editor).
    4. The annoying thing is that the NodeViewWrapper (along with ReactNodeViewWrapper) adds additional `<div>` tags, which can break some of the CSS that we have been doing (e.g. add margin if not first child, etc). A workaround is done inside `src/templates/next/tiptap/index.tsx` which sets the top-level classes to the wrapper to achieve the same outcome.
    5. **For feedback**: Is this approach sane? Alternative would be to rely on Tiptap entirely for the rendering/editing, which means we won't be able to have individual React components.
2. Mixing recursive and 1D tree structure
    1. Previously, we operated on a flat 1D structure where the `content` key will specify the components to render for the page, then the individual components will decide what other components it chooses to nest (e.g. ordered list).
    2. Having a recursive tree structure means that we rely entirely on Tiptap for schema validations, which made it difficult for us to enforce certain constraints (e.g. limits on number of characters).
    3. Mixing the two allows us to cater for certain blocks like ordered lists (which requires a recursive structure) while maintaining complex blocks with a fixed structure (e.g. cards block).
    4. For blocks using the recursive tree structure, the Tiptap editor will perform all the actions required (e.g. sinking list items inside lists).
    5. For blocks using the 1D structure, the user will not be able to edit within the Tiptap editor, and a separate drawer will appear on the right side of the CMS, which will open up a mini form to change the attributes of the block (see the screenshot below).
    6. **For feedback**: Is this approach sane? Or should we go 100% into a certain structure (1D vs recursive) and base our decisions from there?

image


dcshzj:
Points I got from discussions:
- The added coupling between Tiptap and the components may not be ideal, because it ties the longevity of the components library with that of Tiptap. In the event that Tiptap goes into maintenance mode, we will have to rebuild the components library quite significantly to remove this dependency. At that point, the cost of migrating will be even higher.
- In terms of editing experience on the CMS, the form-based editor should not be the first point of contact by the user. Instead, the user should see their rendered preview and click on parts of the page that they want to modify, which we will then open the form for that specific component.
    - If a field in that specific component contains rich text, then we can load the Tiptap editor for that field, which we will then take their JSON output and store as part of the page's JSON.
    - For other fields, we can render the form field depending on the type of data (potentially introducing a custom JSON Schema field to hold this information). So a string type can have a UI that is a date picker.
    - The role of the CMS is to take in the schema and render the UI exactly based on how the schema defines it.
- Rendering of components will mostly be the same, except for props that are rich text, then we will render them recursively. Otherwise, we will still stick to the 1D approach of rendering. This agrees with the point that we should mix recursive and 1D structure for the content tree.

To consider:
- If we want to make the correct form appear when the user clicks on the component within the preview, we might have to "leak" some of the CMS-specific stuff into the component so that the CMS knows exactly which component is being selected, to know which form to render.
- CMS editing experience will be entirely created by us, using the schema that we define explicitly.

kishore03109:
Wah thank you for this long write up

> For feedback: Is this approach sane? Alternative would be to rely on Tiptap entirely for the rendering/editing, which means we won't be able to have individual React components.

src/types/tiptap.ts
```diff
@@ -0,0 +1,15 @@
+import type { IsomerThemes } from "./site"
+
+export interface IsomerTiptapNode {
```
kishore03109, line 3: nit: why not just extend from what tiptap already has tho?

src/templates/next/components/UnorderedList/UnorderedList.tsx
```diff
@@ -1,28 +1,15 @@
...
+import ContentHole from "../../tiptap/ContentHole"
 
-const UnorderedList = ({ items }: UnorderedListProps) => {
+const UnorderedList = ({
+  content,
+  NodeViewWrapper = "ul",
+  NodeViewContent,
+}: UnorderedListProps) => {
   return (
-    <ul className="list-disc ps-8 mt-6">
```
kishore03109, line 8: hmm ok just so my understanding is correct here. If I want to style the elements inside the content hole, I have to create a new element called `li` right? (or simply extend it from tiptap's implementation and add the classes that I require?)

src/templates/next/components/Heading/Heading.tsx
```diff
@@ -1,54 +1,66 @@
...
+        <ContentHole content={content} NodeViewContent={NodeViewContent} />
+      </NodeViewWrapper>
     )
   }
   return (
-    <h6
+    <NodeViewWrapper
+      as="h6"
       id={id}
       className={`${HeadingStyles[6]} text-content [&:not(:first-child)]:mt-7`}
```
kishore03109, line 60: leaving this classes here, i remember in the pr you mentioned that tiptap added new `div`s over, does this not mean that this class names here become "wrong" since the structure of the dom is now different?

kishore03109 (commented):
Wah thank you for this long write up

> For feedback: Is this approach sane? Alternative would be to rely on Tiptap entirely for the rendering/editing, which means we won't be able to have individual React components.

Could you expand on this? I assume that we would still be able to pass in the classes that we require, but for every sub-child that we want to customise, we need to create a separate child for this right? It seems to me that using Tiptap as our render engine might be the fastest to get the prototype out, but iterating upon it would be a hassle correct? 


> If we want to make the correct form appear when the user clicks on the component within the preview, we might have to "leak" some of the CMS-specific stuff into the component so that the CMS knows exactly which component is being selected, to know which form to render.

I don’t see this as a blocker. While it nice to keep the separation of concerns, this particular mental model seems to me like the best experience to the users, and as such I don’t think there is anything wrong in leaking this over. By leaking, may I clarify that we are matching the structure of our complex blocks into the content “holes” that you were talking about right? 

> The added coupling between Tiptap and the components may not be ideal, because it ties the longevity of the components library with that of Tiptap. In the event that Tiptap goes into maintenance mode, we will have to rebuild the components library quite significantly to remove this dependency. At that point, the cost of migrating will be even higher.

Hmm this is not a strong argument imo. Since this is open sourced (MIT Licence), we can always fork our our maintained version in the future and freeze any development.

> For other fields, we can render the form field depending on the type of data (potentially introducing a custom JSON Schema field to hold this information). So a string type can have a UI that is a date picker.

Hmm can’t we encode this directly onto the json schema? Ie declare a type ‘Date’ rather than ‘string’? 

> In terms of editing experience on the CMS, the form-based editor should not be the first point of contact by the user. Instead, the user should see their rendered preview and click on parts of the page that they want to modify, which we will then open the form for that specific component.

This seems ideal, but are we gunning for this for launch ah? 

> Mixing recursive and 1D tree structure

Hmm to anchor ourselves, we know setting constraints on our users is a must. To me, that has direct business impact to the end citizen viewing the site. We also know that we want to enable the use of Tiptap for the quickest dev work when it comes to the CMS side of things. I don’t think we should tie ourselves to one approach then make decisions from there. It seems to me that based on this system level requirements, mixing the two becomes the obvious choice, but we need to have something working out for release first (Ie, just the form based approach). We can always improve the sentiments in the future and add drawers in the future, when we mix the two mental models (recursive vs 1D) together right?   

> The annoying thing is that the NodeViewWrapper (along with ReactNodeViewWrapper) adds additional <div> tags, which can break some of the CSS that we have been doing (e.g. add margin if not first child, etc). A workaround is done inside src/templates/next/tiptap/index.tsx which sets the top-level classes to the wrapper to achieve the same outcome.

Hmm we could also just add <div> tags directly onto our components (ie mock the structure that Tiptap has) and apply the styles directly over there correct? Additionally, we could jump into the source code and remove this from tiptop directly too!

> Converting JSON schema into the Prosemirror schema (currently defined in src/tiptap/components.ts) In theory it should be possible to just convert so just need to create the parser to churn out the Prosemirror schema

Hmm I don’t get this part to be honest, and this sounds like non-trivial work. Why should we go through the trouble of creating a parser here ah? If the AST is already well defined, and we can just apply the rules that we want when walking the tree as shown in `components.ts`, what is the value that creating another parser bring to us ah?

PR #86, opened 2024-05-03
fix: design fixes for table component

dcshzj:
Split from #74, only for table component.

Fixes ISOM-964.

### DIFF
--- a/src/interfaces/native/Text.ts
+++ b/src/interfaces/native/Text.ts
@@ -43,6 +43,6 @@ export type Marks =
 
 export interface TextProps {
   type: "text"
-  marks: Marks[]
+  marks?: Marks[]
   text: string
 }
--- a/src/templates/next/components/internal/Notification/Notification.tsx
+++ b/src/templates/next/components/internal/Notification/Notification.tsx
@@ -14,7 +14,10 @@ const NotificationBanner = ({ content }: NotificationProps) => {
   return (
     isShown && (
       <div className="flex flex-col sm:flex-row justify-between gap-6 bg-site-primary-100 py-4 px-5 sm:px-10">
-        <BaseParagraph className="text-base" content={content}></BaseParagraph>
+        <BaseParagraph
+          className="text-base text-content"
+          content={content}
+        ></BaseParagraph>
         <button
           className="flex flex-row ml-auto items-center gap-1"
           aria-label="Dismiss notification"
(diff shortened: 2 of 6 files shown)


dcshzj:
!run chromatic

src/templates/next/components/internal/Notification/Notification.tsx
```diff
@@ -14,7 +14,10 @@ const NotificationBanner = ({ content }: NotificationProps) => {
   return (
     isShown && (
       <div className="flex flex-col sm:flex-row justify-between gap-6 bg-site-primary-100 py-4 px-5 sm:px-10">
-        <BaseParagraph className="text-base" content={content}></BaseParagraph>
+        <BaseParagraph
+          className="text-base text-content"
+          content={content}
+        ></BaseParagraph>
```
alexanderleegs, line 20: Assuming that this was intended to go in for a different issue!

src/templates/next/components/native/Table/Table.tsx
```diff
@@ -6,16 +6,39 @@ import BaseParagraph from "../../internal/BaseParagraph"
...
+      if (!stickyRowIndexes.includes(index)) {
+        return
+      }
+
+      if (cell.rowSpan) {
+        stickyRowIndexes = stickyRowIndexes.filter(
+          (value) => value <= index || value >= index + cell.rowSpan!,
+        )
+      }
+    })
```
alexanderleegs, line 33: Could you add a comment explaining what this section is doing? It's a little complicated haha

src/interfaces/native/Text.ts
```diff
@@ -43,6 +43,6 @@ export type Marks =
 
 export interface TextProps {
   type: "text"
-  marks: Marks[]
+  marks?: Marks[]
   text: string
```
alexanderleegs, line 47: Assuming this is another separate fix that's supposed to go in!

alexanderleegs (commented):
Left some minor comments!

src/interfaces/native/Text.ts
dcshzj: Oops yup, the marks was marked as required by accident but a normal text node may be a plain one with no marks being applied.

src/templates/next/components/internal/Notification/Notification.tsx
dcshzj: Oh yeah this one was a separate fix for the paragraph one, missed this one when splitting up the original PR.

src/templates/next/components/native/Table/Table.tsx
dcshzj: Added in 39b45ee6bd712996fa47a82207a8789dba0ce2b2!

alexanderleegs (approved):
lgtm

PR #75, opened 2024-04-26
fix: design for navbar and masthead

dcshzj:
Closes ISOM-960.
Closes ISOM-961.

PR #87, opened 2024-05-03
fix: design fixes for masthead component

dcshzj:
Fixes ISOM-960.

Split from #75.


dcshzj:
!run chromatic

dcshzj:
Realised I can't split cos both are interlinked somewhat with the container spacing definition.

PR #96, opened 2024-05-10
Fix/remove indentation

alexanderleegs:
This PR removes indentation for components which were originally built for homepage when used on content pages. As the Homepage layout has styles for the horizontal padding for `ComponentContent` class, the individual component `px-` values are being overridden for homepage, and are unneeded for content pages - as such, they have been removed.

These components have also been added to the storybook for the content layout.

### DIFF
--- a/src/templates/next/components/complex/InfoCards/InfoCards.tsx
+++ b/src/templates/next/components/complex/InfoCards/InfoCards.tsx
@@ -27,7 +27,7 @@ const InfoCards = ({ cards, title, subtitle, variant }: InfoCardsProps) => {
     <section>
       {variant === "side" ? (
         <div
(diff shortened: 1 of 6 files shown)


alexanderleegs:
!run chromatic

kishore03109:
something about this pr feels off, should we get @dcshzj's eye on this? we are removing quite a number of breakpoints, which seemed to have been cleared by design, i wonder if this has side effects on other layout formats

dcshzj (approved):
I think this is okay mainly because the individual components shouldn't be the ones in charge of the padding within the layout (that should be handled by the layout itself through the ComponentContent class).

The removal of the paddings makes sense here since it is now handled by the layouts.

PR #100, opened 2024-05-20
fix(footer): styling fixes

sehyunidaaa:
## Problem

Styling fixes for footer

Closes [ISOM-1058](https://linear.app/ogp/issue/ISOM-1058/footer-design-review)

## Solution

1. Changed site name text to heading-4
2. Changed all text to caption (smaller font size)
3. Changed reach us text style to subhead-1 (new font token)
4. Added fixed width to links (for lg screen only)
5. Added line clamp to long links extending 1 line
6. Adjusted some row/col gaps to fit better with smaller font size

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [x ] No - this PR is backwards compatible

**BEFORE**:
Footer content and gaps were very big; footer was vertically taking up a lot of space
image

**AFTER**:
Very large screen:
Screenshot 2024-05-20 at 11 27 49 PM

Around 1024px breakpoint:
Screenshot 2024-05-20 at 11 27 57 PM

Long links:
Screenshot 2024-05-20 at 11 28 34 PM

Mobile/tablet: 
image

PR #101, opened 2024-05-21
chore: switch to use typography tokens

dcshzj:
Chore to switch all existing typography tokens to native CSS classes.

Will rebase after #100 goes in, so our designer don't need to worry about merge conflicts.

Notes:
- Only switched interpolation way of injecting typography tokens to native CSS classes (i.e. `${Heading[1]}` becomes `text-heading-01`).
- ButtonLink was renamed to use `text-button-link-01` to prevent possible conflicts in the future if we ever have a button-link colour token.
- Will separately discuss with design to remove this screen-based typography, so that we can fully onboard the [Tailwind typography plugin](https://github.com/tailwindlabs/tailwindcss-typography). Seems like our way of doing typography tokens is a little weird because it is dependent on the screen size.


dcshzj:
!run chromatic

seaerchin:
Heya, I'd love to review this but I don't really have a good place to start wrt how tailwind handles this! Are there some docs you could point me to real quick?

src/index.css
```diff
@@ -13,3 +13,114 @@
     font-family: "Lato", system-ui, sans-serif;
   }
 }
+
+@layer utilities {
+  /* Heading typography tokens */
+  .text-heading-01 {
+    @apply text-[2.75rem] font-semibold leading-[3.25rem] -tracking-[0.022em];
```
kishore03109, line 20: can use `tracking-tight` now

kishore03109 (approved):
hihi can run the eslint fixes in a seperate pr first please? the change set is a lot harder to verify when there are additional tailwind eslint rules being applied :(  

may i get context to what is stopping us from moving over to tailwind's typography ah? we can still use container queries right? eg. 
```
<div class="@container">
<article class="prose @md:prose-lg @lg:prose-xl">{{ markdown }}</article>
</div>
```

dcshzj:
> hihi can run the eslint fixes in a seperate pr first please? the change set is a lot harder to verify when there are additional tailwind eslint rules being applied :(

Yeah it's a little annoying, the prettier plugin insists on being used here but sure, will try to find some way to disable it next time.

> may i get context to what is stopping us from moving over to tailwind's typography ah? we can still use container queries right? eg.
> 
> ```
> <div class="@container">
> <article class="prose @md:prose-lg @lg:prose-xl">{{ markdown }}</article>
> </div>
> ```

Addressed in the PR description! Mainly it's because of our weird way of doing typography tokens and it being dependent on the screen size. In tailwind's typography plugin, it assumes that prose-paragraph-1 is always of a certain font size and line height, but in our case it is dependent on the screen size (which is non-standard).

dcshzj:
> Heya, I'd love to review this but I don't really have a good place to start wrt how tailwind handles this! Are there some docs you could point me to real quick?

Oops ps missed this, it should be this one: https://tailwindcss.com/docs/functions-and-directives#layer

dcshzj:
Anyway thanks @kishore03109 and @seaerchin for reviews/attempts to review. This was just a chore work so didn't expect to get reviews, plus my stance is to try to move to tailwind's typography plugin in the future (but that requires some design rework so it will be tackled later).

PR #103, opened 2024-05-21
fix(infocards): styling and a11y fix for infocards

sehyunidaaa:
## Problem

Styling fixes and one a11y fix for infocards

## Solution

1. Removed gap-12, changed to pb-12 (pb-8 for <sm) below title area instead (gap-12 was causing an extra left padding when there is no title)
2. Removed horizontal card variant from the Default info cards layout. Instead, at md-lg range, cards collapse to a 2-col layout. This is after trying out designs at different viewports; turns out vertical variant still fits.
3. Added border-radius of 6px
5. Hover state opacity 80% 
6. <a> is inside the card content instead of the whole card. Whole card is still clickable (ref: [Cards](https://inclusive-components.design/cards/))
7. Fixed image height to h-52
8. Scaled down padding in the text section
9. Added line-clamp for title and paragraph.

## Before & After Screenshots

**BEFORE**:

image

image

**AFTER**:

at w=800px, cards collapse to 2 columns
image

when body text is too long:
image

### DIFF
--- a/packages/components/src/templates/next/components/complex/InfoCards/InfoCards.stories.tsx
+++ b/packages/components/src/templates/next/components/complex/InfoCards/InfoCards.stories.tsx
@@ -26,7 +26,8 @@ Default.args = {
   variant: "top",
   cards: [
     {
-      title: "Testing for a card with a long line length that spans across two lines or more",
+      title:
+        "Testing for a card with a long line length that spans across two lines or more",
       description:
         "Explore Duxton with us and leave with a full belly, tipsy mind, and a happy smile.",
       imageUrl: "https://placehold.co/200x300",
@@ -37,7 +38,7 @@ Default.args = {
     {
       title: "Card with short title",
       description:
-      "In the labyrinthine expanse of Zandoria, an enigmatic government wields authority through a web of intricate bureaucracy and omnipresent surveillance, shaping the lives of its denizens with meticulous precision and unyielding control.",
+        "In the labyrinthine expanse of Zandoria, an enigmatic government wields authority through a web of intricate bureaucracy and omnipresent surveillance, shaping the lives of its denizens with meticulous precision and unyielding control.",
       imageUrl: "https://placehold.co/200x300",
       imageAlt: "alt text",
       buttonLabel: "Explore with us",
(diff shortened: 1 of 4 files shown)


src/templates/next/components/internal/Card.tsx
```diff
@@ -43,26 +43,35 @@ const TextComponent = ({
...
             {text}
           </p>
         )}
       </div>
       {buttonLabel && (
         <div className={`flex items-center gap-1`}>
-          <a href={url}
-      target={url.startsWith("http") ? "_blank" : undefined}
-      rel={url.startsWith("http") ? "noopener noreferrer nofollow" : undefined} className={`text-interaction-link after:absolute after:inset-0 ${ButtonLink[1]}`}>
+          <a
```
sehyunidaaa, line 62: @isomerpages/iso-engineers fyi this is something i found here > https://inclusive-components.design/cards/. not sure if it makes sense or is it worth doing!

packages/components/src/templates/next/components/complex/InfoCards/InfoCards.stories.tsx
```diff
@@ -35,16 +36,18 @@ Default.args = {
       url: "https://www.google.com",
     },
     {
-      title: "A yummy, tipsy evening at Duxton",
+      title: "Card with short title",
       description:
-        "Explore Duxton with us and leave with a full belly, tipsy mind, and a happy smile. Explore Duxton with us and leave with a full belly, tipsy mind, and a happy smile. Explore Duxton with us and leave with a full belly, tipsy mind, and a happy smile.",
+        "In the labyrinthine expanse of Zandoria, an enigmatic government wields authority through a web of intricate bureaucracy and omnipresent surveillance, shaping the lives of its denizens with meticulous precision and unyielding control.",
```
seaerchin, line 41: this hits too close to home

packages/components/src/templates/next/components/complex/InfoCards/InfoCards.stories.tsx
```diff
@@ -35,16 +36,18 @@ Default.args = {
...
       imageUrl: "https://placehold.co/200x300",
       imageAlt: "alt text",
       buttonLabel: "Explore with us",
       url: "/",
     },
     {
-      title: "A yummy, tipsy evening at Duxton",
+      title: "Committee of Supply (COS) 2024",
+      description:
+        "In the kingdom of Veridonia, the government operates as a benevolent monarchy, guided by ancient traditions and the wisdom of its sovereign. Its policies prioritize the welfare of its subjects, fostering prosperity and unity throughout the realm.",
```
seaerchin, line 50: factually correct description of singapore

seaerchin (approved):

src/templates/next/components/internal/Card.tsx
seaerchin: actually that's a great article! i think it'd be worth doing _in the long run_ - not sure now is the right time given time constraints :monkas:

PR #117, opened 2024-05-28
fix: design fixes for navbar component

dcshzj:
Design fixes for Navbar component. Only desktop view is affected.

Note:
- The text colours are hard-coded for now, as no design tokens have been allocated yet. It is still subject to future change.

Figma: https://www.figma.com/design/8TEzSSDbx97gctRJ7nKdqG/Website-Template-Gen-2?node-id=4488-9936&t=RQmuT8l6myyX12S7-4

### DIFF
--- a/packages/components/src/templates/next/components/internal/Navbar/Navbar.tsx
+++ b/packages/components/src/templates/next/components/internal/Navbar/Navbar.tsx
@@ -88,8 +88,8 @@ export const Navbar = ({
             className="mt-2 flex flex-row flex-wrap gap-1"
             ref={navDesktopRef}
           >
-            {items.map(({ name, url, description, items }, idx) => {
-              if (!items || items.length === 0) {
+            {items.map(({ name, url, description, items: subItems }, idx) => {
+              if (!subItems || subItems.length === 0) {
                 return (
                   <li key={`${name}-${idx}`}>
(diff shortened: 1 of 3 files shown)


packages/components/src/templates/next/components/internal/Navbar/Navbar.tsx
```diff
@@ -67,21 +67,122 @@ export const Navbar = ({
...
+                  <li key={`${name}-${idx}`}>
+                    <LinkComponent
+                      className="block px-2 py-1 text-base/5 font-medium text-[#1f2937]"
+                      href={url}
+                    >
+                      {name}
+                    </LinkComponent>
+                  </li>
+                )
+              }
```
seaerchin, line 103: shouldn't this be done outside of the `map`? if `items === undefined`, we have an error here ._.

seaerchin (approved):
Minor issue with `map`, rest lgtm

packages/components/src/templates/next/components/internal/Navbar/Navbar.tsx
dcshzj: Lol shucks variable scoping issue, renamed in 9b3752599270489aa3c949a809e2d2156d54cca3 to be clear what the `items` here mean.

PR #118, opened 2024-05-28
feat(dto): add initial dto for create/edit page

seaerchin:
## Problem
This PR adds the base DTOs in for read/edit page flows. Importantly, this PR makes two key assumptions: 
1. blobs are **separate** from the page metadata. this is due to our db structure separating the two.
2. we model page separately from folders. this **is not true** at present, will have to change at a later date. this PR was created prior to alignment on db structure for pages/folders, which led to this distinction.

Closes ISOM-1085, ISOM-1084, ISOM-1082

## Solution
add in DTOs for frontend -> backend and backend -> frontend. 

1. FE -> BE done via `z.schema`. no validation added for now, we should do that later. 
2. BE -> FE done via the return on `page.router`.


packages/studio/src/server/modules/page/page.select.ts
seaerchin, line 1: can delete this - copied from `starter-kit`

package.json
```diff
@@ -8,7 +8,8 @@
     "clean": "turbo clean",
```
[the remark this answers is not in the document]
harishv7: also for branch names can we copy from linear ticket format?
Screenshot 2024-05-28 at 11 20 34 PM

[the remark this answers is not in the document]
seaerchin: Eh i forgot to mark as draft ._.

packages/studio/src/schemas/page.ts
```diff
@@ -0,0 +1,33 @@
...
+})
+
+export const createPageSchema = z.object({
+  pageName: z.string(),
+  pageTitle: z.string(),
+  // TODO: add the actual layouts in here
+  layout: z.enum(["content"]).default("content"),
+  siteId: z.string(),
+  // NOTE: implies that top level pages are allowed
+  folderId: z.string().nullable()
```
kishore03109, line 32: hmm so we are locking down to two states at the top level ya, one is top level page another is top level folder ya?

packages/studio/src/server/modules/page/page.select.ts
kishore03109: hmm then shall we just delete this ya

packages/studio/src/server/modules/page/page.router.ts
```diff
@@ -0,0 +1,58 @@
...
+    // NOTE: Not returning the `content` first because 
+    // 1. it might potentially be huge
+    // 2. frontend already knows 
+    // ahead of time what is the content (it sent the content over)
+    return {}
+  }),
+  createPage: protectedProcedure.input(createPageSchema).query(async ({ input, ctx }) => {
+    return { pageId: "" }
+  })
+  // TODO: Delete page stuff here
```
kishore03109, line 57: nit: maybe can just add a quick mock mutation here then bah

packages/studio/src/schemas/page.ts
```diff
@@ -0,0 +1,33 @@
...
+export const getEditPageSchema = z.object({
+  pageId: z.string(),
+  siteId: z.string()
+})
+
+export const updatePageSchema = z.object({
+  // NOTE: We allow both to be empty now, 
+  // in which case this is a no-op. 
+  // We are ok w/ this because it doesn't 
+  // incur any db writes
```
kishore03109, line 12: wait i am confused here. isnt every call to the upstream api call `updatePage` updating the db? in what situation will these two be nullable ah?

seaerchin: if your parent id is null, this implies you're a root level page. 

if your `pageName` is null, this implies you're not renaming a page. if this is confusing, i'm ok w/ making it non nullable but that means that the caller always has to pass in these properties. 

the principle for this API call becomes "pass in what you want to update" to "pass in the current state of the world". 

i think both are fine, no point optimising for request size because it's never going to be big but i thought the former was clearer.

packages/studio/src/schemas/page.ts
seaerchin: for now, yes. underlying schema will expose only 1 table still - i think this distinction makes sense because pages/folders are what the frontend sees (and queries), but the internals are db-specific and limited to a single table. 

this also implies that there might be some logical differences between a page-level route/folder-level route, such as (but not limited to) validation of names

packages/studio/src/schemas/page.ts
harishv7: Are we talking about something like a PATCH vs PUT?

harishv7: Also when I update a page schema, why do I need to pass in the parentId? From earlier discussion, it seems that we will derive the parentId from the page's resourceId

turbo.json
```diff
@@ -5,6 +5,7 @@
     "clean": {},
     "setup:test": {},
     "teardown": {},
-    "test-start": {}
+    "test-start": {},
```
harishv7, line 8: whats all this ah? I have already added these commands in turbo.json

harishv7:
@seaerchin might want to rebase on main

turbo.json
seaerchin: There was a dev command missing previously!

package.json
```diff
@@ -11,7 +11,8 @@
     "storybook": "turbo run storybook --parallel",
     "storybook:build": "turbo run storybook:build --parallel",
     "teardown": "turbo teardown",
-    "test-start": "turbo test-start"
+    "test-start": "turbo test-start", 
+    "dev": "turbo dev"
```
harishv7, line 15: this cmd alr exists `"dev": "dotenv -- turbo run dev"` on main branch

turbo.json
harishv7: true, but the latest main alr has these now, so these changes are not required

packages/studio/src/schemas/page.ts
```diff
@@ -0,0 +1,33 @@
...
+  parentId: z.string().nullable(),
+  pageName: z.string().nullable(),
+  siteId: z.string(),
+  pageId: z.string()
+})
+
+export const updatePageBlobSchema = z.object({
```
kishore03109, line 19: hmm actually some of this schema can be extended from one another! but think we can improver this later

packages/studio/src/schemas/page.ts
```diff
@@ -0,0 +1,33 @@
...
+  pageId: z.string(),
+  siteId: z.string(),
+  content: z.string()
+})
+
+export const createPageSchema = z.object({
+  pageName: z.string(),
+  pageTitle: z.string(),
+  // TODO: add the actual layouts in here
+  layout: z.enum(["content"]).default("content"),
```
kishore03109, line 29: nit: we would prob need this in a constants file, shift this up so other components can use/reference this

kishore03109 (approved):

package.json
seaerchin: removed

turbo.json
seaerchin: ok removed

PR #125, opened 2024-05-29
fix: update breadcrumb display

sehyunidaaa:
Update breadcrumb design for three main changes:

- "Home" should be displayed
- Better to not display current page; after doing some competitive analysis, seems like it's rare to show the current page and it serves zero user value as well, especially since we always show the page title top and center. It also messes up the Article layout a bit especially when the current page title is very long
- Last menu in crumbs should not have chevron trailing after it

### DIFF
--- a/packages/components/src/templates/next/components/internal/Breadcrumb/Breadcrumb.tsx
+++ b/packages/components/src/templates/next/components/internal/Breadcrumb/Breadcrumb.tsx
@@ -5,23 +5,19 @@ const Breadcrumb = ({ links, LinkComponent = "a" }: BreadcrumbProps) => {
   return (
     <div className="flex flex-row flex-wrap gap-1">
       {links.map((link, index) => {
-        const isCurr = index === links.length - 1
+        const isLast = index === links.length - 1
         return (
           <div
             key={index}
-            className="flex flex-row items-center gap-1 text-content-medium"
+            className="text-content-medium flex flex-row items-center gap-1"
           >
-            {isCurr ? (
-              <p className="font-semibold text-content">{link.title}</p>
-            ) : (
-              <LinkComponent
-                href={link.url}
-                className="underline underline-offset-2 hover:text-hyperlink-hover active:text-hyperlink"
-              >
-                {link.title}
-              </LinkComponent>
-            )}
-            {!isCurr && <MdChevronRight className="h-auto min-w-6" />}
+            <LinkComponent
+              href={link.url}
+              className="hover:text-hyperlink-hover active:text-hyperlink underline underline-offset-2"
+            >
+              {link.title}
+            </LinkComponent>
+            {!isLast && <MdChevronRight className="h-auto min-w-6" />}
           </div>
         )
       })}
--- a/packages/components/src/utils/getBreadcrumbFromSiteMap.ts
+++ b/packages/components/src/utils/getBreadcrumbFromSiteMap.ts
@@ -7,14 +7,14 @@ export const getBreadcrumbFromSiteMap = (
   sitemap: IsomerSitemap,
   permalink: string[],
 ): BreadcrumbProps => {
-  const breadcrumb = []
+  const breadcrumb = [
+    {
+      title: "Home",
+      url: "/",
+    },
+  ]
   let node = sitemap
   let currentPath = ""
-
-  breadcrumb.push({
-    title: "Home",
-    url: "/",
-  })
 
   for (const pathSegment of permalink.slice(0, -1)) {
     currentPath += "/" + pathSegment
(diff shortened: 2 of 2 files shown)


packages/components/src/templates/next/components/internal/Breadcrumb/Breadcrumb.tsx
```diff
@@ -5,23 +5,19 @@ const Breadcrumb = ({ links, LinkComponent = "a" }: BreadcrumbProps) => {
   return (
     <div className="flex flex-row flex-wrap gap-1">
       {links.map((link, index) => {
-        const isCurr = index === links.length - 1
+        const isLast = index === links.length - 1
         return (
           <div
             key={index}
-            className="flex flex-row items-center gap-1 text-content-medium"
+            className="text-content-medium flex flex-row items-center gap-1"
```
seaerchin, line 12: [no action required] quick question - does order actually matter here?

packages/components/src/utils/getBreadcrumbFromSiteMap.ts
```diff
@@ -11,7 +11,12 @@ export const getBreadcrumbFromSiteMap = (
   let node = sitemap
   let currentPath = ""
 
-  for (const pathSegment of permalink) {
+  breadcrumb.push({
+    title: "Home",
+    url: "/",
+  })
```
seaerchin, line 17: could we do 
```ts
const breadcrumb = [{
    title: "Home",
    url: "/",
  }]
```

looks slightly neater, benefits are that we don't directly mutate the `breadcrumb`

packages/components/src/templates/next/components/internal/Breadcrumb/Breadcrumb.tsx
```diff
@@ -5,23 +5,19 @@ const Breadcrumb = ({ links, LinkComponent = "a" }: BreadcrumbProps) => {
...
-                {link.title}
-              </LinkComponent>
-            )}
-            {!isCurr && <MdChevronRight className="h-auto min-w-6" />}
+            <LinkComponent
+              href={link.url}
+              className="hover:text-hyperlink-hover active:text-hyperlink underline underline-offset-2"
+            >
+              {link.title}
+            </LinkComponent>
```
seaerchin, line 19: i think this might have some differences in styling? previously, the last element has `font-semibold text-content` but this is omitted now and it's styled similarly to every other element in our new rendition. 

both seem fine to me, i'll let you make the call?

seaerchin (commented):
please avoid mutation! i know it's a minor thing but it's honestly one of those things that don't matter until they do

packages/components/src/templates/next/components/internal/Breadcrumb/Breadcrumb.tsx
sehyunidaaa: no, also linter i think!

packages/components/src/templates/next/components/internal/Breadcrumb/Breadcrumb.tsx
sehyunidaaa: yup! now the last element is the current page's parent, so it should be a link and shouldn't have the semibold formatting

packages/components/src/utils/getBreadcrumbFromSiteMap.ts
sehyunidaaa: arhhh thank you! addressed in [632c4ae](https://github.com/opengovsg/isomer/pull/125/commits/632c4ae8d14136ec3859bd3d5b82237629ff20e3)

packages/components/src/utils/getBreadcrumbFromSiteMap.ts
```diff
@@ -7,11 +7,16 @@ export const getBreadcrumbFromSiteMap = (
...
+    {
+      title: "Home",
+      url: "/",
+    },
+  ]
   let node = sitemap
   let currentPath = ""
 
-  for (const pathSegment of permalink) {
+  for (const pathSegment of permalink.slice(0, -1)) {
```
kishore03109, line 19: hmm this one abit error prone.
shall we create a util function called `removeTrailingSlash` that checks for the existence of it and then slices it? this way we are more resilient to errors if upstream callers dont put the slash

sehyunidaaa: oh hmm the `slice` isn't to check for a trailing slash tho, it's to remove the last item in the permalink which is the current page. is it still error prone this way?

kishore03109: hmm if that is the case what happens if i am in the top level ah

sehyunidaaa: Just 'home' would be displayed

kishore03109 (approved):

PR #126, opened 2024-05-29
fix(style): article layout design fixes

sehyunidaaa:
Article layout was looking jank

Key changes:
- Smaller title font size
- Removed box around summary
- Summary font colour is lighter
- Removed divider
- Adjusted some spacing

After images:
image
image


seaerchin (approved):
Lgtm but how should we be thinking about colours?

Noticed we are using palettes here, instead of named variants such as `text.content.dark`.

Either is fine to me but named palettes have more meaning to them as they don't represent just another shade. Additionally, since we attach meaning to the name and we use them with that meaning (eg, I wouldn't use `text.title.heavy` for body), refactorings and changes become easier later

sehyunidaaa:
@seaerchin agreed on the colours! let's figure out next week onwards what's the best practice for this.

PR #127, opened 2024-05-29
ISOM-1116: Create a base setup for isomer studio

harishv7:
## Problem

Has some refactor and setup improvements such as moving `studio` into its own `apps` library.

## Solution

Architecture of the monorepo:
`apps`: contains all the applications we are building. for now, this is just studio
`packages`: shared libs and packages we intend to publish. for now, this is just `isomer-components`

**Breaking Changes**

  - Details ...


apps/studio/package.json
```diff
@@ -28,7 +28,7 @@
     "test-dev:unit": "dotenv -e .env.test vitest",
     "test-start": "start-server-and-test start http://127.0.0.1:3000 test",
     "postinstall": "npm run generate && npm run build:theme",
-    "storybook": "storybook dev -p 6006",
+    "storybook": "storybook dev -p 6007",
```
seaerchin, line 31: Did this conflict with another port?

seaerchin (approved):
Reviewed by commit, seems ok!

apps/studio/package.json
harishv7: yes we have 2 storybooks now - 1 for components, 1 for studio

PR #129, opened 2024-05-29
fix(styling): stats block styling updates

sehyunidaaa:
Tweaks to stats block:
- row gap between title and stats
- adjust max-width
- adjust title size


seaerchin:
i think there's some logic changes so i deferred reviewing. lazy to open and look, lmk if i'm blocking

sehyunidaaa:
@seaerchin no logic changes btw! i removed <section> tag as advised by ZJ, but other than that it should be styling changes!

seaerchin (approved):

PR #130, opened 2024-05-29
fix(style): design fixes for table

sehyunidaaa:
Minor design updates for table:

- Header row text is left-aligned
- Header row text has smaller font size

### DIFF
--- a/packages/components/src/templates/next/components/native/Table/Table.tsx
+++ b/packages/components/src/templates/next/components/native/Table/Table.tsx
@@ -67,7 +67,7 @@ const Table = ({ caption, content }: TableProps) => {
         <caption className="mb-4 caption-top text-left">
           <BaseParagraph
             content={caption}
-            className="text-caption-01 sticky left-0 table-header-group text-balance text-content"
+            className="text-caption-01 text-content sticky left-0 table-header-group text-balance"
           />
         </caption>
         <tbody>
@@ -78,7 +78,7 @@ const Table = ({ caption, content }: TableProps) => {
             return (
               <tr
                 key={index}
-                className="*:first:border-divide-subtle *:first:border-t"
+                className="*:first:border-divide-subtle text-left *:first:border-t"
               >
                 {row.content.map((cell, cellIndex) => {
                   return (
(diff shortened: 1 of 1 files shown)


seaerchin:
!run chromatic

packages/components/src/templates/next/components/native/Table/Table.tsx
```diff
@@ -78,18 +78,18 @@ const Table = ({ caption, content }: TableProps) => {
...
                     <TableCellTag
                       key={cellIndex}
                       colSpan={cell.colSpan}
                       rowSpan={cell.rowSpan}
-                      className={`border-divide-subtle max-w-40 break-words border-b border-r px-4 py-3.5 align-top first:border-l last:max-w-full [&_li]:my-0 [&_li]:pl-1 [&_ol]:mt-0 [&_ol]:ps-5 [&_ol]:text-sm [&_ul]:mt-0 [&_ul]:ps-5 ${
+                      className={`border-divide-subtle max-w-40 break-words border-b border-r px-4 py-3.5 align-top first:border-l last:max-w-full [&_li]:my-0 [&_li]:pl-1 [&_ol]:mt-0 [&_ol]:ps-5 [&_ol]:text-sm [&_p]:text-sm [&_ul]:mt-0 [&_ul]:ps-5 ${
                         cell.type === "tableHeader"
                           ? "bg-utility-neutral"
-                          : "bg-utility-neutral-subtle [&_p]:text-sm"
+                          : "bg-utility-neutral-subtle"
```
seaerchin, line 92: actually just a point here - we now rely exclusively on the background to differentiate between header-row and cell. 

going off notion's one, they seem to have the header row having slightly heavier `font-weight` (and is also what i'd expect off my web browsing habits). wondering how much extra effort is required for this to conform w/ user's expectations
image

seaerchin (approved):
lgtm, 1 minor comment but not a blocker

packages/components/src/templates/next/components/native/Table/Table.tsx
sehyunidaaa: using font-weight to differentiate header as well! this screenshot is from this branch:

image

PR #136, opened 2024-05-29
fix(styling): collections layout design cleanup

sehyunidaaa:
A temporary design cleanup for the Collections layout. **This layout will be going through redesigns soon**, so this cleanup is just to iron out alignment issues to make more presentable, notably:

- Temporarily hide selected filters (it's been commented out)
- Pagination: fix issue where left-arrow was really small and alignment issue
- Vertical alignment of "Filter by" and Results so that the horizontal lines don't look janky
- Filters was taking up too little width, adjusted to take up more space (1/4 in larger screens)
- Search bar is thicker for visual balance
- Page title and subtitle take up limited width, to follow other page layouts & best practices
- Fix missing gap above empty results copy in mobile

Before:
image

After:
image

Mobile:
image

### DIFF
--- a/packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
+++ b/packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
@@ -30,7 +30,7 @@ const ArticleTextComponent = ({
       <div className={`flex flex-col gap-3`}>
         <div className="flex flex-col gap-1 sm:flex-row sm:gap-2">
           <p className="text-caption-01 text-content-medium">{lastUpdated}</p>
-          <p className="text-caption-01 hidden text-content-medium sm:block">
+          <p className="text-caption-01 text-content-medium hidden sm:block">
             |
           </p>
           <p className="text-caption-01 text-content-strong">{category}</p>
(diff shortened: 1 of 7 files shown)


packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
```diff
@@ -30,15 +30,15 @@ const ArticleTextComponent = ({
       <div className={`flex flex-col gap-3`}>
         <div className="flex flex-col gap-1 sm:flex-row sm:gap-2">
           <p className="text-caption-01 text-content-medium">{lastUpdated}</p>
-          <p className="text-caption-01 hidden text-content-medium sm:block">
+          <p className="text-caption-01 text-content-medium hidden sm:block">
```
kishore03109, line 33: hmm wait what why it this hidden ah?

packages/components/src/templates/next/layouts/Collection/CollectionClient.tsx
```diff
@@ -309,7 +309,7 @@ const CollectionClient = ({
                   </>
                 )}
               </p>
-
```
kishore03109, line 312: why ah, do we not want this?

packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
sehyunidaaa: linter! was previously hidden

but it's hidden on mobile cos the category and date collapses into 2 rows

packages/components/src/templates/next/layouts/Collection/CollectionClient.tsx
sehyunidaaa: got design feedback that the pills might not be relevant because you can directly uncheck filters. it makes alignment tricky as well

but just hiding it for now because we are redesigning the layout, might end up keeping it

seaerchin (approved):
lgtm - i'm scared of the comment becoming a fossil in our code but ok

PR #138, opened 2024-05-30
fix(styling): minor design fixes for components

sehyunidaaa:
Adding minor design fixes to single PR because most of these are one-liners per component:

- **Accordion** had focus state which was appearing on active state, removed (focus state is for keyboard manipulation)
- **Button** y-padding needs to be larger
- **Infocards** should use h2 tag (like all other blocks) instead of h3 
- **Infocards** should use outline instead of border

The linter may have caused reordering of classnames

### DIFF
--- a/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
+++ b/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
@@ -4,10 +4,10 @@ import { Prose } from "../../native"
 
 const Accordion = ({ summary, details }: AccordionProps) => {
   return (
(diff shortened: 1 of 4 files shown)


packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
```diff
@@ -4,10 +4,10 @@ import { Prose } from "../../native"
 
 const Accordion = ({ summary, details }: AccordionProps) => {
   return (
-    <details className="border-y border-divider-medium hover:cursor-pointer has-[+_details]:border-b-0 [&_>_summary]:open:font-semibold [&_>_summary_>_svg]:open:rotate-180">
```
kishore03109, line 7: weird. thought i ran linter for the whole project already :/

kishore03109 (approved):

PR #149, opened 2024-06-05
feat: create folder dto

alexanderleegs:
This PR adds the DTO for create folder. Resolves https://linear.app/ogp/issue/ISOM-1083/create-dto-for-create-folder.


apps/studio/src/schemas/folder.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
+
+export const createFolderSchema = z.object({
+  folderTitle: z.string(),
```
harishv7, line 4: do we have restrictions on the length of the fields?

alexanderleegs: we should! will confirm and update the fields

kishore03109 (approved):

apps/studio/src/schemas/folder.ts
alexanderleegs: Added in https://github.com/opengovsg/isomer/pull/149/commits/a915651965c55f32283ea5199230f5057d142d7a

PR #155, opened 2024-06-05
ISOM-1087: Setup ORM

harishv7:
## Problem

StarterKit already comes with Prisma. However, during the last discussion, we talked about how Prisma is limiting and the way it does some of the joins etc. is not efficient. Please point out the exact details here if you know.

Closes ISOM-1087

## Solution

As one of the solutions, we discussed the use of Prisma alongside Kysely. Kysely is a type-safe query builder (not a full-fledged ORM). Hence, for our queries we can use Kysely, and we use Prisma to define our schema and for migrations. We can use the `prisma-kysely` lib to generate the types.

This PR introduces Kysely alongside a simple example.

Note that we can still choose/opt to just go with Prisma alone.

**Breaking Changes**

  - Details ...

## Tests

- [ ] Add a console log after the Kysely example to print out the results
- [ ] Navigate to http://localhost:3000/profile/{userId}/liked on local
- [ ] See if your terminal is printing the output

**New scripts**:

- `db:push` : this is a Prisma specific push command useful during the prototyping phase. Helps to synchronise db schema and prisma schema. Docs [here](https://www.prisma.io/docs/orm/prisma-migrate/workflows/prototyping-your-schema). Note: Not to be used in production!

**New dependencies**:

- `pg` : required to create and initialise pool for Kysely DB in `apps/studio/src/utils/database.ts`
- `kysely`: type-safe query builder
- `prisma-kysely`: helps to generate types from prisma schema

**New dev dependencies**:

- `@types/pg` : for above `pg` module


apps/studio/prisma/generated/types.ts
```diff
@@ -0,0 +1,52 @@
+import type { ColumnType } from "kysely"
```
kishore03109, line 1: nit: use ts-morph, and generate a heading that states not to edit this file direcly as this is auto-generated. 

similar to what kr did (here)[https://github.com/opengovsg/activesg/pull/892/files]

apps/studio/prisma/generated/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  T extends ColumnType<infer S, infer I, infer U>
+    ? ColumnType<S, I | undefined, U>
+    : ColumnType<T, T | undefined, T>
+export type Timestamp = ColumnType<Date, Date | string, Date | string>
+
+export type Accounts = {
+  id: string
+  provider: string
+  providerAccountId: string
+  user_id: string
```
kishore03109, line 12: hmm this means that we will end up using snake case throughout our codebase. Is there a setting we can use to make camelCase instead? (iirc, that is the default, not sure why this is snek case)

apps/studio/prisma/seed.ts
```diff
@@ -3,19 +3,13 @@
...
-    console.error(e)
-    process.exit(1)
-  })
-  .finally(async () => {
-    await prisma.$disconnect()
-  })
+main().catch((e) => {
+  console.error(e)
+  process.exit(1)
+})
```
kishore03109, line 15: shall we .finally(disconnect)? if not failed instances might have dangling connections liao

apps/studio/prisma/generated/types.ts
harishv7: hmm this represents our DB right, which is in snake_case?

apps/studio/prisma/seed.ts
harishv7: oh yes, will add

apps/studio/prisma/generated/types.ts
kishore03109: https://github.com/valtyr/prisma-kysely
think we just need to enable a flag for the camel case plugin bah

apps/studio/prisma/generated/types.ts
seaerchin: I think the point of `ts-morph` was to generate typings automatically cos they were using postgis which wasn't supported.

We can still do it, but given that this type is already automatically generated (and lives in a `generated` folder), there might not be much incremental value

apps/studio/prisma/generated/types.ts
harishv7: i removed the `@map` on the schema level, so all is in camelCamse now

apps/studio/prisma/seed.ts
harishv7: done

apps/studio/prisma/generated/types.ts
harishv7: yea, skipping this for now

apps/studio/.gitignore
```diff
@@ -41,7 +41,7 @@ yarn-error.log*
 playwright/test-results
 
 # env
-.env
+.env*
```
kishore03109, line 44: what about .env.example? we using env-example ah?

kishore03109 (approved):

apps/studio/.gitignore
harishv7: will make it less restrictive first

PR #156, opened 2024-06-06
Chore/migrate playground

alexanderleegs:
This PR migrates jsonforms from playground repo to studio.

The existing implementation for the tiptap editor has been modified slightly to account for the use of design-system instead of material ui. Also note that JSONForms makes use of materialUI under the hood - it's possible to replace this, but it requires us to create a custom renderer for chakra components instead for each component used


harishv7 (approved):
lgtm as this is a portover

harishv7:
@alexanderleegs might want to fix the tests failing, I think need to add types

PR #159, opened 2024-06-10
ISOM 1152 table view for my pages

hanpuliu-charles:
## Problem

Closes [ISOM-1152](https://linear.app/ogp/issue/ISOM-1152/table-view-for-my-pages)
Component needed for dashboard view

## Solution

- Creating a draft of what the dashboard display may look like based on Figma mockups with dummy data. 
- Some changes will have to be made to call backend endpoint, fetch actual data and manipulated to fit expected type(Or change the code for the table to fit the data)

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

image


.gitignore
```diff
@@ -41,3 +41,6 @@ build-storybook.log
 
 # Build artifacts
 dist/
+
+# env file
+.env
```
harishv7, line 46: i think this exists at the root level .env right?

apps/studio/src/pages/dashboard/index.tsx
```diff
@@ -1,17 +1,281 @@
...
-      <h1>DASHBOARD</h1>
+    <>
+      <TableContainer
+        w="100%"
+        borderRadius="8px"
+        border="1px"
+        borderColor="base.divider.medium"
+        bgColor="#FFF"
+      >
+        <Table>
```
harishv7, line 138: Usually we keep the stuff under the `pages` directory lean. It should mostly import components from the root components folder or the `/features` folder

Suggestion is to move this table layout under the `/features/dashboard` folder

apps/studio/src/pages/dashboard/index.tsx
```diff
@@ -1,17 +1,281 @@
...
+                    <Td>
+                      {/* Fill doesn't seem to work with semantic tokens   */}
+                      {element.type === 'page' && element.status == 'draft' && (
+                        <Badge
+                          variant="subtle"
+                          colorScheme="warning"
+                          borderRadius="50px"
+                        >
+                          <BiSolidCircle
+                            fill="#FFDA68"
```
harishv7, line 208: why do we have a fill for the draft circle but not the published one?

hanpuliu-charles: The default color scheme of the badge applies the correct fill. Will put in the fill for the published circle for completeness sake!

apps/studio/src/pages/dashboard/index.tsx
hanpuliu-charles: Understand the above, will shift the layout file to /features/dashboard

.gitignore
hanpuliu-charles: Okay to remove, I think it was autogenerated by something

hanpuliu-charles: Hmm, don't quite get that, possible to clarify? Thanks!

harishv7: Thought this already exists, but saw it was missing.

harishv7 (approved):

apps/studio/src/pages/dashboard/index.tsx
hanpuliu-charles: I have updated to wrap it in Chakra's icon component. Semantic tags now work and we no longer need the inline style tag.

PR #178, opened 2024-06-14
ISOM-1086 DB schema crude draft for comment

hanpuliu-charles:
## Problem

Eventually closes [ISOM-1086](https://linear.app/ogp/issue/ISOM-1086/create-schema-for-database)

prisma-erd (1)

## Solution

Very crude example of what the final schema might look like for comment and feedback, 

**Breaking Changes**

  - Details ...


apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,73 @@ model VerificationToken {
...
-  emailVerified DateTime?
-  image         String?
-  bio           String?
-  posts         Post[]
-  likedPosts    LikedPosts[]
-  accounts      Accounts[]
+  id    String @id @default(cuid())
+  name  String
+  email String @unique
+  phone String @unique
```
dcshzj, line 54: nit: no strong opinion here but I think the phone number doesn't need to be unique, since we are not using it to identify individual users?

dcshzj (changes requested):
Not going in-depth regarding the permissions model yet since it is out of scope of the issue

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,73 @@ model VerificationToken {
...
-  post      Post     @relation(fields: [postId], references: [id], onDelete: NoAction, onUpdate: NoAction)
-  postId    String
-  user      User     @relation(fields: [userId], references: [id], onDelete: NoAction, onUpdate: NoAction)
-  userId    String
-  createdAt DateTime @default(now())
+  site         Site         @relation(fields: [siteId], references: [id])
+  parent       Resource?    @relation("parentRelation", fields: [parentId], references: [id])
+  child        Resource[]   @relation("parentRelation")
+  lastEditUser User?        @relation(fields: [userId], references: [id], onDelete: NoAction, onUpdate: NoAction) // Don't delete the file if the user is deleted
+  blob         Blob?
```
alexanderleegs, line 44: Hmm we might not want to store the blob directly in the resource table! We only really need to retrieve one file's content at a time, and this might end up bloating the table we use primarily to retrieve the structure of the pages

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,73 @@ model VerificationToken {
...
+  Permission   Permission[]
 
-  @@id([postId, userId])
+  @@unique([siteId, parentId, name], name: "uniqueChildNames")
```
alexanderleegs, line 47: Not sure if we really need this to be unique - I think we expect this to be the case but it isn't necessarily required for the table to function properly

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,73 @@ model VerificationToken {
...
+  resourceId String
+  userId     String
+  role       RoleType
+
+  // Delete the row if either resource or user is deleted, default behavior cascade
+  resource Resource @relation(fields: [resourceId], references: [id])
+  user     User     @relation(fields: [userId], references: [id])
+
+  // userId and resource Id should be unique? Each user should have unambiguous access to resource? Or not if we take the strictest or laxest perm for the user
+  @@unique([userId, resourceId], name: "unambiguousPermissionsPerUserResource")
```
alexanderleegs, line 71: I think this depends on whether the roles are strict subsets of each other, from what we discussed last week this seems to be the case, but we can drop it in future if this changes

apps/studio/prisma/migrations/20240614070259_initial_migration/migration.sql
alexanderleegs, line 1: Could we rename this migration to something different to indicate this is for the cms? We already have a initial migration (which came with starterkit)

apps/studio/prisma/schema.prisma
hanpuliu-charles: ah oops, should be blobId?

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,73 @@ model VerificationToken {
...
+model Resource {
+  id             String       @id @default(cuid())
+  name           String
+  siteId         String
+  parentId       String?
+  lastEditTime   DateTime?
+  lastEditUserId Int
+  blobId         String?      @unique
+  variant        ResourceType
+  userId         String
```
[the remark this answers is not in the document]
hanpuliu-charles: Oversight on my part, they should be the same thing

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,73 @@ model VerificationToken {
...
 
-model LikedPosts {
-  post      Post     @relation(fields: [postId], references: [id], onDelete: NoAction, onUpdate: NoAction)
-  postId    String
-  user      User     @relation(fields: [userId], references: [id], onDelete: NoAction, onUpdate: NoAction)
-  userId    String
-  createdAt DateTime @default(now())
+  site         Site         @relation(fields: [siteId], references: [id])
+  parent       Resource?    @relation("parentRelation", fields: [parentId], references: [id])
+  child        Resource[]   @relation("parentRelation")
```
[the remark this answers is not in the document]
hanpuliu-charles: I think that children is more suitable contextually,

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,77 @@ model VerificationToken {
...
-  @@index([createdAt])
+model Resource {
+  id             String       @id @default(cuid())
+  name           String
+  siteId         String
+  parentId       String?
+  lastEditTime   DateTime?
+  lastEditUserId String
+  blobId         String?      @unique
+  variant        ResourceType
```
seaerchin, line 37: suggest removing `variant` here and let the code decide by presence of `blobId`. reasons why this makes sense: 1 less thing to set in our db + product still young, assumptions can change later on.

wdyt? (this isn't a big deal though so happy to keep as well)

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,77 @@ model VerificationToken {
...
+  status         PageStatus // Is this a good place to be storing this?
+  site           Site         @relation(fields: [siteId], references: [id])
+  parent         Resource?    @relation("parentRelation", fields: [parentId], references: [id])
+  children       Resource[]   @relation("parentRelation")
```
seaerchin, line 41: is this a 1 to many relation? so a `page` would never have any `children` + `blobId` is defined?

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,77 @@ model VerificationToken {
...
 
-model Accounts {
-  id                String @id @default(cuid())
-  provider          String
-  providerAccountId String
+model Site {
+  id   String @id @default(cuid())
+  name String @unique
+
+  Resource Resource[]
```
seaerchin, line 77: should `Site` have a 1 to many key to user indicating that the user has access to this site?

apps/studio/prisma/schema.prisma
hanpuliu-charles: Now that you mentioned it, seems like it makes sense to remove variant for now, since there is a direct correlation between blobId being null and variant being folder.
Should also be possible to add such a field in a future migration if needed.

apps/studio/prisma/schema.prisma
hanpuliu-charles: yes, I think that's right, folder's blobId is undefined, may have 0 to n children, but page's blobId is defined, should have 0 children.

apps/studio/prisma/schema.prisma
hanpuliu-charles: hmmm, from your experience, is it more natural to store this relation in the Site table or the User table?

seaerchin: Both are ok, we could even have a join table.

apps/studio/prisma/schema.prisma
hanpuliu-charles: Removing for now

apps/studio/prisma/schema.prisma
hanpuliu-charles: Removing for now

apps/studio/prisma/schema.prisma
hanpuliu-charles: From discussion with Chin, don't lock in to strict subset roles

apps/studio/prisma/schema.prisma
hanpuliu-charles: many to many?

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,86 @@ model VerificationToken {
...
+  lastEditUserId String
+  blobId         String?      @unique
+  status         PageStatus // Is this a good place to be storing this?
+  site           Site         @relation(fields: [siteId], references: [id])
+  parent         Resource?    @relation("parentRelation", fields: [parentId], references: [id])
+  children       Resource[]   @relation("parentRelation")
+  lastEditUser   User?        @relation(fields: [lastEditUserId], references: [id], onDelete: NoAction, onUpdate: NoAction) // Don't delete the file if the user is deleted
+  blob           Blob?
+  permission     Permission[]
+
```
hanpuliu-charles, line 44: Needs relation?

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,86 @@ model VerificationToken {
...
-model Accounts {
-  id                String @id @default(cuid())
-  provider          String
-  providerAccountId String
+model Blob {
+  id      String @id @default(cuid())
+  content Json
 
+  // Should the blob be deleted on deletion of the resource? 
+  resource Resource @relation(fields: [id], references: [blobId])
```
seaerchin, line 85: we should make the be on resource because most (if not all) of the time, the query will be from resource -> blob.

seaerchin (approved):
approving first, only issue is the FK directon.

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,67 @@ model VerificationToken {
...
+  SiteMembers SiteMember[]
 }
 
-model Accounts {
-  id                String @id @default(cuid())
-  provider          String
-  providerAccountId String
+model Blob {
+  id      String @id @default(cuid())
+  content Json
```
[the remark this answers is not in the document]
hanpuliu-charles: Yep, checked the migration sql, it is using jsonb type

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,67 @@ model VerificationToken {
...
+  id       String  @id @default(cuid())
+  name     String
+  siteId   String
+  parentId String?
+
+  blobId     String?      @unique
+  site       Site         @relation(fields: [siteId], references: [id])
+  parent     Resource?    @relation("parentRelation", fields: [parentId], references: [id])
+  children   Resource[]   @relation("parentRelation")
+  blob       Blob?        @relation(fields: [blobId], references: [id])
```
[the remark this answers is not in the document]
hanpuliu-charles: Will make the changes, thanks

apps/studio/prisma/schema.prisma
```diff
@@ -26,68 +26,67 @@ model VerificationToken {
...
-  createdAt DateTime @default(now())
+model User {
+  id    String @id @default(cuid())
+  name  String
+  email String @unique
+  phone String
 
-  @@id([postId, userId])
+  Permission  Permission[]
+  SiteMembers SiteMember[]
```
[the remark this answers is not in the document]
hanpuliu-charles: So I might be wrong here, but the field is relation field for the relationship betweeen User and SiteMember(which is a join table for Users to Sites, that they are a collaborator on). It is plural as a user could be collaborator on many sites(many to many relation here). At least, thats from my understanding of the schema language

apps/studio/prisma/schema.prisma
hanpuliu-charles: Should be rectified!

PR #188, opened 2024-06-18
feat: create form builder for basic input types

dcshzj:
## Problem

Closes ISOM-1159.

## Solution

**Breaking Changes**

**Features**:

- Adds a form builder for basic input types used by the schema

**Notes**:

- Only the very basic input types are supported (dropdown, radio, string, integer, boolean). Things that are not yet supported (but required for MVP are):
    - Support type = 'array', format = 'prose' to be Tiptap editor
    - Support type = 'array' with items of type = 'object' to be sub-form (no drag and drop yet)
    - Support $ref to other parts of the schema
    - Handle grouping of fields
    - Providing default input when editing an existing page
- The schema itself will need further modifications to make the output match the Figma designs

## Before & After Screenshots

Screenshot 2024-06-18 at 12 52 13


apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,165 @@
...
+  Button,
+  FormLabel,
+  Input,
+  NumberInput,
+  Radio,
+  Switch,
+} from '@opengovsg/design-system-react'
+import { IsomerComponent } from '@opengovsg/isomer-components'
+import { useForm } from 'react-hook-form'
+import IsomerSchema from './0.1.0.json'
```
harishv7, line 20: eventually we need to fetch this at runtime from schema.isomer.gov.sg but for dev, can use this first

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,165 @@
...
+              format,
+              enum: options,
+              title,
+              description,
+              // @ts-ignore
+            } = properties[prop]
+
+            if (type === 'string' && options && format === 'select') {
+              return (
+                <div className="py-2">
```
harishv7, line 65: prefer use of `Box` since we use fully chakra components

dcshzj: Ah good point, updated in 7429b65c56669cc4c10c9e532600212c66fee135.

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,166 @@
...
+// - For each prop inside the sub-schema, render the appropriate input field
+//
+// TODO:
+// - Support type = 'array', format = 'prose' to be Tiptap editor
+// - Support type = 'array' with items of type = 'object' to be sub-form (no drag and drop yet)
+// - Support $ref to other parts of the schema
+// - Handle grouping of fields
+export const FormBuilder = ({
+  component,
+  data,
```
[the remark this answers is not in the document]
dcshzj: Yup but it is intended to be used in the future, otherwise users would not be able to edit their existing blocks.

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,166 @@
...
+            const {
+              type,
+              format,
+              enum: options,
+              title,
+              description,
+              // @ts-ignore
+            } = properties[prop]
+
+            if (type === 'string' && options && format === 'select') {
```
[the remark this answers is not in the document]
dcshzj: Fixed in 36755dd499df6d9ff094a0e1ae5042d2a4da8287.

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+            if (type === 'string') {
+              return (
+                <Box py={2}>
+                  <FormControl isRequired={required.includes(prop)}>
+                    <FormLabel description={description}>{title}</FormLabel>
+                    <Input
+                      type="text"
+                      // @ts-ignore
+                      {...register(prop)}
```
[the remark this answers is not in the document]
dcshzj: I think I spent far too long on this part. This is primarily caused by the use of `Object.keys()` which types the result to `string[]`, then trying to use the `string` type to get a specific part of the properties object.

I improved the typings somewhat in 36755dd499df6d9ff094a0e1ae5042d2a4da8287 so we can remove most of the usages of `@ts-ignore`, but there are still 2 instances that remain which is when `prop` is used to access the `properties` object.

I couldn't really figure out a way to do it such that we can completely remove the `@ts-ignore` here, while preserving the fact that this part of the code is intended to be "dynamic". If got suggestions can let me know!

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
seaerchin: pls add a `TODO`

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,166 @@
...
+  component,
+  data,
+}: FormBuilderProps): JSX.Element => {
+  const schema = IsomerSchema.components.complex[component]
+  const { required, properties } = schema
+  const { register, handleSubmit } = useForm<typeof properties>()
+
+  // Utility type to get all keys from a union of objects
+  type UnionKeys<T> = T extends T ? keyof T : never
+  type ComplexComponentProps = UnionKeys<typeof properties>
```
seaerchin, line 49: nit: shift types outside

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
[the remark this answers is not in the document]
seaerchin: add a distinguishing property and make it a tagged union might work? i'm happy to accept this right now if functionality is complete (looks like) but this guarantee should be relooked at prior to rolling out

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
```diff
@@ -0,0 +1,166 @@
...
+// TODO:
+// - Support type = 'array', format = 'prose' to be Tiptap editor
+// - Support type = 'array' with items of type = 'object' to be sub-form (no drag and drop yet)
+// - Support $ref to other parts of the schema
+// - Handle grouping of fields
+export const FormBuilder = ({
+  component,
+  data,
+}: FormBuilderProps): JSX.Element => {
+  const schema = IsomerSchema.components.complex[component]
```
seaerchin, line 43: actually if we can do `as const`, the types might be automati?

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
seaerchin, line 1: i don't really have an issue here - the `FormBuilder` is big but each individual component seems reasonably small and adding an `if` case is clear (albeit unwieldy). 
 
i'm approving this for now because it seems like it's only simple (numeric/string/boolean) types but if we want to commit to bigger components, we might run into issues

seaerchin (approved):

apps/studio/src/features/editing-experience/components/FormBuilder.tsx
dcshzj: Updated in 5374fc108bd5295f6aa946227be03ec544c95ac9.

PR #189, opened 2024-06-18
ISOM-1155: Create a root drawer state

harishv7:
## Problem

We need a root drawer state for the edit page flow. This should show the structure of the page in terms of blocks and allow for drag-and-drop to re-arrange blocks.

Closes ISOM-1155

## Solution

Introduces new drawer state types and implements the root state drawer. The component within allows for drag-and-drop of the "Custom blocks" section as per Figma.

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**AFTER**:

Screenshot 2024-06-18 at 1 14 39 PM

## Tests

- Ensure the sidebar is showing up
- Ensure you can re-arrange the custom blocks

**New dependencies**:

- `@hello-pangea/dnd` : used for drag-and-drop. note that am not using https://github.com/atlassian/react-beautiful-dnd since it is under maintenance


apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { BsPlus } from 'react-icons/bs'
+import { Button } from '@opengovsg/design-system-react'
+import RootStateDrawer, { Block } from './RootStateDrawer'
+
+export type RootDrawerState = {
+  state: 'root'
+  blocks: Block[]
+}
+
+type AddNewBlockState = {
```
seaerchin, line 19: nit: can just use the string as the type 
```ts
type AddNewBlockState = "add-new-block"
```

not a big diff but we don't have to call `.state` to access

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type DrawerState =
+  | RootDrawerState
+  | AddNewBlockState
+  | NativeEditorState
+  | ComplexEditorState
+
+type EditPageDrawerProps = {
+  open: boolean
```
seaerchin, line 38: nit: isOpen

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const onDragEnd = (result: DropResult) => {
+    if (!result.destination) return
+
+    const updatedBlocks = Array.from(currState.blocks)
+    // Remove block at source index
+    const [movedBlock] = updatedBlocks.splice(result.source.index, 1)
+    // Insert at destination index
+    updatedBlocks.splice(result.destination.index, 0, movedBlock!)
+
+    setCurrState({ blocks: updatedBlocks })
```
seaerchin, line 41: did we take this from our existing code?

apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -1,13 +1,25 @@
...
+    <Grid w="100vw" templateColumns="repeat(3, 1fr)" gap={0}>
       {/* TODO: Implement sidebar editor */}
-      <GridItem w="100%" colSpan={1} bg="tomato">
-        <h1>Editor sidebar appears here</h1>
+      <GridItem w="100%" colSpan={1} bg="slate.50">
+        <EditPageDrawer
+          open
+          state={{
+            state: 'root',
+            blocks: [
```
seaerchin, line 15: these are placeholder blocks right

seaerchin (approved):
lgtm, i didn't try out the drag and drop but assuming it's taken from cms, it should be good

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
seaerchin, line 1: just a quick question - we conditionally show each specific drawer based on the state and this was chosen because we want to have the user click in and show different actions based on what they clicked right? eg: initial open -> root drawer -> add new item -> add new item drawer

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
harishv7: what do you mean by existing code? like classic code? nope

apps/studio/src/pages/dashboard/edit/index.tsx
harishv7: yea all placeholders for now

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
harishv7: that's correct!

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: ok sure, could i just check these behaviours:
1. dropping outside of dropzone
2. happy path dnd
3. dropping and releasing into original position

harishv7: @seaerchin no logic for those has been added yet. this is purely UI for displaying page structure.

Currently, dropping outside dropzone does nothing. Same for point (3). 

Happy path is to drag and drop or rearrange the blocks and the preview on the right should update. but these are logic to be tackled later

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
harishv7: ah intentionally left it like this for future extension

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
harishv7: fixed

PR #193, opened 2024-06-19
chore: rename keys from kebab-case to camelCase

dcshzj:
## Problem

Closes ISOM-1196.
Closes ISOM-1197.

## Solution

**Breaking Changes**

**Improvements**:

- Moved image component from native to complex component
- Renamed keys that were using kebab-case to camelCase

**Notes**:

- No changes are needed on mti-corp or other repos as the changes are only internally referenced by the schema. The page schemas uses the `type` key which were not renamed.
- **FOR FEEDBACK**: Decided to keep component names (referenced by the page schemas) to be lowercase (i.e. `unorderedlist` instead of `unorderedList`) for simplicity. Let me know if this approach is sane or we should use camelCase as much as possible.

### DIFF
--- a/docs/next/0.1.0.json
+++ b/docs/next/0.1.0.json
@@ -32,7 +32,7 @@
     }
   ],
   "definitions": {
-    "content-components": {
+    "contentComponents": {
       "type": "array",
       "minItems": 1,
       "items": {
@@ -51,6 +51,9 @@
           },
           {
             "$ref": "#/components/complex/iframe"
+          },
+          {
+            "$ref": "#/components/complex/image"
           },
           {
             "$ref": "#/components/complex/infobar"
@@ -74,9 +77,6 @@
             "$ref": "#/components/native/heading"
           },
           {
-            "$ref": "#/components/native/image"
-          },
-          {
             "$ref": "#/components/native/orderedlist"
           },
           {
@@ -91,7 +91,7 @@
         ]
       }
     },
-    "page-metadata": {
+    "pageMetadata": {
       "type": "object",
       "required": ["title"],
       "properties": {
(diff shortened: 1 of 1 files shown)


seaerchin (approved):

harishv7:
> FOR FEEDBACK: Decided to keep component names (referenced by the page schemas) to be lowercase (i.e. unorderedlist instead of unorderedList) for simplicity. Let me know if this approach is sane or we should use camelCase as much as possible.

@dcshzj with regards to the above, what is the impact of renaming? if it's straightforward, I prefer to have it standardised that when we see keys with "two or more words" they will be in camelCase

dcshzj:
> @dcshzj with regards to the above, what is the impact of renaming? if it's straightforward, I prefer to have it standardised that when we see keys with "two or more words" they will be in camelCase

Mainly we have to rename all the usages of the keys inside the existing repos. I initially hesitated to do this because renaming might require some effort and coordination, and that I feel the names of the components are more of IDs which made sense to be all in lower-case.

harishv7:
> > @dcshzj with regards to the above, what is the impact of renaming? if it's straightforward, I prefer to have it standardised that when we see keys with "two or more words" they will be in camelCase
> 
> Mainly we have to rename all the usages of the keys inside the existing repos. I initially hesitated to do this because renaming might require some effort and coordination, and that I feel the names of the components are more of IDs which made sense to be all in lower-case.

@dcshzj but i saw the following which is no longer standardised right?
Screenshot 2024-06-20 at 2 12 57 PM

dcshzj:
Ok sure, will schedule a time with POEs to perform a migration

harishv7:
> Ok sure, will schedule a time with POEs to perform a migration

Thanks @dcshzj , can merge this and have another issue for the keys requiring migration perhaps

PR #194, opened 2024-06-19
chore(schema): cleanup app

seaerchin:
## Problem
Previously, we had a table called `Posts` inside Prisma which has been dropped recently as a result of our db schema updating. 

This led to migrations + deployments being ran, which causes the typings between frontend and backend to be out of sync (generated typings now different). As a result, we need to tear down stuff related to `Posts`. (however, this turns out to be hugely annoying and i've torn down more stuff to avoid fine-grained but time-wasting deletions)

## Solution
- delete all routers on our backend
- update `AppRouter` injection point to make sure that our typings are correct on the frontend
- remove components on frontend that use said hooks or remove callsites.

## Noteworthy changes
- `auth` integration helper has changed to require a full `User` rather than `Partial<User>`. should have no impact


packages/components/opengovsg-isomer-components-0.0.13.tgz
harishv7, line 1: @seaerchin why do we need to add this binary here?

dcshzj:
I think in theory this is sane but why not completely nuke `src/pages/profile/[username]` and `src/pages/thread` as well?

apps/studio/src/server/modules/post/__tests__/post.router.test.ts
harishv7, line 1: @seaerchin have we also deleted the irrelevant stuff from schema.prisma?

Also this will imply our `initial_migration` file generated for prisma will now be outdated. can we regenerate that?

packages/components/opengovsg-isomer-components-0.0.13.tgz
seaerchin: lemme remove, i auto-added everything cos there were alot of files. think this got added by accident

apps/studio/src/server/modules/post/__tests__/post.router.test.ts
seaerchin: we deleted in hanpu's prior PR, that led to tests failing and hence, why this is required. 

i think the `initial_migration` might have already been updated in hanpu's PR [here](https://github.com/opengovsg/isomer/pull/178/files)

seaerchin:
> I think in theory this is sane but why not completely nuke `src/pages/profile/[username]` and `src/pages/thread` as well?

actually also can hor, i just scared cos each change has ripple effect HAHA

harishv7 (approved):

PR #198, opened 2024-06-19
Feat/add native block editor component

alexanderleegs:
## Problem

This PR introduces the native block editor component, which uses TipTap. For the styling of this component, the styles from the isomer classic cms have been reimported to be used in studio.
Screenshot 2024-06-20 at 12 16 14 AM

Table and link functionality is not fully implemented, pending design

In addition, a basic context has been added to help manage the state of the editor drawer.

Currently, any block will lead to the native block state when clicked - this will be changed in a future PR


apps/studio/src/components/PageEditor/MenuBar.tsx
seaerchin, line 1: skimming, assuming you're taking from existing codebase. please dismiss if not

apps/studio/src/components/PageEditor/MenuBar.tsx
```diff
@@ -179,32 +222,32 @@ const MenuBar = ({ editor, schema }: { editor: Editor; schema: any }) => {
...
     // },
     // {
-    //   type: "item",
+    //   type: 'item',
     //   icon: BiImageAdd,
-    //   title: "Add image",
-    //   action: () => showModal("images"),
+    //   title: 'Add image',
+    //   action: () => showModal('images'),
     // },
```
seaerchin, line 251: we're not deleting cos we want to add images in later?

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
```diff
@@ -9,45 +9,48 @@ import {
...
+          onProceed={(componentType) => console.log(componentType)}
+        />
+      )
+    case 'nativeEditor':
+      return (
+        <TipTapComponent
+          data="test"
+          path="test"
+          handleChange={(path: string, data: any) => console.log(path, data)}
+          onProceed={(path: string, data: any) => console.log(path, data)}
```
seaerchin, line 49: nit: add a `TODO`

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
```diff
@@ -9,45 +9,48 @@ import {
...
 }
 
-export const EditPageDrawer = ({
-  isOpen: open,
-  state,
-}: EditPageDrawerProps) => {
-  const [currState, setCurrState] = useState<DrawerState>(state)
+export function EditPageDrawer({ isOpen: open, state }: EditPageDrawerProps) {
+  const { drawerState: currState, setDrawerState: setCurrState } =
+    useEditorDrawerContext()
```
seaerchin, line 25: looks abit off to me, the `state` is at once derived from the parent and the context?

apps/studio/src/features/editing-experience/components/extensions/Table.tsx
seaerchin, line 1: skimming, assuming taken from FE codebase

apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
```diff
@@ -0,0 +1,185 @@
...
+    icon: BiText,
+    title: 'Paragraph',
+  },
+  image: {
+    icon: BiImage,
+    title: 'Image',
+  },
+}
+
+function TipTapComponent({
```
seaerchin, line 53: this looks vaguely problematic, could we try to add/exit out of the tiptap component? my main concern is that we're re-initialising the editor on every `mount/unmount`, which might cause issues.

alexanderleegs: ah you're right, the existing edits disappear if we exit out of the drawer (e.g. if we want to implement minimizing of the entire drawer, then minimizing the drawer and returning will not preserve the state at current)

I think this might be tackled when we link up the preview and the editor, since this is likely something that should be  handled using localstorage/context?

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
alexanderleegs: yeah whoops this was a hack to just get the state changes working, let me move it out

alexanderleegs: Modified upstream in https://github.com/opengovsg/isomer/pull/219

apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
alexanderleegs: Modified to use context upstream in https://github.com/opengovsg/isomer/pull/219

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
alexanderleegs: Removed upstream

apps/studio/src/components/PageEditor/MenuBar.tsx
alexanderleegs: Ah actually that wouldn't be using the tiptap editor, have removed some stuff that will be unused. Have only preserved the links + files which might be done on tiptap in future

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -80,11 +77,19 @@ export default function RootStateDrawer({ blocks }: RootStateDrawerProps) {
                       index={index}
                     >
                       {(provided) => (
-                        <VStack w="100%" gap={0}>
+                        <VStack
+                          w="100%"
+                          gap={0}
+                          // as={Button}
```
seaerchin, line 83: nit: delete

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -80,11 +77,19 @@ export default function RootStateDrawer({ blocks }: RootStateDrawerProps) {
...
+                          onClick={() => {
```
seaerchin, line 84: nbd, i assume this is just to check that it's clicked?

apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
```diff
@@ -0,0 +1,185 @@
...
+      Text,
+      Underline,
+    ],
+    content: data,
+    onUpdate({ editor }) {
+      handleChange(path, editor.getJSON().content)
+    },
+  })
+
+  if (!editor) return <></>
```
seaerchin, line 113: add a todo for loading states please!

seaerchin (approved):

PR #199, opened 2024-06-20
Rudimentary endpoint for getting a folder/site's content

hanpuliu-charles:
## Problem

Closes [ISOM-1194](https://linear.app/ogp/issue/ISOM-1194/create-api-to-return-the-sites-pages), partially. This is just a simple endpoint that allows for testing of the site dashboard page with DB data. It is missing the following things which I have also added to a new ticket 1203(not linking it as linear might pick up on it): 

1. Currently, no permission checking is done.
2. Since there is no decided solution for storing page status, whether it's published or draft, this is not working yet(just returns published for now)
3. Similar to the above, there is no current decided solution for where the last edited information(datetime and user) is stored. Potentially, this could be stored as the diffs between edits, or somewhere in the db. Dummy data is fielded for this at the moment.

Lastly, in the future, since we are treating both folders and pages as resources, this could be integrated with the router for pages.

## Solution

Modifies the current folder router's readFolderOrTopLevelFolder procedure(tRPC).

**Breaking Changes**

  - Details ...

### DIFF
--- a/apps/studio/src/server/modules/folder/folder.router.ts
+++ b/apps/studio/src/server/modules/folder/folder.router.ts
@@ -47,8 +47,9 @@ export const folderRouter = router({
             id: c.id,
             name: c.name,
             type: 'page',
-            lastEditedDate: new Date(),
-            lastEditedUser: 'Coming Soon',
+            lastEditDate: new Date(),
+            lastEditUser: 'Coming Soon',
+            permalink: '/placeholder',
           }
         }
         return {
@@ -57,6 +58,7 @@ export const folderRouter = router({
           type: 'folder',
           lastEditDate: 'folder',
           lastEditUser: 'Coming Soon',
+          permalink: '/placeholder',
         }
       })
 
--- a/apps/studio/src/server/modules/folder/folder.router.ts
+++ b/apps/studio/src/server/modules/folder/folder.router.ts
@@ -1,3 +1,4 @@
+/* eslint-disable import/prefer-default-export */
 import { protectedProcedure, router } from '~/server/trpc'
 import {
   createFolderSchema,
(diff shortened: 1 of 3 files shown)


seaerchin:
heya @hanpuliu-charles - can i trouble you to update the base branch? the branch now contains alot of extra commits at presetn

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -12,13 +13,85 @@ export const folderRouter = router({
...
+        const { parentId } = folderResult
+        const folderName = folderResult.name
+
+        return {
+          folderName,
+          children,
+          parentId,
+        }
+      }
+      const childrenResult = await ctx.db
```
seaerchin, line 65: i'm abit confused here, why would `resourceId` ever be `null` and in that case, shouldn't we throw an error since this implies that they've tried to fetch a non-existing resource (`resourceId` here maps to `Resource.id`, which is the pkey, no?)

seaerchin (changes requested):

apps/studio/src/server/modules/folder/folder.router.ts
hanpuliu-charles: That I believe was for the case where you only have the site id, and want to get everything in the root level 'folder',i.e. orphan pages and folders. From the discussion on the schema if i understood correctly we wanted one endpoint to get both root 'folders' as well as actual folder contents. If that's not the case let me know and I am happy to break it into 2 cases!

kishore03109:
@hanpuliu-charles hihi can rebase on top of main next time ya, easier to see change diff!

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -1,24 +1,57 @@
...
+    .input(readFolderSchema)
+    .query(async ({ ctx, input }) => {
+      // Things that aren't working yet:
+      // 0. Perm checking
+      // 1. Last Edited user and time
+      // 2. Page status(draft, published)
+
+      const folderResult = await ctx.db
+        .selectFrom('Resource')
+        .select(['name', 'parentId'])
```
seaerchin, line 21: why this optimisation now? i don't foresee `selectAll` being significantly more data or taking longer to query - we can probably `selectAll` first, and save ourselves the grief of manually adding properties in later.

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -1,24 +1,57 @@
...
+        return {
+          id: c.id,
+          name: c.name,
+          type: 'folder',
+          lastEditDate: 'folder',
+          lastEditUser: 'Coming Soon',
+          permalink: '/placeholder',
+          status: 'folder',
+        }
+      })
```
seaerchin, line 50: this seems to be slightly off to me - `lastEditDate/status` for folders isn't very useful.

seaerchin (approved):
No issues functionality wise, some incongruences popping up in code that suggests we might wanna relook at our db model

apps/studio/src/server/modules/folder/folder.router.ts
hanpuliu-charles: thanks for pointing this out. realise they can be different

PR #210, opened 2024-06-21
feat(page-router): get single page

seaerchin:
## Problem
This PR updates the router to return a single page. This is currently missing the site wide configuration, together with footer and navbar.

Closes [insert issue #]

## Solution
- add a new service (`resource.service.ts`)
- add methods for querying db into that service


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    .selectAll()
+    .executeTakeFirstOrThrow()
+}
+
+export const getPageById = (id: string) => {
+  return getById(id)
+    .where('blobId', '!=', null)
+    .selectAll()
+    .executeTakeFirstOrThrow()
+}
```
alexanderleegs, line 39: Hmm when would we need to make use of this endpoint? If we're looking for a quick view this would be retrieved from getPages, and the full view is covered by `getFullPageById`

alexanderleegs (approved):
Assuming most of the resource service is just mocks

lgtm, just would like some clarification on where we intend to use `getPageById`

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: this isn't an endpoint - was thinking if we wanted to just return the meta of teh page without content, such as in a folder view

PR #211, opened 2024-06-21
isom 1143 select a layout for page

hanpuliu-charles:
## Problem

Partially closes ISOM-1143, adds a layout that can be used for selecting the page layout.

- Added Preview component, scaling of the component seems to be off
- **Needs a schema for the page to render, either generated(since this is on page creation flow) or passed in as a prop.** 
- **Need to modify the LayoutSelectionProps, or make use of context to pass in details about the page being created.**
- **Highly likely that have to call backend's CREATE for page once layout is selected.**

## Solution

**Breaking Changes**

  - Details ...

**Features**:

Implements the design on Figma, using a 4 column layout(might need to be changed)

## Before & After Screenshots

**AFTER**:

image


apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
+/* eslint-disable import/prefer-default-export */
```
harishv7, line 1: What are these disables for?

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+  HStack,
+  Image,
+  List,
+  ListItem,
+  Text,
+  VStack,
+  useToken,
+} from '@chakra-ui/react'
+import { Button, BxRightArrowAlt } from '@opengovsg/design-system-react'
+import Preview from '../editing-experience/components/Preview'
```
harishv7, line 22: shall we standardise to import from `~` instead? so e.g. `~/components/...`

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+  layoutDisplayName: string
+  layoutSchemaName: string
+  layoutDescription: string
+  imageSource: string
+}[] = [
+  {
+    layoutDisplayName: 'Default',
+    layoutSchemaName: 'content',
+    layoutDescription: 'This is the most basic layout for your content.',
+    imageSource: '/assets/layout_card/default_layout_card.png',
```
harishv7, line 41: can we name asset folders with `-` instead? `layout-card`

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+  {
+    layoutDisplayName: 'Article',
+    layoutSchemaName: 'article',
+    layoutDescription:
+      'Designed for the perfect reading experience. Use this layout for text-heavy content, such as news, press releases, and speeches',
+    imageSource: '/assets/layout_card/article_layout_card.png',
+  },
+]
+
+const placeholderSchema = {
```
harishv7, line 52: can we extract the schemas out into a different file to export these so that the code is clean

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+                            '.layout-title': {
+                              transition: 'color 300ms ease-out',
+                            },
+                            '.hover-text': {
+                              transition: 'opacity 300ms ease-out',
+                              opacity: 0,
+                            },
+                          }
+                    }
+                    // _hover={{
```
harishv7, line 280: Is this commented block required?

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+                    //     transition: 'opacity 300ms ease-out',
+                    //     opacity: 0,
+                    //   },
+                    //   ':hover .hover-text': {
+                    //     opacity: 1,
+                    //   },
+                    // }}
+                  >
+                    <Box
+                      //    className={selectedLayout==e.layoutName?"":"hover-border"}
```
harishv7, line 300: same here

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+                    >
+                      <Image
+                        src={e.imageSource}
+                        borderRadius="4px 4px 0px 0px"
+                        boxShadow="0px 0px 20px 0px rgba(104, 104, 104, 0.30)"
+                      />
+                      {selectedLayout.layoutSchemaName !=
+                        e.layoutSchemaName && (
+                        <Box
+                          // className="hover-text"
```
harishv7, line 321: is this req?

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+              <Preview schema={placeholderSchema} />
+            </Box>
+          </GridItem>
+        </Grid>
+      </VStack>
+    </>
+  )
+}
+
+// const LinksReportDetails = ({ layoutStyle: String }: {}) => {
```
harishv7, line 426: same here

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,438 @@
...
+  useToken,
+} from '@chakra-ui/react'
+import { Button, BxRightArrowAlt } from '@opengovsg/design-system-react'
+import Preview from '../editing-experience/components/Preview'
+
+import { Icon } from '@chakra-ui/react'
+import { BiLeftArrowAlt, BiShow } from 'react-icons/bi'
+
+interface LayoutSelectionProps {
+  pageName: String
```
harishv7, line 28: i think you can introduce types for the different types of layouts we are supporting, can see on figma

hanpuliu-charles: Not sure if this was what you meant, but let me know if you meant something else!

apps/studio/src/pages/testlayoutselector/index.tsx
```diff
@@ -0,0 +1,9 @@
+import LayoutSelection from '../../features/editing-experience/LayoutSelection'
+
+export default function TestLayoutSelector() {
+  return (
+    <>
+      <LayoutSelection pageName="testpagename" />
+    </>
```
seaerchin, line 7: nit: we can just return this directliy wtihout the react fragments (`<>

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+]
+
+function LayoutSelection(props: LayoutSelectionProps): JSX.Element {
+  // need state here to keep track of selected layout
+  const [selectedLayout, setSelectedLayout] = useState(LAYOUT_DATA[0]!)
+  return (
+    <>
+      {/* This is the floating header */}
+      <VStack position="fixed" h="100vh">
+        <VStack p="1.5rem 2rem" w="100vw" gap="0.5rem" alignItems="flex-start">
```
seaerchin, line 54: whenever i see `100vh/100vw`, i'm always abit hesitant - this forces the element to the full size of the viewport and any wrappers of this will overflow. 

does this component require such a strong guarantee? if instead what we're looking for is "full size/width of container", maybe `w="full"` (and similar) might work well here

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+              {`"`}
+            </Text>
+            <Button size="xs">
+              Start with this layout
+              <BxRightArrowAlt fontSize="1.5rem" />
+            </Button>
+          </HStack>
+        </VStack>
+        <Grid
+          w="100vw"
```
seaerchin, line 72: same comment as above!

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+                              opacity: 1,
+                            },
+                            '.hover-border': {
+                              borderColor: useToken(
+                                'colors',
+                                'interaction.main-subtle.hover',
+                              ),
+                            },
+                          }
+                    }
```
seaerchin, line 105: this chunk suffers from a readability issue as well as a locality issue. what i mean is that the changes that we want to apply (`.layout-title` for example) isn't located close to where the changes are required. 

could we locate this where they are required instead?

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+                        : {
+                            '.layout-title': {
+                              transition: 'color 300ms ease-out',
+                            },
+                            '.hover-text': {
+                              transition: 'opacity 300ms ease-out',
+                              opacity: 0,
+                            },
+                          }
+                    }
```
seaerchin, line 118: same issue as above

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+                      position="relative"
+                      borderColor={
+                        selectedLayout.layoutTypename === e.layoutTypename
+                          ? 'base.divider.brand'
+                          : 'base.divider.medium'
+                      }
+                    >
+                      <Image
+                        src={e.imageSource}
+                        borderRadius="4px 4px 0px 0px"
```
seaerchin, line 135: nit: use the shorthand syntax: `4px 0px` (not entirely sure)

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+                        boxShadow="0px 0px 20px 0px rgba(104, 104, 104, 0.30)"
```
seaerchin, line 136: nit: fien with this `rgba(xxx)` since we're rolling this out but if we have a token, please od use the token!

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+                      {selectedLayout.layoutTypename !== e.layoutTypename && (
+                        <Box
+                          className="hover-text"
+                          position="absolute"
+                          top="0"
+                          left="0"
+                          right="0"
+                          bottom="0"
+                          bg="rgba(172, 199, 250, 0.4)"
```
seaerchin, line 148: nit: pelase use the token if available.

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+                      />
+
+                      {selectedLayout.layoutTypename !== e.layoutTypename && (
+                        <Box
+                          className="hover-text"
+                          position="absolute"
+                          top="0"
+                          left="0"
+                          right="0"
+                          bottom="0"
```
seaerchin, line 147: this is abit problematic in relation to the `position: absolute` being specified above.

what absolute does is to specify co-ordinates w regards to the nearest positioned ancestor (some element with `position` specified). in here, we're saying it should be at the top (and yet at the bottom). this is problematic because of that - should we update the problematic lines?

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+            colSpan={3}
+            bg="grey.100"
+            p="2rem 2rem 0 2rem"
+            overflow="auto"
+          >
+            <Box
+              borderRadius="8px 8px 0px 0px"
+              w="100%"
+              minH="100%"
+              boxShadow="0 0 20px 0 rgba(104,104,104,0.3)"
```
seaerchin, line 206: nit: use token

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+                  </VStack>
+                )
+              })}
+            </VStack>
+          </GridItem>
+          <GridItem
+            w="100%"
+            colSpan={3}
+            bg="grey.100"
+            p="2rem 2rem 0 2rem"
```
seaerchin, line 199: nit: shorthand

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+          >
+            <Box
+              borderRadius="8px 8px 0px 0px"
+              w="100%"
+              minH="100%"
+              boxShadow="0 0 20px 0 rgba(104,104,104,0.3)"
+              position="relative"
+            >
+              <HStack
+                borderRadius="8px 8px 0px 0px"
```
seaerchin, line 210: nit: shorthand

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,243 @@
...
+              </HStack>
+              <Preview schema={placeholderContent} />
+
+              <Box
+                position="absolute"
+                top="0"
+                left="0"
+                w="100%"
+                h="100%"
+                zIndex="10"
```
seaerchin, line 234: using `zIndex: number` is scary because it might be positioned behind/over things unexpectedly. perhaps we could do better by using the z-index export from chakra and using a property there.

alternatively, we could consider using `portals` here - it sounds like a perfect use case! see [here](https://v2.chakra-ui.com/community/recipes/z-index#whats-wrong-with-z-index) for more

seaerchin (approved):
we try to use colour tokens where applicable because it allows us to make theme changes across our codebase at one shot and omits the long tail of outdated colours. we should try to stick to that convention here, if possible. if the token is not available, could i trouble you to ping the wonderful @sehyunidaaa and leave a `// TODO` so we remember to get back to this. 

shorthands i'm not very particular but it's just easier to parse. leave it to your discretion

apps/studio/src/features/editing-experience/LayoutSelection.tsx
isomeradmin: @hanpuliu-charles please tag me on figma (using the `comment` tool :)) if there are places with missing tokens!

sehyunidaaa: (oh shucks I was logged in as isomeradmin to access vercel staging. this is Se Hyun)

apps/studio/src/features/editing-experience/LayoutSelection.tsx
hanpuliu-charles: Realised that by placing the div after the preview component, it is rendered correctly above the preview on the z-axis. Is there a need to still specify a z-index export from chakra in this case?
Also, looked into portals, which seems plausible for this, problem is that the div's width and height will have to be rethought(as well as the anchoring using position=absolute), since it is now placed at a different place in the DOM tree, with different parent.

apps/studio/src/features/editing-experience/LayoutSelection.tsx
hanpuliu-charles: eh, seems like it's already the most compact form, unless i'm missing something here.

apps/studio/src/features/editing-experience/LayoutSelection.tsx
hanpuliu-charles: similar to above, seems like most compact form, since for border-radius 2 value shorthand sets diagonally opposite corners.

apps/studio/src/features/editing-experience/LayoutSelection.tsx
hanpuliu-charles: You're right on this, misunderstood how the position anchor works, keeping just top and left

apps/studio/src/features/editing-experience/LayoutSelection.tsx
hanpuliu-charles: Yes, thanks for pointing this out. didn't consider the case where it might overflow, since initially I was approaching the layout as though it is the root component. But would prefer the safer way here

apps/studio/src/features/editing-experience/LayoutSelection.tsx
seaerchin: ok sgtm

apps/studio/src/features/editing-experience/LayoutSelection.tsx
seaerchin: in this case, no! can i trouble you to remove zindex

apps/studio/src/features/editing-experience/LayoutSelection.tsx
alexanderleegs: Hmm would [_groupHover](https://v2.chakra-ui.com/docs/styled-system/style-props#pseudo) work in this scenario?

hanpuliu-charles: Yes, looks like this is the way to do this. replaced with role:group and _groupHover.

apps/studio/src/pages/testlayoutselector/index.tsx
seaerchin, line 1: quick question, no action needed - what's required for us to delete this file?

seaerchin (commented):
code-wise lgtm but some follow-up questions from me: 
1. on trying it locally, the **preview doesn't actually change**. this is problematic because as a user, _i expect to see some change when i click on the button_. this isn't done at present, so the current experience falls short of what's desired and is potentially confusing to users (at least it was to me). can i check if you'll be doing this in a follow-up? also cc @sehyunidaaa what should be the behaviour when button is clicked?
2. we have incorporated a dummy page `/testlayoutselector`, which i greatly appreciate for testing. i'm fine w/ this for just checking the user experience but our designers (@sehyunidaaa @samanthadotdesign) might appreciate a different means of reviewing design changes, such as using chromatic + storybook. i'll let them decide what's good for them, and if this is ok then no objections from me
3. we actually require an entrypoint to the layout selector. as-is, you've done an excellent job here but it's not quite "complete" in the sense that we can't actually use it in any practical manner. should i be looking to you for the follow-up PRs?

apps/studio/src/pages/testlayoutselector/index.tsx
hanpuliu-charles: we can delete this before the pr is merged, or after we have an entrypoint

sehyunidaaa:
hi @seaerchin thanks for raising this! @hanpuliu-charles the layout preview on the right should change when you click on a layout card.

I see a few styling nits, would be helpful to set up chromatic to give comments on the UI

apps/studio/src/features/editing-experience/LayoutSelection.tsx
```diff
@@ -0,0 +1,251 @@
...
+  // need state here to keep track of selected layout
+  const [selectedLayout, setSelectedLayout] = useState(LAYOUT_DATA[0]!)
+  return (
+    <>
+      {/* This is the floating header */}
+      <VStack position="fixed" h="100%">
+        <VStack p="1.5rem 2rem" w="full" gap="0.5rem" alignItems="flex-start">
+          <Button variant="link">
+            <Icon as={BiLeftArrowAlt} />
+            Back To All Pages
```
sehyunidaaa, line 62: `Back to all pages` instead of Title Case please :)

and the link styling seems off but i think this is a design system problem! let me check with the DS team

sehyunidaaa (changes requested):
added 2 comments on styling - will do an in-depth one when we have the visuals!

hanpuliu-charles:
Follow ups: from page management flow, after user clicks create new page, the page creation modal pops up(page title and url). after that, this layout can appear as a modal(either swapping out the original modal content or a new modal), and once user clicks select this layout, make the call to BE to create the page and redirect user to the page editor. Any comments on this flow?

apps/studio/src/features/editing-experience/LayoutSelection.tsx
sehyunidaaa: actually you can use this variant of the link instead of the link variant of the button component!

https://design.open.gov.sg/?path=/story/components-link--variant-standalone

apps/studio/public/assets/layout-card/article_layout_card.png
karrui, line 1: suggest running all of the images through an optimizer for better loading times. https://squoosh.app/ is a quick one

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+// export const createPageSchema = z.object({
+//   pageName: z.string(),
+//   pageTitle: z.string(),
+//   // TODO: add the actual layouts in here
+//   layout: z.enum(PAGE_LAYOUTS).default('content'),
+//   siteId: z.string(),
+//   // NOTE: implies that top level pages are allowed
+//   folderId: z.string().nullable(),
+// })
```
karrui, line 39: remove unused code

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+        <VStack p="1.5rem 2rem" w="full" gap="0.5rem" alignItems="flex-start">
+          <Button variant="link">
+            <Icon as={BiLeftArrowAlt} />
+            Back to all pages
+          </Button>
+          <HStack p="0" justifyContent="space-between" w="100%">
+            <Text textStyle="h4" color="base.content.default">
+              {`Choose a layout for "`}
+              {props.pageName}
+              {`"`}
```
karrui, line 82: nit: for readability

```suggestion
              {`Choose a layout for "${props.pageName}"`}
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                  <VStack
+                    key={layoutEntry.layoutTypename}
+                    p="0"
+                    gap="0"
+                    cursor="pointer"
+                    onClick={() => {
+                      setSelectedLayout(layoutEntry)
+                    }}
+                    role="group"
+                  >
```
karrui, line 110: this should be a `button` tag instead for semantic purposes. admittedly not important for non-public screens but good to be mindful. Suggest making it a button (but not the Button component) haha.

can be replaced like so:
```suggestion
                  <VStack
                    key={layoutEntry.layoutTypename}
                    p="0"
                    gap="0"
                    // cursor="pointer" // Don't need cursor anymore since button inherently has that style
                    onClick={() => {
                      setSelectedLayout(layoutEntry)
                    }}
                    role="group"
                    as="button"   // This is the change~
                  >
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                                'interaction.main-subtle.hover',
+                              ),
+                              bgColor: useToken(
+                                'colors',
+                                'interaction.main-subtle.default',
+                              ),
+                            }
+                          : {}
+                      }
+                    >
```
karrui, line 154: comment(chakra-specific): Might want to style active state directly instead of doing a bunch of conditionals

see: [chakra docs](https://v2.chakra-ui.com/docs/styled-system/style-props#:~:text=none-,_active,-%26%3Aactive%0A%26%5Bdata%2Dactive)

```suggestion
                    <Box
                      className="hover-border"
                      borderRadius="8px"
                      border="2px"
                      p="1.5rem 1.5rem 0 1.5rem"
                      mb="1.25rem"
                      position="relative"
                      data-active={
                        selectedLayout.layoutTypename ===
                        layoutEntry.layoutTypename
                      }
                      _active={{
                        borderColor: 'base.divider.brand',
                        bgColor: 'interaction.muted.main.active',
                        transition:
                          'border-color 300ms ease-out, opacity 300ms ease-out, background-color 300ms ease-out',
                        _groupHover: {
                          bgColor: 'interaction.main-subtle.default',
                        },
                      }}
                      borderColor="base.divider.medium"
                      bgColor="interaction.muted.main.hover"
                      _groupHover={{
                        borderColor: 'interaction.main-subtle.hover',
                      }}
                    >
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                              transition:
+                                'border-color 300ms ease-out, opacity 300ms ease-out, background-color 300ms ease-out',
+                            }
+                          : {}
+                      }
+                      _groupHover={
+                        selectedLayout.layoutTypename !==
+                        layoutEntry.layoutTypename
+                          ? {
+                              borderColor: useToken(
```
karrui, line 143: any reason why `useToken` was used here too? should be able to use the tokens directly. Should also have intellisense

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                                'colors',
+                                'interaction.main-subtle.default',
+                              ),
+                            }
+                          : {}
+                      }
+                    >
+                      <Image
+                        src={layoutEntry.imageSource}
+                        borderRadius="4px 4px 0px 0px"
```
karrui, line 157: nit: might be clearer using

```suggestion
                        borderRightRadius="4px"
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                            opacity="0.6"
+                            alignItems="center"
+                            w="100%"
+                            h="100%"
+                            borderRadius="6px" // due to this being smaller than the outer Box
+                            bgColor="interaction.main-subtle.default"
+                            sx={{
+                              transition: 'opacity 300ms ease-out',
+                              opacity: 0,
+                            }}
```
karrui, line 180: might not need to mix `sx` and props, slightly affects readability

```suggestion
                            transition="opacity 300ms ease-out"
                            opacity={0}
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                            _groupHover={{
+                              opacity: 0.6,
+                            }}
+                          />
+                          <VStack
+                            sx={{
+                              transition: 'opacity 300ms ease-out',
+                              opacity: 0,
+                            }}
```
karrui, line 189: might not need to mix `sx` and props, slightly affects readability

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                            className="hover-text"
```
karrui, line 190: is className needed here? are y'all mixing css and chakra? this will result in tech debt oh!

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                        boxShadow="md"
+                        alt={`Image of ${layoutEntry.layoutDisplayName} layout`}
+                      />
+
+                      {selectedLayout.layoutTypename !==
+                        layoutEntry.layoutTypename && (
+                        <>
+                          {/* for opacity, since we need the text to be clear */}
+                          <Box
+                            className="hover-blur"
```
karrui, line 167: is className needed here? are y'all mixing css and chakra? this will result in tech debt oh!

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                            bgColor="interaction.main-subtle.default"
+                            sx={{
+                              transition: 'opacity 300ms ease-out',
+                              opacity: 0,
+                            }}
+                            _groupHover={{
+                              opacity: 0.6,
+                            }}
+                          />
+                          <VStack
```
karrui, line 185: if this is a button, it should be using `button` tag. 

[semantic html](https://web.dev/learn/html/semantic-html)

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+                        textAlign="left"
+                        textStyle="h6"
+                        textColor={
+                          selectedLayout.layoutTypename ===
+                          layoutEntry.layoutTypename
+                            ? 'base.content.brand'
+                            : 'base.content.strong'
+                        }
+                        mb="0.5rem"
+                        sx={{ transition: 'color 300ms ease-out' }}
```
karrui, line 222: can just use the prop directly

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+          </GridItem>
+          <GridItem
+            w="100%"
+            colSpan={3}
+            bg="grey.100"
+            p="2rem 2rem 0 2rem"
+            overflow="auto"
+          >
+            <Box
+              borderRadius="8px 8px 0px 0px"
```
karrui, line 249: nit: are designers using radii token? might be good to start so there are less things to hardcode haha. otherwise the [default chakra radii](https://v2.chakra-ui.com/docs/styled-system/theme#border-radius) might be good enough.

```suggestion
              borderRightRadius="lg"
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,281 @@
...
+              w="100%"
+              minH="100%"
+              boxShadow="md"
+              position="relative"
+            >
+              <HStack
+                borderRadius="8px 8px 0px 0px"
+                w="100%"
+                bgColor="slate.200"
```
karrui, line 258: comment: where are these colors coming from? I do not see any theme generated here. Using default chakra colors? Highly suggest using design system tokens or it'll be a PITA in the future to update colors.

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
seaerchin: @sehyunidaaa @samanthadotdesign

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -102,59 +92,45 @@ function LayoutSelection(props: LayoutSelectionProps): JSX.Element {
...
-                              bgColor: useToken(
-                                'colors',
-                                'interaction.main-subtle.default',
-                              ),
-                            }
-                          : {}
-                      }
+                          ? true
+                          : null
+                      } // have to use null so that React will not parse this jsx attribute. having the [data-active] selection attribute will trigger _active.
```
karrui, line 115: ah, I have a handy function to alleviate this, might want to use it here too:

```ts
// utils somewhere
/**
 * Function to return undefined if the value is falsy, otherwise true
 */
export const dataAttr = (value: unknown) => (!!value ? true : undefined);

// Usage:
data-active={dataAttr(selectedLayout.layoutTypename === layoutEntry.layoutTypename)}
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -49,15 +40,15 @@ const LAYOUT_DATA: {
     layoutDisplayName: 'Default',
     layoutTypename: 'content',
     layoutDescription: 'This is the most basic layout for your content.',
-    imageSource: '/assets/layout-card/default_layout_card.png',
+    imageSource: '/assets/layout-card/default_layout_card.webp',
```
karrui, line 43: webp is a newer file type ah, i hope gsibs can load 😆. should be can la according to https://caniuse.com/webp

dcshzj: Checked, GSIBs can load webp! All good

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
seaerchin: for context - we aim to refer by tokens because it provides a centralised place for our theme. the benefit is that when we change 1 colour in our theme, our whole site will be updated accordingly. 

in here, we've used an **inbuilt palette** (noticed how it's suffixed with `.200`). this isn't meant to be used to styled components and when we change our own theme (things like: `bg.focused.strong`), we'll miss out on this `HStack`!

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+export interface LayoutSelectionProps {
+  pageName: string
+  // preparing to make the trpc call to create page. required props
+  pageUrl: string
+  siteId: string
+  folderId: string
+}
+
+const LAYOUT_DATA: {
```
seaerchin, line 32: nit: omit the typing and do a `as const` if this is not mean tot change

seaerchin (commented):
blockers from me: 
1. we should always aim to use tokens from our theme (`base.content.strong` etc) and **not palettes** (`black.200`). a way to think about it is that palettes are meant to provide shades of colours, which our theme then use. for example, we muight want to set our header colour as `black.200` but the token exported for engineering use will be sometihng like `base.content.header`. in the future, if this were to change (eg: darker shade of black), the designer will update the **token but not the palette**
2. we should aim to use semantic html whenever possible. trouble you to read [this](https://web.dev/learn/html/semantic-html) for an explanation as to why.

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
seaerchin: can simplify the assertion down below too if we do this

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,250 @@
...
+                      <Text
+                        className="layout-title"
+                        textAlign="left"
+                        textStyle="h6"
+                        textColor={
+                          selectedLayout.layoutTypename ===
+                          layoutEntry.layoutTypename
+                            ? 'base.content.brand'
+                            : 'base.content.strong'
+                        }
```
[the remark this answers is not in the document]
hanpuliu-charles: yep, good to be consistent with the other components.

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,250 @@
...
+                return (
+                  <VStack
+                    key={layoutEntry.layoutTypename}
+                    p="0"
+                    gap="0"
+                    onClick={() => {
+                      setSelectedLayout(layoutEntry)
+                    }}
+                    role="group"
+                    as="button"
```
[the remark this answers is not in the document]
karrui: `as` is fine here, it's a [ChakraUI thing](https://v2.chakra-ui.com/docs/styled-system/style-props#the-as-prop), which transforms the tag into "button", keeping its semanticness.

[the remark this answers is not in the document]
seaerchin: omg ok TIL

apps/studio/prisma/generated/generatedEnums.ts
```diff
@@ -1,6 +1,6 @@
 export const RoleType = {
-  Admin: "Admin",
-  Editor: "Editor",
-  Publisher: "Publisher",
-} as const
-export type RoleType = (typeof RoleType)[keyof typeof RoleType]
+    Admin: "Admin",
+    Editor: "Editor",
+    Publisher: "Publisher"
+} as const;
```
karrui, line 5: comment: seems like the formatter changed in this whole PR, or somehow your vscode instance got different linter from the rest. I'll review ignoring whitespace, but should be fixed by my future PR when I consolidate the linting and config for the monorepo

apps/studio/src/features/editing-experience/components/LayoutSelection.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Default scenario
+export const Default = Template.bind({})
+Default.args = {
+  pageName: 'Sample Page',
+  pageUrl: '/sample-page',
+  siteId: '1',
+  folderId: '1',
+}
```
karrui, line 25: comment: Might want to use CSFv3 to write stories going forward. (I'll update the rest of the stories in this repo in a future PR)


```suggestion
import { Meta, StoryObj } from '@storybook/react'
import LayoutSelection from './LayoutSelection'

const meta: Meta<typeof LayoutSelection> = {
  title: 'Components/LayoutSelection',
  component: LayoutSelection,
}
export default meta;
type Story = StoryObj<typeof LayoutSelection>

export const Default: Story = {
  args : {
    pageName: 'Sample Page',
    pageUrl: '/sample-page',
    siteId: '1',
    folderId: '1',
    }
  }
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,245 @@
+// Component for layout selection
+
+import React, { useState } from 'react'
```
karrui, line 3: nit: unnecessary React import
```suggestion
import { useState } from 'react'
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,245 @@
...
+    layoutDisplayName: 'Article',
+    layoutTypename: 'article',
+    layoutDescription:
+      'Designed for the perfect reading experience. Use this layout for text-heavy content, such as news, press releases, and speeches',
+    imageSource: '/assets/layout-card/article_layout_card.webp',
+    previewJson: articleLayoutPreview,
+  },
+] as const
+
+const dataAttr = (value: unknown) => (!!value ? true : undefined) // Keeping this here for now, might shift out to a utils file when necessary.
```
karrui, line 48: can just shift to utils file now la haha, confirm use somewhere else in the future one

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,245 @@
...
+// TODO: Make this headless by getting the LAYOUT_DATA from the schema. Find somewhere in the schema for layoutDescription & image(fetch this too or generate it)
+function LayoutSelection(props: LayoutSelectionProps): JSX.Element {
+  const [selectedLayout, setSelectedLayout] = useState<LayoutDataType>(
+    LAYOUT_DATA[0]!,
+  )
+
+  return (
+    <>
+      {/* This is the floating header */}
+      <VStack position="fixed" h="100%">
```
karrui, line 59: nit: instead of using fixed, may be better to use `position="sticky"` (and top=0) instead. Then the grid below will not need to have a `calc(100% - 7rem)` prop. Of course, the parent of this needs to be relative.

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,245 @@
...
+                      border="2px"
+                      p="1.5rem 1.5rem 0 1.5rem"
+                      mb="1.25rem"
+                      position="relative"
+                      borderColor="base.divider.medium"
+                      bgColor="interaction.muted.main.hover"
+                      data-active={dataAttr(
+                        selectedLayout.layoutTypename ===
+                          layoutEntry.layoutTypename,
+                      )} // have to use null so that React will not parse this jsx attribute. having the [data-active] selection attribute will trigger _active.
```
karrui, line 107: nit: unnecessary comment

```suggestion
                      )}
```

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,245 @@
...
+                borderTopRadius="8px"
+                w="100%"
+                bgColor="base.canvas.light"
+                textColor="white"
+                justifyContent="center"
+                gap="0"
+                padding="0.5rem 0.75rem"
+              >
+                <Text textStyle="caption-2">
+                  {`You're previewing the`}&nbsp;
```
karrui, line 225: nit: unnecessary templating

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
```diff
@@ -0,0 +1,245 @@
...
+                </Text>
+                <Text textStyle="caption-1">
+                  {selectedLayout.layoutDisplayName}
+                  {` Layout`}
```
karrui, line 229: nit: unnecessary templating, or if you want you can consolidate them

```suggestion
                  {`${selectedLayout.layoutDisplayName} Layout`}
```

apps/studio/src/theme/foundations/colours.ts
```diff
@@ -0,0 +1,7 @@
+export const colours = {
+  base: {
+    canvas: {
+      light: '#babecb',
```
karrui, line 4: sudden addition of color tokens? can you[ generate isomer's theme](https://github.com/opengovsg/design-system/tree/main/token-gen) from figma tokens instead?
or does not exist yet?

karrui (approved):
lgtm

apps/studio/src/theme/foundations/colours.ts
seaerchin: the colour tokens doesn't exist yet - we're just piggy backing off default theme. i added this so that we can use colour tokens now but when we put our actual tokens inside, it'll be correct also

apps/studio/public/assets/layout-card/article_layout_card.webp
harishv7, line 1: Is webp supported on all modern browsers, esp GSIBs? @dcshzj can do a quick check?

dcshzj: Yup checked, they are viewable on GSIBs.

apps/studio/src/features/editing-experience/components/LayoutSelection.tsx
hanpuliu-charles: Tried to use position="sticky" but can't seem to get it to work with the independent overflow scrolling as I had difficulty forcing the outer container height to be 100%, as it tends to expand to fit contents resulting in non independent scrolling.

seaerchin: i think since this is a nit and you had difficulty getting it to work, feel free to resolve this and merge this PR in! @hanpuliu-charles

PR #212, opened 2024-06-21
chore: combine test editor pages

alexanderleegs:
Removes the `/editor` page and moves preview into `/dashboard/edit` instead
Screenshot 2024-06-21 at 3 46 11 PM

Note that the preview itself doesn't show up properly due to scaling issues, possibly due to not shifting to container queries yet - this should be tackled with a fix to `isomer-components`


harishv7 (approved):
This is just a movement of the pages/code right? If so, lgtm

PR #220, opened 2024-06-24
chore: migrate schema

seaerchin:
## Problem
see [here](https://opengovproducts.slack.com/archives/C06R4DX966P/p1719212071025839)

## Solution
- shift out the table's properties to a jsonb
    - not reusing the existing `Blob` table because the mental model is that db table map to to a specific isomer schema


alexanderleegs (approved):
lgtm

PR #222, opened 2024-06-24
feat(page): update page 

seaerchin:
## Problem
need to implement update endpoint for pages.

Closes ISOM-1094

## Solution
Simple endpoint that is all or nothing to update the page. We wrap the update in a transaction to prevent concurrent updates. since we will only operate on the `page` itself, we only lock a single row

## Notes
- Feels like we need an interface for perms at this point
- hard to test cos no visible UI


apps/studio/src/schemas/page.ts
```diff
@@ -3,7 +3,7 @@ import { z } from 'zod'
 const PAGE_LAYOUTS = ['content'] as const
 
 export const getEditPageSchema = z.object({
-  pageId: z.string(),
+  id: z.string(),
```
kishore03109, line 6: eh, prefer `pageId`, esp when there is another id present

apps/studio/src/schemas/page.ts
```diff
@@ -12,8 +12,8 @@ export const updatePageSchema = getEditPageSchema.extend({
   // in which case this is a no-op.
   // We are ok w/ this because it doesn't
   // incur any db writes
-  parentId: z.string().nullable(),
-  pageName: z.string().nullable(),
+  parentId: z.string().optional(),
+  name: z.string().optional(),
```
kishore03109, line 16: hmm i thought following from the slack thread, there is going to be a different api call for this? no ah?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -40,26 +42,32 @@ export const pageRouter = router({
   updatePage: protectedProcedure
     .input(updatePageSchema)
     .query(async ({ input, ctx }) => {
-      const parentId = ''
-      const pageName = ''
+      // NOTE: Need to do validation like checking for existence of the page
+      // and whether the user has write-access to said page
```
kishore03109, line 46: hmm, why put this comment here ah? 
This seems like code that belongs in the middleware?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -40,26 +42,32 @@ export const pageRouter = router({
...
     .input(updatePageBlobSchema)
     .query(async ({ input, ctx }) => {
+      // NOTE: Need to do validation like checking for existence of the page
+      // and whether the user has write-access to said page
+      await updateBlobById(input)
+
       // NOTE: Not returning the `content` first because
       // 1. it might potentially be huge
       // 2. frontend already knows
       // ahead of time what is the content (it sent the content over)
```
kishore03109, line 67: am ok with returning the input (might be huge, but i think for one agency dont have to think too much about perf for now) 
just remove this no longer relevant comment can already

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -39,6 +45,34 @@ export const getPageById = (id: string) => {
     .executeTakeFirstOrThrow()
 }
 
+export const updatePageById = (
+  page: Partial<Omit<Page, 'id' | 'blobId'>> & { id: string },
```
kishore03109, line 49: hmm, why do we omit the `id` here, only to add `& { id: string }` back again ah?

apps/studio/src/server/modules/page/page.router.ts
seaerchin: i'm open to extending the `protectedProcedure` and putting this there, wdyt

apps/studio/src/schemas/page.ts
seaerchin: that was for folders - the change made as that previously `id` (pkey in db) was optional but now it's not. 

in here, `id` is required

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: `Partial` means that the `id` will be set to optional. we don't want this behaviour so we do a union to let the id be required. another option considered but discarded was having an extra arg called `id: string`. discarded because most callers (atm 1) of `updatePageById` probably already have a object - easier to consume in this way.

apps/studio/src/schemas/page.ts
seaerchin: sure

apps/studio/src/server/modules/page/page.router.ts
seaerchin: done

apps/studio/src/server/modules/page/page.router.ts
seaerchin: shure

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -9,58 +9,54 @@ import {
...
     .input(updatePageBlobSchema)
     .query(async ({ input, ctx }) => {
-      // NOTE: Not returning the `content` first because
-      // 1. it might potentially be huge
-      // 2. frontend already knows
-      // ahead of time what is the content (it sent the content over)
-      return {}
+      await updateBlobById({ ...input, id: input.pageId })
+
+      return input
```
hanpuliu-charles, line 56: will it result in unnecessary network activities, or is there a use case that requires the input back

seaerchin: actually i prefer not returning but change was made due to comment [here](https://github.com/opengovsg/isomer/pull/222#discussion_r1651182789). 

i don't have any real concerns now, i think we will find out when pages start getting big

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -39,6 +44,34 @@ export const getPageById = (id: string) => {
...
+) => {
+  const { id, ...rest } = page
+  return db.transaction().execute((tx) => {
+    return tx
+      .updateTable('Resource')
```
harishv7, line 53: why are we not using constants for the table names?

seaerchin: not doing this because kysely infers typings automatically (see ss below), so in this case, the table is already given as a suggestion. additionally, giving the wrong type (`resource` instead of `Resource`) will be an error.

the value of giving companion objects is 2 fold
- when the argument type is `string` but we want to say "only this subset of strings: `A | B | C`" and avoid errrors
- when we want to also infer a type 

in our case here, neither of this is true because kysely provides typing + auto-completion so i didn't see a big reason to do it + take on maint debt 

image

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -39,6 +44,34 @@ export const getPageById = (id: number) => {
     .executeTakeFirstOrThrow()
 }
 
+export const updatePageById = (
+  page: Partial<Omit<Page, 'id' | 'blobId'>> & { id: string },
+) => {
+  const { id, ...rest } = page
+  return db.transaction().execute((tx) => {
```
hanpuliu-charles, line 51: Just curious, is there a reaction why we wrap what appears to be a single sql statement in a transaction?

seaerchin: defensive programming here - idt the `tx` will do anything different (apart from db sql actually calling `BEGIN TRANSACTION`) but i think having here helps us to avoid potential errors down the line so hapy to keep it

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -9,58 +9,54 @@ import {
   getFooter,
   getFullPageById,
   getNavBar,
+  updateBlobById,
+  updatePageById,
 } from '../resource/resource.service'
 import { getSiteConfig } from '../site/site.service'
 
+// NOTE: Need to do validation like checking for existence of the page
```
dcshzj, line 17: nit: maybe we should differentiate a note from a todo? Seems like they are used interchangeably here but might not be helpful in the future to know exactly what still needs to be done

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -39,6 +44,34 @@ export const getPageById = (id: number) => {
...
+
+export const updateBlobById = (props: UpdateBlobProps) => {
+  const { id, content } = props
+  return db.transaction().execute((tx) => {
+    return (
+      tx
+        .updateTable('Blob')
+        .innerJoin('Resource', 'Resource.id', 'id')
+        // NOTE: This works because a page has a 1-1 relation with a blob
+        .set({ content })
```
dcshzj, line 68: question(blocking): hmm this feels weird, looks like it is only saving a JSON blob with only the `content` key. Should we just directly store the object instead, and not wrap it inside a `content` key?

This means the DB stores the blob content like this:
```
{
  "version": "0.1.0",
  "layout": "article",
  ...
}
```

Instead of:
```
{
  "content": {
    "version": "0.1.0",
    "layout": "article",
    ...
  }
}
```

apps/studio/src/server/modules/page/page.router.ts
seaerchin: good catch, trouble you to make comments like this a non-nit!

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: need your take on this - i'm not quite sure what content implies, i'm taking reference from `placeholder` (see `Preview.tsx`) 

i did this initially because the `placeholder.json` has more properties beyond just `content` (iirc), so in the future if we want to save those properties can just add here

dcshzj: Ah yeah I meant like we should be saving all the props exactly like how `placeholder.json` looks like inside the DB. Right now I'm not sure:

1. What is inside the `content` key (is it `IsomerPageSchema` or `IsomerPageSchema['content']`
2. Are we storing `layout`, `version`, etc. somewhere else or something

seaerchin: ah ok, it's `IsomerPageSchema` - this is confusing because we've chosen `content` to represent our blobs (but at the same time, the page schema itself has another key called `content`)

dcshzj: Can ignore, clarified offline. The `content` here refers to the `content` column inside the DB table.

dcshzj (approved):

PR #224, opened 2024-06-24
feat(db): seed db

seaerchin:
## Problem
At present, we have some type mismatches and also, we have implemented methods to get data from db but no actual data + connection yet. 

## Solution
Seed the db with data

### DIFF
--- a/apps/studio/prisma/generated/generatedTypes.ts
+++ b/apps/studio/prisma/generated/generatedTypes.ts
@@ -13,7 +13,6 @@ export type Blob = {
 }
 export type Footer = {
   id: string
-  name: string
   siteId: string
   content: unknown
 }
--- a/apps/studio/src/schemas/site.ts
+++ b/apps/studio/src/schemas/site.ts
@@ -1,5 +1,5 @@
 import { z } from 'zod'
 
 export const getConfigSchema = z.object({
-  id: z.string(),
+  id: z.string().min(1),
 })
--- a/apps/studio/src/schemas/site.ts
+++ b/apps/studio/src/schemas/site.ts
@@ -1,0 +1,5 @@
+import { z } from 'zod'
+
+export const getConfigSchema = z.object({
+  id: z.string(),
+})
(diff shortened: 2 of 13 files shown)


apps/studio/prisma/generated/generatedTypes.ts
```diff
@@ -13,7 +13,6 @@ export type Blob = {
 }
 export type Footer = {
   id: string
-  name: string
```
kishore03109, line 16: why are we dropping this ah?

apps/studio/src/schemas/site.ts
```diff
@@ -0,0 +1,5 @@
+import { z } from 'zod'
+
+export const getConfigSchema = z.object({
+  id: z.string(),
```
kishore03109, line 4:
```suggestion
  id: z.string().min(1),
```

apps/studio/src/server/modules/site/site.types.ts
```diff
@@ -1,11 +1,11 @@
 export interface Sitemap {
   parentTitle: string
   childrenTitles: string[]
-  sibilngTitles: string[]
+  siblingTitles: string[]
 }
-
+type SiteTheme = 'isomer-classic' | 'isomer-next'
```
kishore03109, line 6: nit: obj this so that in code can just call `SiteTheme.classic`

kishore03109 (approved):

apps/studio/prisma/generated/generatedTypes.ts
seaerchin: this is an extra prop. i referenced the footer props in config: 

```ts

{
  LinkComponent?: any
  isGovernment?: boolean
  siteName: string
  agencyName: string
  lastUpdated: string
}
```

apps/studio/src/schemas/site.ts
seaerchin: updated

apps/studio/src/server/modules/site/site.types.ts
seaerchin: sure

PR #229, opened 2024-06-25
fix: schema

seaerchin:
## Problem
Original schema specifies `cuid`, which is problematic on 2 fronts
- first, cuid is a randomly generated id which is difficult for our users to remember 
- next, cuid is not generated in db (prisma uses js to autogen). this leads to typing on kysely output being `String` instead of `GeneratedAlways<String>`.

## Solution
- change to auto-incrementing ids
    - note that now, the output of kysely types is `GeneratedAlways`
- update migration


apps/studio/src/server/modules/resource/resource.types.ts
```diff
@@ -1,5 +1,9 @@
-export interface NavbarItem {
-  name: string
+import { type IsomerSiteProps } from '@opengovsg/isomer-components'
+
+export type Navbar = { items: IsomerSiteProps['navBarItems'] }
+
+export interface FooterItem {
+  title: string
```
seaerchin, line 6: this is cherry-picked, downstream will refer correctly to `IsomerSiteProps`

apps/studio/prisma/schema.prisma
```diff
@@ -1,4 +1,4 @@
-// This is your Prisma schema file,
+// This is your Prisma schema file,/
```
dcshzj, line 1: nit: extra slash

apps/studio/src/server/modules/resource/resource.types.ts
```diff
@@ -1,5 +1,9 @@
-export interface NavbarItem {
-  name: string
+import { type IsomerSiteProps } from '@opengovsg/isomer-components'
+
+export type Navbar = { items: IsomerSiteProps['navBarItems'] }
+
+export interface FooterItem {
```
dcshzj, line 5: nit: I suspect most of the types that we need on the CMS should be able to be derived from the exported props one way or another, so we might not need to re-define them here.

apps/studio/prisma/seed.ts
```diff
@@ -86,13 +86,8 @@ async function main() {
...
     .values({
-      id: '1',
-      siteId: '1',
-      content: {
-        name: 'navi',
-        url: 'www.isomer.gov.sg',
-        items: NAV_BAR_ITEMS,
-      } satisfies Navbar,
+      siteId: id,
+      content: { items: NAV_BAR_ITEMS } satisfies Navbar,
```
dcshzj, line 90: question(non-blocking): are we able to type `NAV_BAR_ITEMS` to `Navbar` directly when we define it above in line 14?

dcshzj (approved):
No blocking comments

apps/studio/src/server/modules/resource/resource.types.ts
seaerchin: exported for ease of use cos it's very tedious to type `Isomer...['...']` every time ._.

apps/studio/prisma/schema.prisma
seaerchin: idk how that got there but thanks for the catch

apps/studio/prisma/seed.ts
seaerchin: added on otp, still kept check here

apps/studio/prisma/generated/generatedTypes.ts
```diff
@@ -8,43 +8,43 @@ export type Timestamp = ColumnType<Date, Date | string, Date | string>
...
-  resourceId: string
-  userId: string
+  id: GeneratedAlways<number>
+  resourceId: number
+  userId: number
   role: RoleType
 }
 export type Resource = {
-  id: string
+  id: GeneratedAlways<number>
```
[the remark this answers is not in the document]
seaerchin: yes, they've been updated here

PR #231, opened 2024-06-25
feat(preview): connect preview to db

seaerchin:
## Problem
Previously, our data was hardcoded on the frontend. now, we will dynamically get it out from our backend and feed it into the `Preview`

## Solution
- refer to all ids as `number` 
- update seed script to seed db with some mock data 
- retrieve data from db using trpc

### DIFF
--- a/apps/studio/prisma/seed.ts
+++ b/apps/studio/prisma/seed.ts
@@ -85,7 +85,11 @@ async function main() {
       name: 'Ministry of Trade and Industry',
       config: {
         theme: 'isomer-next',
-        sitemap: {
+        siteName: 'MTI',
+        logoUrl: '',
+        search: undefined,
+        lastUpdated: '',
+        siteMap: {
           title: 'Home',
           permalink: '/',
           children: [],
--- a/apps/studio/prisma/seed.ts
+++ b/apps/studio/prisma/seed.ts
@@ -8,6 +8,7 @@ import {
   type Navbar,
   type Footer,
 } from '~/server/modules/resource/resource.types'
+import { type IsomerSitemap } from '@opengovsg/isomer-components'
 import { db } from '../src/server/modules/database'
 
 const NAV_BAR_ITEMS: Navbar['items'] = [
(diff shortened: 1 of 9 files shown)


apps/studio/prisma/seed.ts
```diff
@@ -58,10 +86,13 @@ async function main() {
...
-          childrenTitles: [],
-          parentTitle: '',
-        },
+          title: 'Home',
+          permalink: '/',
+          children: [],
+          layout: 'content',
+          summary: 'something',
+          lastModified: '',
+        } satisfies IsomerSitemap,
```
seaerchin, line 95: NOTE: actually this should be dynamically generated. this is omitted at present.

apps/studio/src/features/editing-experience/components/Preview.tsx
```diff
@@ -13,22 +12,32 @@ export interface PreviewProps {
...
+        theme,
         logoUrl: 'https://www.isomer.gov.sg/images/isomer-logo.svg',
-        isGovernment: true,
+        isGovernment,
         environment: 'production',
         lastUpdated: '3 Apr 2024',
-        navBarItems: navBar,
-        // @ts-expect-error blah
+        navBarItems: navbar.items,
         footerItems: footer,
```
seaerchin, line 41: @dcshzj note the difference in calling `navbar/footer` - i think might make more sense to update exported type

apps/studio/prisma/seed.ts
dcshzj: Yeah hmm actually we shouldn't be storing the sitemap inside the DB since it should be derived when rendering the preview, but yeah we can remove this in a subsequent PR!

apps/studio/src/server/modules/site/site.types.ts
```diff
@@ -1,18 +1,12 @@
...
 
-export const SiteThemes = {
+export const SiteThemes: Record<string, IsomerSiteProps['theme']> = {
   classic: 'isomer-classic',
   next: 'isomer-next',
 } as const
 
-type SiteTheme = (typeof SiteThemes)[keyof typeof SiteThemes]
-
 export interface SiteConfig {
```
dcshzj, line 8: suggestion(blocking): I think `SiteConfig` should derive its types directly from `IsomerSiteProps` using `Omit`. Otherwise it feels like we are duplicating the types definitions.

seaerchin: fixed downstream
/

seaerchin:
tests failing on type error, fixing downstream

dcshzj (approved):

apps/studio/src/server/modules/site/site.types.ts
seaerchin: per offline discussion - `SiteConfig` is site-wide config + footer/navbar. as such, just omitting the two should be ok

PR #232, opened 2024-06-25
feat(page): connect page to db

seaerchin:
## Problem
At present, our page data is still hard coded. We want to retrieve it dynamically from our database.

## Solution
- add mock data for page + users to seed script so that it can be auto-ran
- bugfix for type-inference on `readPageAndBlob` (`await` promise)
- tighten `id` to `Resource.id` in our base query for resources -> required because on `innerjoin`, the other tables also have `id` so the reference to `id` is ambiguous

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,9 @@
         "apps/*",
         "packages/*"
       ],
+      "dependencies": {
+        "sass": "^1.77.6"
+      },
       "devDependencies": {
         "turbo": "^1.13.3"
       }
--- a/package.json
+++ b/package.json
@@ -22,5 +22,8 @@
   "license": "ISC",
   "devDependencies": {
     "turbo": "^1.13.3"
+  },
+  "dependencies": {
+    "sass": "^1.77.6"
   }
 }


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -21,13 +21,14 @@ export const getFolders = () =>
...
-const getById = (id: number) => db.selectFrom('Resource').where('id', '=', id)
+const getById = (id: number) =>
+  db.selectFrom('Resource').where('Resource.id', '=', id)
 
 // NOTE: Throw here to fail early if our invariant that a page has a `blobId` is violated
 export const getFullPageById = (id: number) => {
   return getById(id)
-    .where('blobId', '!=', null)
-    .innerJoin('Blob', 'blobId', 'Blob.id')
+    .where('Resource.blobId', 'is not', null)
```
seaerchin, line 30: bugfix - somehow `!=` doesn't actually get non-nullcols

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -21,13 +21,14 @@ export const getFolders = () =>
     .execute()
 
 // NOTE: Base method for retrieving a resource - no distinction made on whether `blobId` exists
-const getById = (id: number) => db.selectFrom('Resource').where('id', '=', id)
+const getById = (id: number) =>
+  db.selectFrom('Resource').where('Resource.id', '=', id)
```
seaerchin, line 25: bugfix - inner joins have 2 id cols so we need to be explicit which id we're referring to

PR #235, opened 2024-06-26
feat(dnd): add dnd to preview

seaerchin:
## Problem
we want to have drag and drop functionality to re-arrange blocks easily

Closes ISOM-1170.

## Solution
- at present, there are a few stateful variables (multiple calls of `useState` for managing old/new editor state) 
- to have a single centralised place to manage them, i've streamlined them into `editorState/setEditorState` from the `useEditorDrawerContext`
- then, i hooked up this state to the `onDragEnd` call for dnd. 
- the feedback loop is then:
    - i drag something and drop
    - on drag end triggers
    - new state is derived post-drag
    - this new state is saved into `editorState`
    - new state is fed into `Preview`, which triggers a re-render

## Notes
- i'm not quite sure what the `old/newEditorState` is (i've removed both of them) 
- not quite sure what `isCopied` etc is too

## Screenshots/video

https://github.com/opengovsg/isomer/assets/44049504/a3c5d543-c12d-402f-8b7e-0718b08c7969

### DIFF
--- a/apps/studio/src/pages/dashboard/edit/index.tsx
+++ b/apps/studio/src/pages/dashboard/edit/index.tsx
@@ -10,71 +10,9 @@ import { trpc } from '~/utils/trpc'
 
 const ISOMER_SCHEMA_URI = 'https://schema.isomer.gov.sg/next/0.1.0.json'
 
-const placeholder = {
-  version: '0.1.0',
-  layout: 'homepage',
-  page: {
-    title: 'Home',
-  },
-  content: [
-    {
-      type: 'hero',
-      variant: 'gradient',
-      alignment: 'left',
-      backgroundColor: 'black',
-      title: 'Ministry of Trade and Industry',
-      subtitle:
-        'A leading global city of enterprise and talent, a vibrant nation of innovation and opportunity',
-      buttonLabel: 'Main CTA',
-      buttonUrl: '/',
-      secondaryButtonLabel: 'Sub CTA',
-      secondaryButtonUrl: '/',
-      backgroundUrl: 'https://ohno.isomer.gov.sg/images/hero-banner.png',
-    },
-    {
-      type: 'infobar',
-      title: 'This is an infobar',
-      description: 'This is the description that goes into the Infobar section',
-    },
-    {
-      type: 'infopic',
-      title: 'This is an infopic',
-      description: 'This is the description for the infopic component',
-      imageSrc: 'https://placehold.co/600x400',
-    },
-    {
-      type: 'keystatistics',
-      statistics: [
-        {
-          label: 'Average all nighters pulled in a typical calendar month',
-          value: '3',
-        },
-        {
-          label: 'Growth in tasks assigned Q4 2024 (YoY)',
-          value: '+12.2%',
-        },
-        {
-          label: 'Creative blocks met per single evening',
-          value: '89',
-        },
-        {
-          value: '4.0',
-          label: 'Number of lies in this stat block',
-        },
-      ],
-      variant: 'top',
-      title: 'Irrationality in numbers',
-    },
-  ],
-}
-
 const EditPage: NextPageWithLayout = () => {
-  const [, setEditedSchema] = useState<any>(placeholder)
-  const [, setIsJSONValid] = useState(true)
-
-  const [isCopied, setIsCopied] = useState(false)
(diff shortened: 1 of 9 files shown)


apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -118,7 +136,7 @@ const EditPage: NextPageWithLayout = () => {
   }, [isCopied])
 
   const handleEditorChange = (value: any) => {
-    setEditorValue(value)
+    setEditorState(value)
     localStorage.setItem('editorValue', value)
```
dcshzj, line 140: issue: we should remove this local storage as well, since we are pre-loading the page with the contents from the DB now. This is a playground-specific code since playground doesn't have a DB, and was added to prevent any accidental data loss.

apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -69,18 +70,35 @@ const placeholder = {
...
-  const [editorValue, setEditorValue] = useState(
-    JSON.stringify(placeholder, null, 2),
-  )
-  const [newEditorValue, setNewEditorValue] = useState({})
   const [editedSchema, setEditedSchema] = useState<any>(placeholder)
   const [isJSONValid, setIsJSONValid] = useState(true)
   const [isNewEditor, setIsNewEditor] = useState(false)
-  const [isCopied, setIsCopied] = useState(false)
 
+  const [isCopied, setIsCopied] = useState(false)
```
dcshzj, line 77: nit: we can also do away with this copying functionality, it's playground-specific code.

apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -195,7 +205,13 @@ const EditPage: NextPageWithLayout = () => {
       </GridItem>
       {/* TODO: Implement preview */}
       <GridItem colSpan={2} overflow="scroll">
-        <Preview schema={editedSchema} />
+        <Preview
+          schema={{
+            content: pageState,
+            version: page.version,
```
dcshzj, line 211: note: out of scope of this PR but just FYI, the version here should take the version of the schema being used, not the existing version of the page. This is because we should try to "upgrade" the page as much as possible and prevent the editor from saving if the contents of the page violates a future version of the schema.

apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -69,18 +70,35 @@ const placeholder = {
...
+  const {
+    drawerState,
+    setDrawerState,
+    pageState,
+    setPageState,
+    editorState,
+    setEditorState,
+  } = useEditorDrawerContext()
+
+  const [{ theme, isGovernment, sitemap, name }] =
```
dcshzj, line 90: suggestion(non-blocking): maybe consider just passing the props as a single object rather than destructuring here? Otherwise, we will have to make code-level changes if the schema changes.

I think there are some concerns with prop spreading so happy to discuss.

apps/studio/src/pages/dashboard/edit/index.tsx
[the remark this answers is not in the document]
seaerchin: not inclined to - would rather remove in its entirety. see slack thread [here](https://opengovproducts.slack.com/archives/C06R4DX966P/p1719395805334579) as to why

seaerchin: agreed

apps/studio/src/pages/dashboard/edit/index.tsx
seaerchin: agreed

apps/studio/src/pages/dashboard/edit/index.tsx
seaerchin: swapped to props pread

apps/studio/src/pages/dashboard/edit/index.tsx
seaerchin: agree wth you, this is abit annoying to do at present cos of how the query itself is structured. lets do it in a follow-up PR since for now, there's only 1 versoin ._.

seaerchin: added a todo to track this

dcshzj (approved):

PR #236, opened 2024-06-26
feat(page): connect page to db

seaerchin:
## Problem
At present, our page data is still hard coded. We want to retrieve it dynamically from our database.

## Solution
- add mock data for page + users to seed script so that it can be auto-ran
- bugfix for type-inference on `readPageAndBlob` (`await` promise)
- tighten `id` to `Resource.id` in our base query for resources -> required because on `innerjoin`, the other tables also have `id` so the reference to `id` is ambiguous


apps/studio/prisma/seed.ts
```diff
@@ -14,6 +14,76 @@ import {
...
+  'jan',
+  'kishore',
+  'jiachin',
+  'sehyun',
+  'harish',
+  'zhongjun',
+  'hanpu',
+]
+
+const PAGE_BLOB = {
```
dcshzj, line 30: nit: this is quite tied to the schema so might want to consider having this in a central location so we know to update it when the schema is adjusted.

dcshzj (approved):
Left [the same comment](https://github.com/opengovsg/isomer/pull/222/files/3c4bf995e601e0aa8e7a14e12ec0a45aadfd8f93#r1658244361) on the `blob` DB, we can discuss there. No blocking comments.

apps/studio/prisma/seed.ts
seaerchin: i left it here because this is the initial `seed` script ran to populate your db with fake data. 

is there another place you were thinking of?

PR #238, opened 2024-06-26
feat: breakdown site-wide props types

dcshzj:
## Problem

The producer (components) does not differentiate the site-wide props that are generated or user-defined.

## Solution

**Breaking Changes**

**Improvements**:

- The `IsomerSiteProps` are broken down into types that are generated and user-defined, so that consumers know how to differentiate between them.


seaerchin (approved):
lgtm

PR #243, opened 2024-06-28
feat: introduce JSONForms nested object renderer

dcshzj:
## Problem

We do not have a nested object renderer, which makes us unable to render the hero dropdown fields (since it is a nested object within each of the hero variants).

## Solution

**Breaking Changes**

**Features**:

- Added an object renderer to render nested objects specifically. ([Reference here](https://github.com/eclipsesource/jsonforms/blob/master/packages/material-renderers/src/complex/MaterialObjectRenderer.tsx))

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-06-28 at 10 23 57

**AFTER**:

Screenshot 2024-06-28 at 10 21 57

### DIFF
--- a/apps/studio/src/constants/formBuilder.ts
+++ b/apps/studio/src/constants/formBuilder.ts
@@ -4,6 +4,7 @@ export const JSON_FORMS_RANKING = {
   DropdownControl: 2,
   IntegerControl: 4,
   TextControl: 1,
+  ObjectControl: 2,
   OneOfControl: 3,
   ProseControl: 2,
   RadioControl: 3,
--- a/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
@@ -13,6 +13,8 @@ import {
   jsonFormsDropdownControlTester,
   JsonFormsIntegerControl,
   jsonFormsIntegerControlTester,
+  JsonFormsObjectControl,
+  jsonFormsObjectControlTester,
   JsonFormsOneOfControl,
   jsonFormsOneOfControlTester,
   JsonFormsProseControl,
@@ -26,6 +28,7 @@ import {
 } from './renderers'
 
 const renderers: JsonFormsRendererRegistryEntry[] = [
+  { tester: jsonFormsObjectControlTester, renderer: JsonFormsObjectControl },
   { tester: jsonFormsArrayControlTester, renderer: JsonFormsArrayControl },
   { tester: jsonFormsBooleanControlTester, renderer: JsonFormsBooleanControl },
   {
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
@@ -1,0 +1,61 @@
+import {
+  Generate,
+  findUISchema,
+  isObjectControl,
+  rankWith,
+  type RankedTester,
+  type StatePropsOfControlWithDetail,
+} from '@jsonforms/core'
+import { JsonFormsDispatch, withJsonFormsDetailProps } from '@jsonforms/react'
+import { useMemo } from 'react'
+import { JSON_FORMS_RANKING } from '~/constants/formBuilder'
+
+export const jsonFormsObjectControlTester: RankedTester = rankWith(
+  JSON_FORMS_RANKING.ObjectControl,
+  isObjectControl,
+)
+
+export function JsonFormsObjectControl({
+  path,
+  visible,
+  renderers,
+  cells,
+  schema,
+  enabled,
+  uischema,
+  uischemas,
+  rootSchema,
+}: StatePropsOfControlWithDetail) {
+  const detailUiSchema = useMemo(
+    () =>
+      findUISchema(
+        uischemas || [],
+        schema,
+        uischema.scope,
+        path,
+        () =>
+          Generate.uiSchema(schema, 'VerticalLayout', undefined, rootSchema),
+        uischema,
+        rootSchema,
+      ),
+    [uischemas, schema, uischema, path, rootSchema],
+  )
+
+  if (!visible) {
+    return null
+  }
+
+  return (
+    <JsonFormsDispatch
+      visible={visible}
+      enabled={enabled}
+      schema={schema}
+      uischema={detailUiSchema}
+      path={path}
+      renderers={renderers}
+      cells={cells}
+    />
+  )
+}
+
+export default withJsonFormsDetailProps(JsonFormsObjectControl)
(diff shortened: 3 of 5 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/layouts/JsonFormsVerticalLayout.tsx
```diff
@@ -26,8 +26,9 @@ export function JsonFormsVerticalLayoutRenderer({
 
   return (
     <VStack spacing={2}>
-      {elements.map((element) => (
-        <Box key={path} w="100%">
+      {elements.map((element, index) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Box key={`${path}-${index}`} w="100%">
```
hanpuliu-charles, line 31: Is adding the index due to everything within a nested object sharing the same path? and adding the index isn't a problem because we aren't changing the array's ordering or contents

dcshzj: Yeah mainly because we are not going to be re-ordering the form fields, since it is constant as defined by our Isomer Schema.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export function JsonFormsObjectControl({
+  path,
+  visible,
+  renderers,
+  cells,
+  schema,
+  enabled,
+  uischema,
+  uischemas,
```
seaerchin, line 26: why got 2 schemas

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
```diff
@@ -0,0 +1,61 @@
...
+      findUISchema(
+        uischemas || [],
+        schema,
+        uischema.scope,
+        path,
+        () =>
+          Generate.uiSchema(schema, 'VerticalLayout', undefined, rootSchema),
+        uischema,
+        rootSchema,
+      ),
```
seaerchin, line 40: minor nit because i understand this is same as the docs but in here, we're saying that we want the caller to make the determination of the UI schema but i thought actually that should be the responsibility of the component?

not too important to fix since i suspect we'll only have 1 call site

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
dcshzj: The first one is the specific UI schema defined for the particular element that it is rendering. The second one refers to the registry of all UI schema elements (similar to the control renderers registry we define in FormBuilder.tsx).

We don't actually use this, the code is mostly adapted from their material renderer and the `findUISchema` method signature requires it. Basically what this does is that it tries to generate a sub-UISchema within the context of the object that it is operating on, which is then dispatched and handled by JSONForms to render the elements within this object.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
dcshzj: Hmm the UI schema is entirely generated from the original JSON schema, so the responsibility lies within the caller to structure the JSON schema in such a way that influences how the form should look like (e.g. radio vs dropdown, stuff like that).

apps/studio/src/constants/formBuilder.ts
```diff
@@ -4,6 +4,7 @@ export const JSON_FORMS_RANKING = {
   DropdownControl: 2,
   IntegerControl: 4,
   TextControl: 1,
+  ObjectControl: 2,
```
harishv7, line 7: What's this ranking for?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
harishv7: > specific UI schema defined for the particular element that it is rendering

we don't define ui schemas for elements right?

apps/studio/src/features/editing-experience/components/form-builder/renderers/layouts/JsonFormsVerticalLayout.tsx
harishv7: Here the path refers to the path to the element in the schema e.g. `"#/components/native/paragraph"`?

harishv7: Can't we just use one of the `element` properties to craft the key?

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
```diff
@@ -26,6 +28,7 @@ import {
 } from './renderers'
 
 const renderers: JsonFormsRendererRegistryEntry[] = [
+  { tester: jsonFormsObjectControlTester, renderer: JsonFormsObjectControl },
```
harishv7, line 31: Will this be used by all components that define a object?

it's actually very specifc to objects which contain string-type text fileds with limited allowable values (enums) right?

apps/studio/src/constants/formBuilder.ts
dcshzj: It's a JSONForms concept, mainly to determine the order in which the renderers are chosen.

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
dcshzj: Hmm no, it's really all kinds of objects (i.e. with `type` set to `object`). The actual rendering of individual property fields are done via `JsonFormsDispatch`.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
dcshzj: Nope, but JSONForms generates one automatically, which is then used to control the rendering of the form fields and the layouts.

apps/studio/src/features/editing-experience/components/form-builder/renderers/layouts/JsonFormsVerticalLayout.tsx
dcshzj: Hmm no not exactly, it is the path within the context of the parent control element. For instance, when rendering the Hero component, the path is just an empty string `""`, because we are rendering within the context of that specific component. When there are nested objects (like the dropdown within the hero component), then the path would be `dropdown`, relative to the parent.

Example when I print out the path for the hero component just above line 29:

Screenshot 2024-07-02 at 15 26 10

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
seaerchin: gotcha, thanks!

seaerchin (approved):
changes make sense to me, not very sure if @harishv7 has any blocking comments if yes, please dismiss

PR #248, opened 2024-07-01
chore: fix ci deploy

seaerchin:
## Problem
We have a few type errors that prevents us from deploying. This PR fixes those errors. 

## Solution
revert `User.id` from an auto-incrementing `int` to a string `cuid`. prisma actually does a js call to generate this cuid so it isn't autogenerated. also, the existing `auth` helper code doesn't seem to assume this to exist so i'm assuming they used this (`cuid`) for a reason 

update the type inference for set state cos we set it as `null`, which causes ts to infer it as `null`

## Notes
- we had build issues with `trpc/next/ts` etc and to fix this, we locked `ts@5.2.2` and `next@14.2.3`
- other stuff tried: upgrading `trpc`. removing `openApi`, upgrading `next` (in various combinations) but only this works.
- the errors faced were typing related on `trpc` + `openApi`. oddly enough, using `next@14.2.4` seems to have issues too :sadge:


apps/studio/prisma/generated/generatedTypes.ts
```diff
@@ -24,7 +24,7 @@ export type Navbar = {
 export type Permission = {
   id: GeneratedAlways<number>
   resourceId: number
-  userId: number
+  userId: string
```
seaerchin, line 27: required because we reverted from using bigint `id` to using `cuid`. 

this change was made because the `starter-kit` auth helper seem to have been written with nullable `ids`

apps/studio/prisma/seed.ts
```diff
@@ -216,6 +217,7 @@ async function main() {
       return db
         .insertInto('User')
         .values({
+          id: cuid2.createId(),
```
seaerchin, line 220: `cuid` is generated by prisma using js code but `kysely` sees the raw db type - which doesn't guarantee this. 

hence, i have to use `cuid2` to generate it myself on seed (essentially what prisma is doing AHHAHA)

apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -12,7 +13,7 @@ const ISOMER_SCHEMA_URI = 'https://schema.isomer.gov.sg/next/0.1.0.json'
 
 const EditPage: NextPageWithLayout = () => {
   const [, setJsonSchema] = useState(null)
-  const [validate, setValidate] = useState(null)
+  const [validate, setValidate] = useState<ValidateFunction | null>(null)
```
seaerchin, line 16: wrong typing - auto inference to `null`

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -53,7 +53,7 @@ export const updatePageById = (
   })
 }
 
-export const updateBlobById = (props: { id: number; content: Page }) => {
+export const updateBlobById = (props: { id: number; content: string }) => {
```
seaerchin, line 56: changed because `content` (of the blob) should be opaque - this is string because we treat it as json

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -34,7 +38,7 @@ export const pageRouter = router({
         pageName,
         navbar,
         footer,
-        content,
+        content: content as IsomerPageSchema,
```
seaerchin, line 41: sadly a required cast

dcshzj (approved):

PR #254, opened 2024-07-02
fix: add content type header middleware

alexanderleegs:
This PR fixes an issue with Cross-Site Request Forgery (CSRF) protection. While actions that change the application’s state require JSON data in the POST body, the request's content type is not validated. This allows an attacker to craft a cross-site request that the server would accept, leading to a state change in the application.

Currently, because the session cookies do not have the SameSite attribute explicitly set to None, the browser prevents the exploitation of this CSRF vulnerability. However, an XSS on the subdomain could allow the exploitation of this vulnerability.

The fix to this issue is to allow only application/json as a valid content type, to mitigate the risk of CSRF.

### DIFF
--- a/apps/studio/src/server/trpc.ts
+++ b/apps/studio/src/server/trpc.ts
@@ -8,17 +8,17 @@
  * @see https://trpc.io/docs/v10/procedures
  */
 
-import { initTRPC, TRPCError } from "@trpc/server"
-import superjson from "superjson"
-import { ZodError } from "zod"
+import { initTRPC, TRPCError } from '@trpc/server'
+import superjson from 'superjson'
+import { ZodError } from 'zod'
 
-import { APP_VERSION_HEADER_KEY } from "~/constants/version"
-import { env } from "~/env.mjs"
-import { createBaseLogger } from "~/lib/logger"
-import getIP from "~/utils/getClientIp"
-import { type Context } from "./context"
-import { defaultMeSelect } from "./modules/me/me.select"
-import { prisma } from "./prisma"
+import { APP_VERSION_HEADER_KEY } from '~/constants/version'
+import { env } from '~/env.mjs'
+import { createBaseLogger } from '~/lib/logger'
+import getIP from '~/utils/getClientIp'
+import { type Context } from './context'
+import { defaultMeSelect } from './modules/me/me.select'
+import { prisma } from './prisma'
 
 const t = initTRPC.context<Context>().create({
   /**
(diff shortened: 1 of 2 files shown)


apps/studio/src/server/trpc.ts
```diff
@@ -11,9 +11,9 @@
 import superjson from 'superjson'
 import { ZodError } from 'zod'
 import { TRPCError, initTRPC } from '@trpc/server'
+import { type OpenApiMeta } from 'trpc-openapi'
```
harishv7, line 14: understand this is related to the vapt finding of the content header, could i trouble you to add PR description for context for engs?

harishv7 (approved):
otherwise lgtm

apps/studio/src/server/trpc.ts
```diff
@@ -96,6 +96,19 @@ const loggerWithVersionMiddleware = loggerMiddleware.unstable_pipe(
   },
 )
 
+const contentTypeHeaderMiddleware = t.middleware(async ({ ctx, next }) => {
+  if (
+    ctx.req.body &&
+    !ctx.req.headers['content-type']?.startsWith('application/json')
```
seaerchin, line 102: if this says `application/jsononsense` i'll gladly accept it though?

seaerchin: also this is a very narrow `content-type` to accept - how did we make this determination that we'd only accept this?

apps/studio/src/server/trpc.ts
alexanderleegs: Updated!

apps/studio/src/server/trpc.ts
alexanderleegs: This is based on the VAPT finding on KS! For any requests with a body, we always expect a JSON format - this is also the solution provided upstream by starter-kit (as well as for trpc 11)

harishv7: trpc only works with json, so we can't do other things like form-encoded etc.

harishv7: @alexanderleegs But to that earlier point should it be a strict equality check for application/json or a prefix check?

alexanderleegs: Hmm i think either works since there's no other official MIME type beginning with application/json, but i've modified to be a strict equality check in https://github.com/opengovsg/isomer/pull/254/commits/f40692df65e569e5ec912a68dc78821fce8df7b4

harishv7: lgtm

PR #255, opened 2024-07-02
feat: use trpc to fetch Isomer JSON schema

dcshzj:
## Problem

We are fetching the JSON schema outside of trpc context.

## Solution

**Breaking Changes**

**Features**:

- Switch to use trpc to fetch the Isomer JSON schema.
- Also adjusted the exported types of the components package so that we are not dependent on the existence of the schema within studio.
- The schema has also been moved into the components package.

**Notes**:

- We are still duplicating the schema in `docs/` and in the components package but eventually we will remove the version in `docs/` and use a symbolic link. Not doing that right now because [that requires us to convert the GitHub Pages build process into GitHub Actions](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#troubleshooting-publishing-from-a-branch) so that they can support symbolic links.


apps/studio/src/constants/formBuilder.ts
```diff
@@ -11,3 +11,5 @@ export const JSON_FORMS_RANKING = {
   GroupLayoutRenderer: 1,
   VerticalLayoutRenderer: 1,
 }
+
+export const ISOMER_SCHEMA_URI = 'https://schema.isomer.gov.sg/next/0.1.0.json'
```
hanpuliu-charles, line 15: hi! Is this going to be eventually replaced by some way to fetch the latest version of the schema, without hardcoding of the schema version?

seaerchin: i think hard-coding this is ok for now, but if the change is easy to make, we should aim

since the page itself has the version, how do we feel about fetching it at runtime? i'm also ok w/ the build artefact including the various schemas since we can just store it on the backend as a blob.

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
```diff
@@ -53,32 +56,39 @@ const renderers: JsonFormsRendererRegistryEntry[] = [
...
-  const schema = {
-    ...rest,
-    properties: props,
-    components: IsomerSchema.components,
-  }
-
+  const [jsonSchema, setJsonSchema] = useState<IsomerJsonSchema | null>(null)
   const [formData, setFormData] = useState({})
 
+  trpc.page.getIsomerJsonSchema.useQuery(undefined, {
```
seaerchin, line 68: this kind of seems like a code smell to me tbh. this is because we're calling it and using `setState` in a callback of `onSuccess` which:
1. seems very convoluted
2. why not just use the return directly? 

should we be using `useSuspenseQuery` here instead? it returns `[data, query]` directly and allows us to act like the query returns a value instead of a promise. (see [here](https://tanstack.com/query/v4/docs/framework/react/guides/suspense))

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
```diff
@@ -53,32 +56,39 @@ const renderers: JsonFormsRendererRegistryEntry[] = [
...
+    onSuccess({ schema }) {
+      const { properties, ...rest } = schema.components.complex[component]
+      const { type: _, ...props } = properties
```
seaerchin, line 71: not an issue with this pr, but `type` is a keyword so we'll have alot of renaming down the line. easier to solve it by just renaming the property to `kind` or `variant` or smth

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -40,6 +41,13 @@ export const pageRouter = router({
       }
     }),
 
+  getIsomerJsonSchema: publicProcedure.query(async () => {
+    const schema = (await fetch(ISOMER_SCHEMA_URI).then((response) =>
+      response.json(),
+    )) as IsomerJsonSchema
+    return { schema }
+  }),
```
seaerchin, line 49: recently i've been introduced to a library called `ky` - it seems to be the correct library to use here. this is because: 
1. `fetch` doesn't treat 4xx/5xx as errors 
2. `fetch` isn't actually stable (requires a polyfill) 
3. `fetch` doesn't retry so if the problem is ephemeral, this request will die anyway. 

alternatively, we could instead just store the json as a build artifact on backend , which is also ok to me

seaerchin (commented):
hard-coding is problematic but not a big issue for now since our schema is unlikely to change anytime soon. 

however, we might want to relook at:
1. using `useQuery` rather than `useSuspenseQuery` 
2. whether we want to `fetch` at runtime instead of just compiling it as a build artifact

apps/studio/src/constants/formBuilder.ts
harishv7: we can solve the different versions issue when we arrive there

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
harishv7, line 68: is it possible to cache this schema somehow? at least for majority of CMS users, seems like we fetch the same schema again and again

apps/studio/src/constants/formBuilder.ts
dcshzj: We have decided to bundle the schema with the components package, so removed in 71a18021a67d22a779bab79620dc28904bb1c7b1.

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
dcshzj: Moved to the EditorDrawerContext in 71a18021a67d22a779bab79620dc28904bb1c7b1. This fetch will happen once when the user first enters an editing page and I think will be cached automatically by react-query.

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
dcshzj: Removed in 71a18021a67d22a779bab79620dc28904bb1c7b1.

apps/studio/src/server/modules/page/page.router.ts
dcshzj: We resolved to move the schema into the components package, done in 71a18021a67d22a779bab79620dc28904bb1c7b1.

apps/studio/src/types/schema.ts
```diff
@@ -4,14 +4,24 @@ import {
   type UISchemaElement,
   type VerticalLayout,
 } from '@jsonforms/core'
+import { type IsomerComplexComponentProps } from '@opengovsg/isomer-components'
 
-export type IsomerJsonSchema = JsonSchema & {
+export type IsomerExtendedJsonSchema = JsonSchema & {
   groups?: Array<{
```
harishv7, line 10: double checking my understanding here: JsonSchema is the schema definition as given by JsonForms lib. We are extending it to add a new key for `groups`

apps/studio/src/types/schema.ts
```diff
@@ -4,14 +4,24 @@ import {
...
     label: string
     fields: string[]
   }>
 }
 
+export type IsomerJsonSchema = IsomerExtendedJsonSchema & {
+  components: {
+    complex: Record<
```
harishv7, line 18: how about native? Why do we only define selectively?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -1,11 +1,12 @@
-import { protectedProcedure, router } from '~/server/trpc'
+import { schema, type IsomerPageSchema } from '@opengovsg/isomer-components'
```
harishv7, line 1: schema should now be exported from the isomer-components package right? In this PR I still see it within docs folder

apps/studio/src/types/schema.ts
dcshzj: Added native in f0f0d5e486b4f7899c3e54e2d3b9753bbee4a956. Opted not to define more than what we really need because that would start to defeat the purpose of having the schema outside of studio.

Eventually we will solve this with TypeBox when we are able to export the same types and schema from the components package.

apps/studio/src/types/schema.ts
dcshzj: Yup, codifying it here so that we are aware what are the things that were enhanced on top of the JSON Schema specifications.

apps/studio/src/server/modules/page/page.router.ts
dcshzj: Yup, there are some issues with using symbolic links in GitHub Pages. Not doing that right now because [that requires us to convert the GitHub Pages build process into GitHub Actions](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#troubleshooting-publishing-from-a-branch) so that they can support symbolic links.

harishv7: Why do we need symbolic links? Instead of the schema being in the docs, we just move it to the components package and serve that right?

harishv7 (approved):
lgtm, 1 minor qn, can merge if @karrui has no other comments also

apps/studio/src/server/modules/page/page.router.ts
dcshzj: Yeah but the `docs` folder is a special one that GitHub Pages uses to build the site, so the file has to be present in that folder for it to appear in schema.isomer.gov.sg.

We are keeping that schema because that's being used right now by POEs for their migration work, since Studio is not available for them to use yet.

dcshzj:
Going to merge this first since anyway we have some adjustments downstream.

PR #261, opened 2024-07-03
For feedback: create page flow in dashboard

hanpuliu-charles:
## Problem

Starts on ability for users to create a page in the desired folder, with a 2 step process(1. input page name and permalink, 2. choose a layout).
This PR is mainly to get some feedback before handing off the work!

Tried to kept most of the page creation logic within the LayoutSelection component(where it takes in props for the page name and permalink), my rationale being that it would make it easier to reuse this component if we wanted to shift or change the page creation flow.


apps/studio/src/features/dashboard/DashboardTable.tsx
```diff
@@ -25,59 +26,57 @@ import {
...
+  )
 
-  const [pageNumber, onPageChange] = useState(1)
-  const [dataToDisplay, setDataToDisplay] = useState(dummyChildData)
+  useEffect(() => {
+    if (data) {
+      console.log(data)
+      setDataToDisplay(data.children)
+    }
+=  }, [data])
```
karrui, line 68: [you might not need an effect](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state)

apps/studio/src/features/editing-experience/components/PageCreateModal.tsx
```diff
@@ -67,11 +68,18 @@ export const PageCreateModal = ({
     setPageUrlLen(watchAllFields.url?.length || 0)
   }, [watchAllFields])
```
karrui, line 69: for learning/out of scope of this PR, but same comment above. you do not need an effect, can just derive the variable. 

```tsx
const titleLen = watchedFields.title?.length ?? 0
const pageUrlLen = watchedFields.url?.length ?? 0
```

apps/studio/src/features/editing-experience/components/PageCreateModal.tsx
```diff
@@ -67,11 +68,18 @@ export const PageCreateModal = ({
...
+  })
+
+  // This might need to go into the layout component? or we keep state in this component then pass selectedlayout and setselectedlayout into the component
+  // Prefer to keep page creation states within the layout selection modal? More portable in that case if we want to shift the entry point.
+  const onSubmit: SubmitHandler<PageCreateFormFields> = (data) => {
+    setSubmittedData(data)
+  }
 
-  return (
+  let content = (
```
karrui, line 82: slight antipattern, just return directly based on state instead of assigning to a variable.

```tsx
if (submittedData.title) {
  return <xxx>
}

return <yyy>
```

apps/studio/src/features/editing-experience/components/PageCreateModal.tsx
```diff
@@ -82,7 +90,7 @@ export const PageCreateModal = ({
         <ModalCloseButton />
         <form onSubmit={handleSubmit(onSubmit)}>
           <ModalBody>
-            <Stack gap={'1.5em'}>
+            <Stack gap="1.5em">
```
karrui, line 93: nit: try to use `rem` over `em`, ems will take the sizing of the parent instead of the root. This may result in weird values.

apps/studio/src/pages/[siteId]/dashboard/[resourceId]/index.tsx
```diff
@@ -3,7 +3,7 @@ import { Button } from '@opengovsg/design-system-react'
 import { HStack, VStack, Text } from '@chakra-ui/react'
 import { type NextPageWithLayout } from '~/lib/types'
 import _ from 'lodash'
```
Comment (karrui) [inline apps/studio/src/pages/[siteId]/dashboard/[resourceId]/index.tsx:5]: nit: out of scope of this PR, but try to avoid using full lodash import -- it bloats the bundle size (because treeshaking won't work on this).

prefer importing only what you need, and even then via its direct export instead of the barrel export.
aka

```ts
import { mapValues } from 'lodash' // 🚫 
import mapValues from 'lodash/mapValues' // ✅
```

PR #271, opened 2024-07-04
ISOM-1237: Allow edits to be saved as drafts

harishv7:
## Problem

During publishing flow, we need to support 2 modes of a resource: a main blob, a draft blob, and a state of if the resource is published or not.

We also only have an implicit way of knowing if a resource is a folder i.e. the absence of both mainBlob and draftBlob.

Closes ISOM-1237

## Solution

**Note this PR requires DB migration.**

Renames the current `blobId` field to be `mainBlobId` and introduces a new field called `draftBlobId`, both nullable columns.

It also adds a `isPublished` boolean to track if the resource is published. 

It adds a `isFolder` boolean to track if a resource is a folder. This should be a one time flag and hence we should not need to worry about frequent flag management.

We should return draft blobs preferentially to Studio to show the latest edits made by users. In the absence of a draft, we return the published version.

**Breaking Changes**

- [X] Yes - this PR contains breaking changes
  - Details ...


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -2,21 +2,29 @@ import { db } from '../database'
...
 
 export const getPages = () => {
+  // TODO: write a test to verify this query behaviour
   return (
     db
       .selectFrom('Resource')
-      // TODO: write a test to verify this query behaviour
-      .where('blobId', '!=', null)
+      // fetch pages where either a draft exists or a published blob exists
+      .where((eb) =>
```
seaerchin, line 10: could we make this more descriptive? not too clear what `eb` is at first glance

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -2,21 +2,29 @@ import { db } from '../database'
...
 
 export const getFolders = () =>
+  // TODO: write a test to verify this query behaviour
   db
     .selectFrom('Resource')
-    // TODO: write a test to verify this query behaviour
-    .where('blobId', '=', null)
+    // fetch pages where both draft and published blobs are null
+    .where('publishedBlobId', 'is', null)
+    .where('draftBlobId', 'is', null)
```
seaerchin, line 27: i think we might want to check this behaviour here!

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -25,17 +33,27 @@ const getById = (id: number) =>
...
+  if (draftBlob) return draftBlob
+
   return getById(id)
-    .where('Resource.blobId', 'is not', null)
-    .innerJoin('Blob', 'Resource.blobId', 'Blob.id')
+    .where('Resource.publishedBlobId', 'is not', null)
+    .innerJoin('Blob', 'Resource.publishedBlobId', 'Blob.id')
     .selectAll()
     .executeTakeFirstOrThrow()
 }
```
seaerchin, line 50: actually i think we can do this in a single query (and we should). this is because we're doing an `innerJoin` here on our largest table. 

i'm ok w/ accepting this now if it's difficult to do but concretely i think this entails: 
1. omitting the `where draftBlobId is not null` clause
2. return draft blob if exists, else published blob.

also, on further examination, will we ever have a case where we have `draftBlob` **but not** `publishedBlob`? since we likely just copy over, i don't really think this is a case right?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -25,17 +33,27 @@ const getById = (id: number) =>
...
 
 export const getPageById = (id: number) => {
   return getById(id)
-    .where('blobId', '!=', null)
+    .where((eb) =>
+      eb.or([eb('publishedBlobId', '!=', null), eb('draftBlobId', '!=', null)]),
```
seaerchin, line 55: again, could we make this `eb` parameter more descriptive?

seaerchin (commented):
some comments re: the choice of abstraction here
1. we need to update `draft` -> `published` **post-build**. this is kinda problematic because builds can fail for any number of reasons, and if we update erroneously, we won't have the correct state in our db (which we assume to be the sot) 
2. we only ever have 1 draft/published. if we're planning to extend this feature for auditing purposes, this might potentially be problematic 

just throwing out ideas here - how amenable are we to just saving the build artifact on s3? the conceptual model seems simpler here

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -25,17 +33,27 @@ const getById = (id: number) =>
   db.selectFrom('Resource').where('Resource.id', '=', id)
 
 // NOTE: Throw here to fail early if our invariant that a page has a `blobId` is violated
-export const getFullPageById = (id: number) => {
+export const getFullPageById = async (id: number) => {
+  // Check if draft blob exists and return that preferentially
+  const draftBlob = await getById(id)
+    .where('Resource.draftBlobId', 'is not', null)
+    .innerJoin('Blob', 'Resource.draftBlobId', 'Blob.id')
+    .selectAll()
```
karrui, line 41: we should not return everything back to the frontend in case we return more data than necessary. should we select specific columns that the frontend needs to return?

can even extract into something like `defaultDraftSelect`

apps/studio/src/server/modules/resource/resource.service.ts
karrui: might not need to do in a single query if it is hard to read, but suggest using a transaction instead.

karrui (commented):
basically a published state vs a draft state yeah? Feels weird to have the distinction on the resource level. Should the draft state not be on the Site level instead? I think that makes more sense to me. Or even a different table entirely


```
Site:
  status: 'published' | 'draft'  // this tracks whether the site is draft or not
  lastUpdated: Date // This allows tracking of the "version" of the draft / publish
  resources: Resource[] // should have no change
```

edit: every draft should be a new row in the site table. If draft becomes published, then all you need to do is set the status to published. the most up to date published draft follows the "lastUpdated" column maybe?

actually might require an RFC leh, edge cases on draft, versioning, conflict resolution will need handling, and this PR does not cover those.

harishv7:
> basically a published state vs a draft state yeah? Feels weird to have the distinction on the resource level. Should the draft state not be on the Site level instead? I think that makes more sense to me. Or even a different table entirely
> 
> ```
> Site:
>   status: 'published' | 'draft'  // this tracks whether the site is draft or not
>   lastUpdated: Date // This allows tracking of the "version" of the draft / publish
>   resources: Resource[] // should have no change
> ```
> 
> edit: every draft should be a new row in the site table. If draft becomes published, then all you need to do is set the status to published. the most up to date published draft follows the "lastUpdated" column maybe?
> 
> actually might require an RFC leh, edge cases on draft, versioning, conflict resolution will need handling, and this PR does not cover those.

@karrui @seaerchin Different people can edit different parts of the site and publish independently of the other. As such, at any point in time, multiple resources (pages) will have unpublished changes on them.

The way we are publishing is on a page-by-page mode currently - we do not publish on a site level. Due to this, we need to track if there is a draft for a current page the user is accessing.

If there is a draft, we return that to the CMS/Studio. Else, we fetch the published version and return that. If any edits are made we create the draft resource.

Upon publish of the page, we need to move the draft of the page into published state and then kickstart the CodeBuild. Even if CodeBuild fails, we can retry the builds as the DB's state clearly should show what is published and what is in draft state.

apps/studio/src/server/modules/resource/resource.service.ts
harishv7: extracted, currently is all the columns but we can change this later

apps/studio/prisma/schema.prisma
```diff
@@ -32,13 +32,20 @@ model Resource {
...
 
   permission Permission[]
+
+  mainBlob   Blob? @relation("MainBlob", fields: [mainBlobId], references: [id])
+  mainBlobId Int?  @unique
+
+  draftBlob   Blob? @relation("DraftBlob", fields: [draftBlobId], references: [id])
+  draftBlobId Int?  @unique
+
+  isPublished Boolean? @default(false)
```
karrui, line 47: nit: really dislike boolean flag explosion, but CLOSE EYE HERE.

[further reading](https://chughgpt.substack.com/p/dont-use-boolean-arguments-use-enums)

karrui (approved):
lgtm, but pls fix lint haha. can run on root package

```bash
npm run lint:fix && format:fix
```

apps/studio/prisma/schema.prisma
harishv7: i can change to enum with a field called state

harishv7: done

PR #272, opened 2024-07-05
feat: jsonforms image upload renderer

hanpuliu-charles:
## Problem

Closes [ISOM-1216](https://linear.app/ogp/issue/ISOM-1216/create-image-upload-custom-renderer)

**Made changes to the CSP to allow blob: for images, any security implications from this?**

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**AFTER**:

image


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -1,32 +1,31 @@
...
 
 export const jsonFormsImageControlTester: RankedTester = rankWith(
   JSON_FORMS_RANKING.ImageControl,
-  and(
-    schemaMatches((schema) => {
-      return schema.format === 'image'
-    }),
-    isStringControl,
-  ),
+  and(schemaMatches((schema) => schema.format === "image")),
```
dcshzj, line 28: issue: I think this is missing the `isStringControl`?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,126 @@
+// don't use backend to render, proxy once on load, 3 states, cache locally till success
```
harishv7, line 1: what's this comment about?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,126 @@
...
+    // NOTE: Using empty dependency array because we are checking if fetch is needed only upon initial load.
+    // After this load, we are the only editor of this page, and any image url changes are caused by us and we will be caching the file locally.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  // NOTE: Run once only if initial load has non empty data(imageURL).
+  const uploadImageMutation = trpc.page.uploadImageGetURL.useMutation({
+    onSettled(data, error) {
+      if (!!error) {
+        console.log("upload mutation error", error)
```
harishv7, line 49: How are we supposed to handle errors? Is there a modal or toast in design?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
hanpuliu-charles: Will remove,

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -96,4 +99,22 @@ export const pageRouter = router({
...
+    .query(async ({ input, ctx }) => {
+      // NOTE: If image is not publically accessible, might need to add logic to get it
+      const res = await fetch(input.imageUrlInSchema)
+      const blob = await res.blob()
+      const splitUrl = input.imageUrlInSchema.split(".")
+      const extension = splitUrl[splitUrl.length - 1] || ""
+      const base64Data = await BlobToImageDataURL(blob, extension)
+      return { imageDataURL: base64Data }
+    }),
+  uploadImageGetURL: pageProcedure
```
karrui, line 114: commenting only since the image upload portion has not been completed yet. 
feels a little anti-patterny to upload images to the server. What if users uploads a 100mb image etc. 
have we decided how to handle image uploads? the usual flow is to get a presigned url from aws s3 that the user uploads directly to using the client. 

[starter kit has an example flow](https://github.com/opengovsg/starter-kit/blob/main/src/features/posts/api/useUploadImagesMutation.ts) you may want to check out.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -96,4 +99,22 @@ export const pageRouter = router({
       return { pageId: "" }
     }),
   // TODO: Delete page stuff here
+
+  readImageInPage: pageProcedure
```
karrui, line 103: unsure why this needs to be processed on the server. If there is already a URL, shouldn't the client be able to render it? are we using the server as a means to mask the public image URL? suspect this would be very taxing on the server.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
karrui, line 1: implementation of image preview seems convoluted in this PR.
starter-kit has an example of how to locally render attached images (before it is even sent to the server).

[you might want to refer to it](https://github.com/opengovsg/starter-kit/blob/main/src/components/ImageAttachmentButton/AttachmentFileInfo.tsx)

karrui (commented):
if this PR is doing the flow for image upload and preview (without actually sending the image to the server yet), then suggest not sending images to the server, and use a presigned URL instead.

in the meantime, if the purpose is to render pictures locally after selecting files in the file input but before sending to the server, provided a few links to starter kit example code where that is exactly done.

can always ping me on slack if needed. maybe i understand wrongly

karrui:
pls fix types and tests too~

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
hanpuliu-charles: Hey there, the original intent of the convoluted preview code was to only display the preview if the file is successfully uploaded to the backend(since we do not have an explicit button to upload the file), to create a bijection between preview showing and upload flow being complete. But from your experience, is this really a sound model, or should there be another type of explicit feedback that the flow is complete?

apps/studio/src/constants/formBuilder.ts
```diff
@@ -3,6 +3,7 @@ export const JSON_FORMS_RANKING = {
   ArrayControl: 4,
   BooleanControl: 2,
   DropdownControl: 2,
+  ImageControl: 2,
```
harishv7, line 6: can i understand how this rank was chosen?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,96 @@
...
+      const extension = splitUrl[splitUrl.length - 1]
+      const filename = `image.${extension}`
+      setSelectedFile(new File([blob], filename, { type: blob.type }))
+    }
+    if (data) {
+      fetchImage().catch((error) =>
+        console.error('Error in fetching current image:', error),
+      )
+    }
+  }, [data])
```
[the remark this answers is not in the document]
harishv7: Have we discussed this, is it resolved?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,128 @@
...
+        .catch((error) => {
+          console.log("error fetching initial image", error)
+        })
+    }
+    // NOTE: Using empty dependency array because we are checking if fetch is needed only upon initial load.
+    // After this load, we are the only editor of this page, and any image url changes are caused by us and we will be caching the file locally.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+  const getPresignedMutation =
+    trpc.page.getPresignUrlForImageUpload.useMutation()
```
harishv7, line 61: nit: presigned*

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,128 @@
...
+      .put(image)
+      .res()
+    if (response.ok) {
+      setSelectedFile(pendingFile)
+      setErrorMessage("")
+      handleChange(path, fileURL)
+      console.log("new file url", fileURL)
+    } else {
+      setPendingFile(undefined)
+      setErrorMessage(
```
harishv7, line 79: is this a toast?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,128 @@
...
+              setSelectedFile(undefined)
+            }
+          }}
+          onError={(error) => {
+            setErrorMessage("An error occured, please try again: " + error)
+            console.log("File attachment error ", error)
+          }}
+          onRejection={(rejections) => {
+            if (rejections[0]?.errors[0]) {
+              setErrorMessage(rejections[0].errors[0].message)
```
harishv7, line 113: shouldn't this be a toast?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,128 @@
...
+          imagePreview="large"
+          multiple={false}
+          value={selectedFile}
+          onChange={(file) => {
+            console.log(file?.name)
+            if (file) {
+              setPendingFile(file)
+              void uploadImage(file)
+            } else {
+              // NOTE: Do we need to update backend on removal of file?
```
harishv7, line 101: Ideal flow might be to have an assets table with mapping

siteId assetId isDeleted

We can then cleanse our S3 bucket by walking through the isDeleted column

harishv7 (commented):
added some comments

apps/studio/src/constants/formBuilder.ts
hanpuliu-charles: So a higher ranking will be a higher priority. For the image field, since it's also a text field(textcontrol has ranking 1), I gave it the smallest number greater than 1 since we need to have a higher priority than the default text field.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [errorMessage, setErrorMessage] = useState("")
+
+  useEffect(() => {
+    if (!!data) {
+      wretch(data as string)
+        .get()
+        .blob()
+        .then((blob) => {
+          const splitData = (data as string).split("/")
+          const fileName = splitData[-1] || "Current Image"
```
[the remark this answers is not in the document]
hanpuliu-charles: Filename should be keeping the extension, so when generating the key name for the object, we can use a date-time-page format or something similar which includes the file extension at the back. For the image case, by right it should always have a fileName so it splitData[-1] should always be defined. But anyways, it might not be necessary to have the file extension as part of the file name, from my understanding, these infomation are also stored in the file content(file signature) so it should still render correctly.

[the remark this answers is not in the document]
hanpuliu-charles: Filename should be keeping the extension, so when generating the key name for the object, we can use a date-time-page format or something similar which includes the file extension at the back. For the "image" case, by right it should always have a fileName so it splitData[-1] should always be defined. But anyways, it might not be necessary to have the file extension as part of the file name, from my understanding, these infomation are also stored in the file content(file signature) so it should still render correctly.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
hanpuliu-charles: This isn't a toast, but it's a FormErrorMessage. Not too sure about going with a toast, since non of the other renderers are using it at the moment for error flows, so I'm wary that it might be out of place.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
karrui: comment: missed this comment, next time ping me on slack if i do not respond in a timely manner please! onus is on the PR owner to get their PR merged haha.

hmm, might make this PR very big if we implement the full flow. The flow I'd expect is:

**Scenario: User has not uploaded an image before**
1. user drags/uses file selector to select an image file to upload
2. on input change, generate a presigned URL and upload the image to the presigned URL
3. whilst image is uploading to presigned URL, just preview the local image file (since you already have it)
4. if upload fails, show toast and remove local image preview -- allow user to reupload or something
5. if upload success, show toast and continue previewing local image preview 

You probably only need to do steps 1 and 3 in this PR (preview local image). The other steps can be done in future PRs.

**Scenario: User has a prior uploaded image**
1. Client fetches image from remote URL (of uploaded image), and renders that in the preview
2. drag and drop field may need a new state to denote there is a previous upload. unsure if design already has
3. Same 1-5 steps as above in the previous example, only difference is local file will take preview precedence over remote file.

Does this make sense? Probably don't need do the second example in this PR.

hanpuliu-charles: Yep, so concretely this PR should consist of just 
1. user drags/uses file selector to select an image file to upload
2. whilst image is uploading to presigned URL, just preview the local image file (since you already have it)
if I understand correctly? Will try to remove the extra stuff and re-request a review

karrui: yup!

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/utils.ts
hanpuliu-charles, line 1: Can be removed?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
hanpuliu-charles: Done, removed code for the other functionalities not within scope!

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/utils.ts
karrui, line 1: this file is unused now right? should we remove it first

karrui (approved):
lgtm. please start working on the presign-upload flow so this feature works fully.

karrui:
Also realise the image is not previewed on the right side yet. Unsure if it should be in this PR or the next. you decide. just ensure it's not dropped.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/utils.ts
hanpuliu-charles: Yep, removed

PR #273, opened 2024-07-05
feat(tiptap): UI for adding and editing new block

seaerchin:
## Problem
We want to have a simple edit page flow for tiptap, where we are able to update/save the page and have it also update on preview

Closes https://linear.app/ogp/issue/ISOM-1229/add-tiptap-block
Closes ISOM-1229

## Solution
- we create a state variable `currActiveIdx` + its corresponding setter. We need this to set the active element and show the corresponding drawer correctly 
- we snapshot the current page state into another state variable when we enter the tiptap drawer. this is then restored on exit. 
- updated the props for the `isomer-components` to use `attrs`. this was required because prosemirror + tiptap both have an implicit assumption that certain props exist on certain nodes and use the props in their code. 
- update toc generation for new typings on `heading` etc

https://github.com/opengovsg/isomer/assets/44049504/ca0d582e-49fb-4956-b321-667ad75571ca


apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -94,12 +97,33 @@ function BlockItem({
...
+    const newComponent: IsomerComponent = DEFAULT_BLOCKS[sectionType]!
+
+    const nextPageState = [...pageState, newComponent]
+    setPageState(nextPageState)
+    setDrawerState({ state: nextState })
+    setCurrActiveIdx(nextPageState.length - 1)
+    mutate({
+      pageId,
+      content: JSON.stringify({ ...page.content, content: nextPageState }),
+    })
```
seaerchin, line 124: this isn't very clean and we should probably find a better way of doing things.

apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -45,6 +52,8 @@ export function EditorDrawerProvider({ children }: PropsWithChildren) {
       setIsomerJsonSchema,
     }),
     [
+      currActiveIdx,
+      setCurrActiveIdx,
       drawerState,
       setDrawerState,
       pageState,
```
seaerchin, line 59: this context is getting bloated and we should probably break it up

apps/studio/src/features/editing-experience/components/EditPageDrawer.tsx
```diff
@@ -1,23 +1,40 @@
...
 import RootStateDrawer from './RootStateDrawer'
 import TipTapComponent from './TipTapComponent'
 import ComplexEditorStateDrawer from './ComplexEditorStateDrawer'
 
+function narrowToTiptap(
+  component: IsomerComponent,
+): IsomerNativeComponentProps {
+  // TODO: get rid of the cast here
+  return component as IsomerNativeComponentProps
+}
```
seaerchin, line 16: in root state drawer, when we click into a component, we only select `tiptap` if the component is prose. this means that tiptap will always be able handle the content but because our type of `IsomerComponent` is wider, we have to either validate or cast here.

apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
```diff
@@ -20,55 +21,65 @@ import { Text } from '@tiptap/extension-text'
...
       BulletList.extend({
-        name: 'unorderedlist',
-      }),
-      BulletList.configure({
+        name: 'unorderedList',
+      }).configure({
         HTMLAttributes: {
           class: 'list-disc',
         },
       }),
```
seaerchin, line 76: this is a bug fix, previously this returned 2 instances of `BulletList` (1 configured, 1 extended) and the later one won.

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -78,8 +82,15 @@ export default function RootStateDrawer() {
...
-                            console.log('huh')
-                            setDrawerState({ state: 'nativeEditor' })
+                            setCurrActiveIdx(index)
+                            // TODO: we should automatically do this probably?
+                            const nextState =
+                              pageState[index]?.type === 'prose'
+                                ? 'nativeEditor'
+                                : 'complexEditor'
+                            // NOTE: SNAPSHOT
+                            setEditorState(pageState)
```
seaerchin, line 92: note to self - remember to update the exported name

apps/studio/src/features/editing-experience/utils/convert.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const validateAsProse = (
+  content: JSONContent,
+): IsomerNativeComponentProps => {
+  const ajv = new Ajv({ strict: false })
+  const subSchema = {
+    ...schema.components.native.prose,
+    components: schema.components,
+  }
+  const validate = ajv.compile<IsomerNativeComponentProps>(subSchema)
```
seaerchin, line 16: note to reviewers: this is as good as a cast. changing the type here will cause the return type of `validate` to be that type, even if it's wrong. 

we should probably shift to typebox sooner rather than later.

docs/next/0.1.0.json
seaerchin, line 1: mostly format fixes

packages/components/src/interfaces/native/Heading.ts
```diff
@@ -6,7 +6,9 @@ export type HeadingLevel = (typeof HeadingLevels)[number]
 
 export interface HeadingProps {
   type: "heading"
-  id: string // Used for anchor links
   content: TextProps[]
-  level: HeadingLevel
+  attrs: {
```
seaerchin, line 10: adopted tiptap schema instead of bringing it up. this is also a **bugfix** because prosemirror assumes `attrs.<key>`. 

for example, `mergeCells` for table is done via `pm/table` and it uses `attrs.colspan`. hence, we can't destructively edit tiptap's schema. we can only add on

packages/components/src/interfaces/native/Table.ts
```diff
@@ -7,17 +7,19 @@ import type {
...
 }
 
 interface TableCell extends TableBaseCell {
   type: "tableCell"
   content: (
     | DividerProps
     | ParagraphProps
-    | ImageProps
     | OrderedListProps
+    | ImageProps
```
seaerchin, line 22: todo: delete - table cells shouldn't contain `Image`

packages/components/src/interfaces/native/Prose.ts
```diff
@@ -10,15 +10,17 @@ import type { UnorderedListProps } from "./UnorderedList"
 export type ProseContent = (
   | DividerProps
   | HeadingProps
-  | ImageProps
   | ListItemProps
+  | ImageProps
```
seaerchin, line 14: @sehyunidaaa - ok if prose can contain image? i saw in figma that prose shouldn't. i'm happy to edit and make this change

packages/components/src/interfaces/schema.json
seaerchin, line 1: format updates, shifting to adopt tiptap schema, making `content` optional on `Prose`

packages/components/src/interfaces/native/Prose.ts
seaerchin: note: not ok - we need to remove

packages/components/src/interfaces/schema.json
seaerchin: this is required because a full deletion deletes `content` implicitly on `tiptap` cc @dcshzj - can make the change downstream too

packages/components/src/interfaces/native/Table.ts
seaerchin: done by @dcshzj in typebox PR

apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -21,24 +29,30 @@ export function EditorDrawerProvider({ children }: PropsWithChildren) {
   // Current saved state of page
   const [pageState, setPageState] = useState<IsomerComponent[]>([])
   // Current edit view of page
-  const [editorState, setEditorState] = useState<IsomerComponent[]>([])
+  const [snapshot, setSnapshot] = useState<IsomerComponent[]>([])
+  // Isomer page schema
+  const [currActiveIdx, setCurrActiveIdx] = useState(0)
 
   const value = useMemo(
```
karrui, line 36: i don't think you need to memoize any of these. They are controlled by react. You can just return them without memoizing.

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -95,12 +98,33 @@ function BlockItem({
 }
 
 function ComponentSelector() {
-  const { setDrawerState, setPageState, setEditorState } =
+  const { setCurrActiveIdx, pageState, setDrawerState, setPageState } =
     useEditorDrawerContext()
+  const { mutate } = trpc.page.updatePageBlob.useMutation()
```
karrui, line 103: should you handle the loading state too?

apps/studio/src/pages/dashboard/edit/index.tsx
```diff
@@ -1,26 +1,28 @@
...
 
 function EditPage(): JSX.Element {
-  const { setDrawerState, pageState, setPageState, setEditorState } =
-    useEditorDrawerContext()
+  const {
+    setDrawerState,
+    pageState,
+    setPageState,
+    setSnapshot: setEditorState,
+  } = useEditorDrawerContext()
```
karrui, line 15: feels weird that this page itself does not have its own context and has to use a context further up. Any reason why the context is not inside this component and is instead outside? Seems like all the props are initialised in this component.

apps/studio/src/contexts/EditorDrawerContext.tsx
seaerchin: i think it was copy paste code sadly, but sure, will remove!

apps/studio/src/pages/dashboard/edit/index.tsx
seaerchin: sounds good, lemme just shift it in here

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: sure, will add!

apps/studio/src/pages/dashboard/edit/index.tsx
seaerchin: i think this got changed somewhere :monkas:

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: actually going in the opposite direction here - not doing a save at all. the reason why is because we don't want empty states to show up in the end user site - ie, i should save only when the user has content **that they added**. confirmed with @sehyunidaaa already so if they add a block (no content) then refresh, we're ok with this disappearing

dcshzj (approved):

PR #288, opened 2024-07-09
feat: provide default values for constants in schema

dcshzj:
> [!NOTE]
>
> Small PR, don't worry. Lines change are due to changes in the published schema.

## Problem

The schema used by POEs are now disconnected from the actual schema that we use in the components package.

## Solution

**Breaking Changes**

  - The actual schema is being updated, so existing sites need to be migrated.

**Features**:

- The Isomer JSON Schema published at schema.isomer.gov.sg is now auto-generated from the actual schema that we are using to define the public interfaces of the components package.
    - The additional script is structured to fit more of the RJSF form builder that we are letting POEs use.
    - Once Studio is up, this script will no longer be needed, and we do not necessarily need to publish the schema anymore.
- Default values were given for all literals so that they are automatically populated in the form builder.
- Some titles and descriptions were added, but the final copy is not finalised yet with design.
- Some of the changes also involve adding the types, as union types for literals do not have the `type` automatically generated for them.


packages/components/scripts/generateSchema.js
karrui, line 1: should we add some unit tests for this?

packages/components/src/interfaces/native/Prose.ts
```diff
@@ -6,26 +6,31 @@ import { ParagraphSchema } from "./Paragraph"
...
+        {
+          minItems: 1,
+        },
+      ),
     ),
-  ),
-})
+  },
+  {
+    $id: "components-native-prose",
```
karrui, line 29: if `$id` is not needed anymore, should this be here?

packages/components/scripts/generateSchema.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * This script generates a JSON schema file for the Isomer JSON Schema. It is
+ * kept in the docs/next folder so that POEs can use the schema in third-party
+ * form builders when migrating content to Isomer Next.
+ */
+
+const jsonOutput = JSON.stringify(
+  {
+    ...schema,
+    components: {
```
harishv7, line 21: earlier have segmented this components key furthers in complex native internal right? have we removed that?

packages/components/scripts/generateSchema.js
dcshzj: Hmm this script is meant to be temporary and will be removed once Studio is able to be used by POEs directly so they can continue to use [the RJSF form builder](https://rjsf-team.github.io/react-jsonschema-form/) for their migration work, so I opted to not have tests for this output.

packages/components/scripts/generateSchema.js
dcshzj: The segmentation is still maintained in Studio via `IsomerComplexComponentsMap` and `IsomerNativeComponentsMap`! This schema is generated this way to fit into RJSF's form builder, in which they are only able to resolve `$ref`s using paths instead of `$id`.

This functionality is not needed inside JSONForms because `$ref`s are resolved using `$id`s.

packages/components/src/interfaces/native/Prose.ts
dcshzj: Hmm `$id` is needed by TypeBox and JSONForms, but it is not supported by RJSF's form builder (the tool that POEs are using). The script only converts the schema from TypeBox into something that RJSF is able to understand, which they are not able to understand `$ref`s pointing to `$id`s.

dcshzj:
@opengovsg/isomer-engineers migrated the JSON schema and the script to generate it over to playground due to urgency.

seaerchin:
lets schedule some time to run through this tomorrow ba

harishv7 (approved):
lgtm

PR #292, opened 2024-07-10
feat: introduce link custom renderer

dcshzj:
## Problem

We need a custom renderer to insert links.

## Solution

**Breaking Changes**

**Features**:

- Adds a new custom renderer for adding links.
    - Full functionality is not available yet pending design, mainly on getting the page permalink for internal links.
    - Only functionality added was adding email links.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-07-10 at 17 38 49

**AFTER**:

Screenshot 2024-07-10 at 17 38 29


packages/components/src/interfaces/complex/Hero.ts
```diff
@@ -26,7 +26,7 @@ const HeroButtonsSchema = Type.Object({
     Type.String({
       title: "Button destination",
       description: "When this is clicked, open:",
-      format: "link",
+      format: "uri",
```
seaerchin, line 29: what's the difference between `uri` and `link`?

seaerchin (commented):
lgtm code-wise, some follow-up questions below

1. we haven't actually used this anywhere? the PR to use this will be done in a follow-up?
2. have we done validation here? i think the bare minimum is to just conform to the URL format right?

packages/components/src/interfaces/complex/Hero.ts
dcshzj: Hmm no difference, mainly a naming convention. `uri` is one of the formats in the JSON Schema specs, but since we are supporting references as links, it might be more accurate to call it `link` to differentiate. Updated in 0c0bd6f16ce7b993d3249f5a0d70c681819a65ac.

dcshzj:
> 1. we haven't actually used this anywhere? the PR to use this will be done in a follow-up?

This is being used in components like Button, InfoCols, etc. The follow-up is just adding the full functionality to support adding internal links, external links, selecting files and adding validation.

> 2. have we done validation here? i think the bare minimum is to just conform to the URL format right?

Full validation not available yet, but it would not just be the URL format. We have internal links that we store as references rather than as URLs.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,117 @@
...
+    },
+  ]
+  const dataString = data && typeof data === "string" ? data : ""
+
+  return (
+    <Box py={2}>
+      <FormControl isRequired={required}>
+        <FormLabel description={description}>{label}</FormLabel>
+
+        <Tabs variant="group" onChange={() => handleChange(path, "")}>
```
hanpuliu-charles, line 61: Not really an issue, but does this mean that if I had data in a tab, and I clicked on another type of link(maybe to check if this suits the purpose better), and I clicked back my previous data will be lost?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+export function JsonFormsLinkControl({
+  data,
+  label,
+  handleChange,
+  path,
+  description,
+  required,
+}: ControlProps) {
+  const LINK_TYPES = [
```
karrui, line 36: nit: if this is a constant, might want to move it outside of the react component so this is no recreated on every render

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
karrui, line 61: comment: i think tabs might be wrong way to go about this haha, i'd prefer a custom radio for semantic html. but since this is an internal facing application, might be ok

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,117 @@
...
+          </TabList>
+
+          <TabPanels>
+            <TabPanel>
+              <Input
+                type="text"
+                value={dataString}
+                onChange={(e) => handleChange(path, e.target.value)}
+                placeholder="Page permalink"
+              />
```
karrui, line 80: this however is the bigger issue (and thus i'd recommend dynamic state based on selected radio value instead), as the tab panels can desynchronise with `LINK_TYPES` and a specific order is required, which is not obvious if `LINK_TYPE` is ever modified.

might be safer to not use tabs and just render a specific input depending on the selected radio option. to that end, you can rely on compile time typechecking to ensure something is rendered accordingly to the option.

UNLESS the value of the current tab does not matter to you, and only the input value does. then maybe tab can be used.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
dcshzj: Moved out in 950c8ed0b863961599692a2057cc26ee605d469c.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
dcshzj: Switched to radio in 950c8ed0b863961599692a2057cc26ee605d469c.

dcshzj: Yeah, the 4 types are mutually exclusive so the content of one type would not be transferrable to another, hence opted to clear the value to force the user to make the selection first.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
dcshzj: Updated in 950c8ed0b863961599692a2057cc26ee605d469c, agreed that it would better with a more explicit mapping of the link type to the actual rendering.

The value of the current tab would not matter because the 4 link types are mutually exclusive, so it is not possible to translate from one to another. Opted to clear out the value of the input to avoid confusing the user.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,198 @@
...
+      return (
+        <Input
+          type="text"
+          value={data.startsWith("mailto:") ? data.slice("mailto:".length) : ""}
+          onChange={(e) => handleChange(`mailto:${e.target.value}`)}
+          placeholder="test@example.com"
+        />
+      )
+    default:
+      return <></>
```
karrui, line 139: nit: might want to skip default case, or assert `never` for compile-time typecheck

see [knowledge base](https://www.notion.so/opengov/Use-exhaustive-switch-for-enums-e273c275a684492d8aef4789b4882906)

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,198 @@
...
+  selectedLinkType: string
+  data: string
+  handleChange: (value: string) => void
+}
+
+const RadioContent = ({
+  selectedLinkType,
+  data,
+  handleChange,
+}: RadioContentProps) => {
```
karrui, line 100: assert return type for typechecking

```suggestion
}: RadioContentProps): JSX.Element => {
```

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const { getRootProps, getRadioProps } = useRadioGroup({
+    name: "link-type",
+    defaultValue: "page",
+    onChange: handleLinkTypeChange,
+  })
+
+  const dataString = data && typeof data === "string" ? data : ""
+
+  return (
+    <Box py={2}>
```
karrui, line 167: supernit: suggest using `rem` instead of numbers for clarity

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -0,0 +1,198 @@
...
+  handleChange,
+  path,
+  description,
+  required,
+}: ControlProps) {
+  const [selectedLinkType, setSelectedLinkType] = useState("page")
+
+  const handleLinkTypeChange = (value: string) => {
+    setSelectedLinkType(value)
+    handleChange(path, "")
```
karrui, line 155: does this actually do anything? are you supposed to pass in the change into `handleChange` too?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
dcshzj: Fixed in 5e33dc3f352940442a3b6f798764b8dd3fef94d1.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
dcshzj: Updated in 5e33dc3f352940442a3b6f798764b8dd3fef94d1.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
dcshzj: Hmm yeah it is supposed to reset the input field to empty when they change the selected link type. This is because all 4 link types are mutually exclusive of one another, so the input from one does not translate to another, so opted to clear the input field to prevent user confusion.

karrui (approved):
lgtm

PR #293, opened 2024-07-10
feat: Add schema validation middleware for page router

hanpuliu-charles:
## Problem

Closes [ISOM-1232](https://linear.app/ogp/issue/ISOM-1232/validation-of-page-state)

## Solution

1. Added a new procedure based on pageProcedure called validatedProcedure.(removed placeholder for pageProcedure for now since it is unimplemented.)
2. validatedProcedure does a check on the input content(currently, assumed to be entire page schema) using ~TypeBox's type compiler~ajv. Throws TRPC exception if schema is invalid(doesn't match schema).
3. ~**Updated Isomer Components** to remove `@id` and `Type.ref` to enable TypeCompiler to work without error, previously there were errors due to repeated references. Ajv validation hasn't worked since switching over to TypeBox types, and still isn't working.**~ This didn't solve the issue, but fixed in #303 

**Breaking Changes**

  - Details ...

**New dependencies**:

- ~`@sinclair/typebox` : Uses TypeCompiler's Check() function for validation against schema.~


apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -7,18 +10,44 @@ import {
...
   updatePageById,
 } from "../resource/resource.service"
 import { getSiteConfig } from "../site/site.service"
 
+const typeCompiler = TypeCompiler.Compile(IsomerPageSchema)
+
 // TODO: Need to do validation like checking for existence of the page
 // and whether the user has write-access to said page
 const pageProcedure = protectedProcedure
+const validatedPageProcedure = pageProcedure.use(async ({ next, rawInput }) => {
```
karrui, line 29: nit: you can chain directly without declaring multiple variables

```suggestion
const pageProcedure = protectedProcedure.use(async ({ next, rawInput }) => {
```

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -7,18 +10,44 @@ import {
   updatePageSchema,
 } from "~/schemas/page"
 import { protectedProcedure, publicProcedure, router } from "~/server/trpc"
+import { safeJsonParse } from "~/utils/safeJsonParse"
```
karrui, line 13: unused import?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -7,18 +10,44 @@ import {
...
 // and whether the user has write-access to said page
 const pageProcedure = protectedProcedure
+const validatedPageProcedure = pageProcedure.use(async ({ next, rawInput }) => {
+  if (
+    typeof rawInput === "object" &&
+    rawInput != null &&
+    "content" in rawInput
+  ) {
+    // NOTE: content will be the entire page schema for now...
+    if (!typeCompiler.Check(JSON.parse(rawInput.content as string))) {
```
karrui, line 36: does `safeJsonParse` not work for this?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -7,18 +10,44 @@ import {
...
+      throw new TRPCError({
+        code: "BAD_REQUEST",
+        message: "Schema validation failed.",
```
karrui, line 39: do we want to return the exact error for better dev experience (and error debugging when something is thrown)

apps/studio/src/server/modules/page/page.router.ts
hanpuliu-charles: Hmm, believe that pageProcedure is the generic procedure used without schema validation, while validatedpageprocedure is pageProcedure with a schema validation middleware. Both seems to be used so I am thinking that this is okay?

karrui: gotcha. in that case maybe don't need to reassign `protectedProcedure` with `pageProcedure` and use `protectedProcedure` straight in non-validated procedures?

hanpuliu-charles: Okay, might be that I am misunderstanding this, but let me pen down what I think each procedure does?
1. protectedProcedure(declared elsewhere): this has user authentication middleware, used for pages, folders, and everything that isn't public.
2. pageProcedure: currently just protectedProcedure, but eventually when doing up permissioning, will have to add a middleware(so `const pageProcedure = protectedProcedure.use(...)` to enforce permissions, make sure user has perms to CRUD on a page resource(could be folder too, so this procedure might have to shift)
3. validatedPageProcedure: might be bad naming on my part but this is intended to be a pageProcedure with additional middleware for schema val. So what this means is it will have both perms middleware and schema val middleware!

karrui: Yes, but since currently 1 == 2, it is somewhat redundant. Can always add it in the future. Don't code for the future; code for the present haha.

apps/studio/src/server/modules/page/page.router.ts
hanpuliu-charles: Returning the error iterable as cause! Or is there a better way to do it here?

apps/studio/src/server/modules/page/page.router.ts
hanpuliu-charles: Removed in 49d00f8

karrui (approved):

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -16,12 +20,40 @@ import {
...
 // TODO: Need to do validation like checking for existence of the page
-// and whether the user has write-access to said page
-const pageProcedure = protectedProcedure
+// and whether the user has write-access to said page: replace protectorProcedure in this with the new procedure
+
+const validatedPageProcedure = protectedProcedure.use(
+  async ({ next, rawInput }) => {
+    if (
+      typeof rawInput === "object" &&
+      rawInput != null &&
```
karrui, line 33: nit: should use strict equals: `===`

PR #297, opened 2024-07-11
feat: generate better database and add path alias for better dev experience

karrui:
### TL;DR
Generate "selectable" types for generated Prisma models to improve type safety and reduce manual work.
Also add path aliases to database server module for better import paths.

### What changed?
- Updated `generate` script to include `tsx prisma/generate.cts`
- Added new `generate.cts` script to generate TypeScript types
- Updated prisma schema to remove trailing whitespace
- Changed import paths in several TypeScript files to use new path aliases

### How to test?
1. Run `npm run generate` in the `apps/studio` directory.
2. Check if the new TypeScript types are correctly generated in the `generated` folder.
3. Verify that the application builds successfully and linting passes.

### Why make this change?
To automate the generation of TypeScript types for Prisma models, ensuring consistent and error-free type definitions across the project.

---


karrui:
* **#297** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/studio/prisma/generate.cts
```diff
@@ -0,0 +1,136 @@
+import * as fs from "fs/promises"
+import * as path from "path"
+import ts from "typescript"
+
+const TYPE_FOLDER = "./generated"
+// THIS MUST BE PREFIXED WITH A DOT
```
seaerchin, line 6: does this comment apply only to `TYPE_FILE` or to the group of `const`? seems like it applies to all

apps/studio/prisma/generate.cts
seaerchin, line 1: i don't quite understand this so skipping for now

apps/studio/prisma/generated/selectableTypes.ts
seaerchin, line 1: my impression was that the typing was declared using `kysely<DB>` - the import path is still pointing to `generatedTypes#DB`, which uses `generatedTypes` and not `Selectable`. 

do we need to update the original generated file post creation of `selectableTypes` so that `DB` is using the types from here?

seaerchin (commented):
monke brain no uds

apps/studio/prisma/generate.cts
karrui: idk. i copied from activesg implementation. dexter write one LOL

apps/studio/prisma/generated/selectableTypes.ts
seaerchin: from our conversation offline, so that people next time (me next week i bet) can understand why this required. 

given the statement: 
`const permission: Model.Permission = await db.selectFrom("Permission").selectAll().executeTakeFirstOrThrow()`

in here, `permission` has the `Model.Permission` typing, which is the _generated_ typing from kysely. in particular, it has `GeneratedAlways<number>` as the type for `id`. this is to hint to kysely that id is autogenerated and hence, can be omitted/shouldn't be specified when doing creates. however, this means that **post retrieval**, there's a type mismatch! 

specifically, `permission` has `type id: GeneratedAlways<number>` but for the returned value from kysely, it has `id: number`. both of these are not mutually assignable sadly. hence, the `Selectable` type is required to bridge the gap and tell kysely that hey, the values i got from db is a `Selectable<Permission>`. 

because this is onerous to keep doing everytime we want to do a select, we will auto-gen this code using the provided `cjs` file. that being said, i haven't actually read through teh cjs file in detail/for correctness etc

seaerchin (approved):
ok after reading through, i t hink it makes sense. the way that we did the auto-gen is 
1. find the sentinel type `DB` 
2. extract all type members
3. wrwap it in a `Selectable` 

given that the function of `Selectable` is to remove  generation context (see [here](https://kysely-org.github.io/kysely-apidoc/types/Selectable.html)) from the type, i think this is safe to wrap for most, if not all, types otu from kysely.

karrui:
### Merge activity

* **Jul 11, 5:29 AM EDT**: @karrui started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/297).

PR #303, opened 2024-07-11
feat: ref all native components and use getComponentSchema

dcshzj:
## Problem

For some reason builds could work before without using refs, but now it doesn't. It is now time to settle the schemas once and for all.

Concretely, the problem is that the schema (after removal of all the Type.Ref) kept running into issues because of duplicate `$id` definitions. This is because the same component can be referenced multiple times by different components, especially the native ones.

## Solution

**Breaking Changes**

**Features**:

- Make all native components be Type.Ref'd.
- Introduce `getComponentSchema` as the main entry point for component consumers to consume the underlying JSON schemas.
    - This method ensures that all native components are referenced using Type.Ref, then we separately maintain a definitions list which contains the actual definition for those `$id`.
    - This ensures that the `$id` is defined only once in sub-schemas that require them.

### DIFF
--- a/packages/components/src/schemas/main.ts
+++ b/packages/components/src/schemas/main.ts
@@ -1,0 +1,36 @@
+import type { TSchema } from "@sinclair/typebox"
+
+import type { IsomerComponentTypes } from "~/types"
+import { IsomerPageSchema } from "~/types"
+import {
+  IsomerComplexComponentsMap,
+  IsomerNativeComponentsMap,
+} from "./components"
+
+const definitions = {
+  components: {
+    complex: IsomerComplexComponentsMap,
+    native: IsomerNativeComponentsMap,
+  },
+}
+
+export const schema: TSchema = {
+  $schema: "http://json-schema.org/draft-07/schema#",
+  title: "Isomer Next Page JSON",
+  ...IsomerPageSchema,
+  ...definitions,
+}
+
+export const getComponentSchema = (
+  component: IsomerComponentTypes,
+): TSchema => {
+  const componentSchema =
+    component === "prose"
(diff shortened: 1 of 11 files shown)


packages/components/src/schemas/main.ts
```diff
@@ -0,0 +1,36 @@
...
+const definitions = {
+  components: {
+    complex: IsomerComplexComponentsMap,
+    native: IsomerNativeComponentsMap,
+  },
+}
+
+export const schema: TSchema = {
+  $schema: "http://json-schema.org/draft-07/schema#",
+  title: "Isomer Next Page JSON",
```
harishv7, line 19: Page Schema*

packages/components/src/schemas/main.ts
```diff
@@ -0,0 +1,36 @@
...
+  ...IsomerPageSchema,
+  ...definitions,
+}
+
+export const getComponentSchema = (
+  component: IsomerComponentTypes,
+): TSchema => {
+  const componentSchema =
+    component === "prose"
```
harishv7, line 28: We check only for prose, as prose comprises of all the native components right

harishv7 (approved):
lgtm

packages/components/src/schemas/main.ts
dcshzj: Fixed in d819f9d5e2f3211a6951bde3dab63903ec01b7dc.

packages/components/src/schemas/main.ts
dcshzj: Yup, so it will always be complex + 'prose' so we treat prose as a special kind of component.

PR #312, opened 2024-07-15
fix: remove duplicate dropdown custom renderer

dcshzj:
## Problem

A duplicate dropdown renderer appears due to the anyOf renderer and the dropdown custom renderer. This was due to the switch of the schema from being hand-written to TypeBox.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Ensure that the dropdown custom renderer does not appear when there is only a single option available.
- Make the anyOf custom renderer take in the constant value from its children instead of the default label that is auto-generated (which gives something like `anyOf-0`, `anyOf-1`, etc).
- Also prevent the dropdown from selecting the first available option so the user knows what option they are selecting.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-07-15 at 11 02 44

**AFTER**:

Screenshot 2024-07-15 at 11 03 11


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsAnyOfControl.tsx
```diff
@@ -35,25 +36,36 @@ export function JsonFormsAnyOfControl({
     uischemas,
   )
 
-  const variants = anyOfRenderInfos.map((anyOfRenderInfo) => ({
-    label: anyOfRenderInfo.label,
-    value: anyOfRenderInfo.label,
-  }))
+  const options = anyOfRenderInfos.map((anyOfRenderInfo) => {
+    const option = String(anyOfRenderInfo.schema.const || anyOfRenderInfo.label)
```
seaerchin, line 40: why choose `schema.const` over `label`?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsConstControl.tsx
```diff
@@ -0,0 +1,29 @@
...
+  handleChange,
+  path,
+  schema,
+}: ControlProps) {
+  // TODO: Make the default value persist inside JSONForms
+  useEffect(() => {
+    handleChange(path, schema.const)
+  }, [handleChange, path, schema])
+
+  return null
```
seaerchin, line 26: this feels abit sus - what it feels like we're trying to do here is to write through to our underlying data structure (our data) but not render anything.

i think the pattern that we might want to aim for here might be to use a [middleware + reducer](https://jsonforms.io/docs/middleware/) so we fire the event only on update and only to mutate data

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDropdownControl.tsx
```diff
@@ -28,7 +28,12 @@ export function JsonFormsDropdownControl({
 }: ControlProps & OwnPropsOfEnum) {
   const [dropdownValue, setDropdownValue] = useState(data || "")
 
-  if (!options || (options.length === 1 && !!schema.default)) {
+  // Use the default value if it exists
+  useEffect(() => {
+    handleChange(path, schema.default || schema.const)
+  }, [path, schema.default, schema.const, handleChange])
+
```
seaerchin, line 35: what do we feel about setting the default value in teh `useState`

dcshzj: I think the `useState` not too important here cos the component is not rendered but I realised this also will cause changes even if the original data is defined. Fixed in 57174034e4c49db9f3c4d2359d82ad5f65bcc9cd.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsAnyOfControl.tsx
```diff
@@ -35,25 +36,36 @@ export function JsonFormsAnyOfControl({
...
-        <SingleSelect
-          value={variant}
-          name={label}
-          items={variants}
-          isClearable={false}
-          onChange={setVariant}
-        />
-      </FormControl>
+    <>
+      <Box py={2}>
```
karrui, line 57: nit: prefer explicit declaration for consistency.

```suggestion
      <Box py="<xx>rem"> 
```

in the future might have specific spacing tokens haha

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDropdownControl.tsx
```diff
@@ -28,7 +28,17 @@ export function JsonFormsDropdownControl({
...
-  if (!options || (options.length === 1 && !!schema.default)) {
+  // Use the default value if it exists
+  useEffect(() => {
+    if (data !== undefined || data !== "") {
+      return
+    }
+    const value = schema.default || schema.const
+    setDropdownValue(value)
+    handleChange(path, value)
+  }, [path, schema.default, schema.const, handleChange, data])
```
karrui, line 39: [don't sync state, derive it!](https://www.notion.so/opengov/Don-t-sync-state-derive-it-173d074b374644e6bab52e578ad626a3)

feels weird to need this, does JSONForm not update something like `data` when  `handleChange` is triggered? if so, can't you derive the current selected value from it? unsure how it actually works.

```ts
const dropdownValue = data ?? "" // this should always be synchronised with the prop

// unsure if you ever need to call handleChange in an effect?

// then inside singleselect
<SingleSelect
  value={dropdownValue}
  onChange={value => handleChange(path, value)} // < this should update data prop, no?
/>
```

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsAnyOfControl.tsx
dcshzj: Ah I wasn't sure if we should use rem or numbers but agree seems like rem is a lot easier to reason. Fixed in 993b9c92c9f4d1c3a6a5405ef52321df66759a0e!

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDropdownControl.tsx
dcshzj: Ah yup I think I forgot to remove this earlier, done in 993b9c92c9f4d1c3a6a5405ef52321df66759a0e.

karrui (approved):

PR #309, opened 2024-07-11
feat: add new sidebar layout for CMS site page

karrui:
### TL;DR

This pull request introduces a new sidebar layout (`AdminCmsSidebarLayout`) and updates the existing navbar logo with a route to the app dashboard.

### What changed?

1. Added a new CmsSidebar component containing:
   - CmsSidebarContainer
   - CmsSidebarItems
2. Integrated the CmsSidebar in a new layout (AdminCmsSidebarLayout).
3. Replaced the old OGP logo with the Isomer logo in the AppNavbar.
4. Updated various paths to use constants from the configuration (e.g., DASHBOARD).
5. Modified multiple components/pages to align with the new layout and sidebar structure.

### How to test?

Stories, or Vercel

### Why make this change?

This change enhances the user experience by providing a modern sidebar layout and updating the navbar branding.

---

PR #308, opened 2024-07-11
feat: create initial dashboard page

karrui:
### TL;DR
This PR adds a new `/dashboard` page to go to a specific site

### What changed?
1. **Site Listing**:
    - Introduced a new `SiteList` component that lists all sites a user has access to.
    - Added new database query in `site.router.ts` to fetch site list.
2. **Report Issue Button**:
    - Added a button in the AppNavbar to allow users to report issues via the Isomer issue tracker.
3. **Component Updates**:
    - Updated `AppNavbar` to include new functionality and style tweaks.
    - Changed the logo source in `AppNavbar.tsx`.
    - Updated `dashboard.tsx` with new site listing component.
    - Refactored various storybook files to include new handlers and remove unnecessary layouts.

### How to test?
New stories

---

PR #302, opened 2024-07-11
feat: use `chromatic` parameter in stories instead of separate viewport stories

karrui:
This PR includes several updates to the Storybook configuration and related files to improve the development and testing workflow in the Studio application.

### What changed?

1. **Storybook Configuration**: Significant updates have been made to `.storybook/main.ts` and `.storybook/preview.tsx`, including the addition of `mswLoader` and `withChromaticModes` for improved addon integration and testing.
2. **Package Updates**: New packages have been added to `package.json`, including `@isomer/storybook-config`, and corresponding updates have been made to `package-lock.json`.
3. **Component and Story Updates**: Updates to various components and stories, such as `DashboardTable.tsx` and `LandingPage.stories.ts`, to align with the new Storybook and Chromatic configurations.
4. **Tooling**: Added a new `@isomer/storybook-config` package under the `tooling/` directory to centralize Storybook configurations.

### How to test?

1. Checkout this branch.
2. Install the new dependencies: `npm install`.
3. Run Storybook: `npm run storybook`.
4. Verify that all stories load correctly and that Chromatic snapshots are taken as expected.

### Why make this change?

To enhance the developer experience by centralizing Storybook configurations, improving Chromatic integration, and ensuring consistent testing and linting across the project.

---

PR #301, opened 2024-07-11
feat: update next to v14.2.5

karrui:
This pull request updates the Next.js framework from version 14.2.3 to 14.2.5 in several package files, including `package.json`, `package-lock.json`, and various tooling templates. The updates aim to incorporate improvements and bug fixes from the newer version.

PR #313, opened 2024-07-15
feat: use react table to render resources subpage

karrui:
This update introduces a new 'Datatable' component implementation that replaces the prior 'DashboardTable.'

Also update implementation to fit design.

The 'Datatable' leverages the '@tanstack/react-table' library to enhance data table capabilities. It includes features like pagination, loading states, empty placeholders, and customizable cells and headers. Additionally, the PR updates various related file components and ensures that the project dependencies are aligned with these changes.


karrui:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/313?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#313** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#309** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/309?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#308** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/308?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#302** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#301** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/301?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -0,0 +1,167 @@
...
+            top={0}
+            w="100vw"
+            zIndex={2}
+          >
+            <Box m="auto">
+              <Spinner />
+            </Box>
+          </Flex>
+        </>
+      )}
```
seaerchin, line 88: this component has to care about `loading` states but actually what i feel is the right abstraction here would be for the component here to only care about rendering the table and not care about loading. 

`Suspense` feels like a good fit here, we can omit hte `isLoading` and it'll still fit nicely! wdyt?

apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    key={header.id}
+                    pos="relative"
+                    px={0}
+                    style={{
+                      width:
+                        header.getSize() === Number.MAX_SAFE_INTEGER
+                          ? "auto"
+                          : header.getSize() === Number.MIN_SAFE_INTEGER
+                            ? "fit-content"
+                            : header.getSize(),
```
seaerchin, line 110: why does this work ah? this seems like an implementation detail of `header` that i suddenly ahve to care about. actually what we care about here are certain boundary values (eg: 1300px etc) post which we should scale to said factor. 

wdyt about using, for example, `size.lg` etc?

seaerchin: also pls no nested ternary

apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -0,0 +1,167 @@
...
+            {rows.map((row) => {
+              return (
+                <Tr key={row.id} borderBottomWidth="1px">
+                  {row.getVisibleCells().map((cell) => {
+                    return (
+                      <Td
+                        key={cell.id}
+                        verticalAlign="center"
+                        {...tablePropOverrides?.[cell.column.id]?.td}
+                        {...tablePropOverrides?.[cell.id]?.td}
```
seaerchin, line 135: not entirely sure i'm understanding this correctly here, but how would i know ahead of time what's the cell's specific id?

also, why allow an override of `td` props but not other components? i think if the use-case is narrow (only 1/2 places override `td`), we might as well just remove this prop since it's only for styling AHHA

apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -0,0 +1,167 @@
...
+          </Text>
+        )}
+        {pagination && (
+          <Flex ml="auto">
+            <DatatablePagination
+              instance={instance}
+              totalRowCount={totalRowCount}
+            />
+          </Flex>
+        )}
```
seaerchin, line 163: these 2 props (`totalRowCountString` and `totalRowCount`) seem to only be used here and the abstraction of `Datatable` became abit bloated because it now has to handle pagination in component as well. 

what do you feel about just cutting pagination out of datatable and letting consumers choose this behaviour?

apps/studio/src/components/Datatable/Datatable.tsx
seaerchin, line 1: API for datatable here feels abit too bloated tbh - it feels like it's trying to abstract over too many things and i'd rather just having it as 

1. Suspense for loading
2. Datatable for the actual data 
3. DatatablePagination for paginating

apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <Flex
+            // white alpha to denote loading
+            bg="whiteAlpha.800"
+            bottom={0}
+            left={0}
+            p="1rem"
+            pos="absolute"
+            right={0}
+            top={0}
+            zIndex="1"
```
seaerchin, line 71: nit: should be using the `zIndex` export from chakra

apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -0,0 +1,167 @@
...
+          />
+          <Flex
+            bottom={0}
+            flex={1}
+            left={0}
+            pos="fixed"
+            right={0}
+            top={0}
+            w="100vw"
+            zIndex={2}
```
seaerchin, line 81: nit: should be using `zIndex` export from chakra

apps/studio/src/components/Datatable/DatatablePagination.tsx
```diff
@@ -0,0 +1,28 @@
+import { Pagination } from "@opengovsg/design-system-react"
+import { type Table } from "@tanstack/react-table"
+
+export interface DataTablePaginationProps<D> {
+  instance: Table<D>
+  pageSizeOptions?: number[]
```
seaerchin, line 6: nit: should we remove this? doesn't seem to be used

apps/studio/src/components/Datatable/InfoCell.tsx
```diff
@@ -0,0 +1,39 @@
...
+import NextLink from "next/link"
+import { Link, Text, VStack } from "@chakra-ui/react"
+
+export interface InfoCellProps {
+  caption?: string | null
+  subcaption?: string | null
+  href?: string
+}
+
+export const InfoCell = memo(
```
seaerchin, line 11: wow TIL there's such a thing as `memo`

apps/studio/src/features/dashboard/components/ResourceTable/LastEditCell.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <VStack align="start" spacing="0.25rem" textStyle="caption-2">
+      <Text title={email} noOfLines={1} color="base.content.strong">
+        {email}
+      </Text>
+      <Text
+        title={date.toLocaleString()}
+        noOfLines={1}
+        color="base.content.medium"
+      >
+        {formatDistance(date, new Date(), { addSuffix: true })}
```
seaerchin, line 33: wow `formatDistance` is goated

apps/studio/src/features/dashboard/components/ResourceTable/LastEditCell.tsx
```diff
@@ -0,0 +1,37 @@
...
+  date,
+  email,
+}: LastEditCellProps): JSX.Element => {
+  if (date === "folder") {
+    return (
+      <Text as="span" textStyle="caption-2">
+        -
+      </Text>
+    )
+  }
```
seaerchin, line 21: alternatives we can consider: 
1. showing last edited time of direct child pages
2. showing last edited time of all child pages + folder

i think here just using `-` is actually good enough but just cc @sehyunidaaa if you care. i think using `-` is actually really good for me and i hate 2 + dislike 1.

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface ResourceTableMenuProps {
+  title: ResourceTableData["name"]
+  resourceId: ResourceTableData["id"]
+  type: ResourceTableData["type"]
+}
+
+export const ResourceTableMenu = ({ title, type }: ResourceTableMenuProps) => {
+  return (
+    <Menu isLazy size="sm">
```
seaerchin, line 22: how come `Menu` can `isLazy`

seaerchin (commented):
what do we feel about using a bottom up approach to building the types rather than using a top-down approach (everything is using `ResourceTableData`) here - i think this is cos we haven't locked in the types yet so i think it's ok for now

i think the `Datatable` component here might still be doing abit too much here, it feels like we're trying to export a big component that can do everything and i think allowing a bit more flexibility might make more sense here cos it feels like a thin wrapper anw

apps/studio/src/components/Datatable/Datatable.tsx
karrui: cannot use suspense. We do not want the full fallback. We want to show the previous items in the table still, which suspense will not do (since it will swap out the current rendered table items with the suspense item completely). 
That is also the reason why data fetching is still done with `useQuery` and not `useSuspenseQuery`.

The loading state is a semi-transparent overlay.

apps/studio/src/components/Datatable/Datatable.tsx
karrui: copy paste error but actually don't really need in this use case. Will remove. Nested ternaries sometimes no choice inside react components lol. how you going to rewrite this.

but in anycase removing this whole thing cause Isomer's implementation does not need it haha

apps/studio/src/components/Datatable/Datatable.tsx
karrui: removing for now, can add back if we need to in the future. Current isomer's implementation does not need it yet.
(cell id provided by react-table. suggest reading react table docs on styling)

apps/studio/src/components/Datatable/Datatable.tsx
karrui: we not building a design system, so no need for components to be so generic lol. We can always refactor this when we get multiple distinct usecases of datatable (though i suspect we should be consistent with our tables)

apps/studio/src/components/Datatable/Datatable.tsx
karrui: agree to disagree. premature optimisation. we can do it only if we have different tables in design, which seems to not be the case now.

apps/studio/src/components/Datatable/DatatablePagination.tsx
karrui: thx for the catch. removed.

apps/studio/src/components/Datatable/InfoCell.tsx
karrui: suggest using friday learning to read [react docs](https://react.dev/reference/react/memo) hahah. The docs are really very good.

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
karrui: https://v2.chakra-ui.com/docs/components/menu/usage#lazily-mounting-menuitem

suggest reading docs first before raising comments hahaha if not looks like need to spoonfeed :P

seaerchin (approved):
lgtm! some points which i didn't agree with initially but landed on a good middle ground fokr the future: 
1. presenting a powerful API for data-table makes sense here because our use-case here is very narrow. we don't foresee this component changing anytime soon so ok to just hide stuff inside here. 
2. `Suspense` won't quite work here because the loading state which we want to show is dependent on the current data, which `Suspense` doesn't handle well (it wants a `fallback` component). 

again, nbd on both points - lgtm!

karrui:
### Merge activity

* **Jul 18, 11:17 PM EDT**: @karrui started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/313).

PR #315, opened 2024-07-16
fix: decouple template package.json for building sites

dcshzj:
## Problem

Now it is not possible to build the next sites because we have a dependency on the `@isomer/eslint-config` and other stuff.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Decouple this dependency for the time being, since the CodeBuild process we will be using in the future will use the entire monorepo.


harishv7 (approved):

harishv7:
temporarily decoupling till we get codebuild and pipeline up

PR #316, opened 2024-07-16
fix: remove null types from page content

dcshzj:
## Problem

The `Type.Null` is too strongly typed, causing developer grief.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Change all instances of `Type.Null` into `IsomerComponentsSchemas` to standardise with the other page schemas. We still enforce that it is an empty array via the `minItems` and `maxItems` props.

### DIFF
--- a/packages/components/src/types/schema.ts
+++ b/packages/components/src/types/schema.ts
@@ -50,7 +50,7 @@ export const ArticlePageSchema = Type.Object({
 export const CollectionPageSchema = Type.Object({
   layout: Type.Literal(ISOMER_PAGE_LAYOUTS.Collection),
   page: CollectionPageMetaSchema,
-  content: Type.Array(Type.Null(), {
+  content: Type.Array(IsomerComponentsSchemas, {
     title: "Page content",
     description:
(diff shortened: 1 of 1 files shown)


packages/components/src/types/schema.ts
```diff
@@ -50,7 +50,7 @@ export const ArticlePageSchema = Type.Object({
 export const CollectionPageSchema = Type.Object({
   layout: Type.Literal(ISOMER_PAGE_LAYOUTS.Collection),
   page: CollectionPageMetaSchema,
-  content: Type.Array(Type.Null(), {
+  content: Type.Array(IsomerComponentsSchemas, {
```
harishv7, line 53: What's the issue with having it as null?

dcshzj: Cos we would have two types, effectively `IsomerComponents[] | null`, then we need to separately validate in code that the layout is not `collection`, `file` or `link`. Standardising this means that we don't need to perform this validation in code ourselves and we can leave it to Ajv to perform the full validation for us.

PR #317, opened 2024-07-16
feat(frontend): add prose-renderer 

seaerchin:
## Problem
We want something to render a prose block in json forms so that stuff liek accordion can use it. 

Closes [ISOM-1217]

## Solution
1. add a `format: prose` key to our json schema
2. export types from `components` so we can have easier type referencing
3. refactor our tiptap component out into a common shared component so that we can reuse this component across jsonforms and our tiptap component

## Screenshots and video 

https://github.com/user-attachments/assets/61beabf6-609c-4d45-b5c0-fc402b70995d


seaerchin:
cb why format fail

karrui:
will review after you resolve conflicts :P

karrui:
your formatter not working properly. mine is formatting properly when I checkout your branch.

seaerchin:
vercel deploy failing because existing db schema isn't compat with the new one. we need to run a hard reset (or do expand -> contract as suggested by prisma)

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -117,10 +128,15 @@ function ComponentSelector() {
     const nextPageState = !!newComponent
       ? [...savedPageState, newComponent]
       : savedPageState
-    setSavedPageState(nextPageState)
+
     setDrawerState({ state: nextState })
     setCurrActiveIdx(nextPageState.length - 1)
     setPreviewPageState(nextPageState)
+
+    // TODO: Decide if this is a good idea...
```
harishv7, line 136: what's this todo? what are the alternatives to evaluate if this is a good idea?

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -19,15 +19,91 @@ export const DEFAULT_BLOCKS: Record<
       },
     ],
   },
-  accordion: undefined,
-  button: undefined,
-  callout: undefined,
+  accordion: {
```
harishv7, line 22: the purpose of this defintion here is to provide default values?

harishv7: It seems like most of the defaults are empty though. Would a better way to be to see the key e.g. `text` or `title` or `url` and then populate it?

harishv7 (commented):
minor comments

apps/studio/src/components/PageEditor/constants.ts
seaerchin: i think zhong jun had a PR fixing this. will remove on my end. we shouldn't do thsi approach (introspection of the object and then setting based on key) because these defaults are set broadly and are also very far removed from definition. 

i think the correct approach here would be to ask the schema itself for defaults (which it is able to do), so that each schema comes w/ sane defaults.

regardless, no impact here cos removing ._.

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: i'll add more context to the comment (and here): right now, we're doing an `if` condition on `setAddedBlock` because we want the typing to be correct (ie, `addedBlock` is only for complex blcoks, which disallows `prose`) 

what this means practically is that we get alot of `if/else` scattered over the place. if we combine, we'd need to handle the `prose` case for components. 

i decided on this approach cos it's simple for now

karrui:
conflict

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -103,7 +104,17 @@ function ComponentSelector() {
     setDrawerState,
     setSavedPageState,
     setPreviewPageState,
+    setAddedBlock,
   } = useEditorDrawerContext()
+
+  // TODO: get this dynamically
+  const pageId = 1
+  const siteId = 1
```
karrui, line 112: i think we can now pass in these as props for this component from the page that uses this (or put it into `EditorDrawerContext`)

apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -1,14 +1,17 @@
...
 
+import type { SectionType } from "~/components/PageEditor/types"
 import { type DrawerState } from "~/types/editorDrawer"
 
 export interface DrawerContextType {
   currActiveIdx: number
   setCurrActiveIdx: (currActiveIdx: number) => void
   drawerState: DrawerState
   setDrawerState: (state: DrawerState) => void
+  addedBlock: Exclude<SectionType, "prose">
```
karrui, line 13: comment: i feel like we need to rethink having so many derived states using `useState`, but out of scope of this PR.
the context should only hold the editing state, and the SSOT of the current blocks from the server.
editing state like:
- is editing (+ what ID is currently being edited, which can be then retrieved from the blocks)
- is creating

feel like these two states enough already. all the intermediate stuff like drawerState, addedBlock, pageState can theoretically be derived from the server. but nvm ba still ok for now. just very scared of desync from server state

apps/studio/src/features/editing-experience/utils.ts
```diff
@@ -1 +1,24 @@
+import type { IsomerComponent, ProseProps } from "@opengovsg/isomer-components"
+import { getComponentSchema } from "@opengovsg/isomer-components"
+import Ajv from "ajv"
+
 export const dataAttr = (value: unknown) => (!!value ? true : undefined)
+
+const proseSchema = getComponentSchema("prose")
+
+export const inferAsProse = (component?: IsomerComponent): ProseProps => {
```
karrui, line 9: is this used anywhere? if not can delete

karrui (approved):
lgtm

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: updated!

i think the drawer context shouldn't be so bloated. once the 2 PRs from me go in, i'll write a PR to refactor the callsites of `useQueryParse` now so that it'll be done via a context

apps/studio/src/contexts/EditorDrawerContext.tsx
seaerchin: this is actually one of my pain points too - i think we should keep a lean context and try to either derive or use backend state. i'm not quite sure how to do it properly here, but i think this bloat is still ok for now - biggest offenders for me are `preview/savedState`. the UI only states are ok to keep

apps/studio/src/features/editing-experience/utils.ts
seaerchin: removed!

PR #321, opened 2024-07-17
ISOM-1258: add-site-and-move-procedures

harishv7:
## Problem

Currently, no procedure catering for move of resource and setting site configs.
Closes [insert issue #]

## Solution

Adds procedures to move resources (single), and set site config (override fully) and set site notification (minimum requirement of site settings).

TODO: changes in site configs should trigger a immediate publish

**Breaking Changes**

  - Details ...

### DIFF
--- a/apps/studio/src/schemas/site.ts
+++ b/apps/studio/src/schemas/site.ts
@@ -4,7 +4,11 @@ export const getConfigSchema = z.object({
   id: z.number().min(1),
 })
 
+export const getNotificationSchema = z.object({
+  siteId: z.number().min(1),
+})
+
 export const setNotificationSchema = z.object({
-  id: z.number().min(1),
-  notificationStr: z.string(),
+  siteId: z.number().min(1),
+  notification: z.string(),
 })
--- a/apps/studio/src/schemas/site.ts
+++ b/apps/studio/src/schemas/site.ts
@@ -10,5 +10,5 @@ export const getNotificationSchema = z.object({
 
 export const setNotificationSchema = z.object({
   siteId: z.number().min(1),
-  notification: z.string(),
+  notification: z.string().min(1),
 })
--- a/apps/studio/src/schemas/site.ts
+++ b/apps/studio/src/schemas/site.ts
@@ -3,3 +3,8 @@ import { z } from "zod"
 export const getConfigSchema = z.object({
   id: z.number().min(1),
 })
+
+export const setNotificationSchema = z.object({
+  id: z.number().min(1),
+  notificationStr: z.string(),
+})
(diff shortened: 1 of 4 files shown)


apps/studio/src/schemas/site.ts
```diff
@@ -3,3 +3,8 @@ import { z } from "zod"
 export const getConfigSchema = z.object({
   id: z.number().min(1),
 })
+
+export const setNotificationSchema = z.object({
+  id: z.number().min(1),
+  notificationStr: z.string(),
```
seaerchin, line 9: I think just calling it notification is enough, wdyt? 

Also, id is abit ambiguous here but since narrow usage context, still ok with it. I do have to ask - how do we identify what site/resource/user a particular notification belongs to?

Lastly, this seems to suggest we allow empty notifications! Could I double check this assumption?

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -26,3 +26,30 @@ export const getSiteConfig = async (siteId: number) => {
...
+
+// TODO: Should triger immediate re-publish of site
+export const setSiteNotification = async (
+  siteId: number,
+  notificationStr: string,
+) => {
+  return db
+    .updateTable("Site")
+    .set({
+      config: sql`jsonb_set(config, '{"notification"}', to_jsonb(${notificationStr}::text))`,
```
seaerchin, line 51: Since we set the key in place, this might impact @karrui's PR that does the type gen for our JSON blobs. Could we check-in with him on this?

seaerchin (approved):
Minor nits on schema, stuff looks ok otherwise.

Things we've implicitly assumed here are that we don't care about past notifications (we eagerly override) and that we'll only ever have 1 notification at a time.

Since we only have 2 sites ATM and it's only a single key, I'm ok with it for now!

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -22,4 +22,11 @@ export const siteRouter = router({
       const { id } = input
       return getNavBar(id)
     }),
+  setNotification: protectedProcedure
```
hanpuliu-charles, line 25: is there also a way to fetch the current notification(if any) set on the site? getConfig doesn't seem to provide this functionality yet it looks like

apps/studio/src/server/modules/site/site.service.ts
karrui: its ok i can fix it up if it's affected. just merge first if you think its ready i can rebase easily with Graphite TM

apps/studio/src/schemas/site.ts
harishv7: 1. Sure, renaming it
2. I followed the convention in the other methods, we seem to use id to indicate siteId. But agreed, I think it should be explicit, will rename. We can rename the other schemas in a separate PR
3. I thought to allow it, but on hindsight, we should enforce a min validation of at least 1 char for practical usage

apps/studio/src/server/modules/site/site.router.ts
harishv7: Will add it

PR #324, opened 2024-07-18
feat(dnd): allow optimistic loading

seaerchin:
## Problem
We want to allow users to load stuff optimistically so that they have a good editing experience. The way this is done is that we send a tuple `(from, to)` to our backend. We then eagerly update the frontend first and reconcile + shoiw a toast if there are errors on our backend. 

## Solution
1. On drop, we update the `pageState` on the frontend and send a triplet `(from, to, block)` to our backend 
2. On our backend, we validate that **the state we receive is the state that we have** (got other checks but this is the relevant one). Only if they pass this check and have valid bounds for the `from, to` tuple, we save the updated state into our db. 

1. refactoring the `Blob` table into `Blob + Blocks` - this is quite a big change and affects other PRs also. we should only do this change next time, when our content grows to a huge size and users' network cannot support this traffic
2. sending a content hash across vs just sending the raw blob. not very sure if we need the content hash now, the business problem here is that we only want to allow saves iff their current state is updated. we should try this if our network is the limiting factor.

## Screenshots and videos

https://github.com/user-attachments/assets/cb1d4039-2e19-4db6-af5d-ecf02fc900e8


seaerchin:
FORMATTINGF FAILING AGAIN

karrui:
your formatter is not working. Mine is formatting as expected. did you run on root?

```bash
 npm run lint:fix && npm run format:fix
```

Pushing to your branch to fix the formatting.

karrui:
will review when you fix merge conflicts through this chain. if not a lot of unnecessary format changes

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -29,6 +29,7 @@ import {
```
harishv7, line 29: Some qns on UX - shouldn't we be showing a toast when save is successful and when it is not?

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -104,6 +105,14 @@ function ComponentSelector() {
...
+  const { mutate } = trpc.page.updatePageBlob.useMutation()
+  // TODO: get this dynamically
+  const pageId = 1
+  const siteId = 1
+  const [page] = trpc.page.readPageAndBlob.useSuspenseQuery({
+    pageId,
+    siteId
+  })
   const onProceed = (sectionType: SectionType) => {
     // TODO: add new section to page/editor state
```
harishv7, line 117: Sorry didn't understand the TODO here, can explain?

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -104,6 +105,14 @@ function ComponentSelector() {
...
     // NOTE: Only paragraph should go to tiptap editor
```
harishv7, line 118: basically prose right?

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -24,20 +27,59 @@ export default function RootStateDrawer() {
...
+
+    if (
+      from >= savedPageState.length ||
+      to >= savedPageState.length ||
+      from < 0 ||
+      to < 0
+    )
+      return
+
+    console.log("saved", savedPageState)
```
harishv7, line 50: might have forgotten to remove?

apps/studio/src/schemas/page.ts
```diff
@@ -13,6 +13,34 @@ export const getEditPageSchema = z.object({
...
+export const updatePageSchema = getEditPageSchema.extend({
+  // NOTE: We allow both to be empty now,
+  // in which case this is a no-op.
+  // We are ok w/ this because it doesn't
+  // incur any db writes
+  parentId: z.number().min(1).optional(),
+  pageName: z.string().min(1).optional(),
+})
+
+export const updatePageBlobSchema = getEditPageSchema.extend({
```
harishv7, line 39: why does this schema not require a siteId?

apps/studio/src/schemas/page.ts
```diff
@@ -13,6 +13,34 @@ export const getEditPageSchema = z.object({
   siteId: z.number().min(1),
 })
 
+export const reorderBlobSchema = z.object({
+  pageId: z.number().min(1),
+  from: z.number().min(0),
+  to: z.number().min(0),
+  siteId: z.number().min(0),
```
harishv7, line 20: why is this min 0 here?

apps/studio/src/schemas/page.ts
```diff
@@ -13,6 +13,34 @@ export const getEditPageSchema = z.object({
...
+  blocks: z.array(
+    z
+      .object({
+        type: z.string(),
+      })
+      .passthrough(),
+  ),
+})
+
+export const updatePageSchema = getEditPageSchema.extend({
```
harishv7, line 30: Why is siteId required for updatePageBlobSchema but not here?

apps/studio/src/schemas/page.ts
```diff
@@ -13,6 +13,34 @@ export const getEditPageSchema = z.object({
...
+  // NOTE: We allow both to be empty now,
+  // in which case this is a no-op.
+  // We are ok w/ this because it doesn't
+  // incur any db writes
+  parentId: z.number().min(1).optional(),
+  pageName: z.string().min(1).optional(),
+})
+
+export const updatePageBlobSchema = getEditPageSchema.extend({
+  content: z.string(),
```
harishv7, line 40: this is the stringified json?

apps/studio/src/schemas/page.ts
```diff
@@ -13,6 +13,34 @@ export const getEditPageSchema = z.object({
...
+  ),
+})
+
+export const updatePageSchema = getEditPageSchema.extend({
+  // NOTE: We allow both to be empty now,
+  // in which case this is a no-op.
+  // We are ok w/ this because it doesn't
+  // incur any db writes
+  parentId: z.number().min(1).optional(),
+  pageName: z.string().min(1).optional(),
```
harishv7, line 36: is pageName the only thing being modified in this updatePageSchema?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -100,6 +110,79 @@ export const pageRouter = router({
...
+        })
+      }
+
+      const actualBlocks = fullPage.content.content
+
+      if (!isEqual(blocks, actualBlocks)) {
+        throw new TRPCError({
+          code: "CONFLICT",
+          message:
+            "Someone on your team has changed this page, refresh the page and try again",
```
harishv7, line 136: This page has been recently updated. Please refresh the page to fetch the latest changes and try again.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -100,6 +110,79 @@ export const pageRouter = router({
...
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message:
+            "Unable to load content for the requested page, please contact Isomer Support",
+        })
+      }
+
+      const actualBlocks = fullPage.content.content
+
+      if (!isEqual(blocks, actualBlocks)) {
```
harishv7, line 132: I think for now is fine, but in future we might want to explore comparing hashes. Should see which works better with large jsons as well

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -100,6 +110,79 @@ export const pageRouter = router({
...
+        })
+      }
+
+      if (
+        from >= actualBlocks.length ||
+        to >= actualBlocks.length ||
+        from < 0 ||
+        to < 0
+      ) {
+        // NOTE: If this happens, this indicates that either our dnd libary on our frontend has a
```
harishv7, line 146: has a bug*?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -100,6 +110,79 @@ export const pageRouter = router({
...
+        throw new TRPCError({ code: "UNPROCESSABLE_CONTENT" })
+      }
+
+      const [movedBlock] = actualBlocks.splice(from, 1)
+      if (!movedBlock) return blocks
+      if (!fullPage.draftBlobId || !fullPage.mainBlobId) {
+        throw new TRPCError({ code: "NOT_FOUND" })
```
harishv7, line 153: would a msg be useful here?

harishv7: Also generic comment, we explored structuring comments in one of our early RFCs on Isomer. For future implementation, but we can think about having internal error codes which we can map to for errors

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -98,31 +98,42 @@ export const updatePageById = (
...
-  })
+  return db
+    .updateTable("Resource")
+    .set(rest)
+    .where("id", "=", id)
+    .executeTakeFirstOrThrow()
 }
 
-export const updateBlobById = (props: {
+// NOTE: This id here is the page id
```
harishv7, line 108: might be just rename var to be explicit instead?

harishv7: shouldn't we also use siteId for the fetch?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -98,31 +98,42 @@ export const updatePageById = (
...
       .where("Resource.id", "=", id)
+      // NOTE: We update the draft first
+      // Main should only be updated at build
+      .select("draftBlobId")
       .executeTakeFirstOrThrow()
-  )
+
+    if (!page.draftBlobId) {
+      // NOTE: no draft for this yet, need to create a new one
+      return tx.insertInto("Blob").values({ content }).executeTakeFirst()
```
harishv7, line 125: after creating draftBlob, we need to update resource's draftBlobId to this newly created id right?

harishv7 (commented):
added comments

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: this doesn't save

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: oudated, removed

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: yes

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: removed

apps/studio/src/schemas/page.ts
seaerchin: done

apps/studio/src/schemas/page.ts
seaerchin: actually idt we need for both but @karrui added on the `pageBlob` but not here

apps/studio/src/schemas/page.ts
seaerchin: y7es

apps/studio/src/schemas/page.ts
seaerchin: for now yes

apps/studio/src/server/modules/page/page.router.ts
seaerchin: this has been cleared with @sehyunidaaa already, so idt we should change

apps/studio/src/server/modules/page/page.router.ts
seaerchin: we were going to use hashes but 
1. no good shared lib between fe/be (`crypto` only backend) 
2. back of envelope calculation suggests that max size is ~15kb (3000 words each of length 5), which is ~10ms with good wifi (tested against office) 

i think there's a point to be made for hashing but we shuold do it next itme

apps/studio/src/server/modules/page/page.router.ts
seaerchin: rebase error, updated

apps/studio/src/server/modules/page/page.router.ts
seaerchin: sure, added a message.

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: each blob is linked to 1 resource, which is linked to 1 site. don't really see how `siteId` would be used here - could you elaborate?

also renamed to `pageId`

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: good catch, updated

apps/studio/src/components/PageEditor/ComponentSelector.tsx
harishv7: > Only if they pass this check and have valid bounds for the from, to tuple, we save the updated state into our db.

It's mentioned you save the state in db?

apps/studio/src/schemas/page.ts
harishv7: I think we should standardise to query with siteId + resourceId

apps/studio/src/server/modules/page/page.router.ts
harishv7: sgtm

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -98,31 +98,43 @@ export const updatePageById = (
...
+  return db
+    .updateTable("Resource")
+    .set(rest)
+    .where("id", "=", id)
+    .executeTakeFirstOrThrow()
 }
 
-export const updateBlobById = (props: {
-  id: number
+// NOTE: This id here is the page id
```
harishv7, line 108: can remove the comment alr

apps/studio/src/server/modules/resource/resource.service.ts
harishv7: Basically suggestion is to standardise our queries - i.e. to use both siteId and pageId to query in the where clause. 

where siteId = XYZ AND resource.Id = ABC

harishv7 (commented):
minor comment

apps/studio/src/schemas/page.ts
seaerchin: this has `siteId` - `getEditPageSchema` has `siteId + pageId`

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: ok sure

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: done

apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -104,8 +105,15 @@ function ComponentSelector() {
     setSavedPageState,
     setPreviewPageState,
   } = useEditorDrawerContext()
+  const { mutate } = trpc.page.updatePageBlob.useMutation()
+  // TODO: get this dynamically
+  const pageId = 1
+  const siteId = 1
```
karrui, line 111: suggest to take this in as a prop also instead of retrieving from query params, since we don't want to implicitly tie this component to the page.
can see other examples of `useQueryParse` or something i forgot the name

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -24,20 +27,58 @@ export default function RootStateDrawer() {
     setSavedPageState,
     setPreviewPageState,
   } = useEditorDrawerContext()
+  const router = useRouter()
+  const pageId = Number(router.query.pageId)
+  const siteId = Number(router.query.siteId)
```
karrui, line 32: Same as above, suggest to take this in as a prop also instead of retrieving from query params, since we don't want to implicitly tie this component to the page.
can see other examples of `useQueryParse` in the current codebase.

```ts
// in some page/path/to/page.ts
const {pageId, siteId} = useQueryParse(editPageSchema)

// Pass into this RootStateDrawer (or maybe even EditorDrawerContext)
```

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -24,20 +27,58 @@ export default function RootStateDrawer() {
...
+  const { mutate } = trpc.page.reorderBlock.useMutation()
+  const toast = useToast({ status: "error" })
 
   const onDragEnd = (result: DropResult) => {
```
karrui, line 36: nit: feels like a heavy function, may want to use `useCallback`

apps/studio/src/schemas/page.ts
karrui: agree should be both right. each page has pageId and siteId. Even though we know that those ids are all unique, we should still pass it to the server (for future authorisation checks)

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -2,16 +2,25 @@ import type { IsomerSchema } from "@opengovsg/isomer-components"
 import { schema } from "@opengovsg/isomer-components"
 import { TRPCError } from "@trpc/server"
 import Ajv from "ajv"
-import { z } from "zod"
+import { isEqual } from "lodash"
```
karrui, line 5: nit: unsure if this can be treeshaken. Suggest using lodash's explicit export.

```suggestion
import isEqual from "lodash/isEqual"
```

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -100,6 +110,84 @@ export const pageRouter = router({
         ...siteMeta,
       }
     }),
+
+  reorderBlock: protectedProcedure
```
karrui, line 114: suspect this whole function should be wrapped in a transaction. We want to ensure that there is no read/write skew when we are validating the content.

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: updated - using the one you've done in the base page. i felt that made more sense than defining my own because this is a component on the page itself

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: assuming this is for `onDragEnd`

apps/studio/src/server/modules/page/page.router.ts
seaerchin: done

apps/studio/src/server/modules/page/page.router.ts
seaerchin: could you elaborate in abit more detail? not too sure on the portion re: read/write skew - my interpretation is that someone updates between the read and our write.

if that's the case, i'm actually ok not wrapping in tx because i suspect that the time between read/write is going to be so minute (no async calls, pure CPU + it's mostly just sanity checks) that this occurs v rarely

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: not very sure i understand you here - i agree on takign this as a prop but then if we do that, we shouldn't utilise `useQueryParse` right? 

i'll go with ^ and pass the props in via parent

apps/studio/src/components/PageEditor/ComponentSelector.tsx
seaerchin: just selecting the component itself doesn't do anything, we require them to drag/drop in order to display feedback! that part is done via a toast in the mutation itself in `rootstatedrawer`!

apps/studio/src/schemas/page.ts
seaerchin: yeap, i'm fine with having both. i was thinking we could do the retrieval implicitly (ie, if you query for resource A with siteId 2, we assume your `siteId = 2`) but i think having it so that resource authorisation checks can be simplified as middleware makes sense.

my point above was that the schema itslef has `siteId` because it's extending a schema with `pageId + siteId`
image

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -25,20 +30,66 @@ export default function RootStateDrawer() {
...
+
+      if (!!movedBlock) {
+        updatedBlocks.splice(to, 0, movedBlock)
+        setPreviewPageState(updatedBlocks)
+        setSavedPageState(updatedBlocks)
+      }
+
+      // NOTE: drive an update to the db with the updated index
+      mutate({ pageId, from, to, blocks: savedPageState, siteId })
+    },
```
karrui, line 82: i think `onDragEnd` should only call `mutate` instead of adding the stateful logic inside this function. 

For optimistic updates, we should use the `onMutate` function `react-query` explicitly provides. See [docs](https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates).

```suggestion
  const onDragEnd = useCallback(
    (result: DropResult) => {
      if (!result.destination) return

      const from = result.source.index
      const to = result.destination.index

      if (
        from >= savedPageState.length ||
        to >= savedPageState.length ||
        from < 0 ||
        to < 0
      )
        return

      // NOTE: drive an update to the db with the updated index
      mutate({ pageId, from, to, blocks: savedPageState, siteId })
    },
```

apps/studio/src/server/modules/page/page.router.ts
karrui: skew (which i realize just wrapping transactions don't guard against) can occur when:

1. Person 1 reorders block from A to A1 state
  - this function fetch full page A
2. Person 2 reorders block from A to A2 state
  - fetch full page A
  - equality check succeeds
  - page A -> A2 state
3. P1's reorder somehow slower than P2
  - equality check still succeeds, since P1 fetched state A before P2's reordering finished
  - page A2 -> A1 state (invalid, but allowed)

Should have been thrown instead of this invalid state. May cause drifting yeah.

Suggest using a transaction with a [`FOR UPDATE`](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS) clause.

this can be done in kysely just by adding a `.forUpdate` in the query clause when fetching the full page.

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: you're right and i did this somewhere else

seaerchin: misunderstood - you're still right!

seaerchin: i just tried this - this _doesn't_ quite work. this is because the hook gets fired `onDragEnd`. this results in the series of events below: 

(lets assume `[A, B, C]`)
1. user drags and drops a block (now it should logically be `[B, A, C]`)
2. `onDragEnd` gets called to mutate **but state not updated yet**. hence, the state flips back to `[A, B, C]`
3. `onMutate` gets fired 
4. state updated, now we see `[B, A, C]`

practically, this means that there's a moment (albeit very brief) where the order gets messed up before being rearranged. hence, we'd need to perform the state update in the `onDragEnd` hook itself - we cannot do it outside because it'll cause a re-render, leading to the flicker!

i do agree, though, that this isn't ideal. i've committed my last change and i think we can take a look tomorrow or i'll just drop the commit if this is non-ideal

https://github.com/user-attachments/assets/95df71b6-2513-4c2f-8b49-895b4e97ed8c

apps/studio/src/server/modules/page/page.router.ts
seaerchin: i added the `forUpdate` clause! not very sure i understand how it works - read your link and [this](https://www.cockroachlabs.com/blog/select-for-update/) here, it seems like it's basically locking a row for read?

in our case above, wouldn't the read still succeed then the write fail cos they're separate? lets clarify this tomorrow maybe i grug brain

apps/studio/src/components/PageEditor/ComponentSelector.tsx
harishv7: Hmm I am referring to your demo video where you are doing drag and drops of component blocks. Shouldn't each drag and drop show a toast saying the action has saved successfully?

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
harishv7: Why not show a loading state on the Custom blocks section while the mutate is ongoing

apps/studio/src/schemas/page.ts
harishv7: One comment here is that the siteId should come from the user's auth context which defines what sites they have access to, and not from user's input

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -84,6 +85,7 @@ export const getFullPageById = async (args: {
     .where("Resource.mainBlobId", "is not", null)
     .innerJoin("Blob", "Resource.mainBlobId", "Blob.id")
     .select(defaultResourceWithBlobSelect)
+    .forUpdate()
```
harishv7, line 88: under what context are we using forUpdate? Isn't this to be used within a transaction? I don't think this is a transaction right?

harishv7 (commented):
added comments

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: I actually am not sure if `forUpdate` is meant to be used in a tx. 

Trying it on local works but if what you say is correct, then I think the only choice here would be to write a wrapper method with `forUpdate` in a tx

seaerchin: Kysely docs don't provide info on this too

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: Wouldn't the flickering still happen? Unless we have loading state in the whole section, then it's very disruptive

harishv7: While a state change is in progress, we technically do not want the user to perform more changes right? If so, shouldn't we block further drag and drops before the previous one fully resolves? Or is that assumption invalid?

apps/studio/src/server/modules/resource/resource.service.ts
harishv7: Hmm locking mechanics happen within a transaction right

> FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update. This prevents them from being locked, modified or deleted by other transactions until the current transaction ends. 

Mentions "current transaction ends"

@karrui can advise here?

harishv7: So usually with SQL we do smth like 

```
BEGIN tx

Select clause with forUpdate

do the update

COMMIT tx
```

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
seaerchin: i think both ways work! i was thinking we should allow updates optimistically (when you start processing the new update, the old update finished alr) because the network would be the bottleneck here, not our db. 

i wanted to show a toast and rollback if changes don't go through - so that in the majority of cases (assumption: network latency >> db latency), your dnd isn't blocked and it's a seamless exp.

i think if we do the block editing, **we should only block drag and drop** rather than having the whole drawer be blocked. 

for consideration, a 15kb payload took 9ms on office wifi and ~2s on fast 3g.

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: actually lemme double check w/ kr on this - i'll update you later

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -102,6 +112,84 @@ export const pageRouter = router({
...
+        to < 0
+      ) {
+        // NOTE: If this happens, this indicates that either our dnd libary on our frontend has a
+        // bug or someone is trying to mess with our frontend
+        throw new TRPCError({ code: "UNPROCESSABLE_CONTENT" })
+      }
+
+      const [movedBlock] = actualBlocks.splice(from, 1)
+      if (!movedBlock) return blocks
+      if (!fullPage.draftBlobId || !fullPage.mainBlobId) {
```
karrui, line 155: i think this is a bug right. a page may have either a `draftBlobId` or a `mainBlobId`. One or the other may be null. This conditional will be wrongly truthy.

```suggestion
      if (!fullPage.draftBlobId && !fullPage.mainBlobId) {
```

apps/studio/src/server/modules/resource/resource.service.ts
karrui: yes. just follow postgres docs. forUpdate is specifically for a transaction. It locks any rows queries until the transaction ends, which is what we want to prevent other readers from accessing the row, preventing skew.

apps/studio/src/server/modules/page/page.router.ts
sehyunidaaa: ah ok good point, i like your suggestion! with minor lingo tweaks, maybe can do

> "This page was recently updated. Refresh the page to get the latest changes."

@seaerchin

apps/studio/src/server/modules/page/page.router.ts
seaerchin: yes you're right omg LEMME UPDATE

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
karrui: @seaerchin revert ba. Do in onDragEnd ba. Cannot avoid the flicker because it's always one render cycle behind. don't know why formsg no flicker, not worth it to debug now.

seaerchin: re: offline discussion - we're shifting it back to `onDragEnd`

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -24,6 +33,7 @@ const schemaValidator = ajv.compile<IsomerSchema>(schema)
 
 const validatedPageProcedure = protectedProcedure.use(
   async ({ next, rawInput }) => {
+    console.log(rawInput)
```
karrui, line 36: extra console.log

karrui (approved):
lgtm. Repeating the trpc call in so many places feels a little iffy to me but its out of scope of this PR i guess. We can deal with it In The Future™️

apps/studio/src/server/modules/page/page.router.ts
seaerchin: ok i remove rn

PR #329, opened 2024-07-19
feat: wire up all page states in editing experience

dcshzj:
## Problem

The various page states have not yet been wired up.

## Solution

**Breaking Changes**

**Features**:

- Wire up all the page states so the editing flow is more complete.
- What is NOT done/not working:
    - Adding of new blocks (will do in a separate PR, because need to adjust the schema to set default values)
    - Editing hero component (I think the schema is slightly incorrect, may update this PR later or do separately).

### DIFF
--- a/apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
+++ b/apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
@@ -37,13 +37,19 @@ interface TipTapComponentProps {
 }
 
 function TipTapComponent({ content }: TipTapComponentProps) {
-  const { setDrawerState, setPageState, currActiveIdx, snapshot } =
-    useEditorDrawerContext()
+  const {
+    setDrawerState,
+    savedPageState,
+    setSavedPageState,
+    previewPageState,
+    setPreviewPageState,
+    currActiveIdx,
+  } = useEditorDrawerContext()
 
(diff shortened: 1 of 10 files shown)


apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
```diff
@@ -100,6 +112,13 @@ function TipTapComponent({ type, data, path }: TipTapComponentProps) {
     content: data,
     onUpdate({ editor }) {
       // TODO: set editor state - content is retrieved via editor.getJSON().content
+      const { content } = editor.getJSON()
+
+      if (validateFn(content)) {
+        const newPageState = [...savedPageState]
+        newPageState[blockIndex] = content
+        setPreviewPageState(newPageState)
+      }
```
seaerchin, line 121: we don't need this validation here because internally, tiptap's schema conforms to our schema already. 

we also shouldn't do this because we are validating the full content recursively

seaerchin (approved):

apps/studio/src/features/editing-experience/components/TipTapComponent.tsx
dcshzj: Hmm I think we are not entirely guaranteed this fact cos we manually ensure that our schema conforms to Tiptap's schema, but yeah removing it for now cos we are already validating the full content.

PR #340, opened 2024-07-22
fix: enable unsafe-eval for user test

dcshzj:
## Problem

The Ajv compilation to create the validation function requires `unsafe-eval` to be enabled for `script-src`.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Enable `unsafe-eval` for script-src for the time being, so that we can proceed with user testing.
    - Ref: https://ajv.js.org/security.html#content-security-policy


seaerchin (approved):
lgtm but what happens in production next time?

PR #343, opened 2024-07-23
chore: add default blocks contents

dcshzj:
## Problem

The default blocks contents are currently set to undefined.

## Solution

**Breaking Changes**

**Improvements**:

- Add default block contents for all the blocks except for the hero component.
- Also set logger to false for Ajv since we don't use it for debugging (and it is very noisy).

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
@@ -55,6 +55,7 @@ const renderers: JsonFormsRendererRegistryEntry[] = [
     renderer: jsonFormsVerticalLayoutRenderer,
   },
 ]
+const ajv = new Ajv({ strict: false, logger: false })
 
 export default function FormBuilder(): JSX.Element {
   const {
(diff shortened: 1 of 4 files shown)


apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
```diff
@@ -76,7 +76,7 @@ export default function FormBuilder(): JSX.Element {
 
   const subSchema = getComponentSchema(component.type)
   const data = previewPageState[currActiveIdx]
-  const ajv = new Ajv({ strict: false })
+  const ajv = new Ajv({ strict: false, logger: false })
```
karrui, line 79: comment/perf(out of scope): Might want to move this out of the component/memoize this since it does not rely on any props. Might be a heavy operation, though I don’t think so la haha.

karrui (approved):

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
dcshzj: Updated in 572d8e6d22262160b5b8de972f8f729f6709658b

PR #345, opened 2024-07-23
fix: subcomponent overflowing in absolute position

dcshzj:
## Problem

Due to the use of absolute positioning, if the subcomponent is larger than the parent, it will cause the Save button to overlap.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Move the "Remove item" inside the absolute positioned drawer
- Added an overflow scroll behaviour
- Added a z-index on the absolute positioned drawer to hide the original form under it.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-07-23 at 14 48 28\

**AFTER**:

Screenshot 2024-07-23 at 14 47 20


seaerchin:
shouldn't the fix be to remove `absolute`?

dcshzj:
> shouldn't the fix be to remove absolute?

We can't remove it because otherwise, the subcomponent will not be able to be placed above the original drawer.

seaerchin:
some issues i noticed with trying it out: 

1. (unrelated to this PR): the root drawer has a different height from other drawers, such as the add block drawer. this means that when we click from the root drawer that displays our blocks to the add block drawer, there's an overflow and the scrollbar appears, causing layout shift
2. because we render in `absolute`, this drawer has to cover the drawer beneath it totally or otherwise, it'll show the below drawer. if it has greater width than the drawer it's covering, again there's layout shift, which is non-ideal. as such, we only accept if both widths are equal. however, this isn't the case at present (see below ss). additionally, i'd say that this coupling is bad for our application because it allows negative UX to slip in easily, especially if we update the wdith of 1 drawer but not the other.

I'm not super sure what's the ideal solution here - i'll take a stab at this w @karrui tomorrow but i think having another drawer here is definitely not good

image

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -29,7 +29,7 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
   const { title } = getComponentSchema(component.type)
 
   return (
-    <Box position="relative" h="100%" w="100%">
+    <Box position="relative" h="100%" w="100%" overflow="scroll">
```
karrui, line 32: nit: prefer `overflow="auto"` instead of scroll, if not scrollbar will forever be shown even if nothing is overflowing

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -275,27 +290,22 @@ export function JsonFormsArrayControl({
           path={composePaths(path, `${selectedIndex}`)}
           label={label}
           setSelectedIndex={setSelectedIndex}
+          isRemoveItemDisabled={minItems !== undefined && data <= minItems}
+          handleRemoveItem={handleRemoveItem(path, selectedIndex)}
```
karrui, line 294: nit: might want to update `handleRemoveItem` to also have a isDisabled check -- users can easily remove the disabled prop on client and make the button clickable

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -275,27 +290,22 @@ export function JsonFormsArrayControl({
...
-        <Button
-          onClick={handleRemoveItem(path, selectedIndex)}
-          variant="clear"
-          colorScheme="critical"
-          isDisabled={minItems !== undefined && data <= minItems}
-        >
-          Remove item
-        </Button>
+      {selectedIndex === undefined && (
+        <>
```
karrui, line 299: nit: unnecessary fragment

karrui (approved):

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
dcshzj: Updated in 7f2ee305b59a98499427b90037631637e6b43895.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
dcshzj: Updated in 7f2ee305b59a98499427b90037631637e6b43895.

PR #346, opened 2024-07-23
feat: add site settings page for editing site notification

hanpuliu-charles:
## Problem

Adds functionality for admins to edit site notification. 

Closes [ISOM-1258](https://linear.app/ogp/issue/ISOM-1258/edit-site-settings)

## Solution

Add layout and modify backend to retrieve and set notification. Sending empty string to the set notification endpoint will remove notification from the site. 

**Breaking Changes**

  - Details ...

- Details ...

## Before & After Screenshots

**AFTER**:

image
image
image
image


apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -50,15 +50,24 @@ export const getNotification = async (siteId: number) => {
 }
```
karrui, line 50: you can make `getNotification` typesafe, which removes the need for the cast:

```ts
    .select(eb => eb.ref('Site.config', '->').key('notification').as('notification'))
```

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -41,11 +41,12 @@ export const siteRouter = router({
       return getNavBar(id)
     }),
   getNotification: protectedProcedure
+    // type of notification is any.
     .input(getNotificationSchema)
     .query(async ({ input }) => {
       const { siteId } = input
       const notification = await getNotification(siteId)
-      return notification
+      return (notification as string) || ""
```
karrui, line 49: once `getNotification` is typesafe we should remove the cast. 

for learning (not needed in this case): generally if you are casting a primitive, might be better to enforce that primitive with its constructor like so:

```ts
return String(notification ?? "")
```

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,175 @@
...
+import { UnsavedSettingModal } from "~/features/editing-experience/components/UnsavedSettingModal"
 import { type NextPageWithLayout } from "~/lib/types"
 import { AdminCmsSidebarLayout } from "~/templates/layouts/AdminCmsSidebarLayout"
+import { trpc } from "~/utils/trpc"
 
 const SiteSettingsPage: NextPageWithLayout = () => {
-  return <div>Settings page</div>
+  const toast = useToast()
+  const router = useRouter()
+  const params = useParams<{ siteId: string }>()
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:27]: nit: added a `useQueryParse` hook somewhere here to parse queries or throw.
Can probably use it

```ts
// outside component
const siteSettingsSchema = z.object({
  siteId: z.coerce.number(),
})

// usage
const { siteId } = useQueryParse(siteSettingsSchema)
```

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,175 @@
...
+    onSuccess: () => {
+      setNotificationFieldChanged(false)
+      toast({
+        title: "Saved site settings!",
+        description: "Check your site in 5-10 minutes to view it live.",
+        status: "success",
+      })
+    },
+    onError: (error) => {
+      console.log(error)
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:39]: nit: add TODO to remove?

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,175 @@
...
+  }
+
+  const [previousNotification] = trpc.site.getNotification.useSuspenseQuery({
+    siteId: Number(params.siteId),
+  })
+
+  useEffect(() => {
+    setNotificationEnabled(previousNotification !== "")
+    setNotificationText(previousNotification)
+  }, [previousNotification])
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:73]: [don't sync state, derive it](https://www.notion.so/opengov/Don-t-sync-state-derive-it-173d074b374644e6bab52e578ad626a3?pvs=3&qid=)

we should not use useEffect wrongly. should we derive state/initialize state instead:

```ts
// notification text always starts with previousNotification
// initialise state with previousNotification
const [notificatonState, useNotificationState] = useState(previousNotification) // <- initialise state
const isNotificationEnabled = !notificationText // <- derive state


```

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,175 @@
...
+              </>
+            )}
+          </VStack>
+          <HStack justifyContent="flex-end" w="full" gap="1.5rem">
+            {saveEnabled && (
+              <Text textColor="base.content.medium" textStyle="caption-2">
+                Changes will be reflected on your site immediately.
+              </Text>
+            )}
+            <Button onClick={onClickUpdate} isDisabled={!saveEnabled}>
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:165]: actually looking at this flow, it is actually a form right. any reason why react hook form is not used? That gives us validation handling out of the box, which the current implementation does not do yet.

karrui (changes requested):

apps/studio/src/pages/sites/[siteId]/settings.tsx
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:175]: Hmm, no particular reason other than lack of awareness on my side on react hook form. Should I rework this to use react hook form?

Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:175]: Please do!

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,223 @@
...
+                {saveEnabled && (
+                  <Text textColor="base.content.medium" textStyle="caption-2">
+                    Changes will be reflected on your site immediately.
+                  </Text>
+                )}
+                <Button
+                  type="submit"
+                  // onClick={onClickUpdate}
+                  isDisabled={!saveEnabled}
+                >
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:211]: might want to add a `isLoading` prop for when the mutation is running.

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,223 @@
...
+    })
+  }
+
+  const { register, handleSubmit, watch, formState, reset } =
+    useForm<SettingFormValues>({
+      defaultValues: {
+        notificationEnabled: previousNotification !== "",
+        notificationText: previousNotification || "",
+      },
+    })
```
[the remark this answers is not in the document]
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx]: I've put in the changes to make use of zod for validation, but the schema is a little different between the trpc and the form, so I ended up with this 
`schema: setNotificationSchema
      .extend({ notificationEnabled: z.boolean() })
      .omit({ siteId: true })
      .refine((data) => !data.notificationEnabled || data.notification, {
        message: "Notification must not be empty",
        path: ["notification"],
      }),`
Will it be better defining a new schema by itself over keeping this extend, omit?
Or should the trpc endpoint actually take in both values(to keep schema consistent, at a cost of sending redundant data(empty notification already symbolize that notification should be off)?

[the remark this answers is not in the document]
Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx]: it's ok for extending so at least the expected client inputs changes in an expected manner. Can even extract the client schema its own constant if you want to keep it even cleaner:

```ts
const clientSetNotificationSchema = setNotificationSchema 
  .extend({ notificationEnabled: z.boolean() })
  .omit({ siteId: true })
  .refine((data) => !data.notificationEnabled || data.notification, { message: "Notification must not be empty", path: ["notification"], })
```

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,222 @@
...
+                        minLength: {
+                          value: 1,
+                          message: "Notification must not be empty",
+                        },
+                        maxLength: {
+                          value: 100,
+                          message:
+                            "Notification must be 100 characters or less",
+                        },
+                      })}
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:191]: since you are using a zod schema, the validation should be on the schema itself. You should update the schema instead of declaring them here:

```ts
const setNotificationSchema = z.object({
  // ...
  notification.max(100, { message: "Notification must be 100 characters or less" })
```

and the validation errors should appear without any additional config in react hook form:
```ts
{...register("notification")}
```

Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:191]: i realise there is some drift between what the client and backend schema expects. thinking about it more, why not just have them use the same schema? Server can also take in the `notificationsEnabled` boolean and handle accordingly. This may simplify a bunch of things.

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -27,23 +27,35 @@ export const setSiteConfig = async (
 export const getNotification = async (siteId: number) => {
   const result = await db
     .selectFrom("Site")
-    .select(sql`config ->> 'notification'`.as("notification"))
+    .select((eb) =>
+      eb.ref("Site.config", "->").key("notification").as("notification"),
+    )
     .where("id", "=", siteId)
     .executeTakeFirstOrThrow()
```
karrui, line 34: (out of scope, but should we fix this?) we might want to gracefully handle the error by not throwing in the DB and throw a `TRPCError` instead:

```ts
const result = await db...executeTakeFirst()
if (!result) {
  throw new TRPCError({
    code: "NOT_FOUND"
  })
}

return result.notification
```

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -27,23 +27,35 @@ export const setSiteConfig = async (
...
   notification: string,
 ) => {
-  return db
-    .updateTable("Site")
-    .set({
-      config: sql`jsonb_set(config, '{"notification"}', to_jsonb(${notification}::text))`,
-    })
-    .where("id", "=", siteId)
-    .executeTakeFirstOrThrow()
+  if (notification) {
```
karrui, line 46: instead of handling a null string implicitly (users of the function might be unaware of this implicit handling), might we want to separate this function into its `setSiteNotification` and `clearSiteNotification` variants instead? Then we can have the same client and server schema, and the procedure handler calls the appropriate db service function.

karrui (changes requested):

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,223 @@
...
+        notificationEnabled: notificationEnabled,
+        notificationText: notificationText,
+      })
+      toast({
+        title: "Saved site settings!",
+        description: "Check your site in 5-10 minutes to view it live.",
+        status: "success",
+      })
+    },
+    onError: (error) => {
```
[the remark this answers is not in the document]
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:52]: Hi! I've forgotten to comment on the reason for the reset in onSuccess. The reset(with the current form values) is intended to clear the isDirty flag(which is used for displaying the warning modal upon navigating away), else without it, after submitting the form, the isDirty flag will still be true and users still still be shown a warning upon navigating away from the page.

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,222 @@
...
+  const router = useRouter()
+  const trpcUtils = trpc.useUtils()
+  const { siteId } = useQueryParse(siteSettingsSchema)
+
+  const notificationMutation = trpc.site.setNotification.useMutation({
+    onSuccess: () => {
+      reset({
+        notificationEnabled: notificationEnabled,
+        notification: notification,
+      })
```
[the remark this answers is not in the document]
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:44]: Commented on this below(reset to clear isDirty)

apps/studio/src/pages/sites/[siteId]/settings.tsx
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:191]: Okay, I think I will prefer to standardize the backend and client schema. This will also eliminate the need for a clearnotification endpoint(since we explicit pass notificationsEnabled: false to clear the notification.)

apps/studio/src/pages/sites/[siteId]/settings.tsx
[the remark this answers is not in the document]
Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:44]: in that case, suspect we can keep the current values without explicitly passing it in:
`reset({}, { keepValues: true })`

[see docs](https://react-hook-form.com/docs/useform/reset#:~:text=to%20be%20subscribed.-,keepValues,-boolean)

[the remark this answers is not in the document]
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:44]: Was playing with this before commenting above, trying out various ways to write this reset [stackoverflow](https://stackoverflow.com/questions/63953501/react-hook-form-resetting-isdirty-without-clearing-form) but it only seems to work if I explicitly pass in new default values.

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,222 @@
...
+    "notification",
+  ])
+
+  const { isDirty, errors } = formState
+
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const [nextUrl, setNextUrl] = useState("")
+
+  useEffect(() => {
+    const handleRouteChange = (url: string) => {
```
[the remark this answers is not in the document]
Comment (hanpuliu-charles) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:89]: Hi Kar Rui, pardon me but am not too sure how to make use of the routeChangeErrror event, as I don't think I have any special actions to take upon route change failure. Glad to take in any input for this. Thanks!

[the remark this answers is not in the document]
Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/settings.tsx:89]: i think might not need to do anything in this case

apps/studio/src/pages/sites/[siteId]/settings.tsx
```diff
@@ -1,8 +1,210 @@
...
+
+  const [notificationEnabled, notification] = watch([
+    "notificationEnabled",
+    "notification",
+  ])
+
+  const { isDirty, errors } = formState
+
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const [nextUrl, setNextUrl] = useState("")
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/settings.tsx:86]: feels like doing double work. Should the "open state" of the modal be derived from whether `nextUrl` is falsey or not instead? No need for useDisclosure, e.g.:

```ts
const isOpen = !!nextUrl
const onClose = setNextUrl("")
```

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -51,7 +52,11 @@ export const siteRouter = router({
     .input(setNotificationSchema)
     .mutation(async ({ input, ctx }) => {
       const { siteId, notification } = input
-      await setSiteNotification(siteId, notification)
+      if (notification) {
```
karrui, line 55: since you have standardized schema, you can probably use `notificationEnabled` boolean for this instead of a falsey check

karrui (approved):
lgtm

PR #347, opened 2024-07-23
ref: remove component variants and update titles

dcshzj:
## Problem

Having variants for components makes design very complicated and a lot more combinations to polish.

## Solution

**Breaking Changes**

**Improvements**:

- Remove all variants from components.
    - **Removing variants**: Callout (assume info only), Hero (assume gradient only), InfoCards (assume top variant only), Infopic (assume side-by-side variant only), KeyStatistics (assume top variant only)
    - **InfoCols**: Removing option to have a grey background
    - **Infopic**: Removing option to have text on right
    - For Hero component, we are definitely going to have variants in the future, so keeping the key intact for now.
    - For any changes that are not compatible with Classic components, opted to remove them since having a classic theme is not within the scope of MVP.
- Updated usages inside Storybook and seeds.


karrui (approved):
never really review, but assume we are not removing variants that designers have already designed for. we can always add back variants as designers add them as useable components yeah?

seaerchin:
can hold off on merge first? i ask sehyun review

PR #348, opened 2024-07-23
chore: remove render for fallthrough

seaerchin:
## Problem
At present, we have an optional `sectionIdx` prop that's already legacy (gg). This causes visual noise because it shows a little red `No applicable renderer found` text, which is distracting + useless.

## Solution
Add a default fallthrough case with super low priority so that stuff that falls through (specifically this, note the `and` condition from json forms) gets caught and rendered as `null`, preventing the visual noise/

The ideal solution would be to jsut update the schema so section idx doesn't exist. however, i don't know if we can deprecate it yet so i'm still going with this solution for now

### DIFF
--- a/apps/studio/src/constants/formBuilder.ts
+++ b/apps/studio/src/constants/formBuilder.ts
@@ -12,4 +12,5 @@ export const JSON_FORMS_RANKING = {
   RadioControl: 3,
   GroupLayoutRenderer: 1,
   VerticalLayoutRenderer: 1,
+  Catchall: 99999999999
 }
--- a/apps/studio/src/constants/formBuilder.ts
+++ b/apps/studio/src/constants/formBuilder.ts
@@ -12,5 +12,5 @@ export const JSON_FORMS_RANKING = {
   RadioControl: 3,
   GroupLayoutRenderer: 1,
   VerticalLayoutRenderer: 1,
-  Catchall: 99999999999
+  Catchall: -99999999999,
 }
--- a/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
@@ -1,6 +1,6 @@
 import type { JsonFormsRendererRegistryEntry } from "@jsonforms/core"
 import type { IsomerComponent } from "@opengovsg/isomer-components"
-import { and, or, rankWith, schemaTypeIs } from "@jsonforms/core"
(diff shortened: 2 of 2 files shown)


dcshzj:
GG this was alr fixed in #344

seaerchin:
np i did this for `uat` ytd HAHA

dcshzj:
Hmm the condition might be too general, means all integers and numbers types cannot be rendered cos of this catch-all. We got Image component's width that is of type `number`, so that field would not be rendered.

I think a bit more painful for us, but we will need to test every component to make sure that the "No applicable renderer found" message does not appear, rather than having this catch-all.

apps/studio/src/constants/formBuilder.ts
```diff
@@ -12,4 +12,5 @@ export const JSON_FORMS_RANKING = {
   RadioControl: 3,
   GroupLayoutRenderer: 1,
   VerticalLayoutRenderer: 1,
+  Catchall: 99999999999
```
harishv7, line 15: Why do we even have sectionIdx?

harishv7: It's for compat with the classic infobar infocols infocards components?

seaerchin: Yeap! We remove the props in another PR but I reopened this because I wanted to check if there's value in a default case to catch fall throughs

seaerchin:
> is this still needed @dcshzj

up to us to decide, talked to him before i reopened and i think the main value prop is catching fall-through cases, similar to the `default` case for `switch` statements

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
```diff
@@ -1,9 +1,11 @@
+import type { JsonFormsRendererRegistryEntry } from "@jsonforms/core"
 import type { IsomerComponent } from "@opengovsg/isomer-components"
-import { type JsonFormsRendererRegistryEntry } from "@jsonforms/core"
+import { and, or, rankWith, schemaTypeIs } from "@jsonforms/core"
```
dcshzj, line 3: nit: I think some of these are unused?

dcshzj (approved):

apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx
seaerchin: done

PR #350, opened 2024-07-23
fix(preview): add shadows + padding

seaerchin:
## Problem
Previously, the preview took up the whole area and had no discernible border between the sidebar/preview. In order to counteract this, the figma had shadows + padding that wasn't implemented

## Solution
Add 2 `Box`s that sets the `padding` + the `shadow` on the component that uses the preview. Specifically this does:
- set bg to `gray.100`
- add `shadow="md"`
- add `borderRadius=8px`

- didn't set this inside the `Preview` component itself. didn't want to tie the styling too closely with the `Preview`, which is supposed to just be a thin wrapper over `RenderEngine`. this means there's some duplicated code between `Preview` and `PreviewLayout`

### DIFF
--- a/apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
+++ b/apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from "react"
-import { Grid, GridItem } from "@chakra-ui/react"
+import { Box, Grid, GridItem } from "@chakra-ui/react"
 import { z } from "zod"
 
 import { useEditorDrawerContext } from "~/contexts/EditorDrawerContext"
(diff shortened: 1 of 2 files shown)


apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -51,15 +51,17 @@ function EditPage(): JSX.Element {
...
-        {/* and not from the page */}
-        <Preview
-          siteId={siteId}
-          {...page}
-          permalink={permalink}
-          version="0.1.0"
-          content={previewPageState}
-        />
+        <Box p="2rem" bg="gray.100">
+          <Box borderRadius="8px" bg="white" shadow="md" overflow="hidden">
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:55]: will the hidden overflow cause the preview to not be able to be scrolled?
should it be auto instead?

karrui (approved):
lgtm otherwise

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:55]: lemme try this - i put `overflow="hidden"` because otherwise, the overflow would cause the `borderRadius` to not work properly (the masthead wouldn't have border :sadge:)

Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx]: tested against local - `auto` works, i think the reason is just that the default is `visible` and we just need to constrain the masthead into the `Box`. 

thanks for catching!

PR #351, opened 2024-07-23
fix: schema 

seaerchin:
## Problem
1. `callout` has a variant but it shouldn't (i think there's a more general fix going out but just put this here first) 
2. `Heading` schema needs to accept optional content. Otherwise, pressing `#` inside `Tiptap` will crash the editor on validation as this creates an empty heading w/o the content prop. 
3. update prose control testing - we were previously using the `uischema` **and not the `schema`**, which led to the editor not being rendered.

## Solution
1. remove `callout.variant` in `components`
2. set `content` to be `Type.Optional`
3. set prose tester to use `schema` and not `uischema` (argument order)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
```diff
@@ -1,16 +1,16 @@
...
 import { JSON_FORMS_RANKING } from "~/constants/formBuilder"
 
 export const jsonFormsProseControlTester: RankedTester = rankWith(
   JSON_FORMS_RANKING.ProseControl,
-  schemaMatches(
-    (schema) => hasType(schema, "array") && schema.format === "prose",
-  ),
+  (_, schema) => {
+    return schema.format === "prose"
+  },
```
seaerchin, line 13: @dcshzj should i also be using `schemaMatches` here?

packages/components/src/interfaces/native/Prose.ts
```diff
@@ -8,33 +8,45 @@ import { ParagraphSchema } from "./Paragraph"
...
-        },
+const BASE_PROSE_META = {
+
+  title: "Content block",
+  description: "A collection of native content components.",
+  format: "prose",
+
+}
+
+const generateProseSchema = (id?: string) => {
```
seaerchin, line 19: this is done to generate the `ProseSchema` so that there's 1 with an `id` and 1 without. These are **both the same** in terms of validation.

however, the output for `compile` and to `JSONForms` is different - if we use a `Ref`, the schema that `JSONForms` sees **is just the ref** and causes the prose editor to not be rendered.

cc @dcshzj if i missed anything

seaerchin: also if build relies on the `id` implicitly

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
dcshzj: Hmm `schemaMatches` should be correct, this should be referring to the underlying JSON schema and not the uischema. Is the `format: prose` key added to the Prose component schema?

packages/components/src/interfaces/native/Prose.ts
dcshzj: Hmm curious, what if we removed the Type.Ref from the getComponentSchema inside main.ts? That usage means that among all the component schemas, only the prose schema is the one being Type.Ref'd.

https://github.com/opengovsg/isomer/blob/e000312c83dd69d19b5fc1f884adad09dd25b748/packages/components/src/schemas/main.ts#L30

dcshzj: Discussed offline, this wouldn't work cos we still get duplicate $id, one from the actual schema inside `getComponentSchema` and another one from the definitions.

packages/components/src/interfaces/native/Prose.ts
```diff
@@ -8,33 +8,45 @@ import { ParagraphSchema } from "./Paragraph"
...
-          Type.Ref(TableSchema),
-          Type.Ref(UnorderedListSchema),
-        ]),
-        {
-          title: "Content block",
-          description: "A collection of native content components.",
-          minItems: 1,
-        },
+const BASE_PROSE_META = {
+
```
dcshzj, line 12: nit: extra line

packages/components/src/interfaces/native/Prose.ts
```diff
@@ -8,33 +8,45 @@ import { ParagraphSchema } from "./Paragraph"
...
+  title: "Content block",
+  description: "A collection of native content components.",
+  format: "prose",
+
```
dcshzj, line 16: nit: same here

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
seaerchin: yea it is

seaerchin: i'll add back `schemaMatches`

packages/components/src/interfaces/complex/Accordion.ts
```diff
@@ -10,7 +11,7 @@ export const AccordionSchema = Type.Object(
       title: "Accordion summary",
       description: "The summary for the accordion",
     }),
-    details: Type.Ref(ProseSchema),
+    details: InnerProseSchema,
```
dcshzj, line 14: nit: maybe can rename as `BaseProseSchema`? That's kinda what we did for Paragraph so wanted to be consistent in the naming

dcshzj (approved):
Only nits

packages/components/src/interfaces/complex/Accordion.ts
seaerchin: sgtm, idk how to name it so i just gave it as `Inner` LULW

packages/components/src/interfaces/native/Prose.ts
seaerchin: curse the formatting gods

PR #358, opened 2024-07-24
feat: introduce page metadata editor

dcshzj:
## Problem

The page metadata editor cannot be changed currently.

## Solution

**Breaking Changes**

**Features**:

- Introduce a new page metadata editor state to edit the page's metadata.
    - Changed `previewPageState` and `savedPageState` to handle the entire page and not just the content block.
- Also adjusted the FormBuilder to be generic, so that it can be reused in multiple places.
- Also set the Hero component to be the first fixed block if it is the homepage layout.

## Before & After Screenshots
Screenshot 2024-07-24 at 17 13 32

### DIFF
--- a/packages/components/src/schemas/main.ts
+++ b/packages/components/src/schemas/main.ts
@@ -1,8 +1,16 @@
 import type { TSchema } from "@sinclair/typebox"
 import { Type } from "@sinclair/typebox"
 
-import type { IsomerComponentTypes } from "~/types"
-import { IsomerPageSchema } from "~/types"
+import type { IsomerComponentTypes, IsomerSchema } from "~/types"
+import {
+  ArticlePageMetaSchema,
+  CollectionPageSchema,
+  ContentPageMetaSchema,
+  FileRefSchema,
+  HomePageMetaSchema,
+  IsomerPageSchema,
+  LinkRefSchema,
+} from "~/types"
 import {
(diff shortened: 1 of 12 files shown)


seaerchin:
checked homepage only but it looks like the block is gone from the preview?

image

dcshzj:
> checked homepage only but it looks like the block is gone from the preview?

Hmm homepage not supposed to have this page metadata editor, unless we want users to be able to update them for the homepage. Will check with design.

packages/components/src/schemas/main.ts
```diff
@@ -35,3 +43,18 @@ export const getComponentSchema = (
     ...definitions,
   }
 }
+
+const LAYOUT_METADATA_MAP = {
```
seaerchin, line 47: no actionable needed - just voicing my thoughts out loud. wdyt of having a similar API to `JSONForms` here in that we just export `tester + render` and our component in `RootStateDrawer` just uses that to render?

seaerchin (approved):
actually re-reading this, i think it's ok. 4 main changes: 

1. we update via `page.content` rather than the full page
2. we have a metadata drawer that's conditionally rendered via an state change (fixed block)
3. api change on json forms builder

don't think any is blocking; hope 3. can be reverted to old API but ok w/ this for now ba

PR #360, opened 2024-07-24
fix(prose): update typing for `Link`

seaerchin:
## Problem
We need to allow users to set `links` inside our tiptap component. However, tiptap sets teh `href` attribute inside the `attrs` object whereas our validator expects a top level `href` attribute. this causes any link insertion to break on validation. 

## Solution
- Update the `LinkMarkSchema` to be `Unsafe` to avoid deeply nested types
- Update the `LinkMarkSchema` to have a `href` attribute so that validation can pass


tooling/build/scripts/generate-sitemap.js
```diff
@@ -134,7 +134,9 @@ const processDanglingDirectory = async (fullPath, relativePath, name) => {
                   marks: [
                     {
                       type: "link",
-                      href: child.permalink,
+                      attrs: {
+                        href: child.permalink,
+                      },
```
seaerchin, line 139: only relevant change because we update the `href` to be inside `attrs` now. i searched the 2 `.js` files in this directory to check if we needed to update `href`

fact-check me @dcshzj if this is enoguh

packages/components/src/interfaces/native/Text.ts
```diff
@@ -28,14 +28,41 @@ const ItalicMarkSchema = Type.Object(
...
+          Type.Literal("_self"),
+          Type.Literal("_parent"),
+          Type.Literal("_top"),
+        ]),
+        // NOTE: The href given by tiptap here
+        // https://github.com/ueberdosis/tiptap/blob/main/packages/extension-link/src/link.ts
+        // defaults to `null` href but href could also be passed as `undefined`
+        href: Type.Union([Type.String(), Type.Null()]),
+        rel: Type.String(),
+        class: Type.Union([Type.String(), Type.Null()]),
```
dcshzj, line 59: issue: I don't think we should match the schema that Tiptap provides entirely, our schema is intended to be a subset of the schema because we only want to allow certain behaviours instead of all the available options that HTML/Tiptap provides. Adding all here means we lose the signal on what is allowed and what isn't.

In general we override the `rel` and `class` when we output to the user, so those properties will be ignored. For `target` we should only support `_blank`.

seaerchin: done - where do i see the properties that we strip? couldn't find a `Text` component

dcshzj: Mainly in getSanitizedInlineContent.ts:

https://github.com/opengovsg/isomer/blob/c19c5d5cfbc26f8a04ad5052727906ebda5df741/packages/components/src/utils/getSanitizedInlineContent.ts#L5-L42

packages/components/src/interfaces/native/Text.ts
```diff
@@ -28,16 +28,25 @@ const ItalicMarkSchema = Type.Object(
   },
 )
 
-const LinkMarkSchema = Type.Object(
+const BaseLinkMarkSchema = Type.Object(
   {
     type: Type.Literal("link", { default: "link" }),
-    href: Type.String(),
+    attrs: Type.Object({
+      target: Type.Literal("_blank"),
```
dcshzj, line 35: issue: hmm this property should be optional

```suggestion
      target: Type.Optional(Type.Literal("_blank")),
```

seaerchin: i don't think that's quite right because we're using tiptap's link schema. they default it to the `options.target`, which is `_blank` by default, see ss

image

dcshzj: Hmm but we are not using Tiptap's link schema entirely. In HTML it is entirely possible to omit the `target` attribute and the link will have a default behaviour of `_self`. This schema means that all links must have `_blank`, which may not be the case for internal links.

seaerchin: you're right! i misunderstood the schema here because i thought it was only validating tiptap. i'll set it to `optional("_self" | "_blank")` wdyt?

dcshzj (approved):

PR #362, opened 2024-07-25
chore: style fixes for save button

seaerchin:
## Problem
Previously, our save button was very high up and not pinned ot the bottom. this means the UI is very cramped. design has specified that hte save button should be pinned to the bottom and be sticky so that even when there are many items, it'll display correctly.

## Solution
- use `Flex` + `Spacer` so that the `Box` containing the button is at the bottom
- add shadows so that the box containing the save button is visually distinct. this is done via tokens rather than palette + raw css values as we've done before
- ~take note that adding many items will still cause the save button **to be pinned to the bottom of the sidebar**; i think what design wants is for the save button to be pinned **to the bottom of the screen**. this is slightly trickier and might involve `pos: absolute` which might complicate overflows, for example (save button now blocks blocks) <- say that real fast a few times.~
    - this `sticky` behaviour which we have now is actually correct, try it on [this site](https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers)

## Before & After Screenshots
**AFTER**:
image

### DIFF
--- a/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
@@ -79,7 +79,8 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
         bottom={0}
         bgColor="base.canvas.default"
         boxShadow="md"
-        p="1.5rem 2rem 1.5rem 2rem"
+        py="1.5rem"
+        px="2rem"
       >
         <Button
           w="100%"
--- a/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, Heading, HStack, Icon, Spacer } from "@chakra-ui/react"
+import { Box, Flex, Heading, HStack, Icon } from "@chakra-ui/react"
 import { Button, IconButton } from "@opengovsg/design-system-react"
 import { getComponentSchema } from "@opengovsg/isomer-components"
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -68,7 +74,13 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
       <Box px="2rem" py="1rem">
         <FormBuilder />
       </Box>
-      <Box px="2rem" pb="1.5rem">
+      <Box
+        pos="sticky"
+        bottom={0}
+        bgColor="base.canvas.default"
+        boxShadow="md"
+        p="1.5rem 2rem 1.5rem 2rem"
```
dcshzj, line 82: nit: I think `px` and `py` might be easier to read?

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -148,13 +147,13 @@ export default function RootStateDrawer() {
...
         </DragDropContext>
       </VStack>
-      <Spacer />
-      {/* TODO: Add New Block Section */}
       <Box
         w="100%"
-        bgColor="white"
+        bgColor="base.canvas.default"
+        boxShadow="md"
         p="1.5rem 2rem 1.5rem 2rem"
```
dcshzj, line 154: nit: same here

dcshzj (approved):

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
seaerchin: sgtm

PR #365, opened 2024-07-25
update to use bigint for resourceId and blobId

harishv7:
### TL;DR

This pull request updates the ID fields for the `Resource`, and `Blob` models to use `BigInt` instead of `Int`. Corresponding changes have been made to Prisma schema, migration files, TypeScript types, and various parts of the application code to handle this change.

This is in relation to introducing versions later!

### What changed?

- Updated Prisma schema to change ID types.
- Created migration script for the database schema changes.
- Modified TypeScript types to reflect new ID types.
- Updated application code to handle `BigInt` IDs.

Note: 
1. We still use `Number` in JS throughout, so we are technically limited by the max int support in JS which is 2^53 - 1. 
2. If you look at the generatedTypes, you will see that bigints are output as String. This requires a cast to and fro number and string. For example, Zod might validate the incoming payload field as a number. But when we query or mutate DB we will need to convert this into a String.

### How to test?

1. Run the migration script to update the database schema.
2. Ensure that the application builds without errors.
3. Execute the application and ensure that all parts work as expected, particularly those dealing with `Resource`, and `Blob` models.
4. Run existing tests to ensure they pass, and add new tests if necessary.

### Why make this change?

To accommodate larger datasets and future-proof the application, enabling the use of `BigInt` for IDs in critical models like `Resource`, and `Blob`. This change reduces the risk of ID collisions and supports the scaling of the application.

As shared, a quick back-of-envelope calculation shows:

We now have a 1-many relationship for resources and versions. Estimating worst case scenario:
Max running order of integer = 5000 sites, 14k pages each site, each page has 1000 versions = 5k * 14k * 1k = 70,000,000,000
Current max supported int = 4 bytes = 32 bits = 2,147,483,647
But JS supports up to 53 bits (2^53 - 1)
Might be good to consider using BigInt for all our IDs - 8 bytes = 64 bits

But in our app, we will only use a max of 53 bits (JS's max INT value)

### DIFF
--- a/apps/studio/prisma/seed.ts
+++ b/apps/studio/prisma/seed.ts
@@ -205,10 +205,10 @@ async function main() {
     )
     .execute()
 
-  let blobId = 1
+  let blobId = BigInt(1)
   const dedupeBlobId = await db
     .selectFrom("Blob")
-    .where("Blob.id", "=", blobId)
+    .where("Blob.id", "=", String(blobId))
     .select("Blob.id")
     .executeTakeFirst()
   if (!dedupeBlobId) {
@@ -217,13 +217,13 @@ async function main() {
       .values({ content: jsonb(PAGE_BLOB) })
       .returning("id")
       .executeTakeFirstOrThrow()
-    blobId = id
+    blobId = BigInt(id)
   }
 
   await db
     .insertInto("Resource")
     .values({
-      mainBlobId: blobId,
+      mainBlobId: String(blobId),
       permalink: "home",
       siteId,
       type: "Page",
--- a/apps/studio/prisma/generated/generatedTypes.ts
+++ b/apps/studio/prisma/generated/generatedTypes.ts
@@ -9,7 +9,7 @@ export type Generated<T> =
 export type Timestamp = ColumnType<Date, Date | string, Date | string>
 
 export interface Blob {
-  id: GeneratedAlways<number>
+  id: GeneratedAlways<string>
   /**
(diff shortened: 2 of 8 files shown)


harishv7:
* **#365** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/365?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/studio/prisma/seed.ts
```diff
@@ -217,13 +217,13 @@ async function main() {
...
       .executeTakeFirstOrThrow()
-    blobId = id
+    blobId = BigInt(id)
   }
 
   await db
     .insertInto("Resource")
     .values({
-      mainBlobId: blobId,
+      mainBlobId: String(blobId),
```
dcshzj, line 226: question(non-blocking): just wanted to confirm, when inserting we should immediately use string here, rather than passing in the bigint?

dcshzj (approved):
Still not very sure what's the value of using bigint versus using strings directly (since unlikely we will be doing arithmetic operations on the ID), but there might be some other optimisations done by Postgres so I think it should be okay.

apps/studio/prisma/seed.ts
harishv7: Yea, the generatedType for bigint is actually string. So kysely will expect a string to be passed in for this. However, the insert op will then convert this string into a bigint

PR #367, opened 2024-07-25
feat(tiptap): save updates to db + invalidate queries

seaerchin:
## Problem
Previously, `Prose` wasn't saving to db. When `uat` added the save to db, there wasnt the query invalidation portion. This PR adds both functionality in

## Solution
- use the `updatePageBlob` hook in `TiptapComponent`. 
- invalidate queries on save so that we see the new state

## Screenshots and video 

https://github.com/user-attachments/assets/cd4baed3-f3a8-4791-848b-0effee6dee25

### DIFF
--- a/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
@@ -3,8 +3,16 @@ import { Button, IconButton } from "@opengovsg/design-system-react"
 import { getComponentSchema } from "@opengovsg/isomer-components"
 import { BiDollar, BiX } from "react-icons/bi"
 
+import z from "zod"
 import { useEditorDrawerContext } from "~/contexts/EditorDrawerContext"
 import FormBuilder from "./form-builder/FormBuilder"
+import { trpc } from "~/utils/trpc"
+import { useQueryParse } from "~/hooks/useQueryParse"
+
+const complexEditorSchema = z.object({
+  pageId: z.coerce.number(),
+  siteId: z.coerce.number(),
+})
 
 export default function ComplexEditorStateDrawer(): JSX.Element {
   const {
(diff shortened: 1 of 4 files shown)


apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -74,11 +88,36 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
           onClick={() => {
             setDrawerState({ state: "root" })
             setSavedPageState(previewPageState)
+            mutate({
+              pageId,
+              siteId,
+              content: JSON.stringify({
+                ...pageContent,
+                content: previewPageState,
+              }),
```
seaerchin, line 97: having to fetch page content just to stringify here isn't ideal and i'll refactor when all our "content updating" features are in

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -193,7 +193,7 @@ export const pageRouter = router({
     .mutation(async ({ input }) => {
       // @ts-expect-error we need this because we sanitise as a string
       // but this accepts a nested JSON object
-      await updateBlobById({ ...input, pageId: input.pageId })
+      await updateBlobById(db, { ...input, pageId: input.pageId })
```
seaerchin, line 196: error wasn't caught previously due to `@ts-expect-error` .______.

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -1,32 +1,46 @@
...
 import { useEditorDrawerContext } from "~/contexts/EditorDrawerContext"
+import { useQueryParse } from "~/hooks/useQueryParse"
+import { trpc } from "~/utils/trpc"
 import FormBuilder from "./form-builder/FormBuilder"
 
-export default function ComplexEditorStateDrawer(): JSX.Element {
+const complexEditorSchema = z.object({
+  pageId: z.coerce.number(),
+  siteId: z.coerce.number(),
+})
```
dcshzj, line 16: suggestion(blocking): we should use `editPageSchema` here instead

seaerchin: sgtm, copied over the commit from `uat` @__@

dcshzj (approved):

PR #369, opened 2024-07-25
add migration for versions

harishv7:
### TL;DR

Replace `mainBlobId` with `versionId` and introduce `version` table with the associated fields: `id`, `versionNum`, `resourceId`, `blobId`, `publishedAt`. This change includes updates to Prisma schema, generated types, and related service files.

### What changed?
- Removed `mainBlobId`, added `versionId` to `Resource`
- Created `Version` model and table
- Updated Prisma schema and migration files
- Adjusted code to handle new `versionId` logic instead of `mainBlobId`
- Replaced all instances of `mainBlobId` with `versionId`
- Modified tests to reflect changes

### How to test?
1. Run Prisma migrations
2. Verify changes in the database structure
3. Update seed files and run seeding script
4. Check the application for any errors related to resource handling

### Why make this change?

To streamline resource versioning and make the resource model more maintainable by separating version tracking into a distinct table.

Relationships established:
- Resource have a 1 to many relation with Version
- Version has a 1-1 mapping to both a resource and a blob


harishv7:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/369?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#369** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#365** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/365?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/studio/prisma/schema.prisma
```diff
@@ -33,6 +33,16 @@ model VerificationToken {
   expires    DateTime
 }
 
+model Version {
+  id          BigInt     @id @default(autoincrement())
+  versionNum  BigInt
```
dcshzj, line 38: question: why do we need BigInt for this? I suspect that there shouldn't be too many versions for a particular resource, and we will need to perform arithmetic operations on this field.

apps/studio/prisma/schema.prisma
```diff
@@ -33,6 +33,16 @@ model VerificationToken {
...
+  resourceId  BigInt
+  resources   Resource[]
+  blobId      BigInt
+  blob        Blob       @relation(fields: [blobId], references: [id])
+  publishedAt DateTime   @default(now())
+}
```
dcshzj, line 44: suggestion(non-blocking): I think we should also add a publishedBy field to note which user performed the publish

apps/studio/prisma/schema.prisma
```diff
@@ -46,16 +56,25 @@ model Resource {
 
   permission Permission[]
 
-  mainBlob   Blob?   @relation("MainBlob", fields: [mainBlobId], references: [id])
-  mainBlobId BigInt? @unique
+  versionId BigInt?
```
dcshzj, line 59: question: should this also be `@unique` too?

apps/studio/prisma/schema.prisma
```diff
@@ -46,16 +56,25 @@ model Resource {
...
+  version   Version? @relation(fields: [versionId], references: [id])
```
dcshzj, line 60: issue: I'm not sure if the name `version` here would be confusing, because we are actually referring to the latest published version of that resource, not a relation to all the possible versions (that information is maintained within the versions table).

apps/studio/prisma/schema.prisma
```diff
@@ -33,6 +33,16 @@ model VerificationToken {
   expires    DateTime
 }
 
+model Version {
+  id          BigInt     @id @default(autoincrement())
+  versionNum  BigInt
+  resourceId  BigInt
+  resources   Resource[]
+  blobId      BigInt
+  blob        Blob       @relation(fields: [blobId], references: [id])
```
dcshzj, line 42: question: should this be `@unique`?

apps/studio/prisma/schema.prisma
```diff
@@ -46,16 +56,25 @@ model Resource {
...
   state ResourceState? @default(Draft)
   type  ResourceType
 }
 
+model Blob {
+  id            BigInt    @id @default(autoincrement())
+  /// @kyselyType(PrismaJson.BlobJsonContent)
+  /// [BlobJsonContent]
+  content       Json
+  versions      Version[]
```
dcshzj, line 74: thought: the way I see this is that a blob is a unique instance that exists for a single version (if published) or as a draft, so this might refer back to the specific version that the blob is tied to.

apps/studio/prisma/schema.prisma
harishv7: Good point, what should be BigInt is the global id, not the resource-specific one. Will update!

apps/studio/prisma/schema.prisma
harishv7: Each version should be tied to an unique blob ideally. But in case of a rollback, I am thinking:
- We create a new version pointing to the blobId referenced in the version we want to rollback to
- This is to prevent us from tampering old versions. Old versions should be immutable as they preserve things like publishedAt etc.
- This means 2 versions can point to the same blobId

apps/studio/prisma/schema.prisma
harishv7: Qn here from Slack discussion is also separation of concerns bet versions and audit log. But I think for MVP we can keep it in Versions. Will add.

apps/studio/prisma/schema.prisma
harishv7: Each resource to a unique version makes sense, will add!

apps/studio/prisma/schema.prisma
harishv7: Will rename to publishedVersionId

apps/studio/prisma/schema.prisma
harishv7: After offline discussions, it does make sense that we do not immediately create a version during rollbacks. We would likely clone the blobId of the version to rollback -> allow user to preview and then when they confirm to publish we will initiate a new version creation on this blob.

Hence, it would make sense to keep a unique blob to a version.

apps/studio/prisma/schema.prisma
```diff
@@ -75,6 +96,7 @@ model User {
 
   permission  Permission[]
   siteMembers SiteMember[]
+  Version     Version[]
```
dcshzj, line 99: nit: lowerCamelCase

```suggestion
  version     Version[]
```

Also does it make sense for this (plus the siteMembers/permission) to be optional?

apps/studio/prisma/schema.prisma
```diff
@@ -46,16 +58,25 @@ model Resource {
 
   permission Permission[]
 
-  mainBlob   Blob?   @relation("MainBlob", fields: [mainBlobId], references: [id])
-  mainBlobId BigInt? @unique
+  publishedVersionId BigInt?  @unique
+  version            Version? @relation(fields: [publishedVersionId], references: [id])
```
dcshzj, line 62: nit: this should be renamed to `publishedVersion` as well?

```suggestion
  publishedVersion   Version? @relation(fields: [publishedVersionId], references: [id])
```

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -26,7 +26,7 @@ export const folderRouter = router({
         .where("parentId", "=", String(input.resourceId))
         .execute()
       const children = childrenResult.map((c) => {
-        if (c.draftBlobId || c.mainBlobId) {
+        if (c.draftBlobId || c.versionId) {
```
dcshzj, line 29: issue: I think this should be `publishedVersionId`?

```suggestion
        if (c.draftBlobId || c.publishedVersionId) {
```

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -69,7 +69,7 @@ export const pageRouter = router({
           "Resource.id",
           "Resource.permalink",
           "Resource.title",
-          "Resource.mainBlobId",
+          "Resource.versionId",
```
dcshzj, line 72: issue: same here

```suggestion
          "Resource.publishedVersionId",
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -11,7 +11,7 @@ const defaultResourceSelect: SelectExpression<DB, "Resource">[] = [
   "Resource.permalink",
   "Resource.siteId",
   "Resource.parentId",
-  "Resource.mainBlobId",
+  "Resource.versionId",
```
dcshzj, line 14: issue: same here

```suggestion
  "Resource.publishedVersionId",
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -81,8 +81,9 @@ export const getFullPageById = async (args: {
   }
 
   return getById(args)
-    .where("Resource.mainBlobId", "is not", null)
-    .innerJoin("Blob", "Resource.mainBlobId", "Blob.id")
+    .where("Resource.versionId", "is not", null)
+    .innerJoin("Version", "Resource.versionId", "Version.id")
```
dcshzj, line 85: issue: also same here

```suggestion
    .where("Resource.publishedVersionId", "is not", null)
    .innerJoin("Version", "Resource.publishedVersionId", "Version.id")
```

apps/studio/src/server/modules/resource/resource.types.ts
```diff
@@ -12,7 +11,7 @@ export type PageContent = Omit<
 >
 
 // TODO: Technically mainBlobId is not required before 1st publish
```
dcshzj, line 13: nit: I think this TODO can be removed?

apps/studio/prisma/schema.prisma
harishv7: All these are just relations

Or do you mean the user is optional within the Permission and SiteMember models? But both perm and sitemember needs to be associated to an user right?

apps/studio/src/server/modules/resource/resource.service.ts
harishv7: My bad, didn't push the updated types

apps/studio/prisma/schema.prisma
dcshzj: Hmm in this case it means that a user would definitely have permission + site membership, but this might not be entirely true especially for a user that is just created/first log in.

dcshzj: Clarified offline, no need for `undefined` here because it is captured inside the empty array `[]`.

dcshzj (approved):

harishv7:
### Merge activity

* **Jul 31, 4:15 AM EDT**: @harishv7 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/369).

PR #381, opened 2024-07-28
add publish endpoint

harishv7:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


harishv7:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/381?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#381** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#369** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/369?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#365** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/365?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -42,3 +46,49 @@ export const createDefaultPage = ({
     }
   }
 }
+
+export const addNewVersion = async (siteId: number, pageId: number) => {
```
dcshzj, line 50: suggestion: I think this procedure should be kept within versions service and is an implementation that page service should not be concerned with.

Main reason is because I see page service as the one responsible for returning the draft and published page contents. The versions service is the one that supports the page service by providing the latest version, and how this information is stored is the responsibility of the versions service. So if versions service does any incrementing of the version number and stuff, that is up to it.

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -42,3 +46,49 @@ export const createDefaultPage = ({
...
+    if (!page.draftBlobId) {
+      return { error: "No drafts to publish for this page" }
+    }
+
+    let newVersionNum = 1
+    if (page.versionId) {
+      const currentVersion = await getVersionById({
+        versionId: page.versionId,
+      })
+      newVersionNum = Number(currentVersion.versionNum) + 1
```
dcshzj, line 63: issue: continuing with the previous comment, this means that the versions service should perform inserts depending on what it finds to be the current latest version number for the given resource. That way, if we have a race condition, we don't accidentally have two inserts with the same version number into the database.

Effectively what happens is that the versions service is the one that handles the versioning procedure, and the page service is the one that is just concerned with getting the latest version from the versions service.

apps/studio/src/server/modules/version/version.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    blobId: number
+  },
+  tx?: Transaction<DB>,
+) => {
+  const { versionNum, resourceId, blobId } = props
+  const instance = tx ? tx : db
+  const addedVersion = await instance
+    .insertInto("Version")
+    .values({
+      versionNum: String(versionNum),
```
dcshzj, line 34: issue: already mentioned in the other comments but I think that this might potentially be dangerous and opens us up to race conditions, especially when the version number from 2 different requests are the same.

apps/studio/src/components/AppNavbar/PublishButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return (
+    <Button
+      variant="solid"
+      size="sm"
+      onClick={handlePublish}
+      isLoading={isLoading}
+    >
+      Publish
+    </Button>
```
dcshzj, line 49: issue: I think we should either:
1. Disable this publish button if the page has never been saved before (i.e. draftBlobId is null/missing). I think in the procedure you did this, but it is not reflected in the user interface.
2. Immediately publish the unsaved page state, even if the draftBlobId is null/missing, so that this button can do something meaningful.

I'm more in favour of the 1st option because it seems safer and having no draftBlobId usually means an empty page which we would want to avoid publishing, but no strong opinion.

harishv7: Yup makes sense, have implemented option 1 as well

apps/studio/src/components/AppNavbar/PublishButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    onSuccess: (data) => {
+      if (data.versionId) {
+        toast({
+          status: "success",
+          title: "Page published successfully",
+        })
+      } else {
+        toast({
+          status: "error",
+          title: data.error,
```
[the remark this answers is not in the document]
harishv7: fixed

apps/studio/src/components/AppNavbar/AppNavbar.tsx
```diff
@@ -62,6 +66,10 @@ export function AppNavbar(): JSX.Element {
           >
             Report an issue
           </Button>
+          <PublishButton
+            pageId={pathParams.pageId as string}
+            siteId={pathParams.siteId as string}
```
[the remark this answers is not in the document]
harishv7: fixed

apps/studio/src/server/modules/page/page.service.ts
harishv7: yup was pending refactor, extracted now

apps/studio/src/server/modules/version/version.service.ts
harishv7: I have added serializable isolation levels

apps/studio/src/components/AppNavbar/AppNavbar.tsx
```diff
@@ -62,6 +99,12 @@ export function AppNavbar(): JSX.Element {
           >
             Report an issue
           </Button>
+          {publishButtonProps.showPublish && (
+            <PublishButton
+              pageId={publishButtonProps.pageId}
+              siteId={publishButtonProps.siteId}
+            />
+          )}
```
dcshzj, line 107: question(non-blocking): should the button be completely hidden, or should it be shown but hidden?

dcshzj (approved):

apps/studio/src/server/modules/version/version.service.ts
```diff
@@ -0,0 +1,98 @@
...
+}) => {
+  return await db
+    .transaction()
+    .setIsolationLevel("serializable")
+    .execute(async (tx) => {
+      const page = await getPageById(tx, { siteId, resourceId: pageId })
+
+      if (!page.draftBlobId) {
+        return { error: "No drafts to publish for this page" }
+      }
```
seaerchin, line 65: why not just throw here?

apps/studio/src/server/modules/version/version.service.ts
```diff
@@ -0,0 +1,98 @@
...
+    publisherId: string
+  },
+) => {
+  const { versionNum, resourceId, blobId, publisherId } = props
+  const addedVersion = await db
+    .insertInto("Version")
+    .values({
+      versionNum,
+      resourceId: String(resourceId),
+      blobId: String(blobId),
```
seaerchin, line 38: not directly actionable but how will we utilise this `blobId`? the reason i'm asking is because this is a reference so even if we publish, the value of the blob itself can still change.

seaerchin (approved):
some minor comments, happy to approve after!

apps/studio/src/components/AppNavbar/PublishButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+    >
+      Publish
+    </Button>
+  )
+}
+
+const PublishButton = withSuspense(
+  SuspendablePublishButton,
+  <Skeleton width={"100%"} height={"100%"} />,
+)
```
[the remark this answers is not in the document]
seaerchin: @sehyunidaaa

apps/studio/src/components/AppNavbar/AppNavbar.tsx
```diff
@@ -13,10 +14,46 @@ import { BiLinkExternal } from "react-icons/bi"
 import { ADMIN_NAVBAR_HEIGHT } from "~/constants/layouts"
 import { useMe } from "~/features/me/api"
 import { DASHBOARD, SETTINGS_PROFILE } from "~/lib/routes"
+import PublishButton from "./PublishButton"
 
 export function AppNavbar(): JSX.Element {
   const { me, logout } = useMe()
 
+  const pathParams = useParams()
```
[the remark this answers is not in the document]
harishv7: good point, i missed this! thanks

apps/studio/src/components/AppNavbar/AppNavbar.tsx
```diff
@@ -13,10 +14,46 @@ import { BiLinkExternal } from "react-icons/bi"
...
+  }
+
+  interface ShowPublishButton {
+    showPublish: true
+    pageId: number
+    siteId: number
+  }
+
+  type GetPublishButtonResult = HidePublishButton | ShowPublishButton
+
```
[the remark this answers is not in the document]
harishv7: removed since using query parse

apps/studio/src/components/AppNavbar/AppNavbar.tsx
```diff
@@ -13,10 +14,46 @@ import { BiLinkExternal } from "react-icons/bi"
...
+    }
+
+    const parsedSiteId = parseInt(siteId, 10)
+    const parsedPageId = parseInt(pageId, 10)
+
+    // Check if both parsedSiteId and parsedPageId are valid numbers
+    if (isNaN(parsedSiteId) || isNaN(parsedPageId)) {
+      return { showPublish: false }
+    }
+    return { showPublish: true, siteId: parsedSiteId, pageId: parsedPageId }
```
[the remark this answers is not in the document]
harishv7: agreed and done

apps/studio/src/components/AppNavbar/PublishButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+interface PublishButtonProps {
+  pageId: number
+  siteId: number
+}
+
+const SuspendablePublishButton = ({
+  pageId,
+  siteId,
+}: PublishButtonProps): JSX.Element => {
+  const toast = useToast()
```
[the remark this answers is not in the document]
harishv7: done

apps/studio/src/components/AppNavbar/PublishButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+        })
+      }
+      await utils.page.readPage.invalidate({ pageId, siteId })
+    },
+    onError: async () => {
+      toast({
+        status: "error",
+        title: publishFailureMsg,
+      })
+      await utils.page.readPage.invalidate({ pageId, siteId })
```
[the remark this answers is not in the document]
harishv7: The main purpose of publish to update our SOT first -> hence a version will be created and draftId will be set to null. At this point even if codebuild initiation fails, since there is no draft to publish, the button should be disabled. 

To ensure the button is disabled, we should refetch the state

apps/studio/src/components/AppNavbar/PublishButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+    },
+  })
+
+  const handlePublish = () => {
+    const coercedSiteId = Number(siteId)
+    const coercedPageId = Number(pageId)
+    if (coercedSiteId && coercedPageId)
```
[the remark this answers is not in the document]
harishv7: done

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -84,6 +88,15 @@ export const pageRouter = router({
         ])
         .execute()
     }),
+
+  readPage: protectedProcedure
+    .input(getPageSchema)
+    .query(async ({ input: { pageId, siteId } }) => {
+      return db.transaction().execute(async (tx) => {
```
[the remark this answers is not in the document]
harishv7: had a type conflict on DB, updated definition of DB to use local definition from database/types

apps/studio/src/server/modules/version/version.service.ts
harishv7: Once published the blob shouldn't change, it will be tied closely to the version. on the page resource, draftBlobId is immediately set to null so any future edits will be captured as a new blob

apps/studio/src/server/modules/version/version.service.ts
```diff
@@ -0,0 +1,98 @@
...
+    resourceId: number
+    blobId: number
+    publisherId: string
+  },
+) => {
+  const { versionNum, resourceId, blobId, publisherId } = props
+  const addedVersion = await db
+    .insertInto("Version")
+    .values({
+      versionNum,
```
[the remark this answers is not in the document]
harishv7: A resource has multiple versions -> every publish on the resource will increment its version by 1

Version numbering starts from 1

apps/studio/src/server/modules/version/version.service.ts
harishv7: good point, done

apps/studio/src/components/AppNavbar/AppNavbar.tsx
harishv7: Completely hiding might confuse the user as to why they can't publish. I prefer to show but disable and add a Tooltip saying:
"Publish will be enabled if there are unpublished edits"

@sehyunidaaa lmk your thoughts!

harishv7:
### Merge activity

* **Aug 12, 12:44 PM EDT**: @harishv7 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/381).

PR #388, opened 2024-07-29
fix: errorneous string cast

seaerchin:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/apps/studio/src/server/modules/page/page.router.ts
+++ b/apps/studio/src/server/modules/page/page.router.ts
@@ -126,7 +126,7 @@ export const pageRouter = router({
               title,
               permalink,
               siteId,
-              parentId: String(folderId),
+              ...(folderId && { parentId: String(folderId) }),
               draftBlobId: blob.id,
               type: ResourceType.Page,
             })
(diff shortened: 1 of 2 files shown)


dcshzj (approved):

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -126,7 +126,7 @@ export const pageRouter = router({
               title,
               permalink,
               siteId,
-              parentId: String(folderId),
+              ...(folderId && { parentId: String(folderId) }),
```
karrui, line 129: nit: prefer explicit instead of spread:

```suggestion
              folderId: folderId ? String(folderId) : undefined
```

karrui: but hor why need the string cast in the first place?

seaerchin: we migrated to `BigInt` in #365 - this did a wrongful `String` cast. i've went ahead and looked through usages in the codebase and for nullish `stirng | null | undefined`, i went ahead and did this check.

PR #389, opened 2024-07-30
feat: add support for tables in Tiptap editor

dcshzj:
## Problem

We need to be able to add tables within the Tiptap editor.

## Solution

**Breaking Changes**

**Features**:

- Add table support to the Tiptap editor
    - We have an extra caption attribute that Tiptap's base table extension does not support, so it is done by extending the base extension with the additional attribute.
    - A modal was added to allow configuring this attribute.
    - The table header cell only supports paragraphs, so the base extension was also modified (previously could take `block+`).
- Added icons used for table-related editing. These icons were taken from the classic CMS with the exact same functionality.
- Also disallowed any type of marks inside headings

## Screenshots
Screenshot 2024-07-30 at 11 35 05
Screenshot 2024-07-30 at 11 35 33


apps/studio/src/assets/BxAddColLeft.tsx
karrui, line 1: might not want to prefix this file (and the exported component) with `Bx` since they don't seem to come from BoxIcons. Maybe just `IconAddColLeft`?

apps/studio/src/assets/BxAddColRight.tsx
karrui, line 1: might not want to prefix this file (and the exported component) with `Bx` since they don't seem to come from BoxIcons.

apps/studio/src/assets/BxAddRowAbove.tsx
karrui, line 1: Same for the rest of the icons :X

apps/studio/src/assets/index.ts
karrui, line 1: since they are components, should these components live inside `src/components/icons` instead?

karrui (approved):
lgtm

PR #391, opened 2024-07-30
fix(schema): schema validation

seaerchin:
## Problem
We have some bugs - `articleSchema` -> `minItems: 1` causes block creation to fail validation without error message. 

image block is `required` but there's no error message set. additionally , the default is 100 but this isn't obvious from the UI
 
## Solution
- set `minItems: 0` for `articleSchema`  **or** give some UI feedback (for uat, i just chose the easy fix, open to just removing this) 
- give default in schema + show error message

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
@@ -3,7 +3,12 @@ import { useState } from "react"
 import { Box, FormControl, Text } from "@chakra-ui/react"
 import { and, isStringControl, rankWith, schemaMatches } from "@jsonforms/core"
 import { withJsonFormsControlProps } from "@jsonforms/react"
-import { Attachment, FormLabel, useToast } from "@opengovsg/design-system-react"
+import {
+  Attachment,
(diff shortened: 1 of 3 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
```diff
@@ -71,6 +78,11 @@ export function JsonFormsImageControl({
         <Text textStyle="body-2" textColor="base.content.medium" pt="0.5rem">
           {`Maximum file size: ${MAX_IMG_FILE_SIZE_BYTES / 1000000} MB`}
         </Text>
+        {!!errors && (
+          <FormErrorMessage>
+            {label} {errors}
```
karrui, line 83: assuming errors is a string here

karrui (approved):
lgtm

dcshzj (approved):

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsImageControl.tsx
seaerchin: yeap, that's the type given by json forms

PR #392, opened 2024-07-30
fix: prevent redir till save done

seaerchin:
## Problem
This PR fixes an issue where the user could be redirected to the dashboard before the save was completed. This could cause the user to lose their changes if they navigated away from the page before the save was completed.

## Solution
Update the create page wizard so that the user is not redirected until the save is complete.


karrui (approved):
lgtm. lots of repeated code that could realistically live inside a custom hook/context T_T. Should we make a linear ticket to revisit this?

PR #403, opened 2024-07-31
feat(move): move pages functionality

seaerchin:
## Problem
We want to be able to move pages from one locatio to another. This PR implements a simple move modal together with the required backend functionality to move pages.

Closes [insert issue #]

## Solution
- install `jotai`
  - similar api to `react.useState`.
  - it has `atoms` that basically are `state` that live outside of react. 
  - exposes 3 apis (that we use here) to interact with atoms: read/rw/write
- Implement a move modal. This move modal uses `jotai` + an atom to decide when to render - when there's a `MoveItem` (ie, when someone clicks move to...), the `MoveItem` is set and renders the modal. This allows the modal to be fully responsible for its own state.
- implements a move endpoint on our backend
  - currently, this uses `z.coerce` due to issues w/ `BigInt` even though the migration isn't on this branch. will check if this is required after merging changes on `master` into the branch

## Screenshots and videos
Figma link here: https://www.figma.com/design/g7LnLn4IUcUC9dYxgsHEJM/IsomerCMS-V2-(working-title%3A-%22Isomer-Studio%22)?node-id=7393-18008&m=dev

https://github.com/user-attachments/assets/f03af4c5-28f8-4463-8764-b67d2c63d436


apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -0,0 +1,74 @@
...
+    )
+    .query(async ({ input: { resourceId } }) => {
+      const resource = await db
+        .selectFrom("Resource")
+        .where("Resource.id", "=", resourceId)
+        .select(["permalink", "Resource.id", "Resource.title"])
+        .executeTakeFirst()
+
+      if (!resource) {
+        throw new TRPCError({ code: "NOT_FOUND" })
```
harishv7, line 22: why not use executetakefirstorthrow?

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -0,0 +1,74 @@
...
+      }),
+    )
+    .query(async ({ input: { resourceId } }) => {
+      let query = db
+        .selectFrom("Resource")
+        .select(["title", "permalink", "type", "id"])
+        .orderBy("type", "asc")
+        .orderBy("title", "asc")
+
+      if (resourceId === null) {
```
harishv7, line 41: nit: could use `!resourceId`

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -0,0 +1,74 @@
...
+        query = query.where("parentId", "is", null)
+      } else {
+        query = query.where("Resource.parentId", "=", resourceId)
+      }
+
+      return query.execute()
+    }),
+  move: protectedProcedure
+    .input(
+      z.object({
```
harishv7, line 51: should we move out validation schemas in a separate file for consistency?

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -1,5 +1,6 @@
 import { MenuButton, MenuList, Portal } from "@chakra-ui/react"
```
harishv7, line 1: Some lint, typechecks failures in I

seaerchin: Fixed

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -1,5 +1,6 @@
...
 import { IconButton, Menu } from "@opengovsg/design-system-react"
+import { useSetAtom } from "jotai"
```
harishv7, line 3: I saw your description as to what this is, but can share why do we need this to be an atom with the 3 modes?

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: not doing - `null` is meant to actually be the `null` value not any falsy value. ie, an empty string shouldn't be treated the same as `null` here hence the explicit check

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: yes we should - lemme get on it

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: we wnat to return 400 not 5xx!

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -0,0 +1,86 @@
...
+// but validate that the string conforms to bigint.
+// Oddly enough, kysely doesn't allow `bigint` to query
+const bigIntSchema = z
+  // NOTE: A valid `bigint` is one that
+  // begins with a non-zero digit
+  // and has length > 1
+  .string()
+  .min(1)
+  .regex(/^[0-9]+$/)
+  .refine((v) => v.at(0) !== "0")
```
seaerchin, line 17: we shud use `node-pg-types` - see [here](https://github.com/brianc/node-pg-types) so that we can use js `Number` seamlessly without the extra conversion(s)

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
seaerchin: the 3 modes are provided by `jotai`. atoms are the base building blocks from `jotai` so if we use `jotai`, we have to use the atoms

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: done

apps/studio/src/server/modules/resource/resource.router.ts
karrui: yea. Also might have resourceId of 0, etc seeing how it's a big int, may lead to edge cases (though i'm sure they start from 1, but semantically wrong)

apps/studio/src/server/modules/resource/resource.router.ts
karrui: Cannot convert bigint to numbers. it will be lossy. we should just use uuids instead of bigints lol.

seaerchin: i don't convert ah - we're still using bigint. this is just validating that the value is a numerical string that is greater than 0

seaerchin:
* **#436** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#418** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#417** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/417?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#403** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -10,14 +11,24 @@ import {
...
-export const ResourceTableMenu = ({ title, type }: ResourceTableMenuProps) => {
+export const ResourceTableMenu = ({
+  resourceId,
+  title,
+  type,
+  permalink,
+}: ResourceTableMenuProps) => {
+  const setMoveResourceId = useSetAtom(moveResourceAtom)
+  const prepareMoveItem = () =>
+    setMoveResourceId({ resourceId, title, permalink })
```
karrui, line 31: nit: should we rename this for clarity? 

```suggestion
    setMoveResource({ resourceId, title, permalink })
```

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -10,14 +11,24 @@ import {
...
 
-export const ResourceTableMenu = ({ title, type }: ResourceTableMenuProps) => {
+export const ResourceTableMenu = ({
+  resourceId,
+  title,
+  type,
+  permalink,
+}: ResourceTableMenuProps) => {
+  const setMoveResourceId = useSetAtom(moveResourceAtom)
+  const prepareMoveItem = () =>
```
karrui, line 30: nit: prefer standard naming

```suggestion
  const handleMoveResourceClick = () =>
```

apps/studio/src/features/editing-experience/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface MovedItem {
```
karrui, line 1: nit: technically not moved yet

```suggestion
export interface PendingMoveResource {
```

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+      }
+      // NOTE: Set special styling because pages should show normally
+      // but have no onClick
+      // TODO: add permissions for folders that users have no access to
+      // These should have disabled styling
+      _disabled={{ color: "base.content.default" }}
+      pl="2.25rem"
+      onClick={onChangeResourceId}
+    >
+      {icon}
```
karrui, line 57: nit: buttons have a `leftIcon` prop that can be used for the correct styling:

```suggestion
    <Button
      variant="clear"
      w="full"
      justifyContent="flex-start"
      color="base.content.default"
      // NOTE: Homepage is special - it's a page but
      // we also use it to represent site root
      isDisabled={
        (type === "Page" && permalink !== "/") ||
        resourceId === movedItem?.resourceId
      }
      // NOTE: Set special styling because pages should show normally
      // but have no onClick
      // TODO: add permissions for folders that users have no access to
      // These should have disabled styling
      _disabled={{ color: "base.content.default" }}
      pl="2.25rem"
      onClick={onChangeResourceId}
      leftIcon={icon}
    >
```

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+      // we also use it to represent site root
+      isDisabled={
+        (type === "Page" && permalink !== "/") ||
+        resourceId === movedItem?.resourceId
+      }
+      // NOTE: Set special styling because pages should show normally
+      // but have no onClick
+      // TODO: add permissions for folders that users have no access to
+      // These should have disabled styling
+      _disabled={{ color: "base.content.default" }}
```
karrui, line 53: shouldn't a disabled/forbidden move be demarcated as such? i.e. have some disabled state. looks like that in the design too

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+                  w="full"
+                  justifyContent="flex-start"
+                  color="base.content.default"
+                  onClick={onBack}
+                >
+                  <HStack spacing="0.25rem" color="interaction.links.default">
+                    <BiLeftArrowAlt />
+                    <Text textStyle="caption-1">Back to parent folder</Text>
+                  </HStack>
+                </Button>
```
karrui, line 131: supernit: in figma, seems to be using a `Link` component. We can utilise the `as` prop to reduce restyling efforts (and prevent drift):

```suggestion
                <Link
                  size="xs"
                  as="button" // this allows usage of Link styles, but render as button
                  onClick={onBack}
                >
                  <HStack spacing="0.25rem">
                    <BiLeftArrowAlt />
                    <Text>Back to parent folder</Text>
                  </HStack>
                </Button>
```

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+    <Button
+      variant="clear"
+      w="full"
+      justifyContent="flex-start"
+      color="base.content.default"
+      // NOTE: Homepage is special - it's a page but
+      // we also use it to represent site root
+      isDisabled={
+        (type === "Page" && permalink !== "/") ||
+        resourceId === movedItem?.resourceId
```
karrui, line 47: i think we should invert this logic too (in case we add more types, like in my future PR `Collection` and `CollectionPage`. We should only allow movement to only `Folder` types.

But actually I can do it in my PR when the bug surfaces haha

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -0,0 +1,65 @@
...
+        query = query.where("parentId", "is", null)
+      } else {
+        query = query.where("Resource.parentId", "=", String(resourceId))
+      }
+
+      return query.execute()
+    }),
+  move: protectedProcedure
+    .input(moveSchema)
+    .mutation(async ({ input: { movedResourceId, destinationResourceId } }) => {
```
karrui, line 47: we should validate in the backend that the destinationResourceId is a valid type (that can have resources under it).

e.g. check destination type is "Folder"

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -0,0 +1,65 @@
...
+        .selectFrom("Resource")
+        .where("id", "=", String(movedResourceId))
+        .select([
+          "parentId",
+          "Resource.id",
+          "Resource.type",
+          "Resource.permalink",
+          "Resource.title",
+        ])
+        .executeTakeFirst()
```
karrui, line 63: should this be in a transaction? read after write consistency guarantee. idk if can actually be affected or not but no harm just wrapping in tx haha

karrui (approved):
lgtm, but we should make our backend validate possible moves. if not only relying on frontend to prevent a page from being moved under a page.

after launch, we should think about adding constraints directly on the DB layer -- no `parentId` can point to a page

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
seaerchin: good idea updated

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
seaerchin: done

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
seaerchin: done

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
seaerchin: i didn't do because it'll never be disabled until permissions come in - i'll do it here i guess then next time we can just pass the prop!

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
seaerchin: donezxc

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
seaerchin: updated sir

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: donezxc

seaerchin:
### Merge activity

* **Aug 8, 3:23 AM EDT**: @seaerchin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/403).

PR #415, opened 2024-08-02
fix: remove siderail for first-level and fix TOC

dcshzj:
## Problem

Siderail on first-level pages don't really make sense, and the table of contents broke after a previous schema update to the heading attributes.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Removed siderail for first-level pages.
- Fixed anchor ID generation for level 2 headings.

### DIFF
--- a/packages/components/src/templates/next/layouts/Content/Content.tsx
+++ b/packages/components/src/templates/next/layouts/Content/Content.tsx
@@ -115,8 +115,12 @@ const transformContent = (content: ContentPageSchemaType["content"]) => {
             const anchorId = getDigestFromText(
               `${JSON.stringify(component)}_${getRandomNumberBetIntervals(1, 1000)}`,
             )
+            const newAttrs = {
+              ...component.attrs,
+              id: anchorId,
+            }
 
-            return { ...component, id: anchorId }
+            return { ...component, attrs: newAttrs }
           } else {
             return component
           }
@@ -136,7 +140,14 @@ const createContentLayoutStyles = tv({
     container:
       "mx-auto grid max-w-screen-xl grid-cols-12 px-6 py-12 md:px-10 md:py-16 lg:gap-6 xl:gap-10",
     siderailContainer: "relative col-span-3 hidden lg:block",
-    content: "col-span-12 flex flex-col gap-16 lg:col-span-9 lg:ml-24",
+    content: "col-span-12 flex flex-col gap-16 lg:col-span-9",
+  },
+  variants: {
+    isSideRailPresent: {
+      true: {
+        content: "lg:ml-24",
+      },
+    },
   },
 })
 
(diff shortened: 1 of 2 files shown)


packages/components/src/templates/next/layouts/Content/Content.tsx
```diff
@@ -136,7 +140,14 @@ const createContentLayoutStyles = tv({
     container:
       "mx-auto grid max-w-screen-xl grid-cols-12 px-6 py-12 md:px-10 md:py-16 lg:gap-6 xl:gap-10",
     siderailContainer: "relative col-span-3 hidden lg:block",
-    content: "col-span-12 flex flex-col gap-16 lg:col-span-9 lg:ml-24",
+    content: "col-span-12 flex flex-col gap-16 lg:col-span-9",
+  },
+  variants: {
+    isSideRailPresent: {
+      true: {
+        content: "lg:ml-24",
```
karrui, line 148: might want to set `lg:col-span-9` here too so content can take up the full 12 columns if siderail is not present, instead of being on the `content` base slot key

packages/components/src/templates/next/layouts/Content/Content.tsx
```diff
@@ -136,7 +140,14 @@ const createContentLayoutStyles = tv({
     container:
       "mx-auto grid max-w-screen-xl grid-cols-12 px-6 py-12 md:px-10 md:py-16 lg:gap-6 xl:gap-10",
     siderailContainer: "relative col-span-3 hidden lg:block",
-    content: "col-span-12 flex flex-col gap-16 lg:col-span-9 lg:ml-24",
+    content: "col-span-12 flex flex-col gap-16 lg:col-span-9",
+  },
+  variants: {
+    isSideRailPresent: {
+      true: {
```
karrui, line 147: hmmm might we want to ensure siderail is hidden too since now we have this variant prop

```suggestion
      true: {
        siderailContainer: "lg:block col-span-3"
```

then remove that on the root too haha

packages/components/src/templates/next/layouts/Content/Content.tsx
dcshzj: Ah design wants to keep it to 9 columns only, so the right side is whitespace.

Still pending a decision on what to do with the "Back to top" button, might be putting it on the right side for such pages.

karrui (approved):

packages/components/src/templates/next/layouts/Content/Content.tsx
dcshzj: Ok decided not to have the "Back to top" button, will leave the right side purely whitespace for now.

packages/components/src/templates/next/layouts/Content/Content.tsx
dcshzj: Got a conditional render below that renders the siderail only if it is not null

PR #430, opened 2024-08-05
feat(components): Hero banner tokens and styling

sehyunidaaa:
- Apply tokens to hero banner
- Adjust min height and y-padding
- Adjust button gaps and wrapping behaviour
- Adjust gradient for better contrast in middle breakpoints

### DIFF
--- a/packages/components/src/templates/next/components/complex/Hero/Hero.tsx
+++ b/packages/components/src/templates/next/components/complex/Hero/Hero.tsx
@@ -13,25 +13,30 @@ const Hero = ({
 }: HeroProps) => {
   return (
     <section
-      className="flex min-h-[15rem] bg-cover bg-center bg-no-repeat sm:min-h-[22.5rem] md:min-h-[31.25rem]"
+      className="flex min-h-[15rem] bg-cover bg-center bg-no-repeat sm:min-h-[22.5rem] lg:min-h-[31.25rem]"
       style={{
         backgroundImage: `url('${backgroundUrl}')`,
       }}
     >
-      <div className="w-full content-center bg-gradient-to-r from-[rgba(0,0,0,95%)] to-[rgba(0,0,0,70%)] md:from-[rgba(0,0,0,75%)] xl:from-[rgba(0,0,0,95%)]">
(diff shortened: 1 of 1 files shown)


packages/components/src/templates/next/components/complex/Hero/Hero.tsx
```diff
@@ -13,25 +13,30 @@ const Hero = ({
...
+      className="flex min-h-[15rem] bg-cover bg-center bg-no-repeat sm:min-h-[22.5rem] lg:min-h-[31.25rem]"
       style={{
         backgroundImage: `url('${backgroundUrl}')`,
       }}
     >
-      <div className="w-full content-center bg-gradient-to-r from-[rgba(0,0,0,95%)] to-[rgba(0,0,0,70%)] md:from-[rgba(0,0,0,75%)] xl:from-[rgba(0,0,0,95%)]">
+      <div className="w-full content-center bg-gradient-to-r from-[rgba(0,0,0,85%)] to-[rgba(0,0,0,10%)] xl:from-[rgba(0,0,0,100%)]">
         <div
-          className={`${ComponentContent} flex flex-row justify-start p-10 text-start text-white`}
+          className={`${ComponentContent} flex flex-row justify-start py-16 text-start text-base-content-inverse`}
```
karrui, line 23: nit: prefer using twMerge so duplicates can be removed properly

```tsx
import { twMerge } from "~/lib/twMerge"
// ...
twMerge('flex flex-row justify-start p-10 text-start text-white', ComponentContent)
```

But actually I realise that constant is just an extra class lol. But how is it used ah anyone knows? @dcshzj?

karrui (approved):
lgtm

packages/components/src/templates/next/components/complex/Hero/Hero.tsx
dcshzj: Clarified offline, this is the special CSS class so we can let the layout decide the amount of padding that the component should have. By right all components should have it but right now this is limited to only what we think are homepage components.

This was added in #52.

PR #439, opened 2024-08-06
feat: set up S3 connector to get presigned PUT URL

dcshzj:
## Problem

We don't have a proper S3 connector to handle asset uploading.

## Solution

**Breaking Changes**

**Features**:

- Add basic S3 connector to handle uploading of assets. For now it only performs getting a pre-signed PUT URL.
- The filename key is in the format `${siteId}/${UUID}/${fileName}`.


apps/studio/src/server/modules/asset/asset.service.ts
```diff
@@ -0,0 +1,21 @@
...
+import { generateSignedPutUrl } from "~/lib/s3"
+
+export const getFileKey = ({
+  siteId,
+  fileName,
+}: z.infer<typeof getPresignedPutUrlSchema>) => {
+  // NOTE: We're using a random folder name to prevent collisions
+  const folderName = randomUUID()
+
+  return `${siteId}/${folderName}/${fileName}`
```
seaerchin, line 14: hwy not just use the resource id since it's guaranteed to be unique

seaerchin (approved):

apps/studio/src/server/modules/asset/asset.service.ts
dcshzj: Hmm this is the asset itself and it's possible to have multiple images within the same page resource, so will need a more globally unique ID.

PR #442, opened 2024-08-07
feat: add schema migration to add nullable site theme to database

karrui:
currently hardcoded in tooling/template/data/config.json, but we should store it in the DB so each site can have their own colors when building

### DIFF
--- a/apps/studio/prisma/migrations/20240807033704_add_site_theme_column_to_site_table/migration.sql
+++ b/apps/studio/prisma/migrations/20240807033704_add_site_theme_column_to_site_table/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Site" ADD COLUMN     "theme" JSONB;
--- a/apps/studio/prisma/schema.prisma
+++ b/apps/studio/prisma/schema.prisma
@@ -125,6 +125,9 @@ model Site {
   /// @kyselyType(PrismaJson.SiteJsonConfig)
   /// [SiteJsonConfig]
   config      Json
+  /// @kyselyType(PrismaJson.SiteThemeJson)
+  /// [SiteThemeJson]
+  theme       Json?
   navbar      Navbar?
   footer      Footer?
 }
--- a/apps/studio/prisma/types.ts
+++ b/apps/studio/prisma/types.ts
@@ -9,13 +9,16 @@ import type {
   IsomerPageSchemaType as _IsomerPageSchemaType,
   IsomerSchema as _IsomerSchema,
   IsomerSiteConfigProps as _IsomerSiteConfigProps,
+  IsomerSiteThemeProps as _IsomerSiteThemeProps,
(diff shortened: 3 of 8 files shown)


karrui:
* **#443** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/443?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#442** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/studio/prisma/migrations/20240807033704_add_site_theme_column_to_site_table/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Site" ADD COLUMN     "theme" JSONB;
```
seaerchin, line 2: not sure i understand this, why would theme be a jsonb here rather than just an enum? storing jsonb implies there's some config that we are taking into studio and by extension, allowing the user to customise but i thought the point of using the render engine was to have set themes, so allowing customisation on the theme level seems abit surprising to me. not sure if there's something i'm missing? (there probably is)

apps/studio/prisma/types.ts
```diff
@@ -9,13 +9,16 @@ import type {
...
   IsomerSchema as _IsomerSchema,
   IsomerSiteConfigProps as _IsomerSiteConfigProps,
+  IsomerSiteThemeProps as _IsomerSiteThemeProps,
   IsomerSiteWideComponentsProps as _IsomerSiteWideComponentsProps,
 } from "@opengovsg/isomer-components"
 
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace PrismaJson {
+    // TODO: Rename all with XXXYYYJson instead of XXXJsonYYY
```
seaerchin, line 19: agree with this comment, wondering if we shuold just have a nested type like `site.config`?

seaerchin (approved):
lgtm, ignore my comments on why it's a json. noticed that we were injecting colours as an object so it makes perfect sense now since we want to allow theming.

apps/studio/prisma/schema.prisma
```diff
@@ -125,6 +125,9 @@ model Site {
   /// @kyselyType(PrismaJson.SiteJsonConfig)
   /// [SiteJsonConfig]
   config      Json
+  /// @kyselyType(PrismaJson.SiteThemeJson)
+  /// [SiteThemeJson]
+  theme       Json?
```
seaerchin, line 130: nit: wondering if it'd be better to remove this since we can always just give a default theme but not very importatnt since they are functionally equivalent

karrui:
### Merge activity

* **Aug 7, 5:00 AM EDT**: @karrui started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/442).

PR #449, opened 2024-08-07
feat: support uploading of assets to S3

dcshzj:
## Problem

We don't actually persist uploading of assets to the S3 bucket.

## Solution

**Breaking Changes**

**Features**:

- Add support for uploading of assets to the provided S3 bucket in the environment variables.
    - Assets are only uploaded upon saving changes to the complex form editor.
    - NOT SUPPORTED YET:
        - ~Retrieving of the existing image as a File inside the Attachment component.~ (#458)
        - ~Marking removed images as deleted~ (#453)
        - ~Handling of malware scan response~ (#453)
        - ~Display of placeholder image~ (#452)
        - ~Displaying the actual image on preview and final site by adding the domain~ (#452)


apps/studio/next.config.mjs
```diff
@@ -25,7 +25,14 @@ const ContentSecurityPolicy = `
...
   script-src 'self' 'unsafe-eval';
   style-src 'self' https: 'unsafe-inline';
-  connect-src 'self' https://schema.isomer.gov.sg https://browser-intake-datadoghq.com https://*.browser-intake-datadoghq.com https://vitals.vercel-insights.com/v1/vitals;
+  connect-src
+    'self'
+    https://schema.isomer.gov.sg
+    https://browser-intake-datadoghq.com
+    https://*.browser-intake-datadoghq.com
+    https://vitals.vercel-insights.com/v1/vitals
+    https://*.amazonaws.com
```
harishv7, line 34: should this be stricter? with s3.amazonaws?

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -104,6 +111,59 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
+      const updatedBlocks = Array.from(previewPageState.content)
+      const newBlock = _.cloneDeep(updatedBlocks[currActiveIdx])
+
+      if (!newBlock) {
+        return
+      }
+
+      // Upload all new/modified images/files
+      const assetsToUpload = modifiedAssets.filter((asset) => !!asset.file)
+      await Promise.all(
```
harishv7, line 127: What happens if one of the upload fails?

apps/studio/next.config.mjs
dcshzj: The hostname is like `bucket-name.s3.ap-southeast-1.amazonaws.com`, so the `connect-src` will need to be something like `*.s3.*.amazonaws.com` which is not supported because of the double wildcard.

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
dcshzj: All the uploads should fail and the not found placeholder image will be shown. It is done this way because only when everything is successful, then we will update the block in the database.

karrui: you'll need a try catch here i think

apps/studio/next.config.mjs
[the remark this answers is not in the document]
dcshzj: Updated in 89b2d6c9da1c8538c8fd9068966c81970308a10a to use the specific domain that is dependent on the env vars.

dcshzj: Realised that we already have the bucket-name and region information from the env vars so went ahead to use them inside the CSP. Updated in 89b2d6c9da1c8538c8fd9068966c81970308a10a.

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
dcshzj: Added the try-catch + toast message in 89b2d6c9da1c8538c8fd9068966c81970308a10a.

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -104,6 +112,69 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
+              return
+            }
+
+            return uploadAsset({ file }).then((res) =>
+              _.set(newBlock, path, res.path),
+            )
+          }),
+        )
+      } catch (e) {
+        // All the promises are rejected if any one of them fails
```
harishv7, line 141: Why should all the promises be rejected? Should we be using allSettled to upload as many as we can?

harishv7 (approved):
Lgtm with some minor comments/qns

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
dcshzj: Adjusted in 8752fd749c5b7b072e30d76e12b1e3c77dfa1943!

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -104,6 +109,87 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
+          return !results.some(
+            (result) => result.status === "fulfilled" && result.value === path,
+          )
+        })
+
+        if (newModifiedAssets.length > 0) {
+          toast({
+            title: "Error uploading assets",
+            description:
+              "An error occurred while uploading some assets. Please try again later.",
```
harishv7, line 150: For clarity: Should we tell the user something like 2/10 assets failed to upload. Please try again.

cc @sehyunidaaa

harishv7 (approved):
lgtm will minor comment on the user msg

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
dcshzj: Hmm opted not to include this information because there is no further feedback to the user on which assets are the ones that are failing, hence keeping it generic.

sehyunidaaa: when does this error occur? where would a user upload multiple assets?

sehyunidaaa: nits regardless of the context (i assumed we only support 1 image at a time now): 
- would avoid the word "assets" for user-facing copies! files or images if can be specific. if not, `An error occurred while uploading files/images. Try again later.` would also do
- if Harish's recommendation isn't doable, would be great if there's some (visual) indication on the studio which ones failed vs. which ones succeeded! that way the user can recover from the error by retrying the ones that failed

dcshzj: Yup @sehyunidaaa in general we will only have 1 image at a time, but this also applies when the particular component can have multiple images (e.g. multiple infocards with images). I can show the number of failed images but it might not be easy to show exactly which component has errors.

Adjusted in 2fd161656ed3c823494b80645dce02cd1fa8af92!

PR #450, opened 2024-08-07
feat(components): update Accordion styling, use twMerge in Prose

karrui:
### TL;DR

Refactor Accordion component and stories for enhanced functionality and code cleanliness.

### What changed?
- Updated Accordion component to use local state to manage the open/close state of each accordion item.
- Replaced the `BiChevronDown` icon with a dynamic `BiMinus` and `BiPlus` icon for better UX feedback based on accordion state.
- Utilized the `tv` function for creating accordion styles for better maintainability.
- Modified `BaseParagraph` component to use `twMerge` for merging TailwindCSS classes more effectively.
- Improved `Prose` component to use a `ProseComponent` helper for cleaner JSX.

### How to test?
Storybook

### Why make this change?
To improve the readability, maintainability, and user experience of the Accordion component and related stories.

### DIFF
--- a/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
+++ b/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
@@ -1,5 +1,3 @@
-"use client"
-
 import type { VariantProps } from "tailwind-variants"
 import { BiMinus, BiPlus } from "react-icons/bi"
 
--- a/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
+++ b/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
@@ -14,7 +14,7 @@ const createAccordionStyles = tv({
       "group border-y border-divider-medium px-4 py-5 has-[+_details]:border-b-0",
     summary:
       "prose-headline-lg-medium flex list-none flex-row items-center justify-between gap-3 text-base-content-strong outline-none outline outline-0 outline-offset-2 outline-brand-interaction hover:cursor-pointer focus-visible:outline-2",
-    icon: "h-6 w-6",
+    icon: "h-6 w-6 flex-shrink-0",
     content: "pt-5 text-base-content-strong",
   },
 })
--- a/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
+++ b/packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
@@ -1,17 +1,39 @@
-import { BiChevronDown } from "react-icons/bi"
+import { useState } from "react"
+import { BiMinus, BiPlus } from "react-icons/bi"
 
 import type { AccordionProps } from "~/interfaces"
+import { tv } from "~/lib/tv"
 import { Prose } from "../../native"
 
+const createAccordionStyles = tv({
+  slots: {
+    details:
+      "group border-y border-divider-medium px-4 py-5 has-[+_details]:border-b-0",
+    summary:
+      "prose-headline-lg-medium flex list-none flex-row items-center justify-between gap-3 text-base-content-strong outline-none outline outline-0 outline-offset-2 outline-brand-interaction hover:cursor-pointer focus-visible:outline-2",
+    icon: "h-6 w-6",
+    content: "pt-5 text-base-content-strong",
+  },
+})
+
+const accordionStyles = createAccordionStyles()
+
 const Accordion = ({ summary, details }: AccordionProps) => {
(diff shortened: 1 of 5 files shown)


karrui:
* **#450** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/450?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sehyunidaaa:
@karrui we're not changing the schema yet right?

karrui:
> @karrui we're not changing the schema yet right?

No

packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
```diff
@@ -1,17 +1,49 @@
...
+  },
+})
+
+const accordionStyles = createAccordionStyles()
+
+interface AccordionProps
+  extends BaseAccordionProps,
+    VariantProps<typeof createAccordionStyles> {
+  name?: string
+}
```
dcshzj, line 28: question(blocking): this `name` prop cannot be modified directly by the user as it is not part of the schema. Is this setup in preparation for the accordion group?

Also, these props should be part of the interfaces folder so it is easier to abstract out when we split up the package in the future.

packages/components/src/templates/next/components/native/Prose/Prose.tsx
```diff
@@ -7,37 +9,42 @@ import OrderedList from "../OrderedList"
...
+        <BaseParagraph
+          content={getTextAsHtml(component.content)}
+          className="prose-body-base text-base-content"
+        />
+      )
+    case "table":
+      return <Table {...component} />
+    case "unorderedList":
+      return <UnorderedList {...component} />
+  }
```
dcshzj, line 35: nit: should we add the `never` case here?

packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
```diff
@@ -1,17 +1,49 @@
...
+  const [isOpen, setIsOpen] = useState(false)
+  const Icon = isOpen ? BiMinus : BiPlus
+
   return (
-    <details className="border-y border-divider-medium hover:cursor-pointer has-[+_details]:border-b-0 [&_>_summary_>_svg]:open:rotate-180">
-      <summary className="my-[0.125rem] flex list-none flex-row items-center gap-3 px-4 py-5 font-medium text-content outline-none text-paragraph-01">
+    <details
+      name={name}
+      onToggle={(e) => setIsOpen(e.currentTarget.open)}
+      open={isOpen}
```
dcshzj, line 38: question(non-blocking): are we able to utilise just CSS for this and leave it to the browser to handle the disclosure? Was thinking along the lines of `[&_>_summary_>_#plus]:open:hidden [&_>_summary_>_#minus]:open:block`.

packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
karrui: yea, if we want to actually only allow one accordion to open at a time. But actually I'll probably remove this first since that is not an expected feature yet.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#name

packages/components/src/templates/next/components/native/Prose/Prose.tsx
karrui: no need, cause the return type of the component forces a `JSX.Element`. Will fail if the return type is undefined.

packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
karrui: good idea. done in [2505bad](https://github.com/opengovsg/isomer/pull/450/commits/2505bad3ab6c83e6e50f2f5bf642863d2532514d)

packages/components/src/templates/next/components/complex/Accordion/Accordion.tsx
```diff
@@ -1,17 +1,45 @@
-import { BiChevronDown } from "react-icons/bi"
+"use client"
```
dcshzj, line 1: nit: can remove this directive now that we don't use `useState`.

dcshzj (approved):

karrui:
### Merge activity

* **Aug 14, 1:41 AM EDT**: @karrui merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/450).

PR #437, opened 2024-08-06
feat: add folder views

seaerchin:
### TL;DR

- Updated focus rings in `CmsSideNav` component for better accessibility.
- Fixed pagination display issues in `Datatable` component.
- Added new route for folder details page under `/sites/[siteId]/folders/[folderId]/`.
- Adjusted button size in site page to maintain consistency.
- Updated `folder.router.ts` to include title and permalink in the response.

### What changed?

- Changed `p` to `m` in `CmsSideNav` component to ensure proper margin for focus ring.
- Removed unnecessary check on `totalRowCountString` in `Datatable` component.
- Added components and hooks for the new folder details page.
- Modified button size in `/sites/[siteId]/` from `xs` to `md`.
- Updated folder router to return title and permalink.

### How to test?

1. **Focus Ring**: Navigate to the CMS sidebar and check focus ring behavior on navigation items.
2. **Pagination**: Verify pagination works correctly in the `Datatable` component without hidden elements.
3. **Folder Page**: Access a folder using the new route and ensure all elements load correctly.
4. **Button Size**: Check the button size consistency on the site page.
5. **Router**: Confirm the API returns title and permalink along with other details for a folder.

### Why make this change?

- Improve the accessibility and user experience of the navigation components.
- Ensure pagination elements are displayed correctly.
- Introduce the folder details page for enhanced content management.
- Maintain UI consistency across the application.
- Provide more data context in API responses for better frontend handling.

PR #436, opened 2024-08-06
feat: add new site side nav

seaerchin:
This PR implements a new side navigation component (`CmsSideNav`) and integrates it into the CMS layout. The change updates the layout structure within the `CmsSidebarContainer`, adding support for a new sidebar section specifically for navigation purposes. Additionally, the PR includes the necessary modifications to the `AdminCmsSidebarLayout` to include the new side navigation.

PR #418, opened 2024-08-02
chore: shift page.list to resource

seaerchin:
### TL;DR

Consolidates the page listing functionality into the resource module and updates related components.

### What changed?

- Replaced `page.list` calls with `resource.list` calls in various components and modals.
- Updated `resource.router.ts` to include the `list` procedure previously in `page.router.ts`.
- Removed the `list` procedure from `page.router.ts`.
- Adjusted invalidation calls to use `utils.resource.list` instead of `utils.page.list`.

### How to test?

- Go to the base site page
- ensure that you can see pages/folders
- create a new page/folder
- the ui should be updated

### Why make this change?

This change aims to streamline the codebase by consolidating listing functionalities, reducing redundancy, and improving maintainability.

PR #451, opened 2024-08-07
feat: add disclosure for settings

seaerchin:
### TL;DR

This PR introduces a folder settings modal in the folder page.

### What changed?

- Imported necessary components like `Modal`, `Input`, `FormControl`, etc.
- Added folder settings modal state management using `useDisclosure` hook.
- Created a new `FolderSettingsModal` component for folder settings.
- Utilized the `useZodForm` for form handling and validation.
- Implemented the mutation to save folder settings changes to the database.
- Updated schema and router to support editing folder's title and permalink.

### How to test?

1. Navigate to the folder page.
2. Click on the `Folder settings` button.
3. Update the title and permalink in the modal.
4. Click `Save changes` and verify if the changes are reflected.

### Why make this change?

This change adds the ability to edit the folder title and permalink directly from the folder page, enhancing the user experience by providing a straightforward way to manage folder settings.

---

https://github.com/user-attachments/assets/206c6663-663b-4c8d-a002-25dfc2c25a72

### DIFF
--- a/apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
+++ b/apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
@@ -13,11 +13,11 @@ import {
   VStack,
 } from "@chakra-ui/react"
 import { Breadcrumb, Button, Menu } from "@opengovsg/design-system-react"
-import { uniqueId } from "lodash"
 import { BiData, BiFileBlank, BiFolder } from "react-icons/bi"
 import { z } from "zod"
 
 import { MenuItem } from "~/components/Menu"
+import { FolderSettingsModal } from "~/features/dashboard/components/FolderSettingsModal"
 import { ResourceTable } from "~/features/dashboard/components/ResourceTable"
 import { CreateFolderModal } from "~/features/editing-experience/components/CreateFolderModal"
 import { CreatePageModal } from "~/features/editing-experience/components/CreatePageModal"
@@ -27,8 +27,8 @@ import { AdminCmsSidebarLayout } from "~/templates/layouts/AdminCmsSidebarLayout
 import { trpc } from "~/utils/trpc"
 
 const folderPageSchema = z.object({
-  siteId: z.coerce.number(),
-  folderId: z.coerce.number(),
+  siteId: z.string(),
+  folderId: z.string(),
 })
 
 const FolderPage: NextPageWithLayout = () => {
@@ -42,14 +42,18 @@ const FolderPage: NextPageWithLayout = () => {
     onOpen: onFolderCreateModalOpen,
     onClose: onFolderCreateModalClose,
   } = useDisclosure()
+  const {
+    isOpen: isFolderSettingsModalOpen,
+    onOpen: onFolderSettingsModalOpen,
+    onClose: onFolderSettingsModalClose,
+  } = useDisclosure()
 
   const { folderId, siteId } = useQueryParse(folderPageSchema)
 
-  const [{ title, permalink, children, parentId }] =
-    trpc.folder.readFolder.useSuspenseQuery({
-      siteId,
-      resourceId: folderId,
-    })
+  const [{ title, permalink }] = trpc.folder.readFolder.useSuspenseQuery({
+    siteId: parseInt(siteId),
+    resourceId: parseInt(folderId),
+  })
 
   return (
     <>
@@ -78,7 +82,11 @@ const FolderPage: NextPageWithLayout = () => {
             <Spacer />
 
             <HStack>
-              <Button variant="outline" size="md">
+              <Button
+                variant="outline"
+                size="md"
+                onClick={onFolderSettingsModalOpen}
+              >
                 Folder settings
               </Button>
               <Menu isLazy size="sm">
@@ -109,19 +117,27 @@ const FolderPage: NextPageWithLayout = () => {
           </HStack>
         </VStack>
         <Box width="100%">
-          <ResourceTable siteId={siteId} resourceId={folderId} />
+          <ResourceTable
+            siteId={parseInt(siteId)}
+            resourceId={parseInt(folderId)}
+          />
         </Box>
       </VStack>
       <CreatePageModal
         isOpen={isPageCreateModalOpen}
         onClose={onPageCreateModalClose}
-        siteId={siteId}
+        siteId={parseInt(siteId)}
       />
       <CreateFolderModal
         isOpen={isFolderCreateModalOpen}
         onClose={onFolderCreateModalClose}
+        siteId={parseInt(siteId)}
+      />
+      <FolderSettingsModal
+        isOpen={isFolderSettingsModalOpen}
+        onClose={onFolderSettingsModalClose}
         siteId={siteId}
-        key={uniqueId()}
+        resourceId={parseInt(folderId)}
       />
     </>
   )
--- a/apps/studio/src/server/modules/folder/folder.router.ts
+++ b/apps/studio/src/server/modules/folder/folder.router.ts
@@ -1,24 +1,27 @@
-import { createFolderSchema, readFolderSchema } from "~/schemas/folder"
+import {
+  createFolderSchema,
+  editFolderSchema,
+  readFolderSchema,
+} from "~/schemas/folder"
 import { protectedProcedure, router } from "~/server/trpc"
 import { db } from "../database"
+import { defaultFolderSelect } from "./folder.select"
 
 export const folderRouter = router({
   create: protectedProcedure
     .input(createFolderSchema)
-    .mutation(
-      async ({ ctx, input: { folderTitle, parentFolderId, ...rest } }) => {
-        const folder = await db
-          .insertInto("Resource")
-          .values({
-            ...rest,
-            type: "Folder",
-            title: folderTitle,
-            parentId: parentFolderId ? String(parentFolderId) : null,
-          })
-          .executeTakeFirstOrThrow()
-        return { folderId: folder.insertId }
-      },
-    ),
+    .mutation(async ({ input: { folderTitle, parentFolderId, ...rest } }) => {
+      const folder = await db
+        .insertInto("Resource")
+        .values({
+          ...rest,
+          type: "Folder",
+          title: folderTitle,
+          parentId: parentFolderId ? String(parentFolderId) : null,
+        })
+        .executeTakeFirstOrThrow()
+      return { folderId: folder.insertId }
+    }),
   readFolder: protectedProcedure
     .input(readFolderSchema)
     .query(async ({ ctx, input }) => {
@@ -63,5 +66,20 @@ export const folderRouter = router({
         children,
         parentId,
       }
+    }),
+  editFolder: protectedProcedure
+    .input(editFolderSchema)
+    .mutation(async ({ input: { resourceId, permalink, title, siteId } }) => {
+      return db
+        .updateTable("Resource")
+        .where("Resource.id", "=", resourceId)
+        .where("Resource.siteId", "=", Number(siteId))
+        .where("Resource.type", "=", "Folder")
+        .set({
+          permalink,
+          title,
+        })
+        .returning(defaultFolderSelect)
+        .execute()
     }),
 })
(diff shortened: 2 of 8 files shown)


seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/451?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#451** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#437** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/437?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#436** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#418** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#417** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/417?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#403** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/403?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -42,14 +68,18 @@ const FolderPage: NextPageWithLayout = () => {
...
   const { folderId, siteId } = useQueryParse(folderPageSchema)
 
-  const [{ title, permalink, children, parentId }] =
-    trpc.folder.readFolder.useSuspenseQuery({
-      siteId,
-      resourceId: folderId,
-    })
+  const [{ title, permalink }] = trpc.folder.readFolder.useSuspenseQuery({
+    siteId: parseInt(siteId),
+    resourceId: parseInt(folderId),
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:81]: we should refrain from all the casting and transformations imo. easy to desync or lose reasoning on why the types are the way they are. as discussed offline, since we do not actually do any arithmetic on the resource ids, we technically can keep them in their expected types: siteId: int, resourceId: string. Server should be able to pass them to the query without any transformations too

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -109,23 +143,205 @@ const FolderPage: NextPageWithLayout = () => {
...
+  const toast = useToast()
+  const { mutate, isLoading } = trpc.folder.editFolder.useMutation({
+    onSettled: onClose,
+    onSuccess: () => {
+      void utils.site.list.invalidate()
+      void utils.resource.list.invalidate()
+      void utils.resource.getChildrenOf.invalidate({
+        resourceId: parentId ? String(parentId) : null,
+      })
+      void utils.folder.readFolder.invalidate()
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:230]: we should actually await these instead of voiding right

```suggestion
    onSuccess: async () => {
      await utils.site.list.invalidate()
      await utils.resource.list.invalidate()
      await utils.resource.getChildrenOf.invalidate({
        resourceId: parentId ? String(parentId) : null,
      })
      await utils.folder.readFolder.invalidate()
```

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -109,23 +143,205 @@ const FolderPage: NextPageWithLayout = () => {
...
+        title: "",
+        permalink: "",
+      },
+      schema: editFolderSchema.omit({ siteId: true, resourceId: true }),
+    })
+  const [{ title: originalTitle, permalink: originalPermalink, parentId }] =
+    trpc.folder.readFolder.useSuspenseQuery({
+      siteId: parseInt(siteId),
+      resourceId,
+    })
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:218]: since we have the original title and permalink, shouldn't we fetch the data first before setting the default values? 

```suggestion
  const [{ title: originalTitle, permalink: originalPermalink, parentId }] =
    trpc.folder.readFolder.useSuspenseQuery({
      siteId: parseInt(siteId),
      resourceId,
    })
    
  const { setValue, register, handleSubmit, watch, formState, getFieldState } =
    useZodForm({
      defaultValues: {
        title: originalTitle,
        permalink: originalPermalink,
      },
      schema: editFolderSchema.omit({ siteId: true, resourceId: true }),
    })
```

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -109,23 +143,205 @@ const FolderPage: NextPageWithLayout = () => {
...
+
+  useEffect(() => {
+    if (originalPermalink) {
+      setValue("title", originalTitle)
+    }
+
+    if (originalPermalink) {
+      setValue("permalink", originalPermalink)
+    }
+  }, [originalPermalink, originalTitle])
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:269]: if we set the defaults immediately when we fetch the folder metadata, there is no need for this effect anymore

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -109,23 +143,205 @@ const FolderPage: NextPageWithLayout = () => {
...
+        resourceId={parseInt(folderId)}
       />
     </>
   )
 }
 
 FolderPage.getLayout = AdminCmsSidebarLayout
 export default FolderPage
+
+interface FolderSettingsModalProps {
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:175]: should we extract all these non-page related components into their own `feature` directory instead of inlining in this page? We should keep pages scoped to rendering the page.

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -1,24 +1,26 @@
...
+        .insertInto("Resource")
+        .values({
+          ...rest,
+          type: "Folder",
+          title: folderTitle,
+          parentId: parentFolderId ? String(parentFolderId) : null,
+        })
+        .executeTakeFirstOrThrow()
+      return { folderId: folder.insertId }
+    }),
```
karrui, line 23: comment: seems to be formatting desync? does not look like there are changes. Why CI never catch :O

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -64,4 +66,37 @@ export const folderRouter = router({
         parentId,
       }
     }),
+  editFolder: protectedProcedure
+    .input(editFolderSchema)
+    .mutation(async ({ input: { resourceId, permalink, title, siteId } }) => {
+      if (!title && !permalink) {
+        return
+      }
```
karrui, line 74: would this ever happen? we should not need to do anymore validation since we have a zod schema validating input right? if it's not meant to happen maybe we should prevent frontend from submitting this in the first place.

if you need a schema where you must have one of the two fields, we could update the schema:

```ts
z.object({
  resourceId: z.string(),
  permalink: z.string().optional(),
  title: z.string().optional(),
  siteId: z.string(),
})
  .superRefine((data, ctx) => {
  if (!data.permalink && !data.title) {
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      path: ["siteId"],
      message: "Either permalink or title must be provided.",
    })
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      path: ["title"],
      message: "Either permalink or title must be provided.",
    })
  }
})
```

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -64,4 +66,37 @@ export const folderRouter = router({
...
+          })
+        }
+
+        await baseQuery.execute()
+        return db
+          .selectFrom("Resource")
+          .where("Resource.id", "=", resourceId)
+          .executeTakeFirstOrThrow()
+      })
+    }),
```
karrui, line 101: You don't need the conditional, any undefined assignment is stripped. You can also return directly after updating the row with `RETURNING`. This removes the need for a transaction (and multiple queries). Also suggest returning the default select instead of... nothing. 
Actually that will throw runtime error since its invalid SQL. But kysely cannot catch.

```suggestion
        return await db
          .updateTable("Resource")
          .where("Resource.id", "=", resourceId)
          .where("Resource.siteId", "=", Number(siteId))
          .where("Resource.type", "=", "Folder")
          .set({
            permalink,
            title
          })
          .returning(defaultFolderSelect)
          .execute()
```

karrui (changes requested):

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -1,34 +1,60 @@
...
+  MAX_FOLDER_PERMALINK_LENGTH,
+  MAX_FOLDER_TITLE_LENGTH,
+} from "~/schemas/folder"
 import { AdminCmsSidebarLayout } from "~/templates/layouts/AdminCmsSidebarLayout"
 import { trpc } from "~/utils/trpc"
 
 const folderPageSchema = z.object({
-  siteId: z.coerce.number(),
-  folderId: z.coerce.number(),
+  siteId: z.string(),
```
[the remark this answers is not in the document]
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:56]: wa actually i changed it to `string` here - i agree with the reasoning you gave that the `bigint` is just to autoincrement automatically for id, but on the frontend we can just treat as a string since we'll never interact with it.

i think changing to `string` here is ok - i'll have a PR to change number -> string later on wdyt?

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:230]: eh i derped - you're right

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:81]: i agree. how do we feel about a PR later on to tidy these types since they are unlikely to change? i want to do a 1 shot change rather than in phases so0 that post change, it's easier to reference and take code

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:230]: actually it's not jsut here - more widespread! i'll correct it here and elsewhere!

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:218]: thanks for the catch - updated!

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: done

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:269]: done

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:175]: ps i forgot about this - will extract into separate component file

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: not sure what's the intended outcome here - can hint abit?

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: tbh i didn't go down this route because `superRefine` returns an error code and i was thinking this can just be a no-op (if you never give me anything to update, just skip updating)

not very opiniated here - i'm happy to make this change!

karrui (approved):
lgtm after you fix ci

seaerchin:
### Merge activity

* **Aug 12, 11:51 AM EDT**: @seaerchin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/451).

PR #452, opened 2024-08-08
feat: support using assets domain for images

dcshzj:
## Problem

The image component do not support having a specified assets domain when rendering images.

## Solution

**Breaking Changes**

**Features**:

- Add a new `assetsBaseUrl` optional site configuration to allow component users to specify a base URL for the assets domain. If not provided, relative paths will be used instead.
- Added a new `caption` optional prop in the Image schema.
- Added support for automatically scaling up images if the screen size is too small, and the width specified by the user is too small to be reasonably seen when using small devices.
- Centralised the definition of an external URL for easy updating in the future.


packages/components/src/utils/isExternalUrl.ts
```diff
@@ -0,0 +1,9 @@
+// Determine if the provided URL is an external link
+export const isExternalUrl = (url: string) => {
+  return (
+    url.startsWith("http://") ||
```
harishv7, line 4: actually there are a lot more schemes:
https://omni.wikiwand.com/en/articles/List_of_URI_schemes

packages/components/src/templates/next/components/complex/Image/Image.tsx
```diff
@@ -1,26 +1,88 @@
...
-        >
-          <BaseImage src={src} alt={alt} width={width}></BaseImage>
-        </a>
-      ) : (
-        <BaseImage src={src} alt={alt} width={width}></BaseImage>
-      )}
-    </div>
+    <ImageContainer href={href} LinkComponent={LinkComponent}>
+      <img
+        src={isExternalUrl(src) ? src : `${assetsBaseUrl}${src}`}
```
harishv7, line 73: how are you asserting assetsBaseUrl is present for all external URLs?

dcshzj: Hmm we are not asserting that it must be present (hence the undefined). If it is not present, it is treated as if it is not present and rendered as a relative link which is the existing behaviour.

karrui: if `assetsBaseUrl` is undefined, string templating will output the `"undefined"` string, which may not be what we want yeah

dcshzj: Ah okay, updated in e51400d2b8372bc37cc56426a5e9ebd4d168ee18.

packages/components/src/utils/isExternalUrl.ts
dcshzj: Good point, updated to use the internal link definition in d4e8aaea0c44ce5089c9d33c5f2076e68a548192.

karrui (approved):
lgtm

packages/components/src/templates/next/components/complex/Image/Image.tsx
```diff
@@ -1,26 +1,93 @@
...
+// Enforce a certain minimum width of image if screen size is small
+const getWidthSize = (width: ImageProps["width"]) => {
+  if (width === undefined || width >= 66) {
+    return "large"
+  } else if (width >= 40) {
+    return "medium"
+  } else {
+    return "small"
+  }
+}
```
dcshzj, line 38: note to self: to check with design @sehyunidaaa on what the outcome is regarding the sizing, and whether we want to do that change right now or in a future PR

PR #453, opened 2024-08-08
feat: add soft delete functionality for assets

dcshzj:
## Problem

We do not support deleting assets from our S3 bucket when references to it have been removed.

## Solution

**Breaking Changes**

**Features**:

- Added functionality to perform a soft delete on assets. The actual cleanup will happen outside of Studio as part of some cron process which will be implemented later.


karrui:
vercel broke, might need to add the missing env. also add it into our infra yeah

apps/studio/src/constants/asset.ts
karrui, line 1: should we use enums here instead? like more semantic

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -61,6 +61,8 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
 
   const { mutateAsync: uploadAsset, isLoading: isUploadingAsset } =
     useUploadAssetMutation({ siteId })
+  const { mutateAsync: deleteAsset, isLoading: isDeletingAsset } =
```
karrui, line 64: ~~should we use `mutate` over `mutateAsync`? like this if anything throws it will crash the app since errors will not be handled by react-query.~~

~~See [Mastering Mutations in React Query section](https://tkdodo.eu/blog/mastering-mutations-in-react-query#mutate-or-mutateasync)~~

edit: nvm, saw that you are firing off dependent mutations. async seems to be the easiest way for that.

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -137,7 +139,18 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
+      )
+      await Promise.all(
+        assetsToDelete.map(({ src }) => {
+          if (!src) {
+            return
+          }
+
+          return deleteAsset({ fileKey: src })
+        }),
+      )
```
karrui, line 153: I suspect we can update the trpc procedure to take in an array of assets to delete, instead of having this Promise.all.

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -137,7 +139,18 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
+      const assetsToDelete = modifiedAssets.filter(
+        (asset) => asset.src !== undefined,
+      )
+      await Promise.all(
+        assetsToDelete.map(({ src }) => {
+          if (!src) {
+            return
+          }
+
+          return deleteAsset({ fileKey: src })
```
karrui, line 151: since this is a `mutateAsync` call, we should add a try catch ourselves here so we handle the error. Also applies to the above `uploadAsset` call, seems like that was missed in the PR review too.

apps/studio/src/constants/asset.ts
dcshzj: Removed in 8048a8b1cf56252092bca48bbffb1b21d2621957, not doing the handling of the malware scan result right now since we are exploring using S3 access points which in theory seems to be a much cleaner solution than having multiple S3 buckets.

karrui (approved):

PR #454, opened 2024-08-08
fix: remove fetch

seaerchin:
# tl;dr 
- we were refetching on every refocus and this led to the drawer state being reset.
- in order to avoid this, we have to avoid the refetch

# what changed
- removed the fetch in drawers since it was never being used ._.


karrui (approved):
lgtm. Assume there will be future PRs to remove refetch on other invocations of readPageAndBlob yeah?

PR #461, opened 2024-08-12
fix: prevent modal title from overflowing to close button

dcshzj:
## Problem

The modal title sometimes overflows to the close button.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Set a max width to the modal title to prevent it from overflowing to the close button.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-08-12 at 14 07 38

**AFTER**:

Screenshot 2024-08-12 at 14 07 57

### DIFF
--- a/apps/studio/src/features/editing-experience/components/DiscardChangesModal/DiscardChangesModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/DiscardChangesModal/DiscardChangesModal.tsx
@@ -25,9 +25,10 @@ export const DiscardChangesModal = ({
     <Modal isOpen={isOpen} onClose={onClose}>
       <ModalOverlay />
       <ModalContent>
-        <ModalHeader>
+        <ModalHeader maxW="90%">
           Are you sure you want to discard your changes?
         </ModalHeader>
+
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/editing-experience/components/DiscardChangesModal/DiscardChangesModal.tsx
```diff
@@ -25,9 +25,10 @@ export const DiscardChangesModal = ({
     <Modal isOpen={isOpen} onClose={onClose}>
       <ModalOverlay />
       <ModalContent>
-        <ModalHeader>
+        <ModalHeader maxW="90%">
```
karrui, line 28: maybe a padding-right to the icon width + some would be more "correct". If not on small screens the 10% might still still overlap.

dcshzj: Ah right, fixed in b4a9cc906d318474bb8fdb3f133ee99a6b16328d.

karrui (approved):

PR #463, opened 2024-08-12
feat: introduce new index page layout

dcshzj:
## Problem

The current content page layout is not suitable for the index pages.

## Solution

**Breaking Changes**

**Features**:

- Introduce a new index page layout specifically for index pages, which is adapted from the existing content layout but excludes the max-width.

## Before & After Screenshots

Screenshot 2024-08-12 at 19 44 50


karrui (approved):
lgtm. if sehyun approve i also approve 😆

PR #474, opened 2024-08-14
fix: home styling

seaerchin:
### TL;DR

- Added a new `RootpageRow` component to display root page information on the site dashboard.
- Updated the usage of the resource list query to the new `listWithoutRoot` endpoint to exclude root pages from the general resource list.

### What changed?

1. Introduced the `RootpageRow` component to show basic details of the root page, such as title and status (draft or published), on the site dashboard.
2. Modified the `FolderSettingsModal` to wrap `SuspendableModalContent` with a `Suspense` component using a `Skeleton` as a fallback.
3. Changed several instances of `resource.list` to `resource.listWithoutRoot` to exclude root pages from resource listings.
4. Added `getRootPage` query to the `page.router` to support fetching root page details.

### How to test?

1. Navigate to the site dashboard and verify that the root page information is displayed correctly using the new `RootpageRow` component.
2. Check that opening the folder settings modal shows the loading skeleton while content is being fetched.
3. Ensure that resource listings do not include the root page.

### Why make this change?

This change aims to improve the user interface by clearly displaying the root page information on the dashboard and ensuring resource listings are more accurate by excluding root pages.

### DIFF
--- a/apps/studio/src/features/dashboard/components/DeleteResourceModal/DeleteResourceModal.tsx
+++ b/apps/studio/src/features/dashboard/components/DeleteResourceModal/DeleteResourceModal.tsx
@@ -96,7 +96,7 @@ const DeleteResourceModalContent = ({
     onSuccess: async () => {
       // TODO: here and elsewhere, we should aim to simplify our query pattern
       // such that the invalidation logic is clear
-      await utils.resource.list.invalidate()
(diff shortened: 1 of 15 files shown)


karrui (approved):
lgtm, but need to fix stories! I think stories don't have a homepage, that's why it's failing to render. we need to guarantee a root home page when a site is created too.

PR #478, opened 2024-08-15
feat: use starter-kitty for redirect validator

alexanderleegs:
See [thread](https://opengovproducts.slack.com/archives/CK9GGK6EP/p1722841650033399). This PR modifies our redirect url validation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,5 +31,8 @@
   "devDependencies": {
     "turbo": "^2.0.5"
   },
-  "packageManager": "npm@10.2.3"
+  "packageManager": "npm@10.2.3",
+  "dependencies": {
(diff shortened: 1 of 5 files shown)


package.json
karrui, line 1: since this is a monorepo, i think the package should be installed in the correct package!

karrui (changes requested):
@alexanderleegs can help change the directory of the installed package to studio?

package.json
alexanderleegs: whoops, fixed in https://github.com/opengovsg/isomer/pull/478/commits/05bb46285bb68e020a68255e8359d948cdfa9a7e

karrui (approved):
lgtm

PR #485, opened 2024-08-16
feat: implement pagination for resource, collection, and getChildrenOf procedures

karrui:
### TL;DR

Implemented pagination for the Collection and Resource tables, improving performance and user experience for large datasets.

### What changed?

- Added a new `useTablePagination` hook to manage pagination state
- Updated `CollectionTable` and `ResourceTable` components to use pagination
- Modified the backend API to support offset-based pagination
- Added a new `count` endpoint for resources to get the total number of items
- Removed the `totalRowCountString` prop from the `Datatable` component
- Updated the `Datatable` component to use absolute positioning for the loading overlay

### How to test?

1. Navigate to a Collection or Resource table with more than 25 items
2. Verify that only 25 items are displayed initially
3. Use the pagination controls to navigate through the pages
4. Confirm that the total item count is displayed correctly
5. Check that the loading state is shown when fetching new pages

### Why make this change?

This change improves the performance and user experience when dealing with large datasets in Collection and Resource tables. By implementing pagination, we reduce the initial load time and memory usage, while providing a more efficient way to navigate through large sets of data. This also lays the groundwork for future improvements such as sorting and filtering capabilities.

### DIFF
--- a/apps/studio/src/server/modules/resource/resource.router.ts
+++ b/apps/studio/src/server/modules/resource/resource.router.ts
@@ -1,6 +1,8 @@
 import { TRPCError } from "@trpc/server"
 
+import type { ResourceType } from "../database"
 import {
+  countResourceSchema,
   deleteResourceSchema,
   getChildrenSchema,
   getMetadataSchema,
@@ -8,7 +10,7 @@ import {
   moveSchema,
 } from "~/schemas/resource"
 import { protectedProcedure, router } from "~/server/trpc"
-import { db, ResourceType } from "../database"
+import { db } from "../database"
 
 export const resourceRouter = router({
   getMetadataById: protectedProcedure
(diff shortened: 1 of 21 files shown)


karrui:
* **#485** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/485?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

seaerchin (approved):
1 question but overall lgtm. do you think we should standardise the way we return children items so that the user gets a consistent sorting order? don't have to do it in this PR, just raising for discussion

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -26,10 +28,43 @@ export const resourceRouter = router({
...
+  getFolderChildrenOf: protectedProcedure
+    .input(getChildrenSchema)
+    .query(async ({ input: { resourceId, cursor: offset, limit } }) => {
+      let query = db
+        .selectFrom("Resource")
+        .select(["title", "permalink", "type", "id"])
+        .where("Resource.type", "in", ["RootPage", "Folder"])
+        .$narrowType<{
+          type: "Folder" | "RootPage"
+        }>()
```
seaerchin, line 40: do you think we shuold also allow `Collection` here? i saw this is only used for `move`, which ew disallow for collections atm but just food for thought

karrui:
OMG rebase againnizcdzxcdxc

apps/studio/src/server/modules/resource/resource.router.ts
karrui: ya, collections not here solely because you cannot move into collections.

PR #489, opened 2024-08-19
fix: table caption will scroll along with the table on horizontal scroll

karrui:
Update table caption to be outside of table html tag, so the positioning is not affected by the table styles and overflow.

a11y is not compromised; it's an alternative as described in https://www.w3.org/WAI/tutorials/tables/caption-summary/#using-aria-describedby-to-provide-a-table-summary

### DIFF
--- a/packages/components/src/templates/next/components/native/Table/Table.tsx
+++ b/packages/components/src/templates/next/components/native/Table/Table.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { useEffect, useRef, useState } from "react"
+import { useEffect, useId, useRef, useState } from "react"
 
 import type { TableProps } from "~/interfaces"
 import { tv } from "~/lib/tv"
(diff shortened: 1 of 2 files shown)


packages/components/src/templates/next/components/native/Table/Table.tsx
```diff
@@ -85,66 +87,70 @@ const Table = ({ attrs: { caption }, content }: TableProps) => {
...
-        </caption>
-        <tbody>
-          {content.map((row, index) => {
-            const TableCellTag =
-              row.content[0]?.type === "tableHeader" ? "th" : "td"
+    <div className="flex flex-col gap-4">
+      <BaseParagraph
+        id={tableDescriptionId}
+        content={caption}
+        className="prose-label-md-regular text-base-content-subtle [&:not(:last-child)]:mb-0"
```
dcshzj, line 94: I think it will always not be the last child?

```suggestion
        className="prose-label-md-regular text-base-content-subtle mb-0"
```

dcshzj (approved):

packages/components/src/templates/next/components/native/Table/Table.tsx
karrui: no choice. your `BaseParagraph` component has the exact same rule but adds `mb-6` 😆 . I need to override

dcshzj: LOL okok damn

PR #490, opened 2024-08-19
fix(ci): workflow

seaerchin:
## Problem
previous workflow for deploy was missing stuff, this PR adds them in

## Solution


apps/studio/package.json
```diff
@@ -16,7 +16,7 @@
     "predev": "run-s migrate",
     "dev": "next dev",
     "prebuild": "run-s generate",
-    "build": "next build",
+    "build": "SKIP_ENV_VALIDATION=true next build",
```
karrui, line 19: maybe shouldn't right? it should fail properly if the envs are not provided

karrui (approved):

apps/studio/package.json
seaerchin: this implies that all of our env vars are known at build, which isn't true. we only know at runtime, where ssm gets connected ot our task def and the task is deployed.

fwiw, i think morpho + bright both do this and i couldn't find a better way

karrui: morpho doesn't do this. it should work because by the time this `build` script is invoked the environment variables should already have been "injected"

PR #499, opened 2024-08-20
feat(components): add keys where required when mapping over

karrui:
### TL;DR

Added missing `key` props to improve React rendering performance and resolve console warnings.

### What changed?

- Added `key` props to various components in multiple files:
  - `KeyStatistics.tsx`: Added `key={index}` to the mapped div elements
  - `ArticlePageHeader.tsx`: Changed `key={Math.random()}` to `key={index}` for list items
  - `Siderail.tsx`: Added `key={index}` to the `li` element
  - `SiderailList.tsx`: Moved `key={index}` from `SiderailLabel` to the parent `li` element
- Updated `renderComponent` function in `render.tsx` to accept and use an `elementKey` prop
- Modified `renderPageContent` in `renderPageContent.ts` to pass `elementKey={index}` to `renderComponent`

### How to test?

1. Run the application and navigate through pages with components like KeyStatistics, ArticlePageHeader, Siderail, and SiderailList
2. Open the browser console and verify that there are no warnings related to missing `key` props
3. Interact with the components to ensure they still function as expected

### Why make this change?

Adding proper `key` props to elements in loops is a React best practice. It helps React identify which items have changed, been added, or been removed, leading to more efficient updates and better performance. This change also eliminates console warnings about missing keys, improving the overall code quality and developer experience.

---

### DIFF
--- a/packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
+++ b/packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
@@ -26,8 +26,9 @@ const KeyStatistics = ({ title, statistics }: KeyStatisticsProps) => {
         {title}
       </h2>
       <div className="flex flex-col flex-wrap gap-x-8 gap-y-12 md:flex-row">
-        {statistics.slice(0, MAX_ITEMS).map(({ label, value }) => (
+        {statistics.slice(0, MAX_ITEMS).map(({ label, value }, index) => (
           <div
+            key={index}
             className={`flex grow flex-col gap-3 ${
               ITEM_WIDTHS[Math.min(MAX_ITEMS, statistics.length)]
             }`}
--- a/packages/components/src/templates/next/components/internal/ArticlePageHeader/ArticlePageHeader.tsx
+++ b/packages/components/src/templates/next/components/internal/ArticlePageHeader/ArticlePageHeader.tsx
@@ -29,8 +29,8 @@ const ArticlePageHeader = ({
             <BaseParagraph content={summary[0]} />
           ) : (
             <ul className="list-disc ps-7">
-              {summary.map((item) => (
-                <li key={Math.random()} className="pl-0.5 [&_p]:inline">
+              {summary.map((item, index) => (
+                <li key={index} className="pl-0.5 [&_p]:inline">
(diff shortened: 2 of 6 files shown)


karrui:
* **#499** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/499?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

seaerchin (approved):

seaerchin:
fwiw, i reviewed just by checking `keys` - seemed ok. **however**, [react docs](https://react.dev/learn/rendering-lists#why-does-react-need-keys) specify that actually the index is used as the key so i think this only resolves the warning HAHA 

> You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.
>
> Similarly, do not generate keys on the fly, e.g. with key={Math.random()}. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.
>
> Note that your components won’t receive key as a prop. It’s only used as a hint by React itself. If your component needs an ID, you have to pass it as a separate prop: <Profile key={id} userId={id} />.

karrui:
> fwiw, i reviewed just by checking `keys` - seemed ok. **however**, [react docs](https://react.dev/learn/rendering-lists#why-does-react-need-keys) specify that actually the index is used as the key so i think this only resolves the warning HAHA
> 
> > You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.
> > Similarly, do not generate keys on the fly, e.g. with key={Math.random()}. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.
> > Note that your components won’t receive key as a prop. It’s only used as a hint by React itself. If your component needs an ID, you have to pass it as a separate prop: .

ya only for removing warning haha

PR #507, opened 2024-08-22
fix(vapt): update sanitization

seaerchin:
## Problem
Previously, we had a path traversal attack against us where the directory name was not generated correctly in the presence of `../..`.

to counteract this, we adopt a strict validation of only allowing alphanumeric + `-`, similar to pages.

## Solution
1. use same validator throughout fe/be and for pages/collections/folder
2. for frontend, update so that disallowed characters (non alphanum) are automatically replaced with `-`

### DIFF
--- a/apps/studio/src/schemas/common.ts
+++ b/apps/studio/src/schemas/common.ts
@@ -1,0 +1,11 @@
+import { z } from "zod"
+
+export const generateBasePermalinkSchema = (label: string) =>
+  z
+    .string({
+      required_error: `Enter a URL for this ${label}.`,
+    })
+    // Using `*` instead of `+` to allow empty strings, so the correct required error message is shown instead of the regex error message.
+    .regex(/^[a-z0-9\-]*$/, {
+      message: "Only lowercase alphanumeric characters and hyphens are allowed",
+    })
--- a/apps/studio/src/schemas/folder.ts
+++ b/apps/studio/src/schemas/folder.ts
@@ -1,8 +1,16 @@
 import { z } from "zod"
 
+import { generateBasePermalinkSchema } from "./common"
+
 export const MAX_FOLDER_TITLE_LENGTH = 100
 export const MAX_FOLDER_PERMALINK_LENGTH = 200
 export const MAX_FOLDER_DESCRIPTION_LENGTH = 300
+
+const permalinkSchema = generateBasePermalinkSchema("folder")
+  .min(1, { message: "Enter a URL for this folder" })
+  .max(MAX_FOLDER_PERMALINK_LENGTH, {
+    message: `Folder URL should be shorter than ${MAX_FOLDER_PERMALINK_LENGTH} characters.`,
+  })
 
 export const createFolderSchema = z.object({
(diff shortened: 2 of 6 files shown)


apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -76,7 +76,13 @@ export const folderRouter = router({
...
         .where("Resource.id", "=", resourceId)
         .where("Resource.siteId", "=", Number(siteId))
-        .where("Resource.type", "=", "Folder")
+        .where((eb) =>
+          eb("Resource.type", "=", "Folder").or(
+            "Resource.type",
+            "=",
+            "Collection",
+          ),
+        )
```
karrui, line 85: nit: can be simplified with


```suggestion
        .where("Resource.type", "in", ["Folder", "Collection"])
```

karrui (approved):
Should we sanitize `template` too to have a second layer of sanitisation before building the site?

apps/studio/src/schemas/common.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from "zod"
+
+export const generateBasePermalinkSchema = (label: string) =>
+  z
+    .string({
+      required_error: `Enter a URL for this ${label}.`,
+    })
+    // Using `*` instead of `+` to allow empty strings, so the correct required error message is shown instead of the regex error message.
+    .regex(/^[a-z0-9\-]*$/, {
```
spaceraccoon, line 9: Is there a minimum length? Cos this regex allows 0 chars.
```suggestion
    .regex(/^[a-z0-9\-]+$/, {
```

apps/studio/src/schemas/folder.ts
```diff
@@ -1,22 +1,25 @@
...
+  .min(1, { message: "Enter a URL for this folder" })
+  .max(MAX_FOLDER_PERMALINK_LENGTH, {
+    message: `Folder URL should be shorter than ${MAX_FOLDER_PERMALINK_LENGTH} characters.`,
+  })
+
 export const createFolderSchema = z.object({
   folderTitle: z
     .string()
     .min(1, { message: "Enter a title for this folder" })
     .max(MAX_FOLDER_TITLE_LENGTH, {
```
spaceraccoon, line 19: Should traversal validation be added here too? See first step of VAPT poc

spaceraccoon (approved):
Good initial validation with some nits; I know we didn't add this to tooling yet but I think it's still worth doing on both ends of this vulnerability

apps/studio/src/schemas/common.ts
seaerchin: yeap it's actually intended so we show the proper error message! is this a security issue?

apps/studio/src/schemas/folder.ts
seaerchin: lemme clarify this iwth them

seaerchin: image
double checked, the title is irrelevant here, so it's ok!

PR #508, opened 2024-08-23
fix: update directory paths

seaerchin:
## Problem
We have a reverse code execution issue as we allow arbitrary file writes. this fixes the problem by sanitizing the path passed to our build.

## Solution
[normalize](https://nodejs.org/api/path.html#path_path_normalize_path) the filepath first so that only leading `..` is allowed. next, we strip the leading `../` because we'll be doing a `file.join` anyways. 

this still allows hidden files/folders: `/.git` is allowed, for example.

### DIFF
--- a/tooling/build/scripts/publishing/index.ts
+++ b/tooling/build/scripts/publishing/index.ts
@@ -124,14 +124,19 @@ async function writeContentToFile(
       return
     }
 
-    const sanitizedPermalink = path
-      // NOTE: normalization here will remove dual backslashes
-      // and also strip .. filepaths except as a prefix
-      .normalize(fullPermalink)
-      // NOTE: this matches on a leading ../
-      // or a leading ..\
-      // or a plain .. without any paths
-      .replace(/^(\.\.(\/|\\|$))+/, "")
+    // NOTE: do a join with ./ here so that
+    // we don't end up with an absolute path to a special unix folder
+    const sanitizedPermalink = path.join(
+      "./",
+      path
+        // NOTE: normalization here will remove dual backslashes
+        // and also strip .. filepaths except as a prefix
+        .normalize(fullPermalink)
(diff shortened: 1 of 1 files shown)


tooling/build/scripts/publishing/index.ts
```diff
@@ -124,12 +124,21 @@ async function writeContentToFile(
...
     }
 
+    const sanitizedPermalink = path
+      // NOTE: normalization here will remove dual backslashes
+      // and also strip .. filepaths except as a prefix
+      .normalize(fullPermalink)
+      // NOTE: this matches on a leading ../
+      // or a leading ..\
+      // or a plain .. without any paths
+      .replace(/^(\.\.(\/|\\|$))+/, "")
```
spaceraccoon, line 134:
```suggestion
      .normalize(path.join('./', fullPermalink))
```

Avoid regexes generally because they can be quite brittle, and a str replace is not a 1-for-1 match for how the fs or file read API you pass it to will actually interpret the output string; can't find a problem with regex atm tho! Just a more robust implementation imo.

spaceraccoon (changes requested):
Need to make more robust to avoid this bypass:

```
// bad: returns /etc/passwd, a valid POSIX path OUTSIDE current directory
path.normalize('..\\/..//..\\/etc/passwd').replace(/^(\.\.(\/|\\|$))+/, "")
// good: returns ..\\/etc/passwd, a valid POSIX path BUT within current directory
path.normalize(path.join('./', ('..\\/..//..\\/etc/passwd')))
```

tooling/build/scripts/publishing/index.ts
seaerchin: good catch, doing!

spaceraccoon (approved):

PR #515, opened 2024-08-23
chore: update mt

seaerchin:
## Problem
Update the settings page to fit design

Closes ISOM-1359

## Solution
1. minor misc fixes involving design
2. create a new layout that omits the side nav for the settings page
3. update button logic for settings page 

- opted to create a new layout rather than adding a new `prop`. i think having abit of repetition here in exchange for clear roles and not exploding the container component is something that i feel is worth it. the additional clarity i feel is worth the xtra code

### DIFF
--- a/apps/studio/src/templates/layouts/AdminSidebarOnlyLayout.tsx
+++ b/apps/studio/src/templates/layouts/AdminSidebarOnlyLayout.tsx
@@ -9,11 +9,13 @@ import {
   BiLinkExternal,
   BiLogOut,
 } from "react-icons/bi"
+import { z } from "zod"
 
 import type { CmsSidebarItem } from "~/components/CmsSidebar/CmsSidebarItems"
 import { EnforceLoginStatePageWrapper } from "~/components/AuthWrappers"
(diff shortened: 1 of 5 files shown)


apps/studio/src/templates/layouts/AdminSidebarOnlyLayout.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <CmsSidebarWrapper>{page}</CmsSidebarWrapper>
+      </Flex>
+    </EnforceLoginStatePageWrapper>
+  )
+}
+
+// Extracted out since this needs to be a child of EnforceLoginStatePageWrapper
+const CmsSidebarWrapper = ({ children }: PropsWithChildren) => {
+  const router = useRouter()
+  const siteId = String(router.query.siteId)
```
karrui, line 32: Nit: can probably use useQueryParse hook instead

karrui (approved):

apps/studio/src/templates/layouts/AdminSidebarOnlyLayout.tsx
seaerchin: done for this and the original one

PR #520, opened 2024-08-26
fix: last updated

seaerchin:
## Problem
last updated shows invalid date

## Solution
add it to the `page` props being set above

### DIFF
--- a/apps/studio/src/features/editing-experience/components/Preview.tsx
+++ b/apps/studio/src/features/editing-experience/components/Preview.tsx
@@ -35,6 +35,7 @@ function SuspendablePreview({
   const renderProps = merge(props, overrides, {
     page: {
       permalink,
+      // TODO: Fetch from DB in the future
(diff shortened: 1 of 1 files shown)


apps/studio/src/features/editing-experience/components/Preview.tsx
```diff
@@ -48,11 +49,6 @@ function SuspendablePreview({
         siteMap: { title: "Home", permalink: "/", children: [] },
         environment: "production",
         // TODO: Fetch from DB in the future
```
dcshzj, line 51: nit: this comment should move up together with the code change?

dcshzj (approved):

apps/studio/src/features/editing-experience/components/Preview.tsx
seaerchin: sounds good

PR #524, opened 2024-08-26
feat: perf improvements to tiptap menu, allow column headers in table

karrui:
### TL;DR
Enhanced tiptap editor perf using useMemo.
Allow table column headers.

### What changed?

- Added toggle options for header column and header row in the table menu
- Implemented useMemo for performance optimization of menu items
- Updated TableContentRowSchema to allow header cells within table rows
- Added new icons for table header toggles (RiLayoutColumnFill, RiLayoutRowFill)

### How to test?

1. Open the page editor and insert a table
2. Verify that new options for toggling header column and header row are available in the table menu
3. Test the functionality of these new toggle options
4. Create a table with mixed regular and header cells in a single row
5. Ensure that the editor correctly handles and displays the mixed cell types

### Why make this change?

This change improves the flexibility and usability of tables in the page editor. By allowing users to easily toggle header columns and rows, it enhances the ability to create more complex and informative tables. The schema update supports this new functionality by allowing header cells to exist alongside regular cells in table rows. The performance optimization ensures that the menu items are only recalculated when necessary, potentially improving the editor's responsiveness.

Now we can also have column headers. Might be useful for certain table types.

### DIFF
--- a/apps/studio/src/components/PageEditor/MenuBar.tsx
+++ b/apps/studio/src/components/PageEditor/MenuBar.tsx
@@ -1,6 +1,7 @@
 import type { MenuButtonProps, MenuListProps } from "@chakra-ui/react"
 import type { Editor } from "@tiptap/react"
 import type { IconType } from "react-icons/lib"
+import { useMemo } from "react"
 import {
   Box,
   Divider,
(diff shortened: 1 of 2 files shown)


karrui:
setting to draft, seems like the menu does not rerender -- does not update menu buttons.

seaerchin (approved):
lgtm!

documenting here for future: tiptap apis require a function, because it's (editor instance) is continually updating its internal state but the initial `Menubar` is only loaded once. giving it as a static value thus causes the `isHidden` to not work properly and require refreshes to bring it up to date w the UI. 

additionally, the `useMemo` here is to fix the 1 render cycle behind but i don't quite understand _how_ it does so (`useMemo` doesn't actually re-render but does _less_ renders - if deps don't change, it doesn't rerender the memoized item) but it's not really a blocker here

PR #530, opened 2024-08-26
feat: remove blockquote functionality from editor

karrui:
### TL;DR

Removed blockquote functionality from the TipTap editor.

### What changed?

- Removed the "Quote block" option from the MenuBar component in the PageEditor.
- Removed the Blockquote extension from the TipTapEditor component.
- Removed an unused query for page content in the TipTapComponent.

### How to test?

1. Open the PageEditor in the studio app.
2. Verify that the "Quote block" option is no longer present in the text formatting menu.
3. Attempt to create a blockquote in the editor and confirm that it's not possible.
4. Ensure that existing content and other text formatting options still work as expected.

### Why make this change?

This change simplifies the text editor by removing the blockquote functionality. It may have been determined that this feature was not necessary for the current use case or was causing issues in the editor. Removing unused code also helps to maintain a cleaner and more efficient codebase.

### DIFF
--- a/apps/studio/src/components/PageEditor/MenuBar.tsx
+++ b/apps/studio/src/components/PageEditor/MenuBar.tsx
@@ -142,23 +142,6 @@ export const MenuBar = ({ editor }: { editor: Editor }) => {
         },
         {
           type: "item",
-          title: "Quote block",
-          textStyle: "body-1",
-          useSecondaryColor: true,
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
removed from `uat` last time, left it in cos i thoiught we could fix but guess not rofl

PR #531, opened 2024-08-26
fix: accordion

seaerchin:
## Problem
Change the double click flow to accordion dropdown flow as it's better for user. This change also prevents the accordion from appearing if collections/folders **do not have children**.

Closes ISOM-1340
Closes ISOM-1342

## Solution
- Use a box as a wrapper element with `pos="relative"`. 
- Position the chevron (`AccordionButton`) absolutely inside the box
- reserve spacing for the chevron so that it doesn't look too weird
- add a **full width** link so on single click, we redirect that
- additionally, only show the chevon if 
   - `isAllowedToHaveChildren(resourceType) === true` **and**
   -  `isLoading` (we dk how many pages are in this folder, so potentially ambiguous - also disabled said chevron during load) **or**
   - `pages.items.length > 0`

## Screenshots and video

https://github.com/user-attachments/assets/3ec93b73-e87a-431c-b925-75648de1cfc4

### DIFF
--- a/apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
+++ b/apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
@@ -1,3 +1,5 @@
+import { PropsWithChildren } from "react"
+import NextLink from "next/link"
 import { useRouter } from "next/router"
 import {
   Accordion,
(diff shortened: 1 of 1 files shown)


seaerchin:
note that focus is currently wonked (not enough padding) 
image

**but with `p=4px` it looks humongous**
image

since the point of the left nav is for information, i prioritised info density > this edge case

apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
```diff
@@ -118,86 +166,41 @@ const SideNavItem = ({
...
-                    w="full"
-                    color="base.content.default"
-                    alignItems="center"
+                  {permalink}
+                </Text>
+              </SideNavRow>
+              {isAllowedToHaveChildren(resourceType) &&
+                (isLoading || hasChildren) && (
+                  <AccordionButton
+                    disabled={isLoading}
```
karrui, line 194: Nit: use the isDisabled prop instead for consistency? (ignore if it doesn’t exist)

karrui (approved):

apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
seaerchin: sadly doesn't exist
image

PR #535, opened 2024-08-26
fix: don't refetch on focus

seaerchin:
## Problem
previously, we refetched on focus for the root drawer. this led to components rerendering when the window loses focus and then the user tabs back. 

in order to remediate this problem, we just remove the fetch on refocus totally. however, we do need to be sure that we do a revalidate and refetch when we know that our data is stale. (this part isn't quite done yet)

## Solution
- set `refetchOnFocus: false` on the root drawer. compare hte behaviour in the 2 videos below

### With refetch on focus
note the jump back

https://github.com/user-attachments/assets/6027552b-fece-42eb-b75a-429f2924f359

### WithOUT refetch on focus

https://github.com/user-attachments/assets/e5366b20-59ab-4870-b11b-90d1ab6499a6

### DIFF
--- a/apps/studio/src/components/PageEditor/ComponentSelector.tsx
+++ b/apps/studio/src/components/PageEditor/ComponentSelector.tsx
@@ -97,12 +97,7 @@ function BlockItem({
   )
 }
 
-interface ComponentSelectorProps {
-  siteId: number
-  pageId: number
-}
-
-function ComponentSelector({ pageId, siteId }: ComponentSelectorProps) {
+function ComponentSelector() {
   const {
     setCurrActiveIdx,
     savedPageState,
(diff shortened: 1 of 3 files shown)


karrui (approved):
Should automatically refetch when invalidated anywhere. The setting just prevents it from refetching on focus. Your expected behaviour should be already done as long as the proper procedure gets invalidated

seaerchin:
> Should automatically refetch when invalidated anywhere. The setting just prevents it from refetching on focus. Your expected behaviour should be already done as long as the proper procedure gets invalidated

actually wad i was afraid of was the behaviour for invalidation when _others_ edit. since we do a fetch on refocus, this helps to keep data fresh so incidences where this happen is low. but ok not worth this behaviour

PR #536, opened 2024-08-26
refactor: change to skeleton

seaerchin:
## Problem
Design wants to use a skeleton for loading state and not a spiner

Closes ISOM-1426

## Solution
- Swap to skeleton.
- have to use 2 separate skeletons because initially `page` is potentially undefined. 
- i used the base homepage button to calculate a nice looking skeleton for the initial one.

No ss/vid cos i couldn't get the load in for the second skeleton. can try on `uat` or `main` if possible

### DIFF
--- a/apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
+++ b/apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
@@ -139,8 +139,8 @@ const SideNavItem = ({
   const siteProps = useQueryParse(siteSchema)
   const isCurResourceActive = getCurResource(siteProps) === resourceId
 
-  if (!data) {
-    return <Skeleton w="16rem" h="2.25rem" />
+  if (isLoading || !data) {
+    return <Skeleton isLoaded={!isLoading} w="16rem" h="2.25rem" />
   }
 
   const urlType = getResourceType(resourceType)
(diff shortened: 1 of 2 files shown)


apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
```diff
@@ -92,142 +93,140 @@ const SideNavItem = ({
...
   const { pages } = data
 
   return (
-    <Accordion
-      defaultIndex={resourceType === ResourceType.RootPage ? 0 : undefined}
-      allowToggle
-    >
-      <AccordionItem
-        _disabled={{
-          textColor: "interaction.support.disabled-content",
```
karrui, line 112: Typo

apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
```diff
@@ -137,12 +139,8 @@ const SideNavItem = ({
   const siteProps = useQueryParse(siteSchema)
   const isCurResourceActive = getCurResource(siteProps) === resourceId
 
-  if (isLoading || !data) {
-    return (
-      <Flex align="center" height="2rem" pl="2.75rem" pr="1rem">
-        <Spinner />
-      </Flex>
-    )
+  if (!data) {
```
karrui, line 142: this will always be true even if loading, meaning the skeleton below wrapping the component is useless. Might be easier to stick to the old conditional

```suggestion
  if (isLoading || !data) {
```

karrui (approved):

apps/studio/src/components/CmsSidebar/CmsSideNav.tsx
seaerchin: done! didn't check `width` on children cos the load takes forever, so might be the case where, for example, third child is doubly indented (arrow for illustration), but the fourth child (ie child of third child) is full size skeleton while loading: 

```
>> <child />
<Skeleton />
```

PR #537, opened 2024-08-26
chore: add max-w

seaerchin:
## Problem
**single-words** could overflow and cause bad styling. (think like 100++ characters long **without white space**)

## Solution
add a `maxW`

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/TipTapEditor.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/TipTapEditor.tsx
@@ -121,7 +121,7 @@ export function TiptapEditor({ data, handleChange }: TiptapEditorProps) {
   if (!editor) return null
 
   return (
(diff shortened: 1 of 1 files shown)


karrui (commented):
I think the fix for that is to actually use word-break: break-all;

The max width should be 100% on the editor so it fills its parent id think. 

[Read more](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text)

seaerchin:
> I think the fix for that is to actually use word-break: break-all;
> 
> The max width should be 100% on the editor so it fills its parent id think.
> 
> [Read more](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text)

thanks for pointing this out! i learnt about this previously but it somehow didn't cross my mind this time ._. it worked without `w=100%` so i'm omitting it for now!

karrui (approved):

PR #538, opened 2024-08-26
feat: collection settings

seaerchin:
## Problem
We need to have collection settings, which is missing at present from our collections list page.

## Solution
Reuse `FolderModal` and change some invalidation logic. The main chnage to invalidation logic is that we invalidate **specifically the resource id**. we also add invalidation of `collection.getMetadata` as this is what `CollectionsListPage` relies on to fetch.

## Notes
- not a fan of this pattern, where we lump `collection + page` together ;--; 
- the copy needs some change too 
(if got time i'll chnage both)

## Screenshots + video

https://github.com/user-attachments/assets/4d379076-d75a-46ad-8274-1dde38cc083d

### DIFF
--- a/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
@@ -95,12 +95,16 @@ const SuspendableModalContent = ({
   const { mutate, isLoading } = trpc.folder.editFolder.useMutation({
     onSettled: onClose,
     onSuccess: async () => {
-      await utils.site.list.invalidate()
       await utils.resource.listWithoutRoot.invalidate()
       await utils.resource.getChildrenOf.invalidate({
         resourceId: parentId ? String(parentId) : null,
       })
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
```diff
@@ -95,12 +95,16 @@ const SuspendableModalContent = ({
   const { mutate, isLoading } = trpc.folder.editFolder.useMutation({
     onSettled: onClose,
     onSuccess: async () => {
-      await utils.site.list.invalidate()
       await utils.resource.listWithoutRoot.invalidate()
       await utils.resource.getChildrenOf.invalidate({
         resourceId: parentId ? String(parentId) : null,
       })
-      await utils.folder.getMetadata.invalidate()
+      await utils.folder.getMetadata.invalidate({
```
karrui, line 102: Nit: We can probably invalidate everything; no big deal for things to fetch again. Invalidation doesn’t mean it triggers a refetch immediately anyways

karrui (approved):

PR #540, opened 2024-08-27
fix: add env vars for AWS access and secret keys

dcshzj:
## Problem

Vercel requires using AWS access and secret keys to access our S3 bucket.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Add the env vars for AWS access and secret keys.

### DIFF
--- a/apps/studio/src/lib/s3.ts
+++ b/apps/studio/src/lib/s3.ts
@@ -12,10 +12,14 @@ import { getSignedUrl } from "@aws-sdk/s3-request-presigner"
 
 import { env } from "~/env.mjs"
 
-const { NEXT_PUBLIC_S3_REGION } = env
+const { NEXT_PUBLIC_S3_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY } = env
 
 export const storage = new S3Client({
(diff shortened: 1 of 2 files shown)


harishv7 (approved):

apps/studio/src/lib/s3.ts
```diff
@@ -12,10 +12,14 @@ import { getSignedUrl } from "@aws-sdk/s3-request-presigner"
 
 import { env } from "~/env.mjs"
 
-const { NEXT_PUBLIC_S3_REGION } = env
+const { NEXT_PUBLIC_S3_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY } = env
 
 export const storage = new S3Client({
   region: NEXT_PUBLIC_S3_REGION,
+  credentials: {
+    accessKeyId: AWS_ACCESS_KEY_ID ?? "",
```
karrui, line 20: Nit: Should we not even pass in the credentials sub-object if it does not exist? Empty strings feel iffy over undefined.

karrui (approved):

dcshzj:
Too bad, doesn't work. Timeboxing effort, going ahead with using VAPT env instead

PR #544, opened 2024-08-27
fix: datatable

seaerchin:
## Problem
Previously, we showed an empty pagination when there are no items in the collection. To avoid this, we will do a conditional render only if `totalRowCount` is defined and greater than 0.

## Solution
remove pagination if total row count is 0

### Before
image

### After
image

### DIFF
--- a/apps/studio/src/components/Datatable/Datatable.tsx
+++ b/apps/studio/src/components/Datatable/Datatable.tsx
@@ -129,7 +129,7 @@ export const Datatable = <T extends object>({
         </Table>
       </Box>
       <Flex py="1rem" gap="1rem">
-        {pagination && (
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/Datatable/Datatable.tsx
```diff
@@ -129,7 +129,7 @@ export const Datatable = <T extends object>({
         </Table>
       </Box>
       <Flex py="1rem" gap="1rem">
-        {pagination && (
+        {pagination && !!totalRowCount && totalRowCount > 0 && (
```
karrui, line 132: i think the first check is sufficient

```suggestion
        {pagination && !!totalRowCount && (
```

karrui (approved):

PR #546, opened 2024-08-27
fix(ui): update three dots to go to parent

seaerchin:
## Problem
The three dots didn't do anything previously. This PR adds a new functionality to update teh 3 dots to go back to parent. This only works for folders >1 level deep (as if they're 1 level deep, the immediate parent is the root folder, so don't need 3 dots)

Closes ISOM-1433

## Solution
1. add new endpoint (`Resource.getParentOf`) 
2. invalidate `move` endpoint. we only need to invalidate this because this is the only end point that can chnage the parent. 

## Screenshots + video

https://github.com/user-attachments/assets/7cca9ce9-deab-487e-807b-ef6e7e37a34d


karrui (approved):
lgtm, 

comment: but can we actually show the first parent to go back to instead of clicking the triple dots? so fetch one level more for metadata to show on breadcrumb

PR #555, opened 2024-08-29
fix(ui): adding a new block and exiting without saving causes a block to be wrongly created

seaerchin:
## Problem
adding a block to our state and then exiting the editor without doing anything cause a wrong block to be created. this leads to issues later when dragging and dropping as the editor will attempt to update the blob/resourceid of a block that doesnt exist

## Solution
this is a pretty complicated fix so i'll try to explain as much detail as possible. 
1. the root cause of the bug is because we preemptively update the page state in our base `ComponentSelector`. This is to allow our components to use the page state to update the block. because of this, however, when we exit editing, we reset to the state **with the block**. ie, the sequence of events is as follows:
    - page state initially: []
    - page state after clicking on add component: [[]], where the inner array denotes teh initial empty state of the component
    - now we enter into the actual editor 
    - if we exit, we will reset to the state in step 2. ie, we still retain the empty state of the inital added component, which is wrong. 
2. there was an additional bug where we set an inner `text` on the Prose node. this causes the comparison between adding component then immediately exiting and adding component, making some edits then deleting to empty state and exiting to be different. this is because tiptap internally removes the `content` of the paragraph node when it's empty. 
    - i tested that this works via saving when it's empty (backend `schema` accepts this)
3. to make this fix comprehensive, we need to keep this `addedBlockIndex` state in sync across all possible Editor -> Root drawer scenarios. this is done via: 
    - adding the reset to `null` when deleting the block (addedBlock no longer exists)
    - adding the reset to `null` when discarding changes
    - adding the reset to `null` when saving (it's no longer the last added block)

## Before & After Screenshots

**BEFORE**:
see ticket

**AFTER**:

https://github.com/user-attachments/assets/25e5e717-4fdd-4a92-9b51-c84e41cd8b7a

## Notes 
i think the saving/deleting/discarding functionality is shared across all the different types of the editor (complex/metadata/prose) and it's getting unwieldy to keep remembering to sync etc esp when we now have implicit guarantees in place.

### DIFF
--- a/apps/studio/src/components/PageEditor/ComponentSelector.tsx
+++ b/apps/studio/src/components/PageEditor/ComponentSelector.tsx
@@ -29,7 +29,6 @@ import {
 
 import { useEditorDrawerContext } from "~/contexts/EditorDrawerContext"
 import { type DrawerState } from "~/types/editorDrawer"
-import { trpc } from "~/utils/trpc"
 import { DEFAULT_BLOCKS } from "./constants"
 import { type SectionType } from "./types"
 
@@ -104,7 +103,7 @@ function ComponentSelector() {
     setDrawerState,
     setSavedPageState,
     setPreviewPageState,
-    setAddedBlock,
+    setAddedBlockIndex,
   } = useEditorDrawerContext()
 
   const onProceed = (sectionType: SectionType) => {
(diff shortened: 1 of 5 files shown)


apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -129,15 +128,8 @@ function ComponentSelector() {
     setSavedPageState(nextPageState)
     setDrawerState({ state: nextState })
     setCurrActiveIdx(nextPageState.content.length - 1)
+    setAddedBlockIndex(nextPageState.content.length - 1)
```
seaerchin, line 131: the bug only happens when we add a new component and exit. in here. we set the state so that we know (in the editor) that this index is a block that is newly added **but doesnt exist in db**

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -10,12 +10,6 @@ export const DEFAULT_BLOCKS: Record<
     content: [
       {
         type: "paragraph",
-        content: [
-          {
-            type: "text",
-            text: "",
-          },
-        ],
```
seaerchin, line 18: deletion required due to tiptap removing this property when the content of the paragraph is visually empty.

dcshzj (approved):

PR #558, opened 2024-08-30
fix: white bar on bottom

seaerchin:
## Problem
Previously, when scrolling to the bottom, there was a white when we attempted to scroll up. See below video for eg. This wasn't very good UI and led to confusion on users (it only appears when editing a component and doesn't happen at the `RootStateDrawer`)

Closes ISOM-1407
Closes ISOM-1408

## Solution
i _think_ this was caused by the element being sticky and outside of the drawer itself, so the combined scroll led to the white bar being materialised at bottom as the save changes button is `sticky` and hence, overflows.. not super sure, @karrui can help explain?

based off ^, letting the `sticky` element be in the actual drawer would allow the sticky element to scroll relative to the container, which prevents the overflow and hence, whitebar from showing.

### Before

https://github.com/user-attachments/assets/35d286ff-c2f7-45ae-acaa-59691d0497a5

### After
take note, we have quite a few drawers but htis only happens on editing of components so only `Native/Complex/Meta` is affected. 

https://github.com/user-attachments/assets/69df47f8-e714-4509-960d-e9f5bff953ed

https://github.com/user-attachments/assets/3432ce5f-9aa4-494f-9f71-21dfc95d7b2a

## Notes
~the behaviour actually changed - cc @sehyunidaaa, fyi! previously the toolbar always showed on screen with enough space but now it's at the bottom. this is most obvious for components where overflow is the norm (anything using prose like accordion, infocols)~

fixed, now it's correctly sticky at bottom. cause was `overflow: auto` on the containing element


seaerchin:
> Don't think this is the correct fix. In your fix, you are scrolling the whole viewport (including the right side), which will move the preview out of alignment. You might want to constrain the overflow to only the left sidebar.

i updated this to use `overflow: auto` on the bounding `GridItem` so that the sidebar scrolls separately from the preview!. however, the sticky bottom bar still has a few issues  - notably, it causes the text component to have an inner overflow as maxH of text component was newly constrained to `50vh` so it wouldn't take up the whole container and block the bottom chunk of text.

i think @sehyunidaaa is working on a new design to remove the bottom 2 buttons and shift them into the header and she's aware + ok iwth these changes.

sehyunidaaa:
noted this is only fixing the whitespace issue and separate scrolling is on a separate branch

lgtm

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -269,39 +263,39 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
           </HStack>
         </Box>
 
-        <Box px="2rem" py="1rem">
+        {/* NOTE: reserve at least 4.75rem for the bottom bar */}
```
harishv7, line 266: how did we get 4.75rem? feels like a magic number

harishv7 (approved):
lgtm, just 1 nit comment

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
seaerchin: this was gotten by observing the save button - it had 44height, and 24px top and bottom padding, adding to 92 px or 4.75rem

PR #560, opened 2024-09-03
fix: HTML generation from prose JSON content

dcshzj:
## Problem

There is an issue with how the HTML is generated by components when it comes to nested marks.

Closes ISOM-1532

## Solution

**Breaking Changes**

**Bug Fixes**:

- Change the way we generate the HTML from the JSON prose content so that we are able to combine multiple text nodes with the same link mark.

### DIFF
--- a/packages/components/src/utils/getTextAsHtml.ts
+++ b/packages/components/src/utils/getTextAsHtml.ts
@@ -1,3 +1,5 @@
+import isEqual from "lodash/isEqual"
+
 import type { HardBreakProps } from "~/interfaces"
 import type { Marks, TextProps } from "~/interfaces/native/Text"
 
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
     return ""
   }
 
-  return content
-    .map((node) => {
-      if (node.type === "hardBreak") {
-        return "<br />"
-      }
+  const output: string[] = []
+  let existingLinkMark: Marks | undefined = undefined
 
-      if (!node.marks) {
-        return node.text
-      }
+  // At every step, we will close off all marks except for links
+  // First encounter with a link, always open it first before other marks
+  // Close all other marks first before closing the link mark
+  content.forEach((node) => {
+    if (node.type === "hardBreak") {
+      output.push("<br />")
+      return
+    }
 
-      let output = node.text
-      node.marks.forEach((mark) => {
-        if (mark.type === "link") {
-          output = `<${MARK_DOM_MAPPING[mark.type]} href="${
-            mark.attrs.href
-          }">${output}</${MARK_DOM_MAPPING[mark.type]}>`
-          return
+    const newLinkMark = node.marks?.find((mark) => mark.type === "link")
+    const isLinkMarkNew =
+      !!newLinkMark && !isEqual(existingLinkMark, newLinkMark)
+
+    // Close off the existing link mark if it is different
+    if (isLinkMarkNew && !!existingLinkMark) {
+      output.push(`</${MARK_DOM_MAPPING.link}>`)
+      existingLinkMark = undefined
+    }
+
+    // If there are no marks, just push the text
+    if (!node.marks) {
+      output.push(node.text)
+      return
+    }
+
+    if (isLinkMarkNew) {
+      existingLinkMark = newLinkMark
+
+      // Sort such that the link mark is the first item
+      node.marks.sort((a, b) => {
+        if (a.type === "link") {
+          return -1
         }
 
-        output = `<${MARK_DOM_MAPPING[mark.type]}>${output}</${
-          MARK_DOM_MAPPING[mark.type]
-        }>`
+        return 1
       })
 
-      return output
-    })
-    .join("")
+      node.marks.forEach((mark) => {
+        if (mark.type === "link") {
+          output.push(
+            `<${MARK_DOM_MAPPING[mark.type]} target="${mark.attrs.target ?? "_self"}" href="${mark.attrs.href}">`,
+          )
+        } else {
+          output.push(`<${MARK_DOM_MAPPING[mark.type]}>`)
+        }
+      })
+    } else {
+      // Continue with the rest of the marks
+      node.marks
+        .filter((mark) => mark.type !== "link")
+        .forEach((mark) => {
+          output.push(`<${MARK_DOM_MAPPING[mark.type]}>`)
+        })
+    }
+
+    // Push the text
+    output.push(node.text)
+
+    // Close off all marks except for links
+    node.marks
+      .filter((mark) => mark.type !== "link")
+      .forEach((mark) => {
+        output.push(`</${MARK_DOM_MAPPING[mark.type]}>`)
+      })
+  })
+
+  // Close off the last link mark if it exists
+  if (existingLinkMark) {
+    output.push(`</${MARK_DOM_MAPPING.link}>`)
+  }
+
+  return output.join("")
 }
--- a/packages/components/src/utils/getTextAsHtml.ts
+++ b/packages/components/src/utils/getTextAsHtml.ts
@@ -23,25 +23,32 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
   }
 
   const output: string[] = []
-  let existingLinkMark: Marks | undefined = undefined
+  let previousNodeLinkMark: Marks | undefined = undefined
 
   // At every step, we will close off all marks except for links
   // First encounter with a link, always open it first before other marks
   // Close all other marks first before closing the link mark
   content.forEach((node) => {
     if (node.type === "hardBreak") {
+      // Close off the existing link mark if it exists
+      if (previousNodeLinkMark) {
+        output.push(`</${MARK_DOM_MAPPING.link}>`)
+        previousNodeLinkMark = undefined
+      }
+
       output.push("<br />")
       return
     }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+    }
 
-      if (!node.marks) {
-        return node.text
-      }
+    // If there are no marks, just push the text
+    if (!node.marks) {
+      output.push(node.text)
+      return
+    }
```
seaerchin, line 51: nit: for readability reasons, could we put this condition further up

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+
+    if (isLinkMarkNew) {
```
seaerchin, line 53: this overlaps with the earlier condition? is this actually intended? we did a push of the mark earlier and down the line, we will push again ah

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+  // At every step, we will close off all marks except for links
+  // First encounter with a link, always open it first before other marks
+  // Close all other marks first before closing the link mark
+  content.forEach((node) => {
+    if (node.type === "hardBreak") {
+      output.push("<br />")
+      return
+    }
+
+    const newLinkMark = node.marks?.find((mark) => mark.type === "link")
```
seaerchin, line 37: not entirely sure why having a `link` mark implies it's new here - should we just rename htis to `hasLinkMark`?

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
 
-      let output = node.text
       node.marks.forEach((mark) => {
         if (mark.type === "link") {
-          output = `<${MARK_DOM_MAPPING[mark.type]} href="${
-            mark.attrs.href
-          }">${output}</${MARK_DOM_MAPPING[mark.type]}>`
-          return
+          output.push(
+            `<${MARK_DOM_MAPPING[mark.type]} target="${mark.attrs.target ?? "_self"}" href="${mark.attrs.href}">`,
```
seaerchin, line 68: nit: we can just push the type directly (`MARK_DOM_MAPPING.link` rather than go through 1. more leel of indirection

seaerchin: no real actionable but the `??` operator doesn't handle the empty string case (in case we have such a case)

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
         }
+      })
+    } else {
+      // Continue with the rest of the marks
+      node.marks
+        .filter((mark) => mark.type !== "link")
+        .forEach((mark) => {
+          output.push(`<${MARK_DOM_MAPPING[mark.type]}>`)
+        })
+    }
```
seaerchin, line 81: this takes place inside a `forEach`? should we just be doing `output.push` here since the `mark.type !== link` is basically guaranteed

my confusion here stems from the fact that eg: if we have a mark array like `[bold, italics, underline]`, this would first go through the initial array, then push each non link item **at each iteration** such that the end output would be `[bold, italics, underline] * 3`

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+      })
 
-      let output = node.text
       node.marks.forEach((mark) => {
         if (mark.type === "link") {
-          output = `<${MARK_DOM_MAPPING[mark.type]} href="${
-            mark.attrs.href
-          }">${output}</${MARK_DOM_MAPPING[mark.type]}>`
-          return
+          output.push(
```
seaerchin, line 67: why do we not need to set teh `existingLinkMark` here? this might lead to the ending `if (existingLinkMark) push([...])` being unbalanced isn't it

seaerchin (commented):
not 100% sure on this ._.

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+            `<${MARK_DOM_MAPPING[mark.type]} target="${mark.attrs.target ?? "_self"}" href="${mark.attrs.href}">`,
+          )
+        } else {
+          output.push(`<${MARK_DOM_MAPPING[mark.type]}>`)
         }
+      })
+    } else {
+      // Continue with the rest of the marks
```
dcshzj, line 75: need to reverse sort so the closing is correct

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+  let existingLinkMark: Marks | undefined = undefined
+
+  // At every step, we will close off all marks except for links
+  // First encounter with a link, always open it first before other marks
+  // Close all other marks first before closing the link mark
+  content.forEach((node) => {
+    if (node.type === "hardBreak") {
+      output.push("<br />")
+      return
+    }
```
dcshzj, line 35: Also need to check if there is an existing link to close

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,79 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+
+    const newLinkMark = node.marks?.find((mark) => mark.type === "link")
+    const isLinkMarkNew =
+      !!newLinkMark && !isEqual(existingLinkMark, newLinkMark)
```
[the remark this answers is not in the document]
dcshzj: Clarified offline, strict equality is important here because as long as the values are different, the functionality will definitely be different alr (cos there is only `target` and `href`).

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Clarified offline, can't do that here because we need to make sure we close off any existing open link marks first before we can push new texts.

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Clarified offline, we are opening up a new link mark here hence the check. The earlier condition was to check for something else, which is whether the link mark in this iteration's node is different from the one that we are storing as a pointer.

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Updated in 4127bada6311bf562b2500609445d16bbce877dc. This is the current node's link mark if present.

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Updated in 4127bada6311bf562b2500609445d16bbce877dc.

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Clarified offline, this is because our approach is to always open and close all marks other than link (and close in the correct order which was fixed in 4127bada6311bf562b2500609445d16bbce877dc). Hence, whenever we iterate and go into a new text node in the next iteration, we are always starting "fresh" with the exception of the link node which we are tracking separately.

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Clarified offline, this is because we are opening up a new link mark here and the previous one has already been closed.

packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -20,31 +22,92 @@ export const getTextAsHtml = (content?: (HardBreakProps | TextProps)[]) => {
...
+        })
+    }
+
+    // Push the text
+    output.push(node.text)
+
+    // Close off all marks except for links in reverse order
+    const marksToClose = node.marks.filter((mark) => mark.type !== "link")
+    while (marksToClose.length) {
+      const mark = marksToClose.pop()
```
seaerchin, line 96: just fyi, this pops off `node.marks` which might potentially be problematic if we are relying on it downstream or what not. 

could we be abit more paranoid and do a full clone here?

seaerchin (approved):
lgtm, leave it to you whether to do a clone but i suspect if we do more funky stuff with text, we might as well just do it honestly.   

just to bolster our confidence here, can i check if we've tried the below cases (manually is fine): 
1. outer link tag with inner tags
2. outer tags iwth inner links
3. any combination of hard breaks
4. single tag

packages/components/src/utils/getTextAsHtml.ts
dcshzj: Hmm in this case a shallow clone was already performed in line 94, so the pop here would not have any effect on the original `node.marks` array.

PR #566, opened 2024-09-06
feat(page settings): save on blur

seaerchin:
## Problem
This PR adds save functionality **on blur** ot the settings page. the toast shows up on successful save to inform of success and on failures to display the error message.

## Solution
1. add an `onBlur` to the `form` 
2. set the mutation handler to be `onBlur`
3. we disallow the editing of url and title if the resource type is `RootPage`

## Screenshots and videos

https://github.com/user-attachments/assets/b4c5dc92-1f1a-45f9-b3eb-28c988ecfc2b

## Notes 
1. ~@sehyunidaaa the toast is abit disruptive because it lingers for abit too long - shud we shorten it or allow closing?~
    - clarified offline, only 1 toast at any 1 time, can close, 3s
2. ~@sehyunidaaa also the copy for the toast might require some work~
3. saving for `noIndex` and `meta` - where should they go?
    - clarified with @dcshzj, these should go into `Blob.page`
4. there's a bug where our page preview is drawing from the `Blob` rather than the actual column in `Resource`, will file an issue (ISOM-1536)


apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -287,4 +288,23 @@ export const pageRouter = router({
...
       return addedVersionResult
     }),
+
+  updatePageSettings: protectedProcedure.input(pageSettingsSchema).mutation(
+    // TODO: save noIndex and meta to db
+    async ({ input: { pageId, siteId, title, meta, permalink, noIndex } }) => {
+      return db
+        .updateTable("Resource")
+        .where("Resource.id", "=", String(pageId))
+        .where("Resource.siteId", "=", siteId)
```
seaerchin, line 298: need another condition on `Resource.type` to only be `CollectionPage` or `Page`

sehyunidaaa:
@seaerchin 

1. Yes let's make the toast dismissible
2. How's: `Saved page settings. Publish this page for your changes to go live.`

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -117,148 +122,169 @@ const PageEditingView = ({ page, permalink, siteId }: PageEditingViewProps) => {
   )
 }
 
-const pageSettingsSchema = z.object({
-  title: pageTitleSchema,
-  meta: z.string().optional(),
-  permalink: permalinkSchema,
-  noIndex: z.boolean().optional(),
-})
+const THREE_SECONDS = 3000
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:125]: nit: 
```suggestion
const THREE_SECONDS_IN_MS = 3000
```

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -117,148 +122,169 @@ const PageEditingView = ({ page, permalink, siteId }: PageEditingViewProps) => {
...
+  const utils = trpc.useUtils()
+
+  const updatePageSettingsMutation = trpc.page.updatePageSettings.useMutation({
+    onSuccess: async () => {
+      // TODO: we should use a specialised query for this rather than the general one that retrives the page and the blob
+      await utils.page.readPageAndBlob.invalidate()
+      await utils.page.readPage.invalidate()
+      if (!toast.isActive(SUCCESS_TOAST_ID)) {
+        toast({
+          id: "save-page-settings-success",
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:159]: should you use `SUCCESS_TOAST_ID` here too? Since you already have the constant

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -117,148 +122,169 @@ const PageEditingView = ({ page, permalink, siteId }: PageEditingViewProps) => {
...
+        description: error.message,
+        status: "error",
       })
-    }
-  }, [getFieldState, setValue, title])
+    },
+  })
+
+  const onSubmit = handleSubmit(async (values) => {
+    await updatePageSettingsMutation.mutateAsync({ pageId, siteId, ...values })
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:176]: [Prefer mutate over mutateAsync](https://www.notion.so/opengov/Prefer-mutate-over-mutateAsync-ac450f152af84ec880ab520ca1520ebd)

Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:176]: might also want to update the latest rhf state so this does not run if there are no changes. You may want to call `reset` with the values you get here. Then check `isDirty` form state before triggering the mutation, perhaps?

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:125]: i knew i was missing smth...

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:159]: ah oops, i missed this one! thanks for teh catch

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:176]: i did the `isDirty` check, you're right! not too sure on the `reset` part - the reason why is when we submit, we invalidate the cache and cause a refetch which trickles down to this component as `originalPermalink, originalTitle`, which is used as `defaultValues` 

is the `reset` still required here? i'm assuming rhf will use the new values of `originalPermalink` and `originalTitle`

(also made the change to use `mutate`, i don't recall why i used `mutateAsync` honestly)

Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx]: If the form correctly resets with the updated values then sure can skip reset haha

Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx]: ah dang, it doesn't quite get the values correctly. oddly enough, it's still using the old values for the `isDirty` comparison (not very sure why, do you know?) and as a result, i reset the form values now with the `mutate`'s `onSuccess` hook

Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx]: Because the default value of the form hasn't changed when you submit. See [docs](https://react-hook-form.com/docs/useformstate#:~:text=Description-,isDirty,-boolean). rhf's [`reset`](https://react-hook-form.com/docs/useform/reset) will update the default values to the new input (which is the updated values in this case)

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -117,148 +122,177 @@ const PageEditingView = ({ page, permalink, siteId }: PageEditingViewProps) => {
...
       })
+    },
+  })
+
+  const onSubmit = handleSubmit((values) => {
+    if (formState.isDirty) {
+      updatePageSettingsMutation.mutate(
+        { pageId, siteId, ...values },
+        {
+          onSuccess: () => reset(values),
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:181]: nit: should we return the updated page settings from the mutation instead of using the client-side values? Update `trpc.page.updatePageSettings` to return the `title` and `permalink` from the updated row? 

feels a little more correct, though both works.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -287,4 +288,24 @@ export const pageRouter = router({
       await startProjectById(ctx.logger, codeBuildId)
       return addedVersionResult
     }),
+
+  updatePageSettings: protectedProcedure.input(pageSettingsSchema).mutation(
```
karrui, line 292: nit: since we are already in the `page` router, maybe the procedure can be shorter to be less verbose:

```suggestion
  updateSettings: protectedProcedure.input(pageSettingsSchema).mutation(
```

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -287,4 +288,24 @@ export const pageRouter = router({
...
+        .where("Resource.siteId", "=", siteId)
+        .where("Resource.type", "in", ["Page", "CollectionPage"])
+        .set({ title, permalink })
+        .returning([
+          "Resource.id",
+          "Resource.type",
+          "Resource.title",
+          "Resource.permalink",
+          "Resource.draftBlobId",
+        ])
```
karrui, line 307: do we need to return so many things? from what I see you only really need the new title and permalink (and even then in your current implementation don't use it at all). might want to avoid overreturning

karrui (approved):

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx]: i'm abit confused here - after submit, wouldn't the (old) value be invalidated and thus trigger a refetch from db? then, this would cause new data to be sent down to this component and then, get set as the new `defaultValue`?

apps/studio/src/server/modules/page/page.router.ts
seaerchin: going to prefer being abit more explicit here .-.

seaerchin: wait nvm, i realised at the callsite it gets destructured to use `mutate` so this isn't directly relevant anyways

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (karrui) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx]: No, cause the value is cached in rhf. You need to remount the component to use the new prop in that case. You can imagine if the defaultValues are set in a useState in the start. New props won't update the default values since they've already been set in a useState.


```ts
// First render, propDefaultValues = 1
const [defaultValues, _setDefaultValues] = useState(propDefaultValues)
// defaultValues === 1)

// Second render, propDefaultValues = 2
// defaultValues already has a value, nothing will happen. useState will not rerun since the component has not remounted (lifecycle)
// defaultValues still === 1
```

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:181]: i'll do this in a follow up PR, don't want to mix this in here!

karrui (approved):

PR #567, opened 2024-09-06
fix(move): perform UI invalidations correctly

seaerchin:
## Problem
currently, the invalidation logic is wrong for moving items. this results in the ui not updating after we move an item from 1 folder to another. 

Closes ISOM-1483
Closes ISOM-1461

## Solution
1. retrieve parent id from backend and pass it into the table so that each initial click will have access to the parent id
2. update the type of the atom and set parent id on each click of destination so that the move item knows the (original) parent id of the move item 
3. invalidate both destination and origin folder queries so that the ui will appear correct. the origin folder is `moveItem.parentId` and the destination is given by `moveDest.resourceId`

### Invalidation for pages
https://github.com/user-attachments/assets/73d6d807-8844-4b11-b7dd-a3c96a39b1ce

### Invalidation for folders

https://github.com/user-attachments/assets/5073218a-f844-4bf7-b4b6-3f74d9f12533

### DIFF
--- a/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
@@ -59,6 +59,7 @@ const MoveResourceContent = withSuspense(
     const [resourceStack, setResourceStack] = useState<PendingMoveResource[]>(
       [],
     )
+    const moveDest = resourceStack[resourceStack.length - 1]
     const { siteId } = useQueryParse(sitePageSchema)
     const setMovedItem = useSetAtom(moveResourceAtom)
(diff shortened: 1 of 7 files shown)


apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -59,6 +59,7 @@ const MoveResourceContent = withSuspense(
     const [resourceStack, setResourceStack] = useState<PendingMoveResource[]>(
       [],
     )
+    const moveDest = resourceStack[resourceStack.length - 1]
```
harishv7, line 62: Why is the move destination always at the last idx of the resource stack?

harishv7 (approved):

harishv7 (approved):
approved with a minor qn

PR #568, opened 2024-09-06
fix: update move modal to retrieve only folders

seaerchin:
## Problem
Previously, the move modal showed the `RootPage` as being a valid move destination. (refer to ticket for video)

## Solution
1. update `getFolderChildrenOf` so that we don't return the `RootPage` from there 
2. this is **only used in move** so this doesn't cause compat issues iwth other parts of our codebase 

## Screenshots and videos

Uploading Screen Recording 2024-09-06 at 6.15.31 PM.mov…


apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
```diff
@@ -1,33 +1,23 @@
...
 
 type MoveItemProps = Pick<
   RouterOutput["resource"]["getFolderChildrenOf"]["items"][number],
-  "permalink" | "type"
+  "permalink"
 > & {
   onChangeResourceId: () => void
 }
 
 const SuspendableMoveItem = ({
```
dcshzj, line 17: suggestion: I duplicated the functionality of this into ResourceSelector and haven't got the time to merge the two, but I was thinking ideally they use the same one but in this MoveResourceModal, we disable selecting normal pages and only allow users to select folders.

Reason being I think it would be helpful for the user to know the pages that are also present in that particular folder. Might be a design decision so I can follow up with design if needed.

seaerchin: i don't recall if it was a design decision tbh, but lets just ask @sehyunidaaa now HAHAHA

sehyunidaaa: image

This is a good suggestion! I have a state for folders that you don't have access to, maybe can use a similar state for pages so that we can still show them but disable selection?

seaerchin: sounds good, lets agree on:
1. showing pages but they are disabled
2. not showing collections

just fyi, i fixed the caching but might need to recheck after adding pages in. are we ok with doing as a follow up PR?

seaerchin: [chin/isom-1542-move-modal-should-also-show-lla](https://linear.app/ogp/issue/ISOM-1542/move-modal-should-also-show-pages)

PR #575, opened 2024-09-06
fix(folder settings): restrict input fields to max length

seaerchin:
## Problem
previously, the save wasnt disabled when the state of the field was invalid. 

## Solution
1. restrict the 2 fields to their respective max length, so that they cannot exceed the given length
2. for permalink, we already transform every invalid input to `-` so we know that the field's value will always be valid
 
because of this, we dont need to disable save as the permalink and title will always be of a valid length and the permalink will consist only of alphnumeric characters and `-`

## Screenshots

https://github.com/user-attachments/assets/9631c306-571a-423b-8104-5e413fb30328

### DIFF
--- a/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
@@ -158,13 +158,9 @@ const SuspendableModalContent = ({
                   <Input
                     placeholder="This is a url for your folder"
                     {...field}
+                    maxLength={MAX_FOLDER_PERMALINK_LENGTH}
                     onChange={(e) => {
-                      onChange(
-                        generateResourceUrl(e.target.value).slice(
-                          0,
-                          MAX_FOLDER_PERMALINK_LENGTH,
-                        ),
-                      )
+                      onChange(generateResourceUrl(e.target.value))
                     }}
                   />
                 )}
--- a/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
@@ -1,4 +1,4 @@
-import { Suspense, useEffect } from "react"
+import { Suspense } from "react"
 import {
   Box,
   FormControl,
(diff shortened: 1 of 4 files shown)


apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
```diff
@@ -158,7 +159,12 @@ const SuspendableModalContent = ({
                     placeholder="This is a url for your folder"
                     {...field}
```
dcshzj, line 160: issue: should we also add the maxLength here?

seaerchin: actually i tried this before, but having the controller and the maxlength on the input doesn't quite work oddly enough, which is whyi had to slice

seaerchin: wtf i just retried and it works. idk what changd from last time vs now but ok ill do the max length and remove the slice

apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
```diff
@@ -158,13 +158,9 @@ const SuspendableModalContent = ({
...
+                    maxLength={MAX_FOLDER_PERMALINK_LENGTH}
                     onChange={(e) => {
-                      onChange(
-                        generateResourceUrl(e.target.value).slice(
-                          0,
-                          MAX_FOLDER_PERMALINK_LENGTH,
-                        ),
-                      )
+                      onChange(generateResourceUrl(e.target.value))
```
dcshzj, line 163: issue: wait no, we still need the slice (either here or in generateResourceUrl) cos the changes can be coming from outside of the user input in this field.

seaerchin: how can the change come from uotside user input ah?

dcshzj: My bad, got confused, so this is fine.

dcshzj (approved):

PR #577, opened 2024-09-06
feat: dynamically generate the sitemap for preview

dcshzj:
## Problem

The sitemap is not being dynamically generated right now, so our preview's siderails are incorrect.

## Solution

**Breaking Changes**

**Features**:

- Add the ability to generate a sparse sitemap that only includes resources that revolve around the current page that the user is editing.
- For other types of previews (e.g. when creating pages), we show a basic version for now but we can potentially expand to provide a sparse sitemap but with the resource itself being dynamically updated upon user inputs.
- The permalink passed to the preview is also updated to provide the full path from root, instead of only providing the slug of the page.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-09-06 at 20 22 14

**AFTER**:

Screenshot 2024-09-06 at 20 21 40

### DIFF
--- a/apps/studio/src/schemas/site.ts
+++ b/apps/studio/src/schemas/site.ts
@@ -4,6 +4,11 @@ export const getConfigSchema = z.object({
   id: z.number().min(1),
 })
 
+export const getLocalisedSitemapSchema = z.object({
+  siteId: z.number().min(1),
+  resourceId: z.number().min(1).optional(),
+})
+
 export const getNotificationSchema = z.object({
   siteId: z.number().min(1),
 })
--- a/apps/studio/src/server/modules/resource/resource.service.ts
+++ b/apps/studio/src/server/modules/resource/resource.service.ts
@@ -306,7 +306,7 @@ export const getLocalisedSitemap = async (
           if (resource.parentId === null) {
             return fb("Resource.parentId", "is", null)
           }
-          return fb("Resource.parentId", "=", String(resourceId))
+          return fb("Resource.parentId", "=", String(resource.parentId))
         })
         .select(defaultResourceSelect),
     )
--- a/apps/studio/src/server/modules/resource/resource.service.ts
+++ b/apps/studio/src/server/modules/resource/resource.service.ts
@@ -1,7 +1,9 @@
+import type { IsomerSitemap } from "@opengovsg/isomer-components"
 import type { SelectExpression } from "kysely"
 import { type DB } from "~prisma/generated/generatedTypes"
 
 import type { Resource, SafeKysely } from "../database"
+import { getSitemapTree } from "~/utils/sitemap"
 import { db } from "../database"
 import { type Page } from "./resource.types"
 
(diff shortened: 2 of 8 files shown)


apps/studio/src/schemas/site.ts
```diff
@@ -4,6 +4,11 @@ export const getConfigSchema = z.object({
   id: z.number().min(1),
 })
 
+export const getLocalisedSitemapSchema = z.object({
+  siteId: z.number().min(1),
+  resourceId: z.number().min(1).optional(),
```
seaerchin, line 9: i think we shuold actually put these as strings because we don't want to treat them as numbers and we don't actually do any numerical operations on them...

having said that, this isn't a big deal because this inconsistency is littered throughout our codebase so we should fix it in 1 go

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -231,3 +233,147 @@ export const moveResource = async (
...
+    const parent = await getById(db, {
+      resourceId: Number(currentResource.parentId),
+      siteId,
+    })
+      .select(defaultResourceSelect)
+      .executeTakeFirstOrThrow()
+
+    ancestors.push(parent)
+    currentResource = parent
+  }
```
seaerchin, line 278: just fyi, this is non locking so we are allowing the resource (as well as its parents) to change while this method is executing.

since this method is purely a `GET`, i actually dont see a problem but i thought this might be useful to point out

seaerchin: also, i think a better way of doing this might be using recursive ctes  (see [here](https://www.postgresql.org/docs/current/queries-with.html))

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -231,3 +233,147 @@ export const moveResource = async (
...
+      return acc
+    }, {})
+  }
+
+  // Step 6: Construct the localised sitemap object
+  const rootResource = await db
+    .selectFrom("Resource")
+    .where("Resource.siteId", "=", siteId)
+    .where("Resource.parentId", "is", null)
+    .where("Resource.permalink", "=", "")
```
seaerchin, line 335: just double checking - is this supposed to reference the `RootPage` resource? might be clearer if we use `Resource.type = RootPage`

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -231,3 +233,147 @@ export const moveResource = async (
...
+  while (currentResource.parentId !== null) {
+    const parent = await getById(db, {
+      resourceId: Number(currentResource.parentId),
+      siteId,
+    })
+      .select(defaultResourceSelect)
+      .executeTakeFirstOrThrow()
+
+    ancestors.push(parent)
+    currentResource = parent
```
seaerchin, line 372: same comment as earlier, might be useful to do a recursive cte here

apps/studio/src/utils/sitemap.ts
```diff
@@ -0,0 +1,101 @@
...
+  parentId: string | null,
+  path: string,
+): IsomerSitemap[] | undefined => {
+  if (resources.length === 0) {
+    return undefined
+  }
+
+  const children = resources.filter((resource) => {
+    if (parentId === null) {
+      return resource.parentId === null && resource.permalink !== ""
```
seaerchin, line 23: shud just use the type to compare cos that's the authoritative source!

apps/studio/src/utils/sitemap.ts
```diff
@@ -0,0 +1,101 @@
...
+        !(
+          child.type === "Page" &&
+          children.some(
+            (otherChild) =>
+              otherChild.id !== child.id &&
+              otherChild.permalink === child.permalink &&
+              otherChild.type === "Folder",
+          )
+        ),
+    )
```
seaerchin, line 43: can help me understand what this filter is trying to accomplish? at first glance, this is trying to filter out pages that have the same permalink as another folder? but not entirely sure why this filter is required here

apps/studio/src/utils/sitemap.ts
```diff
@@ -0,0 +1,101 @@
...
+    .map((resource) => {
+      const permalink = `${path}${resource.permalink}`
+      if (resource.type === "Folder") {
+        const idOfPage = children.find(
+          (child) =>
+            child.id !== resource.id &&
+            child.permalink === resource.permalink &&
+            child.type === "Page",
+        )?.id
```
seaerchin, line 52: abit confused here, why is the method used `.find`, which suggests that only the first page matters? also, why not just use `child.parentId === resource.id`?

seaerchin: not entirely sure here, but reading at first glance seems to suggest we're trying to find another page iwth the same permalink as the current folder but the earlier filter seems to omit pages that have some folder iwth the same permalink, so not entirely sure why this works

apps/studio/src/utils/sitemap.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        const folderChildren = resources.filter(
+          (item) => item.parentId === resource.id,
+        )
```
seaerchin, line 56: do we not need another conditional here like `resource.type  === Folder`

apps/studio/src/utils/sitemap.ts
```diff
@@ -0,0 +1,101 @@
...
+
+type ResourceDto = Omit<
+  Resource,
+  "id" | "parentId" | "publishedVersionId" | "draftBlobId"
+> & {
+  id: string
+  parentId: string | null
+}
+
+const getSitemapTreeFromArray = (
```
seaerchin, line 12: not entirely sure if this is a bug, but you appear ot have missed out collections and collection pages totally, so can i trouble you to test this behaviour against those 2 items?

seaerchin: also double checking here, it feels like we're retrieving the items from db in a set order (ie, at point of retrieval, the ordering is clear, which must be true because our query is centered around `parentId`) and over here, we're trying to reconstruct the sitemap again?

would it be better to instead just retain the pointers from earlier rather than doing a flat spread `[...ancestors, ...siblings, ...children] or am i misunderstanding something about the sitemap generation?

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Good point, originally wanted to do it step-by-step but doing CTEs can also the steps understandable also. Swapped in bd051a9c8f9d1a82abde4ceb4c108498fb00a10a.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Yup, this is referencing the root page. Updated in bd051a9c8f9d1a82abde4ceb4c108498fb00a10a.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Changed in bd051a9c8f9d1a82abde4ceb4c108498fb00a10a.

apps/studio/src/utils/sitemap.ts
dcshzj: Hmm not exactly, we don't necessarily need them to be given in a particular order for this function to work. Basically, because we are given the array, we want to convert it into a tree object and use the parentId to identify the parent-child relationships.

With the changes in bd051a9c8f9d1a82abde4ceb4c108498fb00a10a, they are now all dumped into a single array.

apps/studio/src/utils/sitemap.ts
dcshzj: Updated in 74f51cb54a95ded6916da5469ee9dfd9e27e55ff.

apps/studio/src/utils/sitemap.ts
dcshzj: Basically right now we always have duplicate resources for the same permalink when it is a folder. However, we don't have such a representation in the sitemap because the SOT is the permalink. So this filter is trying to remove Pages and CollectionPages that have a corresponding Folder or Collection, then re-assigning the resource ID of the Folder or Collection to the ID that was removed.

The reason we do this is because in the actual sitemap, we only are interested in IDs of actual pages (folders is not a concept that exist in the final site). However, our parent-child relationships uses folders (and collections), so we need to perform this re-assignment to ensure that everything in the generated sitemap are actual pages, and the parent-child relationships are still preserved.

apps/studio/src/utils/sitemap.ts
dcshzj: Yup, because there should only be 1 result. The .find is done on the original array, so it still contains the entries with the actual Page.

apps/studio/src/utils/sitemap.ts
dcshzj: Hmm no, we are looking for all resources with that particular parentId, whereby the parentId is a resource that is a Folder. The subsequent recursion does the filtering between the Page and Folder.

seaerchin (approved):
i'm approving this first with caveats:

1. i don't quite understand how the site map generation is done, and the tree -> array -> tree steps really threw me for a loop
2. not too sure if the logic here is correct due to #1 but i tried and it seems to work.

PR #579, opened 2024-09-09
fix: make sure that we get resources only from single site

seaerchin:
## Problem
See ticket for a description and video fo the problem

## Solution
1. add a `where` clause on the query for the `siteId`.

### DIFF
--- a/apps/studio/src/server/modules/resource/resource.router.ts
+++ b/apps/studio/src/server/modules/resource/resource.router.ts
@@ -33,11 +33,12 @@ export const resourceRouter = router({
     }),
   getFolderChildrenOf: protectedProcedure
     .input(getChildrenSchema)
-    .query(async ({ input: { resourceId, cursor: offset, limit } }) => {
+    .query(async ({ input: { siteId, resourceId, cursor: offset, limit } }) => {
       let query = db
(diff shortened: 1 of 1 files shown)


apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -33,11 +33,12 @@ export const resourceRouter = router({
     }),
   getFolderChildrenOf: protectedProcedure
     .input(getChildrenSchema)
-    .query(async ({ input: { resourceId, cursor: offset, limit } }) => {
+    .query(async ({ input: { siteId, resourceId, cursor: offset, limit } }) => {
```
karrui, line 36: do we need to update the callsites? or we have always been passing it in but not using it lol!

karrui (approved):
lgtm

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: passing in but ont used HAHAHAH

seaerchin: image

in `MoveResourceModal`, only place where `getFolderChildrenOf` is used

PR #580, opened 2024-09-09
fix: filter for page and folder

seaerchin:
## Problem
Right now, we can move collections from the dashboard. this should not be allowed as only pages and folders can be moved.

## Solution
1. Filter the `resource.type` on the frontend
2. add a `where` clause on the backend so we only allow move on `Page/Folder`

## Screenshots

https://github.com/user-attachments/assets/22515214-e3e3-4adf-a7d0-c676b0d0b8e0

### DIFF
--- a/apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
+++ b/apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
@@ -69,13 +69,15 @@ export const ResourceTableMenu = ({
               Edit folder settings
             </MenuItem>
           )}
-          <MenuItem
-            as="button"
-            onClick={handleMoveResourceClick}
-            icon={<BiFolderOpen fontSize="1rem" />}
-          >
-            Move to...
-          </MenuItem>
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -69,13 +69,15 @@ export const ResourceTableMenu = ({
...
             </MenuItem>
           )}
-          <MenuItem
-            as="button"
-            onClick={handleMoveResourceClick}
-            icon={<BiFolderOpen fontSize="1rem" />}
-          >
-            Move to...
-          </MenuItem>
+          {(type === "Page" || type === "Folder") && (
```
harishv7, line 72: just to check we do have different types right? Like CollectionPage - are they movable?

seaerchin: No, they are not. The others are Collection, CollectionPage and Rootpage for completeness

PR #581, opened 2024-09-09
feat: use a fake link when rendering page preview

dcshzj:
## Problem

Right now, the links in our preview are legit links, which means the user can potentially open a "mini studio" or go to a 404 page when clicking on them.

## Solution

**Breaking Changes**

**Features**:

- Introduce a fake link component that prevents the default from occurring (which is the actual navigating away).

**Note**:

- The links created in Tiptap have not yet been ported over to use the LinkComponent, cos we are directly generating the HTML. This will be tackled in a future PR.


seaerchin (approved):

seaerchin:
pls fix format before merge

PR #584, opened 2024-09-09
[Template] Edit copy for 404 page

sehyunidaaa:
Edit copy to be more suitable for public-facing sites:

image

### DIFF
--- a/packages/components/src/templates/next/layouts/NotFound/NotFound.tsx
+++ b/packages/components/src/templates/next/layouts/NotFound/NotFound.tsx
@@ -26,8 +26,8 @@ const NotFoundLayout = ({
           site,
           component: {
             type: "infobar",
-            title: "Oh no, we couldn't find that page",
-            description: "404: Page not found",
+            title: "404: Page not found",
+            description: "Sorry, the page you were looking for cannot be found",
             buttonLabel: "Go to homepage",
             buttonUrl: "/",
           },


packages/components/src/templates/next/layouts/NotFound/NotFound.tsx
```diff
@@ -26,8 +26,8 @@ const NotFoundLayout = ({
           site,
           component: {
             type: "infobar",
-            title: "Oh no, we couldn't find that page",
-            description: "404: Page not found",
+            title: "404: Page not found",
+            description: "Sorry, the page you were looking for cannot be found",
```
karrui, line 30: I don't think you need to say "Sorry" that the page does not exist haha. I think better to not be too apologetic in this case.

@shanty-ogp chime in!!!

I read many resources before that say Firm but polite good enough, don't need say sorry.
- [further reading](https://medium.com/honestbee-design/why-not-to-say-sorry-in-error-and-info-messages-11151047f01a)
- [further reading again](https://medium.com/carmax-ux/why-ux-writers-have-to-learn-to-be-less-polite-de2cfb08e912)
- [related-ish reading](https://uxmag.com/articles/are-you-saying-no-when-you-could-be-saying-yes-in-your-web-forms)
-

karrui:
Screenshot 2024-09-09 at 5 34 47 PM

but then amazon does it, but that's following their key value that customer is king haha. maybe not our key value?

packages/components/src/templates/next/layouts/NotFound/NotFound.tsx
sehyunidaaa: Good question! I usually try to refrain from using Sorry, but I thought a 404 page could be a bit more apologetic since the cause might actually be on the site admins' end (broken links etc).

Let me #ask-writers!

PR #585, opened 2024-09-09
fix(hero banner): prevent deletion of hero banner

seaerchin:
## Problem
Currently, our editor allows us the option to delete the `HeroBanner`, which should not be allowed as it is a fixed block for `Homepage`. 

## Solution
1. create another `HeroEditorDrawer` that allows editig only for `Hero` and draws its data from `index: 0`
2. duplicate stuff from `ComplexEditor` but remove the `delete` button

- didn't want to stuff this together into either `ComplexEditor` or `MetadataEditor` cos it's subtly different. `ComplexEditor` allows deletion, and i didn't want to have another `if/else`; `MetadataEditor` sets a fixed layout + writes to diff location


apps/studio/src/features/editing-experience/components/HeroEditorDrawer.tsx
seaerchin, line 1: most code here is copy pasted from `MetadataEditor`!

apps/studio/src/features/editing-experience/components/HeroEditorDrawer.tsx
```diff
@@ -0,0 +1,160 @@
...
+              <Icon
+                as={BiDollar}
+                fontSize="1.5rem"
+                p="0.25rem"
+                bgColor="slate.100"
+                textColor="blue.600"
+                borderRadius="base"
+              />
+              <Heading as="h3" size="sm" textStyle="h5" fontWeight="semibold">
+                Edit page title and summary
```
seaerchin, line 99: @sehyunidaaa any suggestions for wording otherwise i put `Edit hero banner` lel

apps/studio/src/features/editing-experience/components/HeroEditorDrawer.tsx
```diff
@@ -0,0 +1,160 @@
...
+  Heading,
+  HStack,
+  Icon,
+  IconButton,
+  useDisclosure,
+} from "@chakra-ui/react"
+import { Button } from "@opengovsg/design-system-react"
+import { getComponentSchema } from "@opengovsg/isomer-components"
+import Ajv from "ajv"
+import _ from "lodash"
```
karrui, line 14: nit: destructure import instead

karrui (approved):

apps/studio/src/features/editing-experience/components/HeroEditorDrawer.tsx
seaerchin: donezxc

apps/studio/src/features/editing-experience/components/HeroEditorDrawer.tsx
seaerchin: updated, synced offline

PR #587, opened 2024-09-09
fix(cascades): update schema so that deletes are cascaded

seaerchin:
**THIS IS A ONE LINE PR!!! autogen code cause the extra lines**
## Problem
Previously, our deletes are not cascaded - this means that when we delete a folder, for example, the child pages and folders are not deleted. 

## Solution
1. add a constraint to the table's relation so that it does a `CASCADE` on delete

## Screenshots
(note, after running migration)

https://github.com/user-attachments/assets/2e30b61e-4719-44e9-bb12-d20b8f3fc916

## Notes
i **did not** delete the blob as we want to archive it presumably and also for auditing purposes

### DIFF
--- a/apps/studio/prisma/generated/generatedEnums.ts
+++ b/apps/studio/prisma/generated/generatedEnums.ts
@@ -1,19 +1,19 @@
 export const ResourceState = {
-  Draft: "Draft",
-  Published: "Published",
-} as const
-export type ResourceState = (typeof ResourceState)[keyof typeof ResourceState]
+    Draft: "Draft",
+    Published: "Published"
+} as const;
+export type ResourceState = (typeof ResourceState)[keyof typeof ResourceState];
 export const ResourceType = {
-  RootPage: "RootPage",
-  Page: "Page",
-  Folder: "Folder",
-  Collection: "Collection",
-  CollectionPage: "CollectionPage",
-} as const
-export type ResourceType = (typeof ResourceType)[keyof typeof ResourceType]
+    RootPage: "RootPage",
+    Page: "Page",
+    Folder: "Folder",
+    Collection: "Collection",
+    CollectionPage: "CollectionPage"
+} as const;
+export type ResourceType = (typeof ResourceType)[keyof typeof ResourceType];
 export const RoleType = {
-  Admin: "Admin",
-  Editor: "Editor",
-  Publisher: "Publisher",
-} as const
-export type RoleType = (typeof RoleType)[keyof typeof RoleType]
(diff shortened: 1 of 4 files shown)


apps/studio/prisma/generated/generatedEnums.ts
seaerchin, line 1: autogen

apps/studio/prisma/generated/generatedTypes.ts
seaerchin, line 1: autogen

apps/studio/prisma/migrations/20240909103759_cascade_deletions_for_resource/migration.sql
seaerchin, line 1: auto gen

apps/studio/prisma/schema.prisma
```diff
@@ -52,7 +52,7 @@ model Resource {
   siteId    Int
   site      Site      @relation(fields: [siteId], references: [id])
   parentId  BigInt?
-  parent    Resource? @relation("ParentRelation", fields: [parentId], references: [id])
+  parent    Resource? @relation("ParentRelation", fields: [parentId], references: [id], onDelete: Cascade)
```
harishv7, line 55: Are we doing hard delete or soft? Cascade will delete all right?

seaerchin: We've never done soft delete and I'm not entirely sure on this because the extension that I found was using prisma client (which we're not, we're using kysely).

Because of this, I didn't delete the blob for auditing purposes and just kept it as is. 

Soft deletes, at least afaik, are done by the client library by inspecting a special column (`deletedAt` in classic) and hence, save devs the trouble of doing the querying etc themselves. 

We can do it here, but this would involve manual handling I think, and I think I'd rather keep this PR scoped to fixing this first rather than introducing another change!

harishv7: Am fine with this for now, but i do have concerns on hard deletes as we won't be able to surface full audit logs when required. We might want to file a ticket for exploring soft deletes

And yes i think manual handling would be required in this case but I think should be fine - just need to implement the deletedAt column and add a enum state for Resource to be in DELETED state

harishv7 (approved):
lgtm with comments for future tickets

apps/studio/prisma/schema.prisma
seaerchin: filed here: https://linear.app/ogp/issue/ISOM-1541/implement-soft-deletes-for-studio

PR #590, opened 2024-09-10
refactor: persist saves for page title and meta to backend

seaerchin:
-  setting as draft first because our typing for the db types is wrong. oddly enough, it's not picking up the `noIndex` key
- also need to update the data from the return value of `mutation` rather than user input but this is relatively minor since saving is a sync point when successful

## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


seaerchin:
clsoing this as we will instead rework the UI to not show this on the page settings tab

PR #591, opened 2024-09-10
fix: ensure HTML is rendered on server side

dcshzj:
## Problem

The Interweave library was used incorrectly due to missing DOM when building the site.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Add a polyfill as [recommended in their docs](https://interweave.dev/docs/ssr).
- Fixed the tag name to be lower-case instead of upper-case.
- Also added the placeholder image in case the assetsBaseUrl is not available.
- Removed all instances of `noreferrer`. We are no longer preventing the referrer header from being forwarded to third-parties.


seaerchin (approved):
lgtm! up to you whether you want to do but idt it'll add value: should we extract out `noreferrer nofollow` to a const so we can change it easily next time?

but caveat is that this is highly unlikely to change

PR #593, opened 2024-09-11
fix: sticky bottom overflows in edit resource drawers

karrui:
### TL;DR

Improved layout and scrolling behavior of editor drawers.

### What changed?

- Updated the layout of ComplexEditorStateDrawer, HeroEditorDrawer, MetadataEditorStateDrawer, and TipTapProseComponent
- Replaced fixed height with flex layout and added overflow scrolling to content areas
- Removed sticky positioning from bottom action bars

### How to test?

1. Open various editor drawers (Complex, Hero, Metadata, and TipTap Prose)
2. Verify that the content area scrolls independently when content exceeds the drawer height
3. Ensure the bottom action bar remains visible and doesn't overlap content
4. Check that all functionality (editing, saving, deleting) works as expected

### DIFF
--- a/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
@@ -286,8 +286,7 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
           </HStack>
         </Box>
 
-        {/* NOTE: reserve at least 4.75rem for the bottom bar */}
-        <Box h="full" px="2rem" py="1rem" mb="4.75rem">
+        <Box flex={1} overflow="auto" px="2rem" py="1rem">
           <FormBuilder<IsomerComponent>
             schema={subSchema}
             validateFn={validateFn}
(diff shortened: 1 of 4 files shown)


karrui:
* **#593** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/593?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

seaerchin (approved):
lgtm. i checked via going into each component and checking for the overflows on long text. (accordion, text etc) 

just for learning, the way this works is via setting `overflow: auto` so that the wrapper around either `TiptapProse` or `FormBuilder` is the one that decides whether to overflow right? 

the `flex = 1` is set here so that it gets `grow: 1, shrink:1, basis: 0`? from the mdn docs, this then means that the component will grow/shrink baesd on extra space - since there's not enough, this will overflow and cause `overflow: auto` to kick in

karrui:
> just for learning, the way this works is via setting `overflow: auto` so that the wrapper around either `TiptapProse` or `FormBuilder` is the one that decides whether to overflow right?

The layout is currently
```
<parent flex height="100">
  <header />
  <scrollable-content flex={1} overflow="auto" />
  <action-bottom-bar />
</parent>
```

since the parent's height is fixed at 100% (note that it is not min height, nor max height), the children, if overflow is prevented, will at most be the parent's height.
since `scrollable-content` is sandwiched between a non-shrinking/growing `header` and `action-bottom-bar`, given flex 1, it can only fill up the remaining space as constrained by the parent. When it exceeds the space allowed, `overflow: auto` forces the scroll. It still only takes up the height allowed by the sandwich.

[further reading](https://www.joshwcomeau.com/css/interactive-guide-to-flexbox/) (even though it doesn't really cover the above scenario, but might explain to you clearly what flex does)

PR #601, opened 2024-09-11
fix(validation): add validation to the various blocks on editor

seaerchin:
## Problem
Currently, studio does not do validation of any input whatsoever. this allows users to update their fields with content that is more than our max length but because there is no validation performed, this is silently accepted.

## Solution
1. for complex components, the validation logic should come from the schema, so the various max lengths have been added to the schema side of things. these max lengths were obtained from the sheet [here](https://docs.google.com/spreadsheets/d/1VbFp3XPlEVXrDBqpptD3yCKB45BmGTFRr-JC76lOffs/edit?gid=0#gid=0)
2. for native components, the validation should come from tiptap and should be identical to our schema. there is only one such case, which is the `caption`, and i lowered teh existing one to 200 (same as the spreadsheet). sadly, i had to double code because the table schema itself is not accessible by the schema which we export.

## Screenshots

https://github.com/user-attachments/assets/01bd7ee6-d683-47d8-ad03-3ce61f740990


packages/components/src/interfaces/complex/Infobar.ts
```diff
@@ -8,23 +8,27 @@ export const InfobarSchema = Type.Object(
     type: Type.Literal("infobar", { default: "infobar" }),
     title: Type.String({
       title: "Title",
+      maxLength: 100,
```
harishv7, line 11: is there commonality between the various max lengths? e.g. all titles are max len of 100? If so,  I would prefer to extract these into a constants file

packages/components/opengovsg-isomer-components-0.0.13.tgz
harishv7, line 1: Do we still need to commit this binary file? I think for amplify builds it is committed to a separate repo

packages/components/src/interfaces/complex/Infobar.ts
seaerchin: I'll be going through them with @sehyunidaaa later but at present, alot of these weren't present on the excel and i had to make a judgement call, which is why they're left as-is. 

let me finalize them with sehyun and after that, we can extract them out!

packages/components/opengovsg-isomer-components-0.0.13.tgz
seaerchin: idk why i committed this, might have been an accident. lemme remove them

harishv7 (approved):
lgtm apart from existing comments

PR #605, opened 2024-09-11
chore(table): allow tables for accordion and prose

seaerchin:
## Problem
See ticket for a brief overview of the proble

## Solution
1. refactor the various editor hooks so that they call a common base hook that sets baseline functionality that's shared across the various child hooks
2. for accordion and prose, we allow `Table` and their corresponding children. **however**, care needs to be taken here as our schema is quite strict and we want to make sure tiptap's output is a subset of our schema. 
    - previously, allowing both tables and headings was problematic because our schema disallowed headings in table. this caused a run time crash. in order to prevent this problem, we set the tiptap schema for table cells to allow only text and list content and for table headers to allow only text content **no lists**)
    - for completeness, the schemas for isomer components is also shown
 
### table header cell
```ts
 const TableHeaderCellSchema = Type.Object({
  type: Type.Literal("tableHeader", { default: "tableHeader" }),
  attrs: Type.Optional(TableBaseCellSchema),
  content: Type.Array(Type.Ref(ParagraphSchema), {
    title: "Table header cell contents",
    description: "The contents of the table header cell",
    minItems: 1,
  }),
})
 
```

### table cell
```ts
const TableCellSchema = Type.Unsafe<TableCellProps>(
  Type.Object({
    type: Type.Literal("tableCell", { default: "tableCell" }),
    attrs: Type.Optional(TableBaseCellSchema),
    content: Type.Array(
      Type.Union([
        Type.Ref(DividerSchema),
        Type.Ref(ParagraphSchema),
        Type.Ref(OrderedListSchema),
        Type.Ref(UnorderedListSchema),
      ]),
    ),
  }),
)

export const ParagraphSchema = Type.Object(
  {
    content: Type.Optional(
      Type.Array(Type.Union([HardBreakSchema, TextSchema]))
  },
)

export const TextSchema = Type.Object(
  {
    type: Type.Literal("text", { default: "text" }),
    marks: Type.Optional(
      Type.Array(
        Type.Union([
          BoldMarkSchema,
          CodeMarkSchema,
          ItalicMarkSchema,
          LinkMarkSchema,
          StrikeMarkSchema,
          SubscriptMarkSchema,
          SuperscriptMarkSchema,
          UnderlineMarkSchema,
        ]),
      ),
    ),
    text: Type.String(),
  },
  {
    title: "Text content",
  },
)
```
given the above, we see that we only need to support text content, lists and horizontal rules. since tiptap's schema handles unknown nodes by deletion and is safe at runtime, supporting a stricter subset than our schema is ok. 

this does mean that eg: pasting in horizontal rules into the tables **would not work** but since we don't expose a button to click for user, i'm ok with this trade off. concretely, what happens is that the table would get split into 2 blocks (see attached video)

also, since there's [no exposed keybind ](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-horizontal-rule/src/horizontal-rule.ts)for horizontal rule, this is a relatively 'hidden' feature that we can change behaviour of down the line pretty safely.

## Notes
1. pasting lists or selecting a list **will not work** when the currently active cell is a table header. should we just disable the button or remove when in table header? this change was intentional because our schema disallows it and it's also bad ux
2. we can crash by first creating a table then adding a hard break. this splits into 2 tables. for the bottom table, add a list (allowed) then toggle to a header. this crashes because the bottom table is now just table cells, but toggling to a table header now disallows the previously added list.

https://github.com/user-attachments/assets/c3f2f51f-a829-4565-9b1f-b075d2784ee3

## Screenshots

https://github.com/user-attachments/assets/aef297b5-fc51-4213-8f38-7ef287b98054

 
### horizontal rule interaction with table

https://github.com/user-attachments/assets/d70c6048-310f-41f7-80a3-25fc6ac7312c


seaerchin:
Note to reviewers, I'll take out the hard break component for prose and accordion so this bug won't occur

PR #607, opened 2024-09-12
feat: introduce add/edit link functionality in Tiptap

dcshzj:
## Problem

We can't add links in Tiptap right now.

## Solution

**Breaking Changes**

**Features**:

- Add the ability to add and edit links in Tiptap editors.

## Screenshots

Screenshot 2024-09-12 at 11 47 42

image


apps/studio/src/components/PageEditor/MenuBar/AccordionMenuBar.tsx
```diff
@@ -82,9 +92,49 @@ export const AccordionMenuBar = ({ editor }: { editor: Editor }) => {
...
+            .extendMarkRange("link")
+            .deleteSelection()
+            .insertContent(`<a href="${linkHref}">${linkText}</a>`)
+            .run()
+
+          onLinkModalClose()
+        }}
+        onClose={onLinkModalClose}
+        isOpen={isLinkModalOpen}
+      />
```
karrui, line 135: seems like repeated code, might miss one when we change the other next time. Should we extract it out to a reusable hook instead?

```ts
const useLinkEditorModal = (editor) => {
  // setup
  // ...

  return {
    linkText,
    linkHref,
    onSave,
  }
}
```

dcshzj: Switch to the hook method but returns `onLinkModalOpen` and `LinkEditorModal` instead for consumption.

seaerchin:
i merged a PR addinga new base `prose` editor ,could i trouble you to also add this there?

apps/studio/src/features/editing-experience/hooks/useLinkEditorModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+          editor.isActive("link") ? editor.getAttributes("link").href : ""
+        }
+        onSave={(linkText, linkHref) => {
+          editor
+            .chain()
+            .focus()
+            .extendMarkRange("link")
+            .deleteSelection()
+            .insertContent(`<a href="${linkHref}">${linkText}</a>`)
+            .run()
```
seaerchin, line 38: how does this work with text content with 2 marks? a valid use case would be, for example, to make a link bold!

i think this might delete all other marks in this case. maybe a more appropriate command might be `toggleNode`? (see [here](https://tiptap.dev/docs/editor/api/commands/nodes-and-marks/toggle-node))

apps/studio/src/features/editing-experience/hooks/useLinkEditorModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return {
+    onLinkModalOpen,
+    LinkEditorModal: (
+      <LinkEditorModal
+        linkText={
+          editor.isActive("link")
+            ? editor.state.doc.nodeAt(
+                Math.max(1, editor.view.state.selection.from - 1),
+              )?.textContent
+            : ""
```
seaerchin, line 26: i haven't actually taken a look at this, but why would `from` not be sufficient ah? also, if no selection (single I cursor, haven't select anything yet), how would this work?

apps/studio/src/features/editing-experience/hooks/useLinkEditorModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const useLinkEditorModal = ({ editor }: UseLinkEditorModalProps) => {
+  const {
+    isOpen: isLinkModalOpen,
+    onOpen: onLinkModalOpen,
+    onClose: onLinkModalClose,
+  } = useDisclosure()
+
+  return {
+    onLinkModalOpen,
+    LinkEditorModal: (
```
seaerchin, line 19: i don't think this is a common pattern (to return the component) - why this is so is because you're forcing a UI onto your end user. 

i'd rather we define a contract like `LinkEditorModalProps` and this returns a method like `getLinkEditorModalProps` that satisfies the contract. this is done for other libraries like chakra (`getInputProps` for `useRadio`) and rhf (`...register`)

seaerchin (commented):
first glance review, never tried on local. i think the hook might be better off **not** imposing a UI constraint onto the end user, wdyt? 

other minor issue is what happens with toggling links when there are other nodes - not too worried here because the user can rectify and it's easy but if the fix on our end is easy too, i'd ask that you fix it here please!

apps/studio/src/features/editing-experience/hooks/useLinkEditorModal.tsx
dcshzj: Hmm this one I'm not very sure. I went with this option because this seems to be the only way to change the text of the original link with the value of the new link text provided by the user.

Alternative is to build an extension that can handle this for us using ProseMirror's APIs directly, I can maybe follow this up on a separate PR?

apps/studio/src/features/editing-experience/hooks/useLinkEditorModal.tsx
dcshzj: The `from` takes the position just after the actual `from` (implicit +1), so subtracting 1 here to actually get the cursor's position.

apps/studio/src/features/editing-experience/hooks/useLinkEditorModal.tsx
dcshzj: Oh yeah hor I think I went a bit crazy here, simplified the original component so it takes in the simpler props similar to the table settings modal.

seaerchin (approved):
lgtm, i think we should still file a ticket for the issue of nested marks. i think this is a net win though so i'm ok with this going in

PR #609, opened 2024-09-12
feat: fetch last modified date from DB for preview

dcshzj:
## Problem

The last modified date for the preview always shows the current date, which can be confusing for users if they have not edited the page.

## Solution

**Breaking Changes**

**Features**:

- Added created at and updated at columns for most of the tables in the database.
- Fetch the blob's updated at column when retrieving the page's contents, then pass this last modified date to the preview.

**Notes**:

The updatedAt column is not automatically updated by Kysely, so there were 3 ways of going about it: 
- [Write a plugin](https://www.answeroverflow.com/m/1135921621564149870#solution-1136285885063770202) that will automatically update the column during the query
- Have [a Postgres trigger](https://www.answeroverflow.com/m/1135921621564149870#solution-1136285885063770202) to do that automatically
- Add an additional layer above Kysely but below the actual query building

First option is not recommended (shown in the link). Second option was what I went for, and the third option is quite painful to do. Downside is that we have to remember to add this trigger manually whenever we create new tables.

## Screenshots
image


apps/studio/prisma/schema.prisma
```diff
@@ -66,13 +68,20 @@ model Resource {
 
   state ResourceState? @default(Draft)
   type  ResourceType
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
```
harishv7, line 73: the @updatedAt tag is to auto inject this on Prisma ORM level right? but since we are using kysely how does this help?

harishv7 (approved):

apps/studio/prisma/schema.prisma
dcshzj: Yeah there is no functional difference here since we are not using the prisma client and we are only using prisma for the modelling, hence the need to introduce the triggers on the DB level unfortunately.

PR #612, opened 2024-09-12
fix: directly jump to link parent folder in selector

dcshzj:
## Problem

When selecting the link in the link selector, it will always reset to the root folder.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Retrieve the ancestry information about the selected resource, so that we can construct the parentId stack on load.
- Add feedback to the user about the actual page permalink that is being selected.
- Add a maximum height and an overflow for the resource selector.

## Screenshots

image

### DIFF
--- a/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
+++ b/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
@@ -25,10 +25,7 @@ const SuspensableResourceSelector = ({
     resourceId: selectedResourceId,
   })
   const [parentIdStack, setParentIdStack] = useState<string[]>(
-    ancestryStack
-      .map((item) => item.id)
-      .reverse()
-      .slice(0, -1),
+    ancestryStack.map((item) => item.id),
   )
   const currResourceId = parentIdStack[parentIdStack.length - 1] ?? null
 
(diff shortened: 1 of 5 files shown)


apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,41 +1,42 @@
...
+  const [ancestryStack] = trpc.resource.getAncestryOf.useSuspenseQuery({
+    siteId,
+    resourceId: selectedResourceId,
+  })
+  const [parentIdStack, setParentIdStack] = useState<string[]>(
+    ancestryStack
+      .map((item) => item.id)
+      .reverse()
+      .slice(0, -1),
+  )
```
karrui, line 32: since this is the only use case, why not return the shape directly instead of modifying it more on the client?

apps/studio/src/schemas/resource.ts
```diff
@@ -64,3 +64,8 @@ export const listResourceSchema = z
 export const getFullPermalinkSchema = z.object({
   resourceId: bigIntSchema,
 })
+
+export const getAncentrySchema = z.object({
```
karrui, line 68: typo: 
```suggestion
export const getAncestrySchema = z.object({
```

karrui (approved):
lgtm, but strongly suggest returning the shape directly from the server

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
dcshzj: Adjusted in cd38265

dcshzj:
Still got some issue with the modal disappearing when clicking on the parent directory, will investigate

PR #618, opened 2024-09-12
chore(modals): standardise input field

seaerchin:
## Problem
currently, the four modals for creation of pages/folders don't have a standardised way of treating input. this PR fixes that.

## Solution
1. input now has `maxLength` set so users cannot go beyond `maxLength`
2. input now auto transforms illegal characters

## Note
because of #2, if users have illegal characters in the middle, they'll add the characters there, then their cursor will move to the end (eg: aaa -> a<space>aa, cursor at index 1 -> a-aa<cursor>, cursor at ending position)

ok with this trade-off because it standardizes all input field treatment and we don't have error message atm for the input fields hwen illegal characters have been entered.

### DIFF
--- a/apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
@@ -161,7 +161,7 @@ const CreateCollectionModalContent = ({
                 name="permalink"
                 render={({ field: { onChange, ...field } }) => (
                   <Input
-                    placeholder="This is a url for your new collection"
+                    placeholder="This is a URL for your new collection"
                     noOfLines={1}
                     maxLength={MAX_FOLDER_PERMALINK_LENGTH}
                     {...field}
--- a/apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
@@ -22,6 +22,7 @@ import {
   ModalCloseButton,
   useToast,
 } from "@opengovsg/design-system-react"
+import { Controller } from "react-hook-form"
 import { BiLink } from "react-icons/bi"
 
 import { useZodForm } from "~/lib/form"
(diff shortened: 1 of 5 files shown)


apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
```diff
@@ -148,9 +156,25 @@ const CreateCollectionModalContent = ({
...
               </FormLabel>
-              <Input
-                placeholder="This is a url for your new page"
-                {...register("permalink")}
+              <Controller
+                control={control}
+                name="permalink"
+                render={({ field: { onChange, ...field } }) => (
+                  <Input
+                    placeholder="This is a url for your new collection"
```
harishv7, line 164: nit: should url be URL?

apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
```diff
@@ -148,9 +156,25 @@ const CreateCollectionModalContent = ({
...
+                    noOfLines={1}
+                    maxLength={MAX_FOLDER_PERMALINK_LENGTH}
+                    {...field}
+                    onChange={(e) => {
+                      onChange(
```
harishv7, line 169: actually why do we need the slice? if users enter > than the allowed chars we just need to show validation error and prevent them from saving right

apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
seaerchin: @sehyunidaaa

apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
seaerchin: i'm doing this to standardise validation across all our `Input`s. For example, our json forms renderer now also adopts this behaviour so it's just to 
1. have a consistsent behaviour across inputs in frontend
2. don't need to care about `disabled` because [illegal states are not representable](https://ybogomolov.me/making-illegal-states-unrepresentable)

apps/studio/src/features/editing-experience/components/CreateCollectionModal/CreateCollectionModal.tsx
seaerchin: designer say `URL`, i swap to `URL`

harishv7 (approved):

PR #619, opened 2024-09-12
chore: UI cleanup

seaerchin:
## Problem
Right now, we have incomplete features still available in our UI. This PR disables said features. It also comes with enhancements to certain components

## Solution
1. delete or disable features that are not ready
2. input max length have been standardised to prevent anymore input at `maxLength`


karrui (commented):
can we add a TODO referencing a linear ticket so these don't get dropped? we should get the missing features in asap.

seaerchin:
filed as linear issue
https://linear.app/ogp/issue/ISOM-1552/unfinished-scope

karrui:
I'll take over this PR

karrui (approved):

PR #621, opened 2024-09-13
fix: remove image preview and adjust URL

dcshzj:
## Problem

The image preview right now is a little wonky.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Fixed the fetch URL to include the missing `https://`, so the fetching is now from the correct source.
- Disabled the image preview from the Attachment component as it is possible that the uploading/malware scan is still in progress, so there is nothing to show.
    - Opted to use a fallback blob that is empty, so the file size would be inaccurate. Future fix is to directly retrieve this metadata from the S3 bucket.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

### DIFF
--- a/apps/studio/next.config.mjs
+++ b/apps/studio/next.config.mjs
@@ -33,6 +33,7 @@ const ContentSecurityPolicy = `
     https://vitals.vercel-insights.com/v1/vitals
     https://*.amazonaws.com
     https://placehold.co
+    ${env.NODE_ENV === "production" ? "https://isomer-user-content-vapt.by.gov.sg" : "https://*.by.gov.sg"}
     ;
   worker-src 'self' blob:;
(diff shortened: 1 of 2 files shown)


apps/studio/next.config.mjs
```diff
@@ -33,6 +33,7 @@ const ContentSecurityPolicy = `
     https://vitals.vercel-insights.com/v1/vitals
     https://*.amazonaws.com
     https://placehold.co
+    ${env.NODE_ENV === "production" ? "https://isomer-user-content-vapt.by.gov.sg" : "https://*.by.gov.sg"}
```
karrui, line 36: We using vapt domain on prod?

karrui (approved):

harishv7 (approved):

PR #622, opened 2024-09-13
fix seed permalink for root page

harishv7:
### TL;DR

Updated the permalink for the home resource in the seed data.

### What changed?

Changed the permalink value for the home resource from "home" to an empty string ("") in the seed.ts file.

### How to test?

1. Run the database seeding process
2. Check the Resource table in the database
3. Verify that the home resource has an empty string as its permalink

### Why make this change?

This change ensures that the home page has an empty permalink, which is typically the expected behavior for the root page of a website. It allows the home page to be accessed at the root URL (/) rather than at /home, improving URL structure and SEO.

### DIFF
--- a/apps/studio/prisma/seed.ts
+++ b/apps/studio/prisma/seed.ts
@@ -215,7 +215,7 @@ async function main() {
     .insertInto("Resource")
     .values({
       draftBlobId: String(blobId),
-      permalink: "home",
+      permalink: "",
       siteId,
       type: "RootPage",
       title: "Home",


harishv7:
* **#622** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/622?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

seaerchin (changes requested):
Should the permalink be slash instead?

harishv7:
> Should the permalink be slash instead?

should be empty, we do not use "/" for our permalink column

seaerchin (commented):
this change by itself is ok but it leaves the page settings in an ambiguous state as shown below. 
image

could i trouble you to update the `input` in this case?

harishv7:
> this change by itself is ok but it leaves the page settings in an ambiguous state as shown below. 
> image
> 
> could i trouble you to update the `input` in this case? 

Hmm doesn't happen on local, could this have been fixed in recent commits?

seaerchin (approved):
lgtm! this got fixed by another pr HAHAH

PR #623, opened 2024-09-13
fix: form builder fields spacing and missing group

dcshzj:
## Problem

The form builder has weird spacing due to our use of empty elements for constants. The logic to build groups was also slightly flawed due to the use of splice.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Removed the stack gaps and switched to use margin tops instead (unless it is the first element).
- Fixed the logic for building groups to ensure that we always process all the properties.
- Removed DropdownControl and RadioControl. They aren't actually used since we mostly rely on `anyOf`, which leads to the AnyOfControl.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

image


apps/studio/src/constants/formBuilder.ts
```diff
@@ -2,15 +2,14 @@
...
-  DropdownControl: 2,
+  ConstControl: 2,
   ImageControl: 2,
   IntegerControl: 4,
   TextControl: 1,
   ObjectControl: 2,
   AllOfControl: 3,
   AnyOfControl: 3,
   ProseControl: 2,
-  RadioControl: 3,
```
seaerchin, line 13: why did we remove this radio control?

seaerchin (commented):
i'm ok with most of the changes here but i'm quite confused wrt how this actually works. just to ensure my understanding is correct, the dropdown and radio controls are actually rendred by the `anyOf` control and tester and hence, the deleted files were never used?

in addition, we're adding a `const` tester as `typebox` adds a `const` to properties and we want to omit this from being rendered on screen?

happy to approve once doubts cleared

apps/studio/src/constants/formBuilder.ts
seaerchin: answered aboe, i think removed cos functionality subsumed by any of renderer

dcshzj:
> i'm ok with most of the changes here but i'm quite confused wrt how this actually works. just to ensure my understanding is correct, the dropdown and radio controls are actually rendred by the anyOf control and tester and hence, the deleted files were never used?

Yup, the way TypeBox generates the schema is using `anyOf` in JSON Schema. The dropdown and radio controls were from a time where we manually wrote the schema, which we used more of `enum` ([see reference](https://json-schema.org/understanding-json-schema/reference/enum)). We no longer use `enum` and TypeBox doesn't generate such schemas, so there is no need to have these files in the codebase (basically they are dead code).

> in addition, we're adding a const tester as typebox adds a const to properties and we want to omit this from being rendered on screen?

Yup, right now they are being rendered by the catch-all renderer (I think), so we have an explicit one. The catch-all is more for cases where we really didn't plan for, so wanted to make things a bit clearer here.

seaerchin (approved):

PR #625, opened 2024-09-13
feat: fetch proper data for edit page breadcrumb

dcshzj:
## Problem

The edit page breadcrumb currently is using dummy data.

## Solution

**Breaking Changes**

**Features**:

- Fetch the correct data for the edit page breadcrumb (which is the current page title)
- If the page's parent folder is not the root folder, show the immediate parent of that page as a link (this is mainly for collections).

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image


seaerchin (commented):
unfortunately this has a bug. we need  to remember to invalidate the query on move otherwise it'll fetch stale data.

seaerchin:
https://github.com/user-attachments/assets/c267365f-4f44-4feb-b201-300ab8be9a3e

dcshzj:
> unfortunately this has a bug. we need to remember to invalidate the query on move otherwise it'll fetch stale data.

Updated in 45573bbed0aa929304b43fa1c7e1fac2d964c34f.

seaerchin (approved):
lgtm

PR #627, opened 2024-09-13
chore: add filter so that we cannot do self-moves

seaerchin:
## Problem
previously, we didnt do a filter on the data retrieved from backend, which led to self-moves being possible (ie, i move a folder to itself)

## Solution
1. filter by `resourceId` on the frontend
    - this is safe to do because the method invoked on the backend returns strictly `folder`
    - so for pages, the condition will be `false` by default (this is paginated so run time isn't an issue)
    - for folders, we disallow move to self (and children) and because we start move from `root`, we'll never to able to get to the folder and its associated children
2. backend already has an error message here for self moves so it's still disallowed but this error message wouldn't be seen by law abiding citizens who use our frontend

## Screenshots
https://github.com/user-attachments/assets/8bcd295f-6ce8-4cc3-b3a0-416b02dbce81

### DIFF
--- a/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
@@ -191,24 +191,26 @@ const MoveResourceContent = withSuspense(
                 </Flex>
               )}
               {data?.pages.map(({ items }) =>
-                items.map((child) => {
-                  return (
-                    <MoveItem
-                      {...child}
-                      key={child.id}
-                      onChangeResourceId={() => {
-                        setResourceStack((prev) => [
-                          ...prev,
-                          {
-                            ...child,
-                            parentId: curResourceId ?? null,
-                            resourceId: child.id,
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -188,24 +188,26 @@ const MoveResourceContent = withSuspense(
...
-                            parentId: curResourceId ?? null,
-                            resourceId: child.id,
-                          },
-                        ])
-                      }}
-                    />
-                  )
-                }),
+                items
+                  .filter((child) => child.id !== movedItem?.resourceId)
```
karrui, line 192: feel like shouldn't filter it out. can we set it to disabled state for (like for pages) if the resource is moving to itself?

seaerchin: sounds good, lemme make this change!

karrui (approved):

PR #629, opened 2024-09-13
fix(page): derive title from correct location

seaerchin:
## Problem
Previously, we were directly getting the title from the blob which we did not want as the title now lives in the `Resource` table and not the blob. this led to issues with the preview's title being distinctly different from the actual title.

## Solution
1. remove the `title` from the schema and instead, add it to the `additionalProps`. this is required because jsonforms walks our schema and generates the form fields from there. having it on the schema would still lead to the form field being present.
2. change the title to retrieved from the resource table and pipe it down to the `preview`, which does a merge.
3. for creation, this is done using a top level merge with the `title` from the form fields.

## Notes
not really a fan of this solution here due to how we type `layoutPreviewJson` as `IsomerSchema` which **does not have** the `title`. this is injected via a `merge` and also doesn't show up on the type. 

should i: 
1. ~type the `layoutPreviewJson` as `IsomerSchema & BaseItemAddtionalProps`? or~
2. ~is there some type exported form the schema package that encapsulates the additional props?~

updated to `merge` prior to passing it to `Preview`

## Screenshots

https://github.com/user-attachments/assets/3a72d63d-afcd-45d7-8765-932cbe20d52f


seaerchin:
Will fix type errors

PR #632, opened 2024-09-13
fix(page-settings): change field generation

seaerchin:
## Problem
previously, we swapped the method for determining the permalink. unfortunately, this method fetches the full permalink which contains invalid characters (`/`). this led to the save being disabled until the field is typed.

additionally, `isDirty` wasn't quite fitting our use case and we have to check against the `dirtyFields` of the `formState` instead. this showed up as edits not working until the second blur.
image

## Solution
1. change the api of the component to accept a `prefix`. this is the front portion of hte `permalink`
2. update the full permalink retrieved to be 2 portions, the prefix and the current url
3. updat the component so that input edits only the last portion 

## Screenshots

https://github.com/user-attachments/assets/b826990e-9be4-4089-a3c0-70a0d43b85ec

### DIFF
--- a/apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
+++ b/apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
@@ -10,6 +10,7 @@ import {
 } from "@chakra-ui/react"
 import { Infobox, Input, useToast } from "@opengovsg/design-system-react"
 import { ResourceType } from "~prisma/generated/generatedEnums"
+import { isEmpty } from "lodash"
 import { Controller } from "react-hook-form"
 import { BiLink } from "react-icons/bi"
 
@@ -69,7 +70,12 @@ function EditPage(): JSX.Element {
         </EditorDrawerProvider>
       </TabPanel>
       <TabPanel>
-        <PageSettings type={type} permalink={permalink} title={title} />
+        <PageSettings
+          type={type}
+          permalink={permalink.split("/").at(-1) || "/"}
+          prefix={permalink.split("/").slice(0, -1).join("/")}
+          title={title}
+        />
       </TabPanel>
     </TabPanels>
   )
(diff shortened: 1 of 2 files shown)


apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -169,7 +178,7 @@ const PageSettings = ({
   })
 
   const onSubmit = handleSubmit((values) => {
-    if (formState.isDirty) {
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:172]: i think this one is one render cycle behind because formState is a proxy. Might be able to get the latest state if this is invoked before the render instead of inside this function:

```ts
const {
  formState: { isDirty } // invoked here
  // ...
} = useZodForm(...)
```

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -69,7 +70,12 @@ function EditPage(): JSX.Element {
         </EditorDrawerProvider>
       </TabPanel>
       <TabPanel>
-        <PageSettings type={type} permalink={permalink} title={title} />
+        <PageSettings
+          type={type}
+          permalink={permalink.split("/").at(-1) || "/"}
+          prefix={permalink.split("/").slice(0, -1).join("/")}
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:76]: should the trpc procedure return the `relativePermalink` and `prefix` instead of doing this frontend conversion? if not might desync.

apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx
```diff
@@ -179,6 +188,9 @@ const PageSettings = ({
     }
   })
 
+  const fullPermalink =
+    type === ResourceType.RootPage ? "/" : `${prefix}/${permalink}`
```
Comment (karrui) [inline apps/studio/src/pages/sites/[siteId]/pages/[pageId]/index.tsx:192]: nit: Might be safer (and make more sense) to use `String#join` instead so don't need the conditional.

```suggestion
  const fullPermalink = [prefix, permalink].join('/')
```

karrui (approved):

karrui (changes requested):
tests failed for some reason, might want to fix

karrui:
fixed the tests in #638

karrui (approved):

PR #633, opened 2024-09-13
fix(overflow): breadcrumbs and titles can overflow leading to bad ux

seaerchin:
## Problem
previously, titles and breadcrumbs can overflow, which blocks off content, leading to bad ux

Closes ISOM-1468
Closes ISOM-1480

## Solution
1. define overflow styling and a max width for breadcrumbs so that on overflow, the last breadcrumb will truncate nicely
2. set overflow styling on the container for the title

https://github.com/user-attachments/assets/9acdcdb1-662e-40e2-ac0b-c67e8bdb5636

### DIFF
--- a/apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
+++ b/apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
@@ -142,7 +142,7 @@ const FolderPage: NextPageWithLayout = () => {
             </BreadcrumbItem>
           </Breadcrumb>
           <Flex w="full" flexDir="row">
-            <HStack overflow="auto" gap="0.75rem" flex={1}>
+            <HStack mr="1.25rem" overflow="auto" gap="0.75rem" flex={1}>
               <Box
                 aria-hidden
                 bg="brand.secondary.100"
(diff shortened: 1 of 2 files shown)


sehyunidaaa:
@seaerchin lgtm! but i think can still increase the gap to 20px? wdyt

karrui:
will review after rebase

apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx
```diff
@@ -109,29 +109,40 @@ const FolderPage: NextPageWithLayout = () => {
     <>
       <VStack w="100%" p="1.75rem" gap="1rem">
         <VStack w="100%" align="start">
-          <Breadcrumb size="sm">
+          <Breadcrumb size="sm" w="100%">
```
Comment (harishv7) [inline apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:112]: should it be 100% though? Never seen such a long breadcrumb either

i thought we could do something like:

folder123 > superlooooo .... oooonnnng > anotherrrrr...rrrlongone >

basically should each section have a max length

Cos in this case if the first one is super long, then it itself will take up the whole. space right?

Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:112]: since our designer is ok with this current implementation, should we merge this first and file a follow up issue? i don't really have an issue iwth hte breadcrumb taking up the whole area because that's what the whitespace is for. 

i made the choice to truncate on the last breadcrumb because it's the one that has redundant information - the title already show the information on the breadcrumb and the breadcrumb is also non-clickable so it doesn't contribute anything. 

cc @sehyunidaaa should the max length of hte breadcrumbs be constrained perhaps to `50%`? (put another way, should the breadcrumb be allowed to take the full whitespace)

Comment (harishv7) [reply apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx:112]: @seaerchin I am good to merge and file a follow up issue!

PR #634, opened 2024-09-13
chore: add banner

seaerchin:
## Problem
add simple banner for users

## Solution

add banner

https://github.com/user-attachments/assets/28a3f17a-25c3-4d17-b1ac-da9a4ff2dc0a


apps/studio/src/pages/_app.tsx
```diff
@@ -33,6 +33,10 @@ const MyApp = ((props: AppPropsWithAuthAndLayout) => {
             <ErrorBoundary FallbackComponent={DefaultFallback}>
               <Suspense fallback={<Skeleton width="100vw" height="100vh" />}>
                 <Stack spacing={0} minH="$100vh">
+                  <Banner variant="warn">
+                    Isomer Next is currently in Beta. To manage site settings
+                    that are not displayed here, contact Isomer Support.
```
karrui, line 38: since we have growthbook liow. I think we can take this opportunity to wire up the banner text to growthbook. Can see [ActiveSG implementation](https://github.com/opengovsg/activesg/blob/main/packages/ui/src/components/AppBanner.tsx#L36)

seaerchin: sounds good, lemme add it in

package.json
```diff
@@ -31,5 +31,8 @@
   "devDependencies": {
     "turbo": "^2.0.5"
   },
-  "packageManager": "npm@10.2.3"
+  "packageManager": "npm@10.2.3",
+  "dependencies": {
+    "@growthbook/growthbook-react": "^1.2.0"
```
karrui, line 36: wrong place, put inside studio repo pls

apps/studio/next.config.mjs
```diff
@@ -33,6 +33,7 @@ const ContentSecurityPolicy = `
     https://vitals.vercel-insights.com/v1/vitals
     https://*.amazonaws.com
     https://placehold.co
+    https://cdn.growthbook.io/api/features/${env.NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY}
```
karrui, line 36: now we have two env vars that are the same. do you want to remove the (non-public) `GROWTHBOOK_CLIENT_KEY` env var with this public one?

remember update iac also

seaerchin: actually i think these are different! not very sure _how_ they're different tbh and growthbook doesn't make it very clear ._.

i was under the impression that you'd need to create 1 for be, 1 for fe and 1 for each env?

image

package.json
seaerchin: oopsie ok

apps/studio/next.config.mjs
karrui: don't need actually, but up to you haha. one is enough

karrui:
fix lint and tests pls

seaerchin:
gg lemem fix it later. the `lint` is actually detrimental here - it's complaining that the interface is not indexable by a `string` type but changing the `interface` to a `type` declaration works

apps/studio/next.config.mjs
harishv7: @karrui I think the FE only is read-only, the BE only is read+write. so should have separate key if we are exposing to FE

harishv7: At least thats what it was in the old version of GB, can double check if the latest release is still the case

karrui: okie. leave it up to chin!

karrui (approved):
lgtm, but got merge conflict hehe

PR #641, opened 2024-09-16
feat: add page settings route and separate edit/settings views, link to page in dashboard

karrui:
### TL;DR

Added a new page settings view and improved navigation between edit and settings pages.

### What changed?

- Created a new `TabLink` component for consistent tab-like navigation
- Added a new `/settings` page for editing page settings
- Moved page settings form from the edit page to the new settings page
- Updated the `SiteEditNavbar` to use `TabLink` for navigation between edit and settings views
- Implemented `getResourcePermalinkTree` and `getPermalinkTree` API endpoints
- Updated resource utility functions for better link generation

### How to test?

1. Navigate to a page editing view
2. Verify that the "Edit" and "Page Settings" tabs are present and functional
3. Click on "Page Settings" to access the new settings page
4. Test editing page title and URL on the settings page
5. Verify that changes are saved and reflected in the UI
6. Test navigation between edit and settings views

### Why make this change?

This change improves the user experience by separating the page editing and settings functionalities into distinct views. It provides a cleaner interface for managing page settings and allows for easier navigation between editing content and modifying page metadata. The new structure also lays the groundwork for potential future expansions of page settings options.

PR #642, opened 2024-09-16
feat: add first cut indices for resource table, unique constraint on permalinks

karrui:
Has breaking changes!!!!

Also closes ISOM-1559

## Solution

**Breaking Changes**

May fail to apply on any databases that has duplicate permalinks. 

**Improvements**:

- Added indices to the Resource table in the database schema to optimize query performance
- Created a unique constraint on the `siteId` and `permalink` columns of the Resource table

### DIFF
--- a/apps/studio/prisma/schema.prisma
+++ b/apps/studio/prisma/schema.prisma
@@ -72,7 +72,10 @@ model Resource {
   createdAt DateTime @default(now())
   updatedAt DateTime @default(now()) @updatedAt
 
-  @@unique([siteId, parentId, permalink], name: "unique_site_permalink")
+  // This unique index is subsequently replaced with a custom index that
+  // treats nulls as not distinct.
+  // This is required so prisma does not attempt to drop the custom index.
+  @@unique([siteId, parentId, permalink])
   @@index([siteId])
   @@index([siteId, id])
   @@index([siteId, id, parentId])
--- a/apps/studio/prisma/schema.prisma
+++ b/apps/studio/prisma/schema.prisma
@@ -44,7 +44,7 @@ model Version {
   publishedBy String
   publisher   User       @relation(fields: [publishedBy], references: [id])
 
-  updatedAt   DateTime   @default(now()) @updatedAt
+  updatedAt DateTime @default(now()) @updatedAt
 }
 
 model Resource {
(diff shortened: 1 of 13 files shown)


karrui:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/642?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#642** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/642?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#641** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/641?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

seaerchin (commented):
Don't we need it to be unique by parent too?

karrui:
I forgot. Yup need. Let me rebase

apps/studio/prisma/schema.prisma
```diff
@@ -71,16 +71,21 @@ model Resource {
 
   createdAt DateTime @default(now())
   updatedAt DateTime @default(now()) @updatedAt
+
+  @@unique([siteId, permalink], name: "unique_site_permalink")
```
harishv7, line 75: caveat here is that this means within the same siteId, the permalink must be unique. But what we are ensuring is the full qualified permalink - which might be expensive to re-create recursively, but given that create/edit page settings is a relatively less frequent operation.

Am okay with this for now, but is something we should take note of

harishv7, line 75: if create page/edit page settings fails, will the user know it is due to the permalink? we should return an appropriate error back

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -20,6 +23,15 @@ export const folderRouter = router({
...
         .executeTakeFirstOrThrow()
+        .catch((err) => {
+          if (get(err, "code") === "23505") {
+            throw new TRPCError({
+              code: "CONFLICT",
+              message: "A resource with the same permalink already exists",
+            })
+          }
+          throw err
+        })
```
seaerchin, line 34: we need to duplicate this for `collection.router` otherwise, we get an ugly error message

seaerchin (changes requested):
i found 2 issues:
1. when creating a collection that has the same permalink as another collection/folder/page, we get an ugly toast
2. when creating a collection page that has the same permalink as another item in the collection, there's no error message
https://github.com/user-attachments/assets/0b8f9033-4ac0-417f-a641-3f6dc2994467


https://github.com/user-attachments/assets/477306ed-8e52-46ca-ae92-201ccea47c44

apps/studio/src/server/modules/folder/folder.router.ts
karrui: fixed

seaerchin (approved):
lgtm. a minor nit here is that the code `23505` is an error code used by pg to signal violation of a unique index, which might not be immediately obvious here. we could extract this out to a `const` for maintainability but functionality wise this works

PR #646, opened 2024-09-16
chore(arraycontrol): add extra tester so we have a renderer

seaerchin:
## Problem
Currently, when we have an article, theere is no applicable renderer for the heading as it is an array of primitives.

## Solution
1. add a or statement so that object or primitive arrays are both caught by this tester.

## Screenshots

https://github.com/user-attachments/assets/b28cdb7c-9b27-42d3-b81c-e244ef6ee671

## Note
pending design to confirm if this is what they want...

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
@@ -15,6 +15,7 @@ import {
   createDefaultValue,
   findUISchema,
   isObjectArrayControl,
+  isPrimitiveArrayControl,
   rankWith,
   Resolve,
 } from "@jsonforms/core"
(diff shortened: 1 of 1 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -37,7 +38,9 @@ import DraggableDrawerButton from "./DraggableDrawerButton"
 
 export const jsonFormsArrayControlTester: RankedTester = rankWith(
   JSON_FORMS_RANKING.ArrayControl,
-  isObjectArrayControl,
+  (uischema, schema, context) =>
+    isObjectArrayControl(uischema, schema, context) ||
+    isPrimitiveArrayControl(uischema, schema, context),
```
dcshzj, line 43: issue: I think this component wasn't designed for array of primitives, so might need to have a separate renderer

seaerchin: it actually worked (refer to video) so i think it might be fine tbh

dcshzj (approved):
Approving first, let me redo this later

PR #660, opened 2024-09-18
ref: separate metadata from page object

dcshzj:
## Problem

The page object is mixed with some page metadata, which is difficult to separate on the Studio.

## Solution

**Breaking Changes**

    - No schema changes are needed because the new `meta` object is optional and hence, backward compatible.

**Improvements**:

- Refactor all page metadata into a new `meta` object, which is exposed by `getLayoutMetadataSchema`. This object is set to be optional for all page layouts for now to minimise the need to upgrade existing schemas.
- The remaining `page` object is exposed by `getLayoutPageSchema`, which replaces the one currently used in MetadataEditorStateDrawer.
- Additionally, if the page meta description is not provided, the article/content page summary is used for those layouts.

**Notes**:

- At this stage, the page metadata cannot be edited. They will need to be added into the page settings screen.

## Screenshots

image


karrui (approved):

sehyunidaaa:
Sounds good! No concerns from me, definitely makes more sense to separate meta fields out to settings

PR #662, opened 2024-09-19
docs(readme): remove references to Vercel

LoneRifle:
## Problem

Despite moving away from Vercel, Isomer's README still makes reference to the platform. This confuses new users who might think that Vercel is needed.

## Solution

Rework relevant text in README


apps/studio/vercel.json
karrui, line 1: we actually still use vercel for hosting our preview branches. so shouldn't remove this file

seaerchin (approved):

PR #663, opened 2024-09-19
fix: refactor and upload/delete hero banner

dcshzj:
## Problem

Fixes ISOM-1585.

## Solution

**Breaking Changes**

**Improvements**:

- The assets uploading logic has been abstracted into its own utility function to facilitate reuse.

**Bug Fixes**:

- The hero banner is now correctly being uploaded to the S3 bucket on save.
- Marking assets as deleted on the server is now correctly done, as previously they were trying to delete the local blob instead of the actual path to the file in the S3 bucket.
- The placeholder image is now correctly skipped from deletes.
- The preview state is now updated once the file keys are available, so users will be shown the placeholder image as the image is being scanned for malware on the S3 bucket.

## Before & After Screenshots

https://github.com/user-attachments/assets/b268fa90-385b-4fac-8e45-ceb6961ccba7


apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -200,26 +169,30 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
         return
       }
 
-      deleteAssets({ fileKeys: assetsToDelete })
+      // Delete the original assets for those that have been modified
+      // This is done by deleting the file key stored in the src attribute, as
+      // it would have been replaced by new file keys after uploading
+      const assetsToDelete = modifiedAssets
+        .map(({ src }) => src?.slice(1))
+        .filter((src) => src !== PLACEHOLDER_IMAGE_FILENAME)
```
karrui, line 187: as discussed offline, maybe we should have a hardcoded default image (served from our own public folder) as a fallback instead of using a hosted image haha. But that can come in a later PR. I worry because then this is a coupled requirement for all our image/ image upload components.


something like
```ts
import { useState } from 'react';
import Image from 'next/image';

export const ImageWithFallback = ({ src, fallbackSrc = "./assets/fallbackimg", ...rest }) => {
    const [imgSrc, setImgSrc] = useState(src);

    return (
        <Image
            {...rest}
            src={imgSrc}
            onError={() => {
                setImgSrc(fallbackSrc);
            }}
        />
    );
};
```

karrui (approved):

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
dcshzj: Yeah the fallback image is quite straightforward to do inside the components, but right now the placeholder image is being injected by our component selector default blocks definition, and omitting the value is not allowed since the prop is required.

Probably when we can make it optional, then we can remove the need for this placeholder.

PR #665, opened 2024-09-19
feat: introduce big add block button when empty

dcshzj:
## Problem

Fixes ISOM-1581.

## Solution

**Breaking Changes**

**Features**:

- Add the "Add a new block" big button when the page state is empty.

## Screenshots

image


karrui (approved):
can also add one empty block story to track regressions

dcshzj:
> can also add one empty block story to track regressions

Hmm the default state is the empty state so it should keep track

PR #666, opened 2024-09-20
feat: provide layout info to components

dcshzj:
## Problem

We don't provide layout information to the components currently, but there is a growing need to render components slightly differently depending on the layout that it is being rendered in (e.g. homepage vs content page.

## Solution

**Breaking Changes**

**Features**:

- Provide layout information to the components so they are aware of the layout context they are rendering in.

**Improvements**:

- Standardised all LinkComponent and ScriptComponent to not provide a fallback value, and leave it to `renderLayout` to provide one.


karrui (approved):
just to clarify; this PR does nothing except to pass layout to components yeah? in the future the layout prop may be used inside components for different behaviour? if so, lgtm.

PR #672, opened 2024-09-20
chore: update workflow for production

seaerchin:
## Problem
currently, we release using a label. this is not ideal because we have to raise a pr with no meaningful changes and then attach a label.

## Solution
update so that we now deploy to production on release rather than using a pr label. this is preferred because we don't have to make empty commits and a fake pr to create a release.

removed the action to comment on pr because no easy out of the box functionality was found. we can add it back next time if we feel that it's needed

## Notes
chose `released` over `published.` see here: https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release for the differences but i think both are fine, just that it's more semantic to publish releases as a version bump rather than attaching to the draft release. this means that we'll only deploy to prod when we publish a release!

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -5,31 +5,14 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  pull_request:
+  release:
     types:
-      - labeled
+      - released
 
 # NOTE: This is actually using our federated isomer-production account
 jobs:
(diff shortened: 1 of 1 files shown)


karrui (approved):
lgtm. so the new way is just to create a release?

seaerchin:
@karrui yes, and it should trigger the workflow in the background!

PR #674, opened 2024-09-22
fix(components): standardise hover interactions for links with arrows on megamenu

sehyunidaaa:
## Problem

Closes [ISOM-1357]

Currently, all links with arrows on the Isomer Next template share an interaction where the arrow inches by 4px to the right with an ease-in transition. The megamenu was the only place where this interaction was different. This causes an inconsistency in interactions which impacts usability of the site.

## Solution

Changed megamenu item interactions from underline to nudging the arrow and getting an interaction-hover colour. This won't take effect on external links.

**Breaking Changes**

  - Details ...
- [x ] No - this PR is backwards compatible

## Before & After Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/191e54af-58d6-438d-8e32-e8e3127302ca

**AFTER**:

https://github.com/user-attachments/assets/6a14db76-afe0-4b1a-9cd3-94fa99329644

### DIFF
--- a/packages/components/src/templates/next/components/internal/Navbar/NavItem.tsx
+++ b/packages/components/src/templates/next/components/internal/Navbar/NavItem.tsx
@@ -145,11 +145,11 @@ const Megamenu = ({
                         isExternal={isExternal}
                         showExternalIcon={isExternal}
                         href={subItem.url}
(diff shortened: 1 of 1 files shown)


karrui:
need to rebase so the new CI runs

PR #675, opened 2024-09-22
fix(components): add background to On this page/ToC for visual balance and cohesion

sehyunidaaa:
## Problem

Previously, the ToC ("On this page") lacked background colour which made it look floaty. 

Closes [insert issue #]

## Solution

This PR adds a background to the component so that all the links under a ToC seem like they are in the same group. This should bring more attention to the ToC, helping site visitors to find the exact subtopic they want.

The component also has an 8px rounded corner to make the component look modern and softer, which is the same value that we use for callouts.

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**BEFORE**:

Previously, the ToC was floating around, and top-alignment with the siderail was off:
image

**AFTER**:

View chromatic.


seaerchin (approved):

karrui:
need to rebase so the new CI runs

PR #676, opened 2024-09-22
fix(components): footer enhancements

sehyunidaaa:
## Problem

This PR addresses some styling and copy enhancements for footer.

Previously, the footer wasn't collapsing right on the tablet-small desktop breakpoints. It had a larger x-padding which made the footer contents misaligned with the rest of the homepage content.

Also we had comments from migration sync that the capitalisation for "Contact Us" was inconsistent, and the term "Reach us" clashes with "REACH". 

Closes ISOM-1356

## Solution

- Changed padding to be consistent at the lg-xl breakpoint at 40px. 
- Also minimised some minor gaps that were very glaring between sections.
- Copy changes.

## Before & After Screenshots

**BEFORE**:

Notice the footer contents are misaligned with the side rail:
https://github.com/user-attachments/assets/bbd33e6e-c1cd-4122-88fe-dc3291afade0

**AFTER**:

Footer collapses consistently with homepage contents:
https://github.com/user-attachments/assets/51e54733-0836-47f7-8aa8-158b05f0a05e

### DIFF
--- a/packages/components/src/templates/next/components/internal/Footer/Footer.tsx
+++ b/packages/components/src/templates/next/components/internal/Footer/Footer.tsx
@@ -379,7 +379,7 @@ const FooterDesktop = ({
   siteMapLink,
 }: FooterProps) => {
   return (
-    <div className="hidden px-[4rem] py-14 lg:block">
+    <div className="hidden px-10 py-14 lg:block">
       <div className="mx-auto flex max-w-[72.5rem] flex-col gap-6">
         <SiteNameSection siteName={siteName} />
         <div className="grid-cols-[1fr_min-content] grid-rows-[1fr_min-content] gap-x-8 gap-y-14 lg:grid">
--- a/packages/components/src/templates/next/components/internal/Footer/Footer.tsx
+++ b/packages/components/src/templates/next/components/internal/Footer/Footer.tsx
@@ -96,7 +96,7 @@ const NavSection = ({
   customNavItems,
 }: Pick<FooterProps, "LinkComponent" | "siteNavItems" | "customNavItems">) => {
   return (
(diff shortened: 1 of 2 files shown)


packages/components/src/templates/next/components/internal/Footer/Footer.tsx
```diff
@@ -379,7 +379,7 @@ const FooterDesktop = ({
   siteMapLink,
```
harishv7, line 379: This vertical gap between the site name and the botton seems to be quite large

harishv7: Or is it because we didn't add footer items in storybook?

sehyunidaaa: Yeah, but will fix this!

sehyunidaaa:
I will work on this tomorrow and remove draft status

packages/components/src/templates/next/components/internal/Footer/Footer.tsx
sehyunidaaa: @harishv7 This wasn't fixable on my end, the div itself is created even if there are no navigation items so the padding still appears. It looks like we can't create a footer without navigation items so for now we'll assume this doesn't happen, need engineer's help later on for cases where the site's navbar is empty (not priority now?)

But I adjusted some other gaps like the one between the site name and the 1st nav item in https://github.com/opengovsg/isomer/pull/676/commits/e866254fc35cfe8c80e0f0a0bea2dc937bc01582

karrui (approved):
ownself review ownself on chromatic k. if you leave it unapproved, next time the changes will propagate to the other person's PR. we need to keep the baselines updated (if they are correct)

sehyunidaaa:
Thanks @karrui will do for all my PRs!

PR #677, opened 2024-09-23
fix: adjust homepage blocks spacing on content page

dcshzj:
> [!NOTE]
>
> This is still pending design adjustments.

## Problem

The homepage blocks have very large spacing when used on content pages.

## Solution

**Breaking Changes**

**Improvements**:

- Adjust the spacing of homepage blocks when they are used on content pages.


packages/components/src/templates/next/components/complex/InfoCols/InfoCols.tsx
```diff
@@ -75,14 +107,32 @@ const InfoCols = ({
   title,
   subtitle,
   infoBoxes,
+  layout,
   site,
   LinkComponent,
 }: InfoColsProps) => {
+  const simplifiedLayout = layout === "homepage" ? "homepage" : "default"
```
karrui, line 114: Should we extract this to a reusable function instead? So the layout prop can be changed in a single place if we ever want to add more layout types in the future

karrui (approved):

PR #681, opened 2024-09-24
feat: add db schema for resource permissions

seaerchin:
## Problem
we have no resource schema for permissions.

## Solution
add a db schema for resouce permissions. we track the `(siteid, userid, resourceid)`. this returns a `role`, which can be one of `publisher, editor, admin`. 

at present, we will take `resourceId === null` as meaning that this user has site wide permissions

### DIFF
--- a/apps/studio/prisma/migrations/20240923090529_add_resource_perms/migration.sql
+++ b/apps/studio/prisma/migrations/20240923090529_add_resource_perms/migration.sql
@@ -1,0 +1,18 @@
+-- CreateTable
+CREATE TABLE "ResourcePermission" (
+    "userId" TEXT NOT NULL,
+    "siteId" INTEGER NOT NULL,
+    "resourceId" BIGINT NOT NULL,
+    "role" "RoleType" NOT NULL,
+
+    CONSTRAINT "ResourcePermission_pkey" PRIMARY KEY ("siteId","userId","resourceId")
+);
+
+-- AddForeignKey
+ALTER TABLE "ResourcePermission" ADD CONSTRAINT "ResourcePermission_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ResourcePermission" ADD CONSTRAINT "ResourcePermission_siteId_fkey" FOREIGN KEY ("siteId") REFERENCES "Site"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ResourcePermission" ADD CONSTRAINT "ResourcePermission_resourceId_fkey" FOREIGN KEY ("resourceId") REFERENCES "Resource"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
--- a/apps/studio/prisma/schema.prisma
+++ b/apps/studio/prisma/schema.prisma
@@ -70,8 +70,9 @@ model Resource {
   state ResourceState? @default(Draft)
   type  ResourceType
 
-  createdAt DateTime @default(now())
-  updatedAt DateTime @default(now()) @updatedAt
+  createdAt          DateTime             @default(now())
+  updatedAt          DateTime             @default(now()) @updatedAt
+  ResourcePermission ResourcePermission[]
 
   // This unique index is subsequently replaced with a custom index that
   // treats nulls as not distinct.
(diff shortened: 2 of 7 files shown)


apps/studio/prisma/schema.prisma
```diff
@@ -198,6 +201,19 @@ model SiteMember {
   @@id([siteId, userId])
 }
 
+model ResourcePermission {
```
harishv7, line 204: this seems to track perms on a resource-level?

for a site-wide perms ie. a user can access all resources in this site as a Admin / Editor / Publisher, how will we represent this?

harishv7, line 204: Since we might need to extend this to future extended perms, what might be confusing is:

is null on resourceId = no resource that the user has access to 
or
is it to all resources (which is the case here) -> implicit allowance on a site-level

would it be better to introduce a enum that makes this entry clear

permsType - site vs resource

Under Site model, maybe we can add the perms type configured for the site - site wide vs resource level

seaerchin: i think `resource === null` to refer to the root resource is pretty much locked in and we've quite a few places that use this convention too. i think the confusion part isn't as bad because if the user has no access, they won't even have the permission in the first place.

additionally, adding the `enum` complicates quite a few things
1. flipping from `site` -> `root` level permissions (another bit to flip)
2. checks next time against whether a user has access to x resource cannot be done with a single query to `ResourcePermission` but needs an additional query (and logic) against `Site` 

the permission checking stuff is hidden behind a library (casl) w. a conditional, so most of the time devs won't be interacting with this

harishv7: I was referring to resourceId === null being interpreted as site-wide perms. Is an implicit convention

Regarding complications:
1. Actually this bit will at most flip once (manually) when a site moves to extended perms, so not really complicated
2. This I agree as it adds 1 more DB touchpoint from Site table which will increases our authz latency

I think it's a fair point this interpretation is abstracted away by our library, hence am okay with this approach esp for point (2) you raised

harishv7 (approved):

harishv7 (approved):

apps/studio/prisma/schema.prisma
```diff
@@ -198,6 +201,19 @@ model SiteMember {
...
+  id         BigInt     @id @default(autoincrement())
+  userId     String
+  user       User     @relation(fields: [userId], references: [id])
+  siteId     Int
+  site       Site     @relation(fields: [siteId], references: [id])
+  resourceId BigInt?
+  resource   Resource? @relation(fields: [resourceId], references: [id])
+  role       RoleType
+
```
dcshzj, line 213: issue: we should add the createdAt and updatedAt fields for all tables
```suggestion

  createdAt DateTime @default(now())
  updatedAt DateTime @default(now()) @updatedAt

```

apps/studio/prisma/migrations/20240923090529_add_resource_perms/migration.sql
```diff
@@ -0,0 +1,18 @@
+-- CreateTable
+CREATE TABLE "ResourcePermission" (
+    "userId" TEXT NOT NULL,
+    "siteId" INTEGER NOT NULL,
+    "resourceId" BIGINT NOT NULL,
+    "role" "RoleType" NOT NULL,
+
+    CONSTRAINT "ResourcePermission_pkey" PRIMARY KEY ("siteId","userId","resourceId")
+);
+
```
dcshzj, line 10: issue: the trigger for the updatedAt needs to be manually added (once you add the field in the prisma schema)

```suggestion

-- AlterTable
CREATE TRIGGER update_timestamp BEFORE UPDATE ON "ResourcePermission" FOR EACH ROW EXECUTE PROCEDURE moddatetime("updatedAt");

```

seaerchin: oh this is abit interesting - why do we need the trigger ah? went to peek at the prisma [docs](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#updatedat) and they said that it's automatically added! 

let me try running this to the schema and check if it's updated

seaerchin: eh cool, it doesn't work without the trigger. i tried with kysely and manual updates but `updatedAt` doesn't work. going to do the `trigger`

dcshzj: It's cos we are using Kysely as the query builder and only using Prisma for the models and db client. Usually on Prisma with the `@updatedAt` decorator, Prisma (as the ORM) will automatically perform the update on the updatedAt field, but we are not using Prisma in our queries here so we have to do it ourselves.

Can see #609 for more context.

dcshzj (approved):

PR #684, opened 2024-09-24
feat: add native image component for Tiptap

dcshzj:
## Problem

We changed the image component into a complex component, but we have cases where we need to support the image component natively inside a Tiptap editor.

## Solution

**Breaking Changes**

**Features**:

- Add a new TiptapImage component that is functionally the same as a normal Image component, but can be rendered within a Prose component.


karrui (approved):

karrui:
actually, how would the image upload work? we just store in the DB? might not make sense right?

dcshzj:
> actually, how would the image upload work? we just store in the DB? might not make sense right?

Hmm we store the image in a separate assets bucket, then store the URL in the component. The final site will load from this external URL as if it is loading any other image.

PR #685, opened 2024-09-24
feat: add ability to add image in accordions

dcshzj:
## Problem

We are not able to add images inside accordion.

## Solution

**Breaking Changes**

**Features**:

- Add ability to add images inside accordion.

## Screenshots

image
image


sehyunidaaa:
How/when did we decide to do this?

Also, will users be able to edit the caption/size/alt text after they've inputted it?

dcshzj:
> How/when did we decide to do this? (Asking because on Monday's migration sync we agreed to look for other ways to present information, especially because the examples raised didn't fully justify the use of images in an accordion)

There might be some miscomm cos I might have interpreted that it is something that the migrators would need. If it's not required yet then we can drop this, no worries!

> 
> A few qns:
> 
> * Will users be able to edit the caption/size/alt text after they've inputted it?

Yup, the functionality is very similar to the existing link editor modal whereby:
1. The user will select the image on the Tiptap editor
2. The image icon will appear as active
3. The user can click on the image icon to edit attributes of the image node. Those field will be auto-populated inside the modal.

> * Will clicking on the `Upload image` button upload the image, or will upload happen as soon as you select an image on the attachment modal? If it's the previous, what happens if upload fails?

I went for a simpler approach here but it is possible to extend this further. Right now, the upload happens after clicking on the "Upload image" button at the bottom of the modal. There are two types of failures:
1. The image upload itself fails. In this case, the modal will not close and a box will appear at the bottom of the modal before the button (see screenshot below).
2. The image upload passes but the malware scanning fails. In this case, there is no feedback to the user (similar to the existing behaviour on complex blocks).

image

In the second case and in the happy path, the image on the Tiptap editor will first appear with the placeholder image. Once the malware scanning is complete, the image will appear on the next re-render (usually when they continue typing on the editor).

The extension to this approach is to prevent the modal from closing until the image itself is ready to be shown to the user. This is more complex because it involves the user's browser polling the image URL until the image becomes available (or after some x number of retries, whichever comes first). This would make the experience more complete for the user as we will be able to provide feedback for both types of failures.

dcshzj:
Not going to have an accordion inside the Tiptap editor and instead, will be going with creating a new class of components called wrapper components instead.

PR #686, opened 2024-09-25
fix: sanitize link href inputs in components

dcshzj:
## Problem

Fixes ISOM-1441.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Add sanitisation to all links before they are rendered to the user.


packages/components/src/templates/next/components/complex/Infopic/Infopic.tsx
```diff
@@ -69,7 +73,9 @@ export const Infopic = ({
           <div className={compoundStyles.button()}>
             <LinkButton
               LinkComponent={LinkComponent}
-              href={getReferenceLinkHref(buttonUrl, site.siteMap)}
+              href={getSanitizedLinkHref(
+                getReferenceLinkHref(buttonUrl, site.siteMap),
+              )}
```
karrui, line 78: since all `getReferenceLinkHref`s use `getSanitizedLinkHref`, should it be built into the function instead

dcshzj: Oh ya hor, fixed in 7a87b8606c09bc7f280cb9675120eb57a8b047b1.

packages/components/src/utils/getReferenceLinkHref.ts
```diff
@@ -1,15 +1,35 @@
...
+  if (url === undefined) {
     return undefined
   }
 
+  const dirty = document.createElement("a")
+  dirty.setAttribute("href", url)
+  dirty.setAttribute("id", ELEMENT_ID)
+  const clean = DOMPurify.sanitize(dirty, { RETURN_DOM_FRAGMENT: true })
+
+  const sanitizedUrl = clean.getElementById(ELEMENT_ID)?.getAttribute("href")
```
karrui, line 19: might be able to be simplified (without needing a separate `ELEMENT_ID` const):

```ts
const elem = document.createElement("a")
elem.setAttribute("href", url)
DOMPurify.sanitize(elem, { IN_PLACE: true })
const sanitizedUrl = elem.getAttribute("href")
```

packages/components/src/utils/getReferenceLinkHref.ts
```diff
@@ -31,3 +51,16 @@ export const getReferenceLinkHref = (
...
+
+export const getReferenceLinkHref = (
+  referenceLink: string | undefined,
+  sitemap: IsomerSitemap,
+) => {
+  if (!referenceLink) {
+    return undefined
+  }
+
+  const actualLink = getActualLinkFromReference(referenceLink, sitemap)
```
karrui, line 63: should we inline this function? feels unnecessary to extract it out when its only used in a single place

karrui (approved):
lgtm

packages/components/src/utils/getReferenceLinkHref.ts
dcshzj: Ah missed that option, updated in 11d37516985edebd25a933b0943a26e1b2686eaa.

PR #683, opened 2024-09-24
feat(resource): basic implementation and tests

seaerchin:
## Problem
this pr adds tests and logic (using casl) for basic resource permissions for editors and admins

## Solution
1. use `casl` to define permissions for admin and editors
    - for editors, they can perform every action on non-root pages but are unable to create/delete root pages. editors can stiil r/w on root pages 
    - for admins, they can perform read/writes on root pages.

## Notes
how this is implemented is non-overlapping, so at run time, we will construct the permission based on what access the user has to the resource.

PR #693, opened 2024-09-26
feat: add permissions checking on router

seaerchin:
## Problem
This PR adds some router changes to block creation on backend together with frontend related changes to remove the button when they are not permitted

## Solution
1. added a router change backend for `pages.router` so that we check the capability of the user to do the respective action
2. on frontend, we use the `Can` component from `casl` to do the respective action.
    - note that this is only done for the `/[siteId]` route **and not for individual folders** because we don't have root level permissions (ie, all permissions apply throughout to each folder). because we make no distinction on a per folder basis, all actions are permitted for folders. 

## Screenshots and videos

https://github.com/user-attachments/assets/105d83b5-7138-40cd-8eff-5ed4307f74ab


seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/693?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#693** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#683** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#681** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/studio/src/pages/sites/[siteId]/index.tsx
```diff
@@ -22,12 +26,22 @@ import { CreatePageModal } from "~/features/editing-experience/components/Create
...
 })
 
+const getPermissions = (roles: { role: RoleType }[]) => {
+  const builder = new AbilityBuilder<ResourceAbility>(createMongoAbility)
+  roles.forEach(({ role }) => {
+    buildPermissionsFor(role, builder)
+  })
+  return builder.build({ detectSubjectType: () => "Resource" })
+}
+
```
Comment (seaerchin) [inline apps/studio/src/pages/sites/[siteId]/index.tsx:44]: from karrui: we might be able to jsut return the `ability` straight from backend and skip all these hassle

Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/index.tsx:44]: we can try `pack/unpackRules`, i only tried sending `rules` across but faced difficulties

apps/studio/src/server/modules/page/page.router.ts
seaerchin, line 1: think about how to interact with pagination for permissions because eg: for `list`, we only check that you have perms to a certain resource

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -36,9 +41,40 @@ import { createDefaultPage } from "./page.service"
 const ajv = new Ajv({ allErrors: true, strict: false, logger: false })
 const schemaValidator = ajv.compile<IsomerSchema>(schema)
 
+const validateUserPermissions = async ({
```
seaerchin, line 44: from karrui: we can consider consolidating all the resources that the user has access to, and then set the ability itself on the context. 

the benefit is that the callsite is simpler + the logic is consolidated into one place

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -73,7 +109,13 @@ export const pageRouter = router({
...
       }),
     )
-    .query(async ({ input: { siteId, resourceId } }) => {
+    .query(async ({ ctx, input: { siteId, resourceId } }) => {
+      await validateUserPermissions({
+        userId: ctx.user.id,
+        siteId,
+        action: "read",
+      })
+
```
seaerchin, line 118: next time with root permissions, we might not be able to do this chekc. this check now works by assuming that `resourceId` is `null` (and hence can check for site wide permissions)

apps/studio/src/pages/sites/[siteId]/index.tsx
```diff
@@ -22,12 +26,22 @@ import { CreatePageModal } from "~/features/editing-experience/components/Create
...
 import { type NextPageWithLayout } from "~/lib/types"
+import { buildPermissionsFor } from "~/server/modules/permissions/permissions.util"
 import { AdminCmsSidebarLayout } from "~/templates/layouts/AdminCmsSidebarLayout"
+import { trpc } from "~/utils/trpc"
 
 export const sitePageSchema = z.object({
   siteId: z.coerce.number(),
 })
 
+const getPermissions = (roles: { role: RoleType }[]) => {
```
Comment (harishv7) [inline apps/studio/src/pages/sites/[siteId]/index.tsx:37]: why are we doing this logic on FE ah?

apps/studio/src/pages/sites/[siteId]/index.tsx
```diff
@@ -22,12 +26,22 @@ import { CreatePageModal } from "~/features/editing-experience/components/Create
 import { MoveResourceModal } from "~/features/editing-experience/components/MoveResourceModal"
 import { useQueryParse } from "~/hooks/useQueryParse"
 import { type NextPageWithLayout } from "~/lib/types"
+import { buildPermissionsFor } from "~/server/modules/permissions/permissions.util"
```
Comment (harishv7) [inline apps/studio/src/pages/sites/[siteId]/index.tsx:29]: feels weird to import a server module on FE

rather BE should be returning a fully-formed response than FE doing this

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -264,7 +340,13 @@ export const pageRouter = router({
 
   getRootPage: protectedProcedure
     .input(getRootPageSchema)
-    .query(async ({ input: { siteId } }) => {
+    .query(async ({ ctx, input: { siteId } }) => {
+      await validateUserPermissions({
```
harishv7, line 344: can this authorization be a middleware? if we have a map of the procedures to the required perms?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -264,7 +340,13 @@ export const pageRouter = router({
...
+        userId: ctx.user.id,
+        siteId,
```
harishv7, line 346: Should we inject siteid into the session context when users login and pick it up from there?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -307,86 +390,105 @@ export const pageRouter = router({
...
+              .returning([
+                "Resource.id",
+                "Resource.type",
+                "Resource.title",
+                "Resource.permalink",
+                "Resource.draftBlobId",
+              ])
+              .executeTakeFirstOrThrow()
+              .catch((err) => {
+                if (get(err, "code") === "23505") {
```
harishv7, line 457: what's this code? is it defined somewhere as a constant?

apps/studio/src/server/modules/permissions/permissions.type.ts
```diff
@@ -0,0 +1,20 @@
+import type { PureAbility } from "@casl/ability"
+
+import type { Resource as RawResource } from "../database"
+
+type Resource = Pick<RawResource, "parentId">
+
+type AllowedResourceActions = (typeof ALL_ACTIONS)[number]
+export type CrudResourceActions = (typeof CRUD_ACTIONS)[number]
```
harishv7, line 8: nit: can define these 2 after the respective declarations

apps/studio/src/pages/sites/[siteId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/index.tsx:37]: this is just the builder! we need it because i tried sending the `builder.rules` across but it didn't work. i'll try the `pack/unpack` method in a follow-up

apps/studio/src/pages/sites/[siteId]/index.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/index.tsx:29]: this is technically a util method - i agree iwth you and i tried sending a json of the `builder.rules` but i ran into type issues generating the ability and went down this route.

if it's the import, i can shift it into a top-level `/utils/permissions` file? i think the important thing for me is using a shared method to generate the ability.

apps/studio/src/server/modules/permissions/permissions.type.ts
seaerchin: sgtm!

apps/studio/src/server/modules/page/page.router.ts
seaerchin: there were a few issues i faced with this (the middleware approach) - we need to tell the middleware _ahead of time_ what permissions are required (`read/create` etc) and also, the middleware requires a few things on the `input` itself (`siteId, resourceId, userId`) which is abit hard because the middleware will impact the downstream typing 

i think what we needed here is a [standalone middleware](https://trpc.io/docs/v10/server/middlewares#experimental-standalone-middlewares) but i didn't do it here cos it's marked as experimental. 

karrui pointed out that we can just use it first (and `starter-kit` does use `unstable_pipe`) but i'm of the opinion that we should get it working w tests first before shifting! wdyt?

apps/studio/src/server/modules/page/page.router.ts
seaerchin: we _can't_ do this at the point of login because a user can have access to >1 site, which is also why the query itself requires a `siteId`.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -307,86 +390,105 @@ export const pageRouter = router({
...
-    .mutation(async ({ input: { pageId, siteId, title, permalink, meta } }) => {
-      return db.transaction().execute(async (tx) => {
-        const fullPage = await getFullPageById(tx, {
-          resourceId: pageId,
+    .mutation(
+      async ({ ctx, input: { pageId, siteId, title, permalink, meta } }) => {
+        await validateUserPermissions({
+          userId: ctx.user.id,
           siteId,
+          action: "update",
```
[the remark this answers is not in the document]
seaerchin: yeap!

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -30,11 +36,112 @@ import { getSiteConfig, getSiteNameAndCodeBuildId } from "../site/site.service"
 import { incrementVersion } from "../version/version.service"
 import { createDefaultPage } from "./page.service"
 
+const ajv = new Ajv({ allErrors: true, strict: false, logger: false })
```
dcshzj, line 39: issue: suspect there might be some rebasing issue here. This `ajv` should be conflicting with the `ajv` import above at line 20.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -30,11 +36,112 @@ import { getSiteConfig, getSiteNameAndCodeBuildId } from "../site/site.service"
...
   readPage: protectedProcedure
     .input(basePageSchema)
-    .output(readPageOutputSchema)
-    .query(async ({ input: { pageId, siteId } }) => {
+    .query(async ({ ctx, input: { pageId, siteId } }) => {
+      await validateUserPermissions({
+        userId: ctx.user.id,
+        siteId,
+        action: "read",
+      })
```
dcshzj, line 143: thought(non-blocking): actually why do we not want to just pass in the pageId/resourceId here, and leave it to just `validateUserPermissions` to determine if it wants to make use of the information?

Not a big deal tbh so don't need to make changes at least for launch, just maybe wondering if I should adjust my thought process a little bit here regarding the permissions model.

apps/studio/src/server/modules/page/page.router.ts
seaerchin: yea i fixe d it, was planning to check psot rebase :sadge:

dcshzj (approved):

seaerchin:
### Merge activity

* **Oct 22, 1:43 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/693).

PR #697, opened 2024-09-26
fix(rate-limit): add rate limiting and update otp logic

seaerchin:
## Problem
we currently do not have rate limiting and users are allowed to ddos each other. 

## Solution
1. add rate limiting. this is heavily taken from activesg's rate limiter middleware with minor changes.
2. do not create a new otp if there are already entries in the otp table for the existing user. we will instead reuse the old otp and hint to the user to use the previous one (identified via the prefix). to facilitate this, we will also add a new column in the `VerificationToken` table for the `prefix`


.env.test
```diff
@@ -1,7 +1,7 @@
 # Specific environment variables when running tests
 # Make sure to sync this file with turbo.json so that CI can pass
 NODE_ENV=test
-DATABASE_URL=postgres://root:root@localhost:5432/test
+DATABASE_URL=postgres://root:root@localhost:5432/app
```
harishv7, line 4: was this change intentional?

apps/studio/prisma/schema.prisma
```diff
@@ -30,6 +30,7 @@ generator kysely {
 model VerificationToken {
   identifier String   @id
   token      String
+  prefix     String
```
harishv7, line 33: are these the alphabets in front of the OTP numbers?

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -45,37 +46,83 @@
...
+                prefix: otpPrefix,
+                expires,
+              })
+              .execute()
+
+            return {
+              savedPrefix: otpPrefix,
+              newlyCreated: true,
+            }
+          } else if (existing.expires < new Date(Date.now())) {
```
harishv7, line 76: Should we have a buffer for this? Should be at least valid for 30 seconds? Else by the time the user checks and inputs it's expired

apps/studio/src/server/modules/rate-limit/rate-limit.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  const windowMs = rateLimitOptions.windowMs ?? 1000
+
+  const store = new RateLimiterPrisma({
+    storeClient: prisma,
+    points: max,
+    duration: windowMs / 1000, // in seconds
+    insuranceLimiter: rateLimiterMemory,
+  })
+
+  const fingerprint = getRateLimitFingerprint({ type: "IP", req })
```
harishv7, line 38: whats this fingerprint?

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -45,37 +46,83 @@ export const emailSessionRouter = router({
...
       <p>If your OTP does not work, please request for a new one.</p>`,
           recipient: email,
-        }),
-      ])
-      return { email, otpPrefix }
+        })
+      } else {
+        await sendMail({
+          subject: `Sign in to ${url.host}`,
+          body: `Please use your existing OTP beginning with ${savedPrefix} to login to your account.
```
karrui, line 116: What if the first email does not get sent due to some issue? Does that mean they need to wait out 10 minutes for the token to expire before getting a chance to get another email with the token?

apps/studio/src/server/modules/rate-limit/rate-limit.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  prisma,
+}: {
+  rateLimitOptions: RateLimitMetaOptions
+  req: NextApiRequest
+  prisma: PrismaClient
+}) {
+  const max = rateLimitOptions.max ?? 5
+  const windowMs = rateLimitOptions.windowMs ?? 1000
+
+  const store = new RateLimiterPrisma({
```
karrui, line 31: i think we can store this in memory; don't need to put in the DB. See morpho codebase!

karrui (commented):
might want to rethink our OTP approach. Rate limiting might be "good enough"

apps/studio/src/server/modules/auth/auth.service.ts
```diff
@@ -30,10 +30,6 @@ export const verifyToken = async (
       throw new VerificationError("Token is invalid or has expired")
     }
 
-    if (verificationToken.attempts > 5) {
```
karrui, line 33: this one also cannot kill. will get new vuln due to bruteforcing haha

seaerchin: This got removed cos it's redundant, ther s an identical block above oddly enough HAHA

.env.test
seaerchin: Yeap! Oddly enough, it was trying to connect to `test` and failing. This change made it run on local but I think this shouldn't be needed because we mock prisma w pglite. Need to investigate more

apps/studio/prisma/schema.prisma
seaerchin: Clarifying here, are you suggesting that we store the unhashed OTP or the hashed  version? I didn't want to store the unhashed version, hence I stored the prefix as an identifier

apps/studio/src/server/modules/rate-limit/rate-limit.service.ts
seaerchin: Key to store in our table for this particular rate limit

apps/studio/prisma/schema.prisma
karrui: i'm saying maybe accept the DOS risk and just keep old behavior. Just add rate limiting to mitigate DOS ability

apps/studio/src/server/modules/rate-limit/rate-limit.service.ts
seaerchin: i took this with reference to morpho's codebase - in there, they used a redis instance, which is also shared memory between their compute, with the insurance limiter being in memory.

i think a shared storage is needed because we want to avoid the case where different requests go to different instances and lead to different attempts being logged in the memory of each instance.

for example, a user attempts to verify their otp and it goes to instance A. it fails and they retry, causing the second attempt to go to instance B. in this case, if we use a memory backed rate limiter, the isntances would each have count: 1 rather than the actual count of 2

apps/studio/prisma/schema.prisma
seaerchin: i'm ok with this. in this case, i'll revert the changes in my last commit. under this model, it's possible for a malicious actor to prevent a good user from logging in purely by persistently re-requesting otp. if we're ok with this trade-off, i'll make the changes

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -45,37 +46,83 @@
...
           )}.
       Please use this to login to your account.
       <p>If your OTP does not work, please request for a new one.</p>`,
           recipient: email,
-        }),
-      ])
-      return { email, otpPrefix }
+        })
+      } else {
+        await sendMail({
```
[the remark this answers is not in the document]
seaerchin: yeap! lmk if this is too generous, we can always change

seaerchin:
@karrui and @harishv7 i removed the email otp changes and only added the rate limiter in. i'll do the otp fixes in a follow up

apps/studio/src/server/modules/auth/email/email.router.ts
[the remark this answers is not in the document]
harishv7: sounds okay to me, if rate limit is via IP+email

apps/studio/src/server/modules/rate-limit/rate-limit.service.ts
harishv7: yes needs a shared instance - hence a cache or DB is required

karrui (commented):
the prisma schema changes seems to have been discarded. Will need it for rate-limiter-flexible to work yeah?

seaerchin:
> the prisma schema changes seems to have been discarded. Will need it for rate-limiter-flexible to work yeah?

oops, thanks for the catch!

apps/studio/prisma/generated/generatedTypes.ts
```diff
@@ -8,7 +8,7 @@ export type Generated<T> =
     : ColumnType<T, T | undefined, T>
 export type Timestamp = ColumnType<Date, Date | string, Date | string>
 
-export interface Blob {
+export type Blob = {
```
harishv7, line 11: Why do we always have this conflict? Sometimes it generates interfaces, sometimes types

seaerchin: i honestly have no clue

harishv7 (approved):
lgtm, pls resolve threads and comments before merge

apps/studio/src/server/modules/auth/email/email.router.ts
seaerchin: removed this

apps/studio/src/server/modules/auth/email/email.router.ts
seaerchin: removed this change

PR #699, opened 2024-09-27
fix(image-upload): update image upload validator

seaerchin:
## Problem
right now, our validation is too loose for the filename of the uploaded images. this leads to issues where the image can be `../../image`, leading to  issues with url traversal, where an attacker can replace a valid asset

## Solution
1. stricten the initial validation for the presigned url validation.
2. add this new validation to our image component. because we always upload to s3, we can safely place this in the json forms attachment componnet. add the validator and feedback via a toast the error message

## Screenshots

https://github.com/user-attachments/assets/75203107-51a3-46fe-879e-59455c77642e


karrui (approved):
lgtm

PR #700, opened 2024-09-29
feat(component): Table enhancements and border fixes

sehyunidaaa:
## Problem

- Making last column fill up the width of a table was an assumption that was not applicable to many use cases we saw; we decided to remove this feature and let site admins decide the width of a column on their own
- When multiple rows were spanned, the bottom border would disappear; example on: https://staging.dorwsv8lwiupx.amplifyapp.com/lifestage/5-seniors/

Closes [ISOM-1594]

## Solution

- Added border to table
- Removed max-w-full on the last column of a table


karrui (approved):

karrui:
To note that admins "deciding the width on their own" is just them merging/dividing cells in the tiptap editor

PR #702, opened 2024-09-29
fix(component): add margin-top to iframe/embeds when it's not a first child

sehyunidaaa:
## Problem

Embed/iframes don't have a margin-top when placed into pages, so it sticks to the element previous like the following: 
image

Closes [ISOM-1595]

## Solution

- Add top margin (28px, standard that we use for most elements) when the iframe is not a first child
- Added an iframe to the Content page story to test it out

## After

image

Mobile:
image

### DIFF
--- a/packages/components/src/templates/next/components/complex/Iframe/Iframe.tsx
+++ b/packages/components/src/templates/next/components/complex/Iframe/Iframe.tsx
@@ -30,7 +30,7 @@ const Iframe = ({ title, content }: IframeProps) => {
   const iframeUrl = sanitizedIframe.getAttribute("src")
 
   return (
-    <section className={ComponentContent}>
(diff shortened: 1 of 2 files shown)


karrui (approved):

sehyunidaaa:
Added delay to Chromatic, re-requesting review! Will check Chromatic myself once build is done

sehyunidaaa:
@karrui hey, I added the delay but chromatic snapshots are still not showing the iframes. Is this okay?

karrui:
> @karrui hey, I added the delay but chromatic snapshots are still not showing the iframes. Is this okay?

Might just be inherent in how they snapshot then. Guess we can close one eye unless it starts to flake! Probably can remove the delay if it doesn't help

sehyunidaaa:
@karrui okie! i'm thinking of just removing the embed from the story because it only validates whether the embed as a top margin. we don't need to have every block in a layout.

PR #705, opened 2024-09-30
chore: increase accordion title content length

dcshzj:
## Problem

The accordion title content length of 100 characters is too limiting.

## Solution

**Breaking Changes**

**Improvements**:

- Increase the max length limit from 100 to 300 characters.


dcshzj:
> lgtm. Should we add a story on how it would look like for that many characters?

Yeah we have one under the Long Content story.

PR #709, opened 2024-09-30
feat: implement nanoid for User model and adjust User schema

karrui:
## Solution

**Breaking Changes**

  - The `User` model has been modified, removing the `preferredName` field and making the `phone` field optional

**Features**:

- Added `nanoid()` and `nanoid_optimized()` functions to generate compact, URL-friendly unique identifiers in Postgres

**Improvements**:

- Updated the `User` model to use `nanoid()` for ID generation instead of `cuid()`
- Removed the `@paralleldrive/cuid2` dependency

**Bug Fixes**:

- None

## Before & After Screenshots

N/A

## Tests

- Run database migrations to apply the new schema changes
- Test user creation and retrieval to ensure the new ID generation works correctly
- Verify that existing user-related functionality still works with the updated `User` model

**New scripts**:

None

**New dependencies**:

None

**New dev dependencies**:

None

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -66,7 +66,6 @@
     "@opengovsg/isomer-components": "*",
     "@opengovsg/sgid-client": "^2.2.0",
     "@opengovsg/starter-kitty-validators": "^1.1.0",
-    "@paralleldrive/cuid2": "^2.2.2",
     "@prisma/client": "5.10.2",
     "@sendgrid/mail": "^8.1.3",
     "@tanstack/match-sorter-utils": "^8.15.1",
--- a/apps/studio/prisma/generated/generatedTypes.ts
+++ b/apps/studio/prisma/generated/generatedTypes.ts
@@ -85,7 +85,7 @@ export interface SiteMember {
   updatedAt: Generated<Timestamp>
 }
 export interface User {
-  id: string
+  id: GeneratedAlways<string>
   name: string
   email: string
   phone: string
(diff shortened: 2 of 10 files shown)


karrui:
* **#709** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

seaerchin (commented):
why use `nanoid` over `uuidv4`, which is already [implemented](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#uuid) in prisma ah?

seaerchin:
based off what i found, the main advantage is size? but not too sure how important this is .-.

image

karrui:
> why use `nanoid` over `uuidv4`, which is already [implemented](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#uuid) in prisma ah?

Because we are moving away from prisma right? Prisma's `uuid()` function is app-level, means if we use kysely it won't actually be generated. This PR adds in a DB-level generation

karrui:
but actually i think i'll just use pg's `gen_random_uuid` instead of mucking around with a custom extension haha. let me update this PR

karrui:
actually dropping it, sian now i feel the pain of using pglite. When need to add extensions you cry. They don't support pgcrypto yet. let me just update user schema to not have the fields that we don't expect.

PR #713, opened 2024-09-30
feat: introduce database content layout

dcshzj:
> [!NOTE]
> 
> I haven't test this fully yet!

## Problem

Closes ISOM-1592.

## Solution

**Breaking Changes**

**Features**:

- Introduce a new SearchableTable component that creates a table that comes with a search filter and pagination controls.
- Introduce a new database layout which locks the SearchableTable component at the bottom of the page.

## Screenshots

image


packages/components/src/templates/next/components/internal/SearchableTable/SearchableTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+  )
+  const filteredItems = getFilteredItems({
+    items,
+    searchValue: search,
+  })
+  const paginatedItems = getPaginatedItems({
+    items: filteredItems,
+    currPage,
+    itemsPerPage: PAGINATION_MAX_ITEMS,
+  })
```
karrui, line 75: memo!! seems heavy to fetch every render (especially when the search value changes)

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+        emptyStateTitle: "text-base-content-subtle",
+      },
+    },
+  },
+})
+
+const compoundStyles = createSearchableTableStyles()
+
+const SearchableTable = ({ title, headers, items }: SearchableTableProps) => {
+  const [search, setSearch] = useState("")
```
karrui, line 58: might want to use `useDeferredValue` for filtering items for better performance.

See https://react.dev/reference/react/useDeferredValue

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+  })
+
+  const isInitiallyEmpty = items.length === 0 || maxNoOfColumns === 0
+  const isFilteredEmpty = items.length !== 0 && filteredItems.length === 0
+
+  const sectionTopRef = useRef<HTMLDivElement>(null)
+  const onPageChange = () => {
+    sectionTopRef.current?.scrollIntoView({
+      block: "start",
+      behavior: "smooth",
```
karrui, line 84: nit(a11y): might want to change the behaviour depending on user's `prefers-reduced-motion` setting; should be instant if true.

Can probably use a [hook](https://github.com/anatoliygatt/use-prefers-reduced-motion/blob/master/src/usePrefersReducedMotion.ts) to retrieve that value

[Further reading ](https://www.joshwcomeau.com/react/prefers-reduced-motion/)(or use his [hook](https://www.joshwcomeau.com/snippets/react-hooks/use-prefers-reduced-motion/) if you want, but feels old lol)

packages/components/src/templates/next/components/internal/SearchableTable/constants.ts
```diff
@@ -0,0 +1,6 @@
+// This is the maximum number of columns that the table can have
+// 10 was decided because at 1240px, each column will only have 124px which is
+// the minimum width for a column to be readable
+export const MAX_NUMBER_OF_COLUMNS = 10
```
karrui, line 4: in mobile will still look weird right haha. Might as well set min-width on each cell. Can always enhance it (and our other tables?) with something like https://www.npmjs.com/package/react-table-column-resizer (or write our own/vendorize) to allow users to resize columns.

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
```diff
@@ -0,0 +1,15 @@
...
+  items: SearchableTableProps["items"]
+  searchValue: string
+}
+
+export const getFilteredItems = ({
+  items,
+  searchValue,
+}: GetFilteredItemsParams) => {
+  return items.filter((item) =>
+    item.join(" ").toLowerCase().includes(searchValue.toLowerCase()),
```
karrui, line 13: nit/perf: do we want to precache each "row" before searching them to make searching slightly more performant?
I'd think the items will never change since it is static

```ts
const cache = items.map(item => ({ ...item, key: item.join(" ").toLowerCase() }))

const getFilteredItems = ({cache, searchValue}) => {
  return cache.filter(item => item.key.includes(searchValue.toLowerCase()))
}
```

karrui (commented):
initial review.

should the searchtable also be useable on other layouts (i.e. is database layout necessary)?

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTable.tsx
dcshzj: TIL, updated in 44298b012cbed4c09b6d65838834415abe98f1f6.

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTable.tsx
dcshzj: Removed this, want to standardise this with the behaviour currently in the collections layout.

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
dcshzj: Split into server and client parts, with the server part doing this pre-cache step in 44298b012cbed4c09b6d65838834415abe98f1f6.

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTable.tsx
dcshzj: Updated in 44298b012cbed4c09b6d65838834415abe98f1f6.

packages/components/src/templates/next/components/internal/SearchableTable/constants.ts
sehyunidaaa: If we set min-width (e.g., 124px) on the cols, will a table with 10 cols overflow and hori scroll on w=1025px (desktop)?

dcshzj: I think with the borders likely will overflow, so maybe we can use a smaller number such that the min-width of all columns + borders is <= 1024px?

sehyunidaaa: Hmm I don't think overflow is necessarily bad when you have 10 (!!) columns, because at that point your data table is so big, that it's more important for us to guarantee that the column texts are readable than trying to avoid overflow. I'd rather the table be scrollable than we try to squeeze 10 columns into {1024px - margins} which might break words at awkward places.

I guess no issue in setting a min-width on columns, but should we be setting a max-width on the entire table instead? So we let the columns resize depending on content but prevent the scroll being ridiculously long, like 2* of the screen width

sehyunidaaa:
@dcshzj sorry to bother you again! just fyi got some small items pending on chromatic, after you push again please let me know i'll approve chromatic

packages/components/src/templates/next/components/internal/SearchableTable/CellContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+  sitemap,
+  LinkComponent,
+}: CellContentProps) => {
+  if (
+    typeof content === "string" &&
+    content.startsWith(HYPERLINK_EXCEL_FUNCTION) &&
+    content.endsWith(")")
+  ) {
+    const link = content.slice(HYPERLINK_EXCEL_FUNCTION.length, -1)
+    const [linkHref, linkText] = link.split(",")
```
seaerchin, line 24: this is abit confusing to me - i'm not really clear what `link` here refers to! on initial glance, i assumed that this wuold be `www.isomer.gov.sg` (for example), so i didn't quite understand how the `,` got in there

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
```diff
@@ -0,0 +1,221 @@
...
+  LinkComponent,
+}: SearchableTableClientProps) => {
+  const [_search, setSearch] = useState("")
+  const search = useDeferredValue(_search)
+  const [currPage, setCurrPage] = useState(1)
+  const titleId = useId()
+
+  const maxNoOfColumns = Math.min(
+    headers.length,
+    ...items.map((row) => row.row.length),
```
seaerchin, line 80: what happens if we have a case where the header length is shorter than the max row length? do we have to handle overflow? (never check)

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
```diff
@@ -0,0 +1,221 @@
...
+      getPaginatedItems({
+        items: filteredItems,
+        currPage,
+        itemsPerPage: PAGINATION_MAX_ITEMS,
+      }),
+    [currPage, filteredItems],
+  )
+
+  const isInitiallyEmpty = items.length === 0 || maxNoOfColumns === 0
+  const isFilteredEmpty = items.length !== 0 && filteredItems.length === 0
```
seaerchin, line 103: nit: should we be using `!!search` rather than `items.length`?

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  const sectionTopRef = useRef<HTMLDivElement>(null)
+  const onPageChange = () => {
+    sectionTopRef.current?.scrollIntoView({
+      block: "start",
+    })
+  }
```
seaerchin, line 110: i think the scroll behaviour defaults to `auto` - not sure if this shuold be a smooth scroll (again, never tried this on local)

seaerchin: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  return (
+    <div className={compoundStyles.container()} ref={sectionTopRef}>
+      {title && (
+        <h2 id={titleId} className={compoundStyles.title()}>
```
seaerchin, line 115: (this is purely cos i'm curious) - why does this `h2` need an id ah

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
```diff
@@ -0,0 +1,221 @@
...
+          {title}
+        </h2>
+      )}
+
+      <div className={compoundStyles.search()}>
+        <SearchField
```
seaerchin, line 121: can consider adding [aria role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role)

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
```diff
@@ -0,0 +1,14 @@
...
+  items: SearchableTableClientProps["items"]
+  searchValue: string
+}
+
+export const getFilteredItems = ({
+  items,
+  searchValue,
+}: GetFilteredItemsParams) =>
+  items
+    .filter((item) => item.key.includes(searchValue.toLowerCase()))
```
seaerchin, line 13: if we want a case insensitive search, might as well lowercase the key!

packages/components/src/utils/getTransformedPageContent.ts
```diff
@@ -0,0 +1,41 @@
+import type { IsomerComponent } from "~/types"
+import { getDigestFromText } from "./getDigestFromText"
+import { getRandomNumberBetIntervals } from "./getRandomNumber"
+
+// if block.id is not present for heading level 2, we auto-generate one
+// for use in table of contents anchor links
+export const getTransformedPageContent = (content: IsomerComponent[]) => {
```
seaerchin, line 7: food for thought: this seems to be catching hte case for prose blocks w/ content -> h2 that lack id. we can also add on studio side since tiptap can guarantee this property so that we can treat these similarly - wdyt?

seaerchin (commented):
first pass review! some questoins on my end but overall seems ok. what's the point of the sentinel `=HYPERLINK(`? are we going to use it internally for github stored pages?

also, do we need an equivalent on studio for the layotu here?

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
seaerchin: saw kr's earlier comment, resolving

packages/components/src/templates/next/components/internal/SearchableTable/CellContent.tsx
dcshzj: This is an Excel-style of hyperlinks: https://support.microsoft.com/en-us/office/hyperlink-function-333c7ce6-c5ae-4164-9c47-7de9b76f577f

So to add a link, the cell contents will be like `=HYPERLINK(https://www.isomer.gov.sg,Isomer)`, and it will be rendered as [Isomer](https://www.isomer.gov.sg).

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
dcshzj: This checks every row and finds the minimum number of columns. So if we have header row with 6 columns, 1st row with 6 columns and 2nd row with 3 columns, the max we should display is 3.

Overflow is handled automatically via scroll.

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
dcshzj: Hmm but if the table is initially empty, then using the search value means the user types in a search value will display as "no results found" but actually there is no results at the beginning.

Was thinking of sticking to displaying the message that there is no results in the beginning, regardless of the inputs by the user.

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
dcshzj: Hmm SearchField is a react-aria component so it comes with the aria role attached.

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
dcshzj: Hmm the key is lowercased in the server component so the client does lesser stuff: https://github.com/opengovsg/isomer/pull/713/files#diff-5c209342b9152356c25bfdfc004d41eeb44c795172426b1bc17569d3b0e31bdb

packages/components/src/utils/getTransformedPageContent.ts
dcshzj: Hmm I was thinking of not letting studio handle this, because design is thinking of expanding this table of contents to also capture headings from other complex blocks. This guarantee is more important for components rather than on studio hence this check is here.

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
dcshzj: Ah this is for aria, there is an `aria-describedby` below.

Also realised that it might be undefined, updated in 0b847a2504f7a3830e4f8dae996146dc0cc572af.

dcshzj:
> first pass review! some questoins on my end but overall seems ok. what's the point of the sentinel =HYPERLINK(? are we going to use it internally for github stored pages?

This is to support links. The intent is for people to create their table in Excel, export as CSV then upload to Studio in the future.

> also, do we need an equivalent on studio for the layotu here?

Nope, editing on Studio for this layout is deprioritised for now, cos quite complicated to implement.

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
seaerchin: ah ok, so to double check what happens here - we only display 3 and there's a scrollbar allowing us to horizontally scroll and see the other 6? ok with this, i wanted to double check that we're not just removing the content

packages/components/src/templates/next/components/internal/SearchableTable/SearchableTableClient.tsx
seaerchin: yea i think both are fine! was just thinking that having a search term means that we shuold reflect that, regardless of collection state. this is very minor though, happy to take either

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
seaerchin: not very good to assume here, we should just lowercase if it's cheap to do so. i think the runtime here won't be anyting significant

packages/components/src/utils/getTransformedPageContent.ts
seaerchin: ok, if this is for future proofing then sgtm!

seaerchin:
the logic here doesn't quite work! this is because we parse the text **after** we cache it. because of this, a false search of `hyperlink` will cause the result involving the link to show up!

image

in order to circumvent this, we should probably parse first prior to doing the cache construction so that we get a repsentation of the textual content.

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
dcshzj: Hmm this is done as part of a performance improvement suggested here: https://github.com/opengovsg/isomer/pull/713#discussion_r1782061929

I'm more inclined to reduce the work that the client does, especially since our users may have devices that are extremely slow.

dcshzj:
> the logic here doesn't quite work! this is because we parse the text **after** we cache it. because of this, a false search of `hyperlink` will cause the result involving the link to show up!

Updated in 79a1f8135b6ffd8cb295ceb598f835b7ae10d64e.

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
seaerchin: ok! can add a comment on the studio side then, so people are aware of this!

seaerchin (approved):

packages/components/src/templates/next/components/internal/SearchableTable/getFilteredItems.ts
dcshzj: Eh hmm there's no studio equivalent here, so the input is fine but on build time the key is generated with the lowercase.

PR #718, opened 2024-10-02
chore: exclude components package from git

dcshzj:
## Problem

We keep accidentally committing this package into the repository.

## Solution

**Breaking Changes**

**Improvements**:

- Add this package into gitignore


seaerchin (approved):
actually can we do `**/*.tgz` so that we ignore all artifacts?

dcshzj:
> actually can we do `**/*.tgz` so that we ignore all artifacts?

e747b27554e184704769eaa0f1fb0771816a64c5

PR #719, opened 2024-10-02
fix: standardise date formats displayed on site

dcshzj:
## Problem

The way we display dates on the site is very different, and can lead to date parsing issues.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Standardise the way we display dates on the site, and provide a range of date formats to parse the date string using dayjs.

### DIFF
--- a/packages/components/src/utils/getFormattedDate.ts
+++ b/packages/components/src/utils/getFormattedDate.ts
@@ -1,7 +1,13 @@
-export const getFormattedDate = (date: string) => {
-  return new Date(date).toLocaleDateString(undefined, {
-    day: "2-digit",
-    month: "short",
-    year: "numeric",
-  })
-}
+import dayjs from "dayjs"
+import customParseFormat from "dayjs/plugin/customParseFormat"
+
+dayjs.extend(customParseFormat)
+
+// Standardise the format of dates displayed on the site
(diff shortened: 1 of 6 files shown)


packages/components/src/utils/getFormattedDate.ts
```diff
@@ -1,7 +1,13 @@
...
-    year: "numeric",
-  })
-}
+import dayjs from "dayjs"
+import customParseFormat from "dayjs/plugin/customParseFormat"
+
+dayjs.extend(customParseFormat)
+
+// Standardise the format of dates displayed on the site
+export const getFormattedDate = (date: string) =>
```
seaerchin, line 7: what's the string format that we expect here btw!

seaerchin (approved):
in here, we've specified a few formats for dates, but i'm wondering fi we should similarly save dates to our db in a specific format so that it's always guaranteed. 

right now, if i save as (for example), DD/MM/YY, it might not work for this display ._.

dcshzj:
> in here, we've specified a few formats for dates, but i'm wondering fi we should similarly save dates to our db in a specific format so that it's always guaranteed.

Yeah we will need to standardise on Studio's end as well but I think we haven't done any standardisations yet. Doing this for now so we can have some baseline on the template's end first.

PR #720, opened 2024-10-02
fix: page router validations and add test cases

karrui:
This PR starts the chain of PRs for adding test cases for all our procedures in the Studio application so we can be more confident when shipping code.

## Solution

**Breaking Changes**

**Features**:

- Details ...

**Improvements**:

- Add Tagged types to `jsonb` columns so we remember to call `jsonb()` function before insertion into the database. This prevents runtime errors.
- Update `updatePageBlobSchema` to validate the type of `content` in the schema itself.
- Update all `*select` constants to use `satisfies` keyword for better type inference when used.
- Throw 404 errors when resource is not found in all `page.router` procedures
- Validate whether `folderId` is a folder when calling `createPage` procedure and passing the `folderId` arg.
- use transaction when updating page in `updatePage` procedure
- add foreign key violation error handling

**Bug Fixes**:

- use correct `fullPage.publishedVersionId` check in `readPageAndBlob` procedure

## Tests

Tests are added for all procedures in the page router.

A few tests regarding permissions are skipped until permissions are in cc @seaerchin


karrui:
* **#720** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/720?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/studio/src/schemas/page.ts
```diff
@@ -44,17 +51,19 @@ export const reorderBlobSchema = z.object({
...
-  content: z.string(),
+  content: z.string().transform((value, ctx) => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const parsed = safeJsonParse(value)
+    if (schemaValidator(parsed)) {
+      return parsed
+    }
+    ctx.addIssue({
+      code: z.ZodIssueCode.custom,
+      message: "Invalid page content",
```
harishv7, line 63: we are trying to abort early right, do we need the FATAL: true flag?

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -38,7 +38,7 @@ export const setSiteConfig = async (
 ) => {
   return db
     .updateTable("Site")
-    .set({ config })
+    .set({ config: jsonb(config) })
```
harishv7, line 41: previously when we didn't cast it as jsonb it still worked right? how does the cast help now?

tooling/eslint/base.js
```diff
@@ -87,6 +87,7 @@ export default tseslint.config(
     files: ["**/*.spec.ts", "**/*.spec.tsx", "**/*.test.ts", "**/*.test.tsx"],
     rules: {
       "@typescript-eslint/no-unsafe-assignment": "off",
+      "@typescript-eslint/no-empty-function": "off",
```
harishv7, line 90: why do we allow empty functions? do the tests require this?

apps/studio/src/utils/ajv.ts
```diff
@@ -0,0 +1,3 @@
+import Ajv from "ajv"
+
+export const ajv = new Ajv({ allErrors: true, strict: false, logger: false })
```
harishv7, line 3: maybe not related to this PR, why do we use strict false?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -134,7 +140,8 @@ export const getPageById = (
       ]),
     )
     .select(defaultResourceSelect)
-    .executeTakeFirstOrThrow()
+    .executeTakeFirst()
```
harishv7, line 143: why do we not opt to throw here? wouldn't this be a 404?

apps/studio/tests/integration/helpers/seed/index.ts
```diff
@@ -0,0 +1,264 @@
+import { db, jsonb } from "~server/db"
+import { nanoid } from "nanoid"
+
+import type { ResourceState, ResourceType } from "~server/db"
+
+export const setupSite = async (siteId?: number, fetch?: boolean) => {
+  if (siteId !== undefined && fetch) {
+    return db.transaction().execute(async (tx) => {
+      const site = await tx
```
harishv7, line 9: these 3 ops seems to be fetching the existing site, nav, footer configs. Why are they within the setupSite function?

apps/studio/tests/integration/helpers/seed/index.ts
```diff
@@ -0,0 +1,264 @@
...
+    return db
+      .selectFrom("Blob")
+      .where("id", "=", blobId)
+      .selectAll()
+      .executeTakeFirstOrThrow()
+  }
+  return db
+    .insertInto("Blob")
+    .values({
+      content: jsonb({
```
harishv7, line 134: would it make sense to extract all of the values into constants to keep these queries clean (similar to curr seed script)

apps/studio/prisma/types.ts
```diff
@@ -12,15 +12,22 @@ import type {
...
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace PrismaJson {
     // TODO: Rename all with XXXYYYJson instead of XXXJsonYYY
-    type SiteJsonConfig = _IsomerSiteConfigProps
-    type SiteThemeJson = _IsomerSiteThemeProps
-    type BlobJsonContent = _IsomerSchema
-    type NavbarJsonContent = _IsomerSiteWideComponentsProps["navBarItems"]
-    type FooterJsonContent = _IsomerSiteWideComponentsProps["footerItems"]
+    type SiteJsonConfig = Tagged<_IsomerSiteConfigProps, "JSONB">
```
harishv7, line 21: Can share why using tagged/unwraptagged is better here?

apps/studio/src/schemas/page.ts
karrui: it's ok, this will still cause the validation to fail. FATAL just means zod stops validating immediately. No big deal for it to gather all validation errors before returning to the user

apps/studio/src/server/modules/site/site.service.ts
karrui: It still worked because of luck. we did not trigger any ambiguity on the pg engine. In more complex shapes, postgres may throw. So this is insurance.

tooling/eslint/base.js
karrui: some tests, especially skipped tests, will trigger it. No need for lint to be super strict on tests too

apps/studio/src/utils/ajv.ts
karrui: cause there may be more metadata on the input than exactly the schema, which causes ajv to throw. this one need ask @dcshzj maybe is code smell.

apps/studio/src/server/modules/resource/resource.service.ts
karrui: I waffling between trpc errors being thrown inside `service` files or in the router files directly. Might want the caller to take responsibility instead (which the caller actually does in this case)

apps/studio/tests/integration/helpers/seed/index.ts
karrui: a site is always the triple <site, footer, navbar>. so if fetching we fetch lo; if not we create

apps/studio/tests/integration/helpers/seed/index.ts
karrui: next time ba. i think not unwieldy yet and still readable. I favor readability over everything for tests

apps/studio/prisma/types.ts
karrui: Added an ADR 
https://www.notion.so/opengov/Explicitly-tagging-Kysely-jsonb-types-11577dbba78880c9917cd65dfe148e07?pvs=4

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+  setupSite,
+} from "tests/integration/helpers/seed"
+
+import type { reorderBlobSchema, updatePageBlobSchema } from "~/schemas/page"
+import { createCallerFactory } from "~/server/trpc"
+import { db } from "../../database"
+import { pageRouter } from "../page.router"
+import { createDefaultPage } from "../page.service"
+
+const createCaller = createCallerFactory(pageRouter)
```
seaerchin, line 23: til -this is a really cool helper function!

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+  let caller: ReturnType<typeof createCaller>
+  const session = await applyAuthedSession()
+
+  beforeAll(() => {
+    caller = createCaller(createMockRequest(session))
+  })
+
+  describe("readPage", () => {
+    beforeEach(async () => {
+      await resetTables("Site")
```
seaerchin, line 35: actually this isn't quite right - we should also reset the `Resource` table to avoid inter-test pollution right? 

we don't want to have a later test being reliant on an earlier one

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+      })
+
+      // Assert
+      await expect(result).rejects.toThrowError(
+        new TRPCError({ code: "NOT_FOUND", message: "Resource not found" }),
+      )
+    })
+  })
+
+  describe("reorderBlock", () => {
```
seaerchin, line 250: sorry for burden  - this test case isn't required for this pr but if you decide to not do it, could you file an issue and assign it to me?

we should check the behaviour in the event of value equality but different ordering of properties. as an example, consider the state `[{a:1, b:2}, some_other_block]`. if this gets changed to `[{b:2, a:1}, some_other_block]`, should we consider the state as being changed? to the user, this will appear exactly the same (same properties on block and block is same index) but i think this will fail atm cos we do a string comparison

seaerchin (dismissed):
lgtm! actually 1 thing that's risky right now but untested is the behaviour of js coercion to pg queries. 

an example we have atm, is that we coerce `Number(some_str)`, which might result in `NaN`. the converse, `String(some_number)` is safer but do you think we should add in a test case for the former or even both?

seaerchin:
eh approve too early omegalul

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+      const result = await caller.updatePageBlob(pageUpdateArgs)
+
+      // Assert
+      const actual = await db
+        .selectFrom("Blob")
+        .where("id", "=", pageToUpdate.draftBlobId)
+        .select("content")
+        .executeTakeFirstOrThrow()
+      expect(actual.content).toEqual(result.content)
+    })
```
seaerchin, line 543: @harishv7 trouble you to relook again at the draft logic. we should aim to have our invariants verified through test

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+        .select([
+          "Blob.content",
+          "Resource.publishedVersionId",
+          "Resource.draftBlobId",
+        ])
+        .executeTakeFirstOrThrow()
+      expect(actual).toMatchObject({
+        content: result.content,
+        publishedVersionId: publishedPageToUpdate.publishedVersionId,
+        draftBlobId: expect.any(String),
```
seaerchin, line 573: leave it to you but shuold we check if this id is different from the published blob id? this is to check that we don't override already published blobs

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+      // Assert
+      await expect(result).rejects.toThrowError(
+        new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Folder not found or folderId is not a folder",
+        }),
+      )
+    })
+
+    it("should throw 400 if folderId is not a Folder resource", async () => {
```
seaerchin, line 790: actually our invariant is that pages can only be created if your parent is `null` or a folder, so we need to check that `folderId` is not a collection too

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+      )
+    })
+
+    it.skip("should throw 403 if user does not have access to site", async () => {})
+
+    it.skip("should throw 403 if user does not have read access to root", async () => {})
+  })
+
+  // TODO: Implement tests when publish works
+  describe.skip("publishPage", () => {})
```
seaerchin, line 890: @harishv7 - i think it'd be good for us to write **test descriptors**, wdyt? can you write skipped tests for behaviour that's important to verify here?

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+
+    // TODO: Implement tests when permissions are implemented
+    it.skip("should throw 403 if user does not have access to site", async () => {})
+
+    it.skip("should throw 403 if user does not have write access to folder", async () => {})
+
+    it.skip("should throw 403 if user does not have write access to root", async () => {})
+  })
+
+  describe("getRootPage", () => {
```
seaerchin, line 822: low value case but should we check behaviour of what happens when the `parentId` of rootpage is non-null?

technically this should return a 5xx but i think it'll pass at present. although it's worth noting that actually users don't get to edit this property so it's really really really very minor and i'm happy to skip

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+
+    it.skip("should throw 403 if user does not have access to site", async () => {})
+
+    it.skip("should throw 403 if user does not have read access to root", async () => {})
+  })
+
+  // TODO: Implement tests when publish works
+  describe.skip("publishPage", () => {})
+
+  describe("updateSettings", () => {
```
seaerchin, line 892: should we also check that permalink cannot have special characters?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -265,15 +232,21 @@ export const pageRouter = router({
...
+        // TODO: Only return sites that the user has access to
+        .where("Resource.siteId", "=", siteId)
+        .where("Resource.type", "=", "RootPage")
+        .select(["id", "title", "draftBlobId"])
+        .executeTakeFirst()
+
+      if (!rootPage) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Root page not found",
```
seaerchin, line 246: wdyt about "The homepage for your site could not be found"? `rootpage` is internal to isomer and we should keep terminology consistent cc @sehyunidaaa

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -405,7 +374,7 @@ export const pageRouter = router({
       if (isEmpty(permalinkTree)) {
         throw new TRPCError({
           code: "NOT_FOUND",
-          message: "No permalink could be found for the given page",
+          message: "No permalink tree could be found for the given page",
```
seaerchin, line 377: sadly we should revert this because this error is likely going to be seen by the user and the place where it shows is in `/settings` for the permalink .-.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -156,42 +163,52 @@ export const updatePageById = (
...
     .select("draftBlobId")
-    .executeTakeFirstOrThrow()
+    .executeTakeFirst()
+
+  if (!page) {
+    throw new TRPCError({
+      code: "NOT_FOUND",
+      message: "Resource not found",
+    })
+  }
```
seaerchin, line 191: no real opinion here - noticed previous call didn't throw error but this now does. not sure how muych value consistency brings here but if nextjs has good error handling layer, i think it's ok to throw in services

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -360,36 +377,48 @@ export const getResourcePermalinkTree = async (
...
-    .execute()
+  return db.transaction().execute(async (tx) => {
+    // Guard against invalid resource
+    const resource = await getById(tx, {
+      siteId,
+      resourceId,
+    }).executeTakeFirst()
+    if (!resource) {
+      return []
+    }
```
seaerchin, line 388: how would we distinguish between a resource being empty and a resource not existing in this case?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -360,36 +377,48 @@ export const getResourcePermalinkTree = async (
...
 
 export const getResourceFullPermalink = async (
   siteId: number,
   resourceId: number,
 ) => {
   const permalinkTree = await getResourcePermalinkTree(siteId, resourceId)
-
+  if (permalinkTree.length === 0) {
+    return null
+  }
```
seaerchin, line 422: just a sanity check to ensure that this `null` value is handled properly on frontend

apps/studio/src/utils/ajv.ts
seaerchin: i think this is because we need to have extra properties on the item that ajv doesn't support. i could be wrong here thoguh

seaerchin (approved):
lgtm! i think making progress here is better than making no progress - for the test cases that i've raised, let me know if you are going to be doing it here or file issue. both are fine but we should just track them somewhere if it's hindering progress

seaerchin:
actually just fix the copy before merge is ok already

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
seaerchin: https://github.com/opengovsg/isomer/pull/770

karrui:
### Merge activity

* **Oct 11, 5:55 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/720).

PR #723, opened 2024-10-03
fix(permalinks): update sitemap generation for index pages

seaerchin:
## Problem
previously, index pages were identified by a sibling page with the same permalink as the folder. however, as we now have a unique constraint to prevent resources with the same permalink, this is no longer true and we need to update the index page generation.

Closes ISOM-1470

## Solution
1. update index pages to be identified using the special sentinel `_index`. this is true for now because **no permalink can contain an underscore `_`**. additionally, users are unable to create this index page themselves and hence, this sentinel is valid for our use case
2. for templates, we need to read from the actual `_index.json`, as publishing scripts will write to the file dictated by the `permalink`. however, for sitemap generation for **index pages,** we will **substitute** the permalink given by our db with the user's url. consider the index page for `/folderA` - this has db permalink: `/folderA/_index` but at runtime, the user will enter hte url `/folderA` expecting to see this page. hence, we need to keep the existing url **without the index**.  
3. for publishing scripts, the `writeToFile` method uses the full permalink out from the db (see point 2), but at the point of sitemap generation (this is what's being fed to the template), we need to remove `_index` if it exists. 
4. for studio, we need to update how the localised sitemap is generated, and process index pages differently. if we encounter an index page for the folder, we'll have to return the index page's id rather than the folder id.additionally, a db migration is added for this page + we disallow updating the url of the index page + moving the index page. **take note**: for now, users cannot create an index page so this is a preventive measure.

## Screenshots

https://github.com/user-attachments/assets/becaa8df-4838-4655-9bff-4290f9c44020


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -120,7 +122,7 @@ export const getFullPageById = async (
 }
 
 // There are 3 types of pages this get query supports:
```
dcshzj, line 124: nit: 4 now

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -359,30 +362,39 @@ export const getLocalisedSitemap = async (
...
             .selectFrom("Resource")
             .where("Resource.siteId", "=", siteId)
             .innerJoin("Ancestors", "Ancestors.parentId", "Resource.id")
-            .select(defaultResourceSelect),
+            .select(["Resource.id", "Resource.permalink", "Resource.parentId"]),
         ),
     )
     .selectFrom("Ancestors")
-    .select("Ancestors.permalink")
+    .select(["Ancestors.id", "Ancestors.permalink", "Ancestors.parentId"])
```
dcshzj, line 384: nit: I think we don't need to select more than the permalink, since it is discarded anyway in the map below.

apps/studio/src/utils/sitemap.ts
```diff
@@ -23,79 +25,51 @@ const getSitemapTreeFromArray = (
...
-        (child) =>
-          child.id !== resource.id &&
-          child.permalink === resource.permalink &&
-          child.type ===
-            (resource.type === "Folder" ? "Page" : "CollectionPage"),
-      )?.id
+  return children.map((resource) => {
+    const permalink = `${path}${resource.permalink}`
 
+    console.log("permalink", permalink)
```
dcshzj, line 37: nit: can remove this console.log

dcshzj (approved):

harishv7 (approved):

apps/studio/src/utils/sitemap.ts
seaerchin: done

PR #724, opened 2024-10-04
feat(permissions): folder permissions

seaerchin:
## Problem
This pr adds permissions checking for permissions on folders and collections **on the backend**. do note that this is only a backend change because the user is restricted from creating/deleting the folder/collectoin already on frontend as the `ability` restricts them.

## Solution
1. shift shared permissions check out to `permission.service` . this is not done for `move` yet, because the api is slightly different. i will aim to get behaviour correct first before shifting to another implementation (via `experimentalMiddleware`)
2. add permissions check to the methods on `collection.router` and `folder.router`

### DIFF
--- a/apps/studio/src/server/modules/resource/resource.router.ts
+++ b/apps/studio/src/server/modules/resource/resource.router.ts
@@ -20,7 +20,10 @@ import { protectedProcedure, router } from "~/server/trpc"
 import { publishSite } from "../aws/codebuild.service"
 import { db, sql } from "../database"
 import { PG_ERROR_CODES } from "../database/constants"
-import { definePermissionsFor } from "../permissions/permissions.service"
+import {
+  definePermissionsFor,
+  validateUserPermissions,
+} from "../permissions/permissions.service"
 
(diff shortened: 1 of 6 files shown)


apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -35,7 +38,7 @@ const fetchResource = async (resourceId: string | null) => {
   )
 }
 
-const validateUserPermissions = async ({
+const validateUserPermissionsForMove = async ({
```
dcshzj, line 41: nit: actually since you already moved `validateUserPermissions` into the permissions.service file, I think this function should be moved there as well so we can consolidate all functions related to permissions validation into a single place.

dcshzj (approved):

seaerchin:
### Merge activity

* **Oct 23, 5:15 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/724).

PR #726, opened 2024-10-07
fix: remove react-aria from Link component

dcshzj:
## Problem

The react-aria-components version of Link brings about some issues regarding how links are handled in mobile devices, mainly taps are not being recognised correctly.

Fixes ISOM-1605.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Completely remove react-aria from handling links within the components package.
- Use index for generating table of contents instead of a random number to ensure stability between renders.
- Remove the `use client` directive from files that no longer need them.
- Removed use of `useMemo` from Siderail as it is a server component.


karrui (commented):
hehe chromatic many regressions, pls fix!

dcshzj:
> hehe chromatic many regressions, pls fix!

Fixed!

sehyunidaaa:
@dcshzj reviewed the single snapshot, as discussed we'll check the rest on the output sites yeah

packages/components/src/templates/next/components/internal/Link/Link.tsx
```diff
@@ -111,46 +15,47 @@ const linkStyles = tv({
...
-  withFocusVisibleHighlight?: boolean
-}
-
-export function Link({
+export const Link = ({
+  href,
+  current,
+  isDisabled,
+  isExternal,
+  isWithFocusVisibleHighlight,
```
karrui, line 31: should this default to true (for a11y)? only false if overriding with something else perhaps. Then all the links don't need to care

karrui (approved):
lgtm, trusting chromatic that there are no regressions in behaviour.

packages/components/src/templates/next/components/internal/Link/Link.tsx
dcshzj: Hmm I was thinking for future cases, usages of the Link will be more customised rather than a direct text link? So as long as BaseParagraph is used for text content, it will automatically come with the focus visible stuff.

PR #727, opened 2024-10-07
fix(component): change placeholder copy on collections search to be more generic

sehyunidaaa:
## Problem

The default placeholder copy on Collections search was too specific with a piped text, so it read weirdly depending on the page title.

Closes [ISOM-1611]

## Solution

Change copy to say "Start typing to search"


sehyunidaaa:
thank you @adriangohjw!

adriangohjw (approved):

PR #730, opened 2024-10-07
fix: email checking

seaerchin:
## Problem
previously our email check was case sensitive, which led to validation failing when the users input an upper cased character

## Solution
lowercase the email so that we perform an unconditional check. ~do note that this assumes the email stored in grwoth book is also lowercased, other wise this comparison will no longer be valid~

### DIFF
--- a/apps/studio/src/server/trpc.ts
+++ b/apps/studio/src/server/trpc.ts
@@ -125,9 +125,11 @@ const baseMiddleware = t.middleware(async ({ ctx, next }) => {
 
 const authMiddleware = t.middleware(async ({ next, ctx }) => {
   const defaultWhitelist: string[] = []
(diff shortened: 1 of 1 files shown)


adriangohjw (commented):
> do note that this assumes the email stored in grwoth book is also lowercased, other wise this comparison will no longer be valid

if the goal is to make the check less strict, perhaps we should also lowercase he email in `whitelistedUsers` before comparing it?

seaerchin:
> > do note that this assumes the email stored in grwoth book is also lowercased, other wise this comparison will no longer be valid
> 
> if the goal is to make the check less strict, perhaps we should also lowercase he email in `whitelistedUsers` before comparing it?

i derped, i lowercased it manually oops! lemme fix real quick

adriangohjw (approved):
LGTM!

PR #732, opened 2024-10-08
fix: improve date parsing for collection items

dcshzj:
## Problem

Some of the dates in collections show up as invalid date because the allowed date format is very specific.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Improve the date parsing of collection items.

### DIFF
--- a/packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
+++ b/packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
@@ -20,7 +20,7 @@ const COLLECTION_ITEMS: IsomerSitemap[] = flatten(
       layout: "article",
       summary:
         "We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year. We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year.",
-      date: "07/05/2024",
+      date: "2024-05-07",
       category: "Category Name",
     },
     {
@@ -31,7 +31,7 @@ const COLLECTION_ITEMS: IsomerSitemap[] = flatten(
       layout: "file",
       summary:
         "This is supposed to be a description of the hero banner that Isomer uses on their official website.",
-      date: "07/05/2024",
+      date: "2024-05-07",
       category: "Category Name",
       ref: "https://www.isomer.gov.sg/images/Homepage/hero%20banner_10.png",
       fileDetails: {
@@ -47,7 +47,7 @@ const COLLECTION_ITEMS: IsomerSitemap[] = flatten(
       layout: "link",
       summary:
         "Have a look at the Isomer guide to understand how to use the Isomer CMS.",
-      date: "12/08/2023",
+      date: "2023-08-12",
       category: "Category Name",
       ref: "https://guide.isomer.gov.sg",
     },
@@ -452,7 +452,7 @@ export const FilteredEmptyResults: Story = {
                 layout: "file",
                 summary:
                   "This is supposed to be a description of the hero banner that Isomer uses on their official website.",
-                date: "07/05/2025",
+                date: "2025-05-07",
                 category: "Category Name",
                 ref: "https://www.isomer.gov.sg/images/Homepage/hero%20banner_10.png",
                 fileDetails: {
--- a/packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
+++ b/packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
@@ -20,7 +20,7 @@ const COLLECTION_ITEMS: IsomerSitemap[] = flatten(
       layout: "article",
       summary:
         "We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year. We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year.",
-      date: "2024-05-07",
+      date: "07/05/2024",
       category: "Category Name",
     },
     {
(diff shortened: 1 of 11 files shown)


packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
```diff
@@ -20,7 +20,7 @@ const COLLECTION_ITEMS: IsomerSitemap[] = flatten(
       layout: "article",
       summary:
         "We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year. We’ve looked at how people’s spending correlates with how much microscopic plastic they consumed over the year.",
-      date: "2024-05-07",
+      date: "07/05/2024",
```
karrui, line 23: hmmmm still think ISO Date format is more standard. Can we not use that standard and transform it properly instead?

packages/components/src/templates/next/layouts/Collection/utils.ts
```diff
@@ -1,6 +1,11 @@
+import dayjs from "dayjs"
+import customParseFormat from "dayjs/plugin/customParseFormat"
+
 import type { AppliedFilter, Filter as FilterType } from "../../types/Filter"
 import type { CollectionCardProps } from "~/interfaces"
 
+dayjs.extend(customParseFormat)
```
karrui, line 7: feels overkill to use dayjs just to format dates. dayjs also somewhat dangerous; any `extend` call will affect every other dayjs instance.

should we use [date-fns](https://date-fns.org/v4.1.0/docs/format) instead? might make more sense (and is lighter)

karrui (commented):
suggest keeping ISO format instead of the new dd/mm/yyyy format. please use date-fns instead!

packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
dcshzj: Hmm yeah true, reverted in 3b67c53d9a03540807094d60ce84e34ef8e0d835.

packages/components/src/templates/next/layouts/Collection/utils.ts
dcshzj: Updated in df8e63fa23a3d0ebda68d44c5e4c7b650d3158d9.

seaerchin (commented):
do you think we should have an equivalent function for dates in studio so that this implicit contract is kept?

packages/components/src/utils/getParsedDate.ts
```diff
@@ -0,0 +1,37 @@
...
+      return parse(dateString, "dd MMMM yyyy", new Date())
+    }
+
+    if (isMatch(dateString, "yyyy-MM-dd")) {
+      return parse(dateString, "yyyy-MM-dd", new Date())
+    }
+
+    if (isMatch(dateString, "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")) {
+      return parse(dateString, "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", new Date())
+    }
```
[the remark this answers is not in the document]
dcshzj: Ah yes, good idea, updated in 2daaaf2a62d74b125ad0aa22c79cf93e98e11c26.

packages/components/src/utils/getFormattedDate.ts
```diff
@@ -1,13 +1,14 @@
...
 
 // Standardise the format of dates displayed on the site
-export const getFormattedDate = (date: string) =>
-  dayjs(date, [
-    "DD/MM/YYYY",
-    "D MMM YYYY",
-    "DD MMM YYYY",
-    "YYYY-MM-DDTHH:mm:ss.SSSZ",
-  ]).format("D MMMM YYYY")
+export const getFormattedDate = (dateString?: string) => {
```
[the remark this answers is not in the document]
dcshzj: I think usually the result is directly used and not further processed, but added in 2daaaf2a62d74b125ad0aa22c79cf93e98e11c26 for forward compatibility.

packages/components/src/utils/getParsedDate.ts
```diff
@@ -0,0 +1,39 @@
...
+    const parsedDate = SUPPORTED_DATE_FORMATS.reduce<Date | undefined>(
+      (acc, format) => {
+        if (acc) {
+          // Date has already been parsed by an earlier format
+          return acc
+        }
+
+        if (isMatch(dateString, format)) {
+          return parse(dateString, format, new Date())
+        }
```
seaerchin, line 24: nit: we should shift the `try/catch` here since this is the only thing that can fail

packages/components/src/utils/getParsedDate.ts
```diff
@@ -0,0 +1,39 @@
...
+  "dd/MM/yyyy",
+  "d MMM yyyy",
+  "d MMMM yyyy",
+  "dd MMM yyyy",
+  "dd MMMM yyyy",
+  "yyyy-MM-dd",
+  "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
+]
+
+export const getParsedDate = (dateString: string) => {
```
seaerchin, line 13: actually the flow i had in mind was `reduce(function, new Date()`

then in our reducer function:
```ts
if (isMatch(date)) { // parse } 
```

and we can put supported date formats in increasing order of priority, so that the code looks cleaner overall. not a big deal though, just that this was not quite what i had in mind

seaerchin (approved):
lgtm. we should shift the `try/catch` inside cos that impacts readabiliy and the wideness of the `try/catch` makes it difficult to tell exactly what we're catching 

leave it to you if we want to refactor the `reduce` orders

packages/components/src/utils/getParsedDate.ts
dcshzj: Done in 715e22c6828872fb0861f310c808324f983f2993.

packages/components/src/utils/getParsedDate.ts
dcshzj: Hmm I a bit confused, I think that's what I'm doing right now? The reducer function is indeed going by priority (from first to last element) and it starts with `undefined` to differentiate between whether the date string has been parsed or not. If parsed then the type is of Date, otherwise it will remain as undefined.

PR #735, opened 2024-10-08
fix: tooling pipeline

seaerchin:
## Problem
tooling build depends on components but the ordering isn't clear, resulting in failed builds

## Solution
add dependency for `tooling` on `isomer-components` using `turbo.json`


karrui (approved):
@karrui approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/735)

PR #736, opened 2024-10-08
fix(cards): remove default from components package

seaerchin:
## Problem
right now, when we delete the card, it uses the schema default, resulting in the delete not going through and showing the default. 

## Solution
1. remove `default` from the schema 
    - by right, the schema should tell us what the default is. at present, this is being done by studio. another alternative _could_ be to just use the schema default and remove the constants in studio but the drawback is that it doesnt add much value as of now, is quite some drudgery and teh behaviour changes (see below):
    - if we use `schema.default`, what happens is that when you clear the input, it'll use the default value from the schema leading to potentially unintuitive behaviour. (tbh, i'm ok with this but we need to check if this is obvious to the user)
    - [relevant issue](https://github.com/rjsf-team/react-jsonschema-form/issues/2980)

## Screenshots

https://github.com/user-attachments/assets/185b01c3-146e-4600-820a-bce2a9a846bb

### DIFF
--- a/apps/studio/public/assets/css/preview-tw.css
+++ b/apps/studio/public/assets/css/preview-tw.css
@@ -1524,10 +1524,6 @@ video {
   height: 500px;
 }
 
-.h-\[4\.25rem\] {
-  height: 4.25rem;
-}
-
 .h-\[68px\] {
   height: 68px;
 }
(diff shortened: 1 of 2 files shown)


dcshzj (approved):
> by right, the schema should tell us what the default is. at present, this is being done by studio. another alternative could be to just use the schema default and remove the constants in studio but the drawback is that it doesnt add much value as of now, is quite some drudgery and teh behaviour changes (see below):

I think this might be more of a bug from JSONForms so ok with this approach of defining the defaults in studio for now.

PR #744, opened 2024-10-08
isom-1479 - fix: use `lg` for all ModalCloseButton

adriangohjw:
## Problem

ModalCloseButtons have inconsistent styling with Figma

Closes https://linear.app/ogp/issue/ISOM-1479/all-modals-modal-close-buttons-should-be-lg

## Solution

Add `lg` size to all of them

**Breaking Changes**

  - Details ...


seaerchin (approved):
This PR looks good but the issue i find with doing this is that the styles might be missed on a subsequent PR (and it's very easy to miss too!) 

we actually use `chakra-ui` under the hood for design system, and [it allows us ](https://v2.chakra-ui.com/docs/styled-system/customize-theme) to set the default theme props (search for `defaultProps`). 

because the themes are iterativeliy merged, our styles will still be applied! i think this might be a good way of ensuring consistent styling across the codebase but it's not super important for now (and we can always do it later) so i'm happy to accept this!

PR #745, opened 2024-10-09
isom-1477 - fix: add `mr` 3.5rem to ModalHeader

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1477/edit-folder-title-eats-into-close-modal-button

## Solution

- add a 3.5rem margin-right to the ModalHeader
  - or margin-left if the close button is on the left
  - don't add if there's no close button e.g. details modal

image

Note: 
- as not be 8px for everything for now due to wrong close button size, but that's fixed independently in https://github.com/opengovsg/isomer/pull/744

**Breaking Changes**

  - Details ...

### DIFF
--- a/apps/studio/src/components/VersionWrapper/VersionModal.tsx
+++ b/apps/studio/src/components/VersionWrapper/VersionModal.tsx
@@ -31,8 +31,8 @@ export const VersionModal = ({ isOpen, onClose }: VersionModalProps) => {
     <Modal isOpen={isOpen} onClose={onClose} size={modalSize}>
       <ModalOverlay />
       <ModalContent>
+        <ModalHeader mr="3.5rem">Update Available</ModalHeader>
         <ModalCloseButton size="lg" />
-        <ModalHeader ml="3.5rem">Update Available</ModalHeader>
         <ModalBody>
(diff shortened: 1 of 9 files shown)


apps/studio/src/components/VersionWrapper/VersionModal.tsx
```diff
@@ -32,7 +32,7 @@ export const VersionModal = ({ isOpen, onClose }: VersionModalProps) => {
       <ModalOverlay />
       <ModalContent>
         <ModalCloseButton size="lg" />
-        <ModalHeader>Update Available</ModalHeader>
+        <ModalHeader ml="3.5rem">Update Available</ModalHeader>
```
seaerchin, line 35: double checking here if this should be `mr`

seaerchin (approved):
lgtm with 1 comment - please double check cos i think the `ModalButton` is aboslutely positioned

apps/studio/src/components/VersionWrapper/VersionModal.tsx
adriangohjw: yup you are right, thanks for spotting
have fixed 👍

PR #748, opened 2024-10-09
fix(input): typing quickly causes skipped characters

seaerchin:
## Problem
Currently, when we type extremely quickly on the Homepage, occasionally the input field might eat the characters leading to cases where teh input isn't properly formed. 

As an example, consider the string `The quick brown` -> typing this quickly (last char is `n`) might case the `n` to be eaten. This leads to the UI showing `The quick brow` instead of `The quick brown`.

## Solution
1. use `flushSync` (read [here](https://react.dev/reference/react-dom/flushSync)), which is a **band-aid** but does fix the issue. this issue happens because we have 2 updates , which **need to be in the same update cycle** but are not. the first update is when we write to the component in`JsonForms#handleChange` and the second one is when we call `ComplexEditorStateDrawer` to use the updated state to update the overall preview state. 
2. how `flushSync` works is that we flush all the pending work (at the call site and inclusive of what's inside `flushSync`) and forces react to reconcile + update the dom. this forces the 2 updates above to be part of the same update cycle, fiximg the issue 

## Screenshots

https://github.com/user-attachments/assets/702c0d3f-dcea-4c91-ae2d-8dffc8bc8e25

1. `useTransition` - 
image

2. `useDeferredValue` 
image
this causes the render to be 2 cycles behind...

3. `debounce` 
"works" but the updates are still queued, jsut that each update must be >debounce value. this means that the bug still exists if users update close to the debounce value (still 2 renders) 
4. `atoms` for the state - couldn't quite get it working as we need an initialiser and the atom still has the bug
5. use function for updating state - didn't work


apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -55,12 +56,21 @@ export function EditorDrawerProvider({
...
     useState<IsomerSchema>(initialPageState)
   // State of the page to render in the preview
-  const [previewPageState, setPreviewPageState] =
+  const [previewPageState, _setPreviewPageState] =
     useState<IsomerSchema>(initialPageState)
   // Holding state for images/files that have been modified in the page
   const [modifiedAssets, setModifiedAssets] = useState<ModifiedAsset[]>([])
   const [addedBlockIndex, setAddedBlockIndex] = useState<number | null>(null)
 
+  const setPreviewPageState = useCallback(
```
karrui, line 65: might want to add a comment here saying why this is done.

karrui (approved):
lgtm

apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -1,6 +1,7 @@
 import type { IsomerSchema } from "@opengovsg/isomer-components"
 import type { Dispatch, PropsWithChildren, SetStateAction } from "react"
-import { createContext, useContext, useState } from "react"
+import { createContext, useCallback, useContext, useState } from "react"
+import { flushSync } from "react-dom"
```
harishv7, line 4: hmm, a bit concerned that the docs mention:

> Using flushSync is uncommon and can hurt the performance of your app.

> Most of the time, flushSync can be avoided. Use flushSync as last resort.

Actually thought this could be just a debounce issue. Was this the only solution of the alternatives mentioned?

seaerchin: yea this was the only thing that worked :sadge:

apps/studio/src/contexts/EditorDrawerContext.tsx
seaerchin: added huge paragraph!

PR #751, opened 2024-10-09
isom-1348 - move folder single click and double click pattern

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1348/my-pages-move-folder-single-click-and-double-click-interactions

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- added a `isHighlighted` UI state for `MoveItem`
- use a state `isResourceHighlighted` to keep track if the current resource is being highlighted - this help us identify what view to render
  | view | screenshot |
  | - | - |
  | selected and highlighted but still in parent folder | image |
  | selected but in the selected folder itself | image |
- pass in `handleOnClick` instead of `onChangeResourceId` into `MoveItem` to reduce its scope of props and context needed
- minor refactoring to improve readability and duplication e.g. renaming `moveDest?.resourceId` to `curResourceId` as it has been defined previously
- minor UI fix - increase the gap between the text in the bottom feedback component to be consistent with FIgma

## Before & After Screenshots

| scenario | video |
| - | - |
| moving a page into a folder and subfolder | https://github.com/user-attachments/assets/93d06b2d-8fd2-45e3-8dac-d8709ad9ef53 |
| moving a folder | https://github.com/user-attachments/assets/8c649555-a706-41fd-939b-ea5b98c0e2c2 |

## Note

- resource cannot be moved into root folder. this is an issue(?) that isn't addressed in this ticket - please let me know if this is an issue or intended design


apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
```diff
@@ -11,24 +11,54 @@ type MoveItemProps = Pick<
...
       justifyContent="flex-start"
-      color="base.content.default"
+      color={buttonProps.color}
+      bg={buttonProps.bg}
+      {...(buttonProps._hover && {
+        _hover: {
+          color: buttonProps._hover.color,
+          bg: buttonProps._hover.bg,
+        },
+      })}
```
seaerchin, line 58: nit: i think just spreading `buttonProps` will work here

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -53,23 +53,28 @@ export const MoveResourceModal = () => {
...
+      useState<boolean>(true)
     const { siteId } = useQueryParse(sitePageSchema)
     const setMovedItem = useSetAtom(moveResourceAtom)
     const [{ title }] = trpc.resource.getMetadataById.useSuspenseQuery({
       resourceId,
     })
-    const curResourceId = resourceStack[resourceStack.length - 1]?.resourceId
+    const moveDest = resourceStack[resourceStack.length - 1]
+    const parentDest = resourceStack[resourceStack.length - 2]
+    const curResourceId = moveDest?.resourceId
```
seaerchin, line 70: tbvh i think at this point we should make this into a custom hook cos there's alot of logic, but i'm ok with merging this first - more important to get it working than to aim for perfection

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -191,21 +200,42 @@ const MoveResourceContent = withSuspense(
...
-                          },
-                        ])
+                      {...item}
+                      key={item.id}
+                      isDisabled={isItemDisabled}
+                      isHighlighted={isItemHighlighted}
+                      handleOnClick={() => {
+                        if (isItemDisabled) {
+                          return
+                        }
```
seaerchin, line 218: nit: i think we can omit this case cos if it's disabled the `onClick` for the button shouldn't fire

seaerchin (approved):
lgtm, i tested it locally and it works well! the only issue is that the code in the `MoveResourceModal` file is getting abit bloated - i didn't investigate too deeply, but we might be able to simplify by shifting uot some logic to a custom hook. nbd though, if we can file a 1-2 liner issue so we don't lose track, that'd be great

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
adriangohjw: yeah i tot so too but it overwrite the default _hover state which lead to a weird UI

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
adriangohjw: yeah true, i added it because it will be less confusing to future code maintainer by explicitly accounting for all states

adriangohjw:
NOTE to self - there are other instances of similar file selection thing, will be great to port over the functionality over for better and more consistent UX

PR #752, opened 2024-10-10
feat: support site publishing via CodeBuild

dcshzj:
## Problem

We are not using AWS Amplify for hosting our sites anymore, instead using CloudFront with an S3 bucket.

## Solution

**Breaking Changes**

**Features**:

- Removed uploading to Amplify during a deployment on CodeBuild and instead just upload directly to S3.
- After the upload is complete, update the CloudFront origin path and create an invalidation.
- Also perform publishes for implicit actions such as:
    - Creates: Folder/Collection
    - Updates: Renaming of folder, moving of folder/pages
    - Deletes: Pages/Folder/Collection
- A new build is not triggered if there is a currently running build that was started within the last 60 seconds of the time of publish.
    - Reason: CodeBuild doesn't have the functionality to queue builds, so if we want to avoid the starvation of builds, we will have to implement our own queueing system which would be quite an involved effort. For context, starvation of builds happen because we originally kill off the existing running build when starting new ones.
    - 60 seconds is a number decided based on tests, whereby the publisher script takes about a minute to clone this repo and build the components package on a powerful machine (70 GB memory, 36 vCPUs) before it runs the script to pull the site data from the database. So any publishes that happen within this time period would have been captured by this earlier running build (since the DB fetch has not happened yet).
    - Race conditions can still happen as there is no locking mechanism implemented, but willing to take a wait-and-see approach first before diving into more involved engineering solutions.


tooling/build/scripts/publishing/index.ts
```diff
@@ -289,24 +292,21 @@ async function writeContentToFile(
...
-        // NOTE: normalization here will remove dual backslashes
-        // and also strip .. filepaths except as a prefix
-        .normalize(fullPermalink)
-        // NOTE: this matches on a leading ../
-        // or a leading ..\
-        // or a plain .. without any paths
-        .replace(/^(\.\.(\/|\\|$))+/, ""),
-    )
+    const sanitizedPermalink = !fullPermalink
+      ? "index"
```
seaerchin, line 298: double checking - iosn't this supposed to be `_index`?

seaerchin (commented):
lgtm with 1 question. i tried to check through the cases that you listed for implicit publish and they look rouighly correct but good to be sure by adding tests

apps/studio/src/features/editing-experience/components/PublishButton.tsx
```diff
@@ -50,19 +50,20 @@ const SuspendablePublishButton = ({
     if (coercedSiteId && coercedPageId)
       mutate({ pageId: coercedPageId, siteId: coercedSiteId })
   }
+
+  const isChangesPendingPublish = !!currPage.draftBlobId
```
dcshzj, line 54: note to self: because we are using this to check, we need to do an implicit publish when page settings are changed (but we won't publish the resource so that's ok)

tooling/build/scripts/publishing/index.ts
dcshzj: Ah right yes, updated in 33df722fd3307f6cad2229dc88269df08f82a63d.

packages/components/src/templates/next/components/internal/Navbar/Navbar.tsx
```diff
@@ -6,7 +6,7 @@ import { useResizeObserver } from "usehooks-ts"
 
 import type { NavbarProps } from "~/interfaces"
 import { tv } from "~/lib/tv"
-import { isExternalUrl } from "~/utils"
+import { getReferenceLinkHref, isExternalUrl } from "~/utils"
```
seaerchin, line 9: not sure if bug but i expected to see `getReferenceLinkHref` being used but it isn't

seaerchin: ah ok it got shifted into navbar item

tooling/build/scripts/publishing/index.ts
```diff
@@ -80,6 +86,7 @@ async function main() {
       if (
         (resource.type === "Page" ||
           resource.type === "CollectionPage" ||
+          resource.type === "IndexPage" ||
```
seaerchin, line 89: do we need to accuont for `CollectionLink`? i remember we rendered an empty layout but double checking here

seaerchin (commented):
some minor questions but ok to approve once answered

tooling/build/scripts/publishing/index.ts
dcshzj: Yup we wil need to, but right now cos the `CollectionLink` type is still in your PR (sad) so I can't type it here.

packages/components/src/utils/getReferenceLinkHref.ts
```diff
@@ -22,41 +22,62 @@ const getSanitizedLinkHref = (url?: string) => {
...
   sitemap: IsomerSitemap,
+  assetsBaseUrl: string | undefined,
 ) => {
   if (!referenceLink) {
     return undefined
   }
 
-  const actualLink = getActualLinkFromReference(referenceLink, sitemap)
+  const assetLink = convertAssetLinks(referenceLink, assetsBaseUrl)
+  const actualLink = convertReferenceLinks(assetLink, sitemap)
```
seaerchin, line 80: **no actionable required**. structure here feels abit off - i thought this would accept a `ReferenceLink` (or do some check to prove that it's a reference link) but the logic here is quite difficult to follow. 

i think we can fix this next time ba but we should probably have a top level function that accepts raw strings and spits out `reflink/extlink` etc as template types

seaerchin (approved):

packages/components/src/templates/next/components/complex/Hero/Hero.tsx
```diff
@@ -38,7 +38,11 @@ const HeroGradient = ({
             {buttonLabel && buttonUrl && (
               <div className="flex flex-col justify-start gap-x-5 gap-y-4 sm:flex-row">
                 <LinkButton
-                  href={getReferenceLinkHref(buttonUrl, site.siteMap)}
+                  href={getReferenceLinkHref(
+                    buttonUrl,
+                    site.siteMap,
+                    site.assetsBaseUrl,
+                  )}
```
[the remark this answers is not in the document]
dcshzj: Yup, this is handled by the function and only converts if it matches a reference link or an asset link.

PR #754, opened 2024-10-10
feat - Add intercom to CMS

adriangohjw:
## Problem

Intercom is needed for v1 and is a migration blocker
- https://linear.app/ogp/project/isomer-studio-v1-665b82dbc7c7
- https://www.notion.so/opengov/Isomer-s-focus-Q3-Q4-2024-10577dbba78880d78ae6efec05110246?pvs=4#11b77dbba7888065a956e35b7bb45b86

## Solution

Added Intercom integration to the CMS, allowing for real-time support interaction.
- Included the Intercom package as a dependency.
- Added the necessary environment variable for Intercom App ID.
- Updated Content Security Policy to allow Intercom widgets.
- Initialized Intercom in the application.

Note: cannot test in development due to localhost but it does not throw error but it's implemented based on https://www.intercom.com/help/en/articles/168-install-intercom-for-users-on-web#install-the-javascript-code

**Breaking Changes**

  - Details ...


seaerchin (approved):

apps/studio/src/pages/_app.tsx
```diff
@@ -38,7 +39,15 @@ void gb.init({
   // Optional, enable streaming updates
   streaming: true,
 })
+
 const MyApp = ((props: AppPropsWithAuthAndLayout) => {
+  if (env.NEXT_PUBLIC_INTERCOM_APP_ID) {
+    Intercom({
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
karrui, line 46: this should not have an eslint error since the env should be typed properly

adriangohjw: yeah i tot so too but it's showing as an issue on my editor

image

karrui: what's the error? sometimes need restart eslint after adding env

adriangohjw: you are right - it was gone after i restart my code editor and restart eslint
thanks for pointing it out and sharing! :D

.github/workflows/aws_deploy.yml
```diff
@@ -167,6 +167,7 @@ jobs:
             NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME=${{ inputs.app-s3-assets-domain-name }}
             NEXT_PUBLIC_S3_ASSETS_BUCKET_NAME=${{ inputs.app-s3-assets-bucket-name }}
             NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY:${{ secrets.GROWTHBOOK_CLIENT_KEY }}
+            NEXT_PUBLIC_INTERCOM_APP_ID:${{ secrets.INTERCOM_APP_ID }}
```
seaerchin, line 170: just fyi, in order for this to work, we need to tell Docker that this is a build arg. in order to do that, we need to pass this as a build arg + env in docker!

just passing to docker isn't quite enough because docker doesn't know that this is supposed to be a build arg. hence, when we deploy this to envs with this secret, it won't quite work (but won't crash the app)

in order to get this picked up by docker as a build arg, we need `ARG NEXT_PUBLIC_INTERCOM_APP_ID` and for it to be available to the app as an env we also need to pass `ENV NEXT_PUBLIC_INTERCOM_APP_ID = $NEXT_PUBLIC_INTERCOM_APP_ID`

adriangohjw: @seaerchin okay done, should i do the same for `NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY` in another separate PR?

PR #758, opened 2024-10-10
fix: remove unused tables

seaerchin:
## Problem
the permissions feature uses a different table and schema from the one we have. this is because our approach now is to store a role in the database and retrieve it at run-time.

## Solution
1. delete the table from schema


adriangohjw (approved):

sehyunidaaa:
@seaerchin I see some regressions? on Chromatic for this PR e.g., the copy returning to old one

PR #759, opened 2024-10-10
fix(collections): remove selector step

seaerchin:
## Problem
1. we now don't have a dedicated pdf page. rather, we upload pdf to s3 and show them as normal links

## Solution
remove this step in our modal

## Screenshots

https://github.com/user-attachments/assets/b440ac98-853d-4290-b4d8-c3d47545d615


adriangohjw:
will we be supporting PDF file in the future?

if so, there might still be value in showing the selector page but instead add a "coming soon" to the PDF file option - this will improve users confidence in the product and the product desirability, thus reducing friction in onboarding new agencies

dcshzj:
> will we be supporting PDF file in the future?

Yeah actually @seaerchin we still need this selector, just that it is not a new page but rather a new item that references the file, so the page editor will not have a preview. Editors can edit this item by uploading a file, editing the title and category.

seaerchin:
> > will we be supporting PDF file in the future?
> 
> Yeah actually @seaerchin we still need this selector, just that it is not a new page but rather a new item that references the file, so the page editor will not have a preview. Editors can edit this item by uploading a file, editing the title and category.

are we not uploading from the page directly? ie, the flow is that they create a new page, select a `pdf` component then we upload. if this isn't the case and we're still preserving the pdf, i'm ok with closing this

dcshzj:
> are we not uploading from the page directly? ie, the flow is that they create a new page, select a `pdf` component then we upload. if this isn't the case and we're still preserving the pdf, i'm ok with closing this

Hmm I think no, cos we will still need the "edit item" page for editors to change existing items in collections, so we should funnel to the same editor. The file will be part of the page's metadata rather than a separate pdf component.

PR #762, opened 2024-10-10
isom-1621 - remove type from collections

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1621/remove-type-from-collections

## Solution

- remove `variant` from filtering utils
- other cleanup: remove unused `IsomerSitemap` import

**Breaking Changes**

  - Details ...
- [x]  No - this PR is backwards compatible

## Before & After Screenshots

Screenshot 2024-10-11 at 6 19 34 AM


packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
```diff
@@ -469,6 +469,9 @@ export const FilteredEmptyResults: Story = {
   play: async ({ canvasElement }) => {
     const screen = within(canvasElement)
     await userEvent.click(screen.getByText(/2025 \(1\)/i))
-    await userEvent.click(screen.getByText(/Link \(10\)/i))
+    const searchElem = screen.getByRole("searchbox", {
+      name: /Start typing to search/i,
+    })
+    await userEvent.type(searchElem, "anything")
   },
 }
```
adriangohjw, line 477: issue with this is that it is effectively testing 2 things - the filters and searchbar

im thinking if this should be rewritten to change the `COLLECTION_ITEMS` such that i can always have a unique set of filter permutations that return 0 results?

dcshzj: If just doing filters alone I think we can have a new single collection item with a different category from the rest, then use that category with the year to get an empty filtered result

dcshzj (approved):

PR #772, opened 2024-10-11
feat(file-upload): implement file uploads on tiptap 

seaerchin:
## Problem
Users should be able to upload their pdf to the page.

## Solution
1. add a `FileAttachment` component that does the attachment. I used a `Skeleton` and set the string to `""` so that we are disabling the save
2. we add a validation mode of `onChange` so that we'll start validation immediately. this is because the default mode is `onSubmit`, which won't trigger until we try to submit the form (and we'll have no error messages) 

## Screenshots

https://github.com/user-attachments/assets/dd7c35c2-a171-41e6-ae8d-5e5314f7c3af

### DIFF
--- a/apps/studio/src/components/PageEditor/FileAttachment.tsx
+++ b/apps/studio/src/components/PageEditor/FileAttachment.tsx
@@ -1,0 +1,78 @@
+import { useEffect, useState } from "react"
+import { FormControl, Skeleton, Text } from "@chakra-ui/react"
+import { Attachment, FormErrorMessage } from "@opengovsg/design-system-react"
+
+import { MAX_PDF_FILE_SIZE_BYTES } from "~/features/editing-experience/components/form-builder/renderers/controls/constants"
+import { useUploadAssetMutation } from "~/hooks/useUploadAssetMutation"
+import { getPresignedPutUrlSchema } from "~/schemas/asset"
+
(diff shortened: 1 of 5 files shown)


apps/studio/src/components/PageEditor/FileAttachment.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const [file, setFile] = useState<File | undefined>(undefined)
+  // TODO: Add a mutation for deletion next time of s3 resources
+  const { mutate: uploadFile, isLoading } = useUploadAssetMutation({
+    siteId,
+  })
+
+  useEffect(() => {
+    // NOTE: The outer link modal uses this to disable the button
+    if (isLoading) setHref("")
+  }, [isLoading, setHref])
```
adriangohjw, line 33: do we need this to be a `useEffect`?

adriangohjw (approved):

apps/studio/src/components/PageEditor/FileAttachment.tsx
seaerchin: yes sadly. this is because the `useEffect` sync the `isLoading` with the external state. because the `mutate` runs as a side-effect, we need something to sync the `loading` state to whether the button is disabled

PR #774, opened 2024-10-13
fix(components): collectionCard spacing and line truncation enhancements

sehyunidaaa:
## Problem

During migration, we observed enhancement opportunities to the CollectionCard:

- We clamp the title and description at 2 lines, which results in critical content becoming obscured (especially when there is an image or the screen is small) 
- There is a gap that appears when there is no summary (`description` field) in an article because a `<span>` is created regardless of whether the field is empty
- Image is too wide, so content is obscured even further

Closes [ISOM-1600]

## Solution

- Change line clamping of title to line-clamp-3
- Change description <span> to only show when there is a description present
- Change width of image to be responsive (25% of card in desktop) and have a minimum height on tablet
- Add a story that shows CollectionCards with image

### Before
image

### After
CollectionCard with image, desktop:
image

No major changes on mobile:
image

### DIFF
--- a/packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
+++ b/packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
@@ -12,7 +12,7 @@ type CollectionCardProps = BaseCollectionCardProps
 
 const collectionCardLinkStyle = tv({
   extend: focusVisibleHighlight,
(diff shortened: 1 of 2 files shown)


sehyunidaaa:
@adriangohjw could I please get your help to review this if possible! 🙏

adriangohjw (approved):

PR #778, opened 2024-10-15
isom-1586 - establish our standards of support (browser versions)

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1586/establish-our-standards-of-support

## Solution

- check for user agent from the browser
- display the `<Infobox>` in sites page

**Breaking Changes**

  - Details ...

image

### DIFF
--- a/apps/studio/src/components/OutdatedBrowserWarning.tsx
+++ b/apps/studio/src/components/OutdatedBrowserWarning.tsx
@@ -5,7 +5,7 @@ import { isBrowserOutdated } from "../utils/isBrowserOutdated"
 export const OutdatedBrowserWarning: React.FC<
   Partial<React.ComponentProps<typeof Infobox>>
 > = (props) => {
-  if (isBrowserOutdated())
+  if (isBrowserOutdated({ userAgent: window.navigator.userAgent }))
     return (
       <Infobox variant="warning" textStyle="body-2" size="sm" {...props}>
         Your browser is outdated. Please update it for the best experience.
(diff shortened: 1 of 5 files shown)


sehyunidaaa:
Oh @adriangohjw the ticket was for end sites, not the Studio!

adriangohjw:
oh noo hhaha will fix

sehyunidaaa:
Seems like wrongly tagged! I think the goal of the ticket was to:
1. first establish what are the browser version(s) we officially don't support 
2. understand what % of viewership it is, [based on proxy data](https://opengovproducts.slack.com/archives/C06R4DX966P/p1727173531720739) 
3. then explore using a banner (easiest way could be using Site Notification, but otherwise we can have a custom component just for this purpose that is more attention-grabbing) or a fallback page that says "this browser is not supported'

Between a banner and a fallback page, I'm not sure what is more ideal/feasible. Maybe we can start there?

For design, if we go with a banner I would expect it to look like this. Different colour, non-dismisslble, and sits above the site notification/navbar.
image

adriangohjw:
@sehyunidaaa i think the above design looks good!

I was wondering if we should have a page that states the min. supported browser version. However, people who are using unsupported browser version are probably very non-tech-savvy and saying "Chrome min. 90" probably won't help. But WDYT?

sehyunidaaa:
@adriangohjw Yeah I think that would be great too, maybe we can put up a page on the Isomer Guide (or somewhere else that has wider browser support than us) and link out to there?

Also agree that versioning info is necessary, but not sufficient for the less tech-savvy users. I've seen cases where there are 1) official links to download the most updated version and 2) saying things like "use a modern browser".

Let me try to identify where we can place this info, and I can take the copywriting for this before launch. How are we going to approach the banner? I'm thinking maybe we could create a variant of the site notification that is non-dismissible and uses a more striking colour scheme, but def open to other approaches.

PR #779, opened 2024-10-15
Fix link editor bugs

adriangohjw:
## Problem

- Closes https://linear.app/ogp/issue/ISOM-1627/[bug]-link-editor-resets-state-after-selecting-item
- Closes https://linear.app/ogp/issue/ISOM-1626/[bug]-links-over-highlighted-text-causes-a-replacement
- Closes https://linear.app/ogp/issue/ISOM-1629/[bug]-toggling-between-states-in-link-editor-doesnt-reset-button-state

## Solution

**Breaking Changes**

  - Details ...

### recordings

| feature/bug | video |
| - | - |
| do not start modal with "add link" button enabled | (no need) |
| clicking "link" over text does not cause replacement + autofill in the text <br>https://linear.app/ogp/issue/ISOM-1626/[bug]-links-over-highlighted-text-causes-a-replacement | https://github.com/user-attachments/assets/13066fb6-bc6e-4f27-84cd-fa791dd04b55 |
| general validation works fine (previously was generally pretty buggy)<br>https://linear.app/ogp/issue/ISOM-1627/[bug]-link-editor-resets-state-after-selecting-item | https://github.com/user-attachments/assets/2d84de95-083d-4a15-a49c-d458ef7a8b11 |
| reset button state when toggling between tabs<br>https://linear.app/ogp/issue/ISOM-1629/[bug]-toggling-between-states-in-link-editor-doesnt-reset-button-state | https://github.com/user-attachments/assets/ddd8b59d-08b0-46a0-b3b8-80b07aec2721 |
| saved state for each link type so that it doesn't clear between switching link type | https://github.com/user-attachments/assets/5897c17b-44e2-4b54-b83a-8762f3a08fc2 |

### general approach
- add `linkValueHistory` to keep track of states
- using `isFirstRender` to 1) not display error message and state even if it fails validation, while 2) disabling button
- retrieve text selected by users when opening the modal and pre-populate link text with that

### other refactoring
- use constant name instead of hardcoded values e.g.`external` -> `LINK_TYPE_EXTERNAL` for better single source of truth and reduced error
- move `PageLinkElement` and `FileAttachment` into `LinkTypeRadioContent` by default
- make use of design system `isInvalid` and `showFileSize` instead of rebuilding those component again
- move components into standalone file to reduce LOC for each file and easier to maintain

### other fixes
- error messages not showing up as they aren't in `FormControl` (according to OGP design system)
- email field not keeping the value as `mailto:` when it's empty and thus triggering and failing error validation

### notes
- this only partially fix the validation issue but if `LinkHrefEditor` are used in other places, it does not replace their existing UI and UX of their FileAttachment and Page selector (re: https://opengovproducts.slack.com/archives/C06R4DX966P/p1729026621225809)


apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -102,7 +62,15 @@ const LinkEditorModalContent = ({
     onSave(linkText, linkHref),
   )
 
-  const { siteId } = useQueryParse(editPageSchema)
+  // hacky solution to allow setting the correct state of "Add link" CTA button
+  // while not showing error message on initial render for better UX
+  const [isFirstRender, setIsFirstRender] = useState(true)
+  const showLinkTextErrorState: boolean =
+    !isFirstRender && !!errors.linkText?.message
```
seaerchin, line 69: i think i might have fixed this in my pr (or at least for collections) - this triggers because we have a few required properties that are not set. instead of doing this, we can add an empty string though i have to note that this is technically not ideal (user can save w/ empty input)

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -102,7 +62,15 @@ const LinkEditorModalContent = ({
...
+
+  useEffect(() => {
+    void trigger()
+  }, [trigger])
```
seaerchin, line 73: nit: `useLayoutEffect` - this makes it so that this function before repaint and is less likely to mess with other stateful stuff

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -125,48 +93,37 @@ const LinkEditorModalContent = ({
...
-                  }}
-                />
+              errorMessage={errors.linkHref?.message}
+              setErrorMessage={(errorMessage) =>
+                setError("linkHref", {
+                  type: "custom",
+                  message: errorMessage,
+                })
               }
+              clearErrorMessage={() => clearErrors("linkHref")}
```
seaerchin, line 125: sorry for this but i removed all these props from `FileAttachment` so might no longer be required here

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -33,21 +42,40 @@ export const LinkHrefEditor = ({
...
   fileLinkElement,
 }: LinkHrefEditorProps) => {
   const linkType = getLinkHrefType(value)
   const [selectedLinkType, setSelectedLinkType] = useState(linkType)
+  const [linkValueHistory, setLinkValueHistory] = useState({
+    ...INITIAL_LINK_VALUE_HISTORY,
+    ...{
+      [linkType]: value,
+    },
+  } as LinkValueHistory)
```
seaerchin, line 58: not sure why we need the cast here, do you mind explaining it abit more?

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -33,21 +42,40 @@ export const LinkHrefEditor = ({
...
+    errorMessage !== "" &&
+    errorMessage !== undefined
 
   const { getRootProps, getRadioProps } = useRadioGroup({
     name: "link-type",
     defaultValue: linkType,
-    onChange: handleLinkTypeChange,
+    onChange: (newLinkType) => {
+      setIsLinkTypeInitialRender(true)
+      setSelectedLinkType(newLinkType)
```
seaerchin, line 73: this might have been fixed by the other pr when i removed the switch case but i can't confirm

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -33,21 +42,40 @@ export const LinkHrefEditor = ({
...
+      onChange({
+        value: linkValueHistory[newLinkType as keyof LinkValueHistory],
```
seaerchin, line 75: nit: why do we need the cast here?

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -72,8 +100,32 @@ export const LinkHrefEditor = ({
...
+            onChange({
+              value: newLinkTypeContentValue,
+              shouldValidate,
+            })
+            setIsLinkTypeInitialRender(false)
+          }}
+          shouldShowErrorState={shouldShowErrorState}
+          errorMessage={errorMessage}
+          setErrorMessage={setErrorMessage}
+          clearErrorMessage={clearErrorMessage}
```
seaerchin, line 128: we should be able to remove most of these

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -72,8 +100,32 @@ export const LinkHrefEditor = ({
...
         <LinkTypeRadioContent
           selectedLinkType={selectedLinkType}
-          data={value}
-          handleChange={onChange}
+          data={
+            value === ""
+              ? linkValueHistory[selectedLinkType as keyof LinkValueHistory] ||
+                ""
+              : value
+          }
```
seaerchin, line 108: nit: `value || linkValueHistory[selectedLinkType] || ""`

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
adriangohjw: wasn't familiar with best practise of TS - skill issue, will remove haha

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
adriangohjw: wasn't familiar with best practise of TS - skill issue, will remove haha

adriangohjw:
@seaerchin thanks for the review! I think i might create a fresh new branch from `main` and only add the changes, too painful to deconflict with all the changes 👍

PR #781, opened 2024-10-15
fix: build steps for turbo

seaerchin:
## Problem
previously, we used to run the `template#build` command for builds of the template package. however, this is actually wrong. this is because the `build` command will be executed locally (and elsewhere) when we invoke `turbo build`. as part of the publishing flow, this package should only be built **after** the pages have been retrieved from teh database and written to disk.

because of this, we cannot actually use the `build` command to build this package. to preserve the correct dependency graph (and yet have this agnostic of `build`), i changed teh build command for templates to `build:template`

## Solution
1. update the `turbo.json` for `template` so that the task is `build:template` and update the `package.json` accordingly so that the `build` command is removed
2. update root `turbo.json` to add this as a command + also update the `package.json` so that turbo executes the `build:template` command

PR #780, opened 2024-10-15
chore: add build step to ci

seaerchin:
## Problem
previously, the index pages PR was merged it but it had a failing build due to a dynamic import, which wasn't caught. this cuold have been solved w/ a build step

## Solution
1. add build step to ci

PR #782, opened 2024-10-16
feat(collections): add pdf collection item

seaerchin:
## Problem
users need to be able to upload pdf files and have them appear as collection items. users also need to be able to show these collection items on the end user site

## Solution
**Type/data level changes**
1. add a new `CollectionLink` type to our db schema, reflecting that this is a **special** kind of page
2. for components, i updated the `BaseRefPageSchema` which is used for `link` layouts to generate the collection items for links. previously, they extended the `BaseCollectionItemPageSchema` but that schema has quite a few extra options which are never used for the link, which is why i opted to update it to be a simple type. fact check me pls @dcshzj 
3. also export the `LAYOUT_PAGE_MAP` - don't really understand why this isn't exported because the `getLayoutPageSchema` is returning the map's access but as a more general type, which is less ideal

**Logic/code level changes**
1. minor typing fixes to account for this new `CollectionLink` type
2. **big decision** - decided to make a new page + associated endpoints rather than reusing the old one. this is because the old `/page` layout couldn't quite fit this paradigm. this is because the link layout only has a known universe of items, and more importantly, **we disallow people from adding blocks to this page** - which is why it was opted to be a new page rather than reusing existing one.
3. minor refactors for `Preview` and related components. shifted out `ViewportContainer` for reuse across different `Previews`. additionally, shifted out the base `Preview` that accepts a sitemap and one that fetches a sitemap. this allows for us to inject our own sitemap if required, which is needed for the preview to show the collection link being updated. additionally, this allowed us to strengthen our backend typing for `resourceId` on teh `getFullPermalink` (iirc) endpoint so that `resourceId` isn't undefined
4. for jsonforms, added a new `Ref` renderer that uses the existing `LinkEditorModal` together with the styling from `Figma`
5. for the `LinkEditorModal`, in lieu of a proper refactor, i've just extracted out the `editor` dependency and made a new component so that the various tiptap components can draw from the same place
6. for the `FileAttachment` component, i removed the errors and instead, relied on the component's internal error mechanism to display error statse out to the user
7. for the actual `EditLinkPage`, i decided to use an atom to synchronise state across the preview and the drawer. this isn't quite ideal because i do the set state manually on fetch success. instead, we could do something like [this](https://jotai.org/docs/extensions/trpc) but due to time constraint + uncertainty, i chose to do this stupid method first 

## Screenshots 
wip lemme add in sometime later sorrys

## Design TODOs 
(note: this won't impact your review and is mostly styling/copy changes)
1. use col span and grid layout for the previews
2. update copy on pdf
3. (check with @sehyunidaaa if we want to keep the `FormLabel` - i think it's good to keep because our behaviour so far has been to have a label for every block and it helps to hint to the user what this particular upload chunk is used for)
4. (shud we standardise UI for uploading @sehyunidaaa? imo yes - consistent ux is important but we should do this post launch)


seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/782?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#782** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#781** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/781?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#780** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

apps/studio/src/features/editing-experience/components/form-builder/renderers/layouts/JsonFormsVerticalLayout.tsx
```diff
@@ -104,7 +105,7 @@ export function JsonFormsVerticalLayoutRenderer({
   const newElements = getUiSchemaWithGroup(schema, elements)
 
   return (
-    <>
+    <Box w="100%">
```
seaerchin, line 108: bugfix so item will be full width, not sure if there are any downstream impact (ie, across our other jsonforms component) but based off just clicking aorund the UI, it seems ok

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -258,20 +259,8 @@ export const moveResource = async (
...
-    // Provide a default sitemap with a single resource
-    return {
-      id: "0",
-      layout: "content",
-      title: "Root",
-      summary: "",
-      lastModified: new Date().toISOString(),
-      permalink: "/",
-    }
-  }
```
seaerchin, line 273: case removed because we know ahead of time on the frontend if we need a `sitemap` and if we don't, i passed this exact same one (or a collection one).

this case only occurs for page creation, so i strengthened this argument to take `resourceId`

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -97,7 +97,12 @@ export const resourceRouter = router({
           .selectFrom("Resource")
           .where("siteId", "=", Number(siteId))
           .where("id", "=", String(resourceId))
-          .where("Resource.type", "=", "Folder")
+          .where((eb) =>
+            eb.or([
+              eb("Resource.type", "=", "Folder"),
+              eb("Resource.type", "=", "Collection"),
+            ]),
+          )
```
seaerchin, line 105: bugfix - oddly enough, we weren't getting the children previously

seaerchin:
note to self to check and invalidate stale data as required

packages/components/src/types/page.ts
```diff
@@ -33,16 +33,21 @@ const BaseCollectionItemPageSchema = Type.Object({
...
   }),
-])
+  summary: Type.Optional(
+    Type.String({
+      title: "Summary",
+      description: "This will appear on the collection",
+      format: "textarea",
+      maxLength: 120,
+    }),
+  ),
```
dcshzj, line 49: issue: need to check this but I think this might not work because we generate the sitemap using either `subtitle` or `description`.

https://github.com/opengovsg/isomer/blob/bc3e2f2ee00518c686d2da488de5ed7ac2f05149/tooling/build/scripts/generate-sitemap.js#L56-L63

This should be called `description` instead.

apps/studio/src/features/editing-experience/atoms.ts
```diff
@@ -3,3 +3,12 @@ import { atom } from "jotai"
 import type { PendingMoveResource } from "./types"
 
 export const moveResourceAtom = atom<null | PendingMoveResource>(null)
+
+export interface CollectionLinkProps {
+  ref: string
+  summary?: string
```
dcshzj, line 9: issue: same comment as the one in the schema, should be called `description` so that it can be read by the parent collection page.

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/CreateCollectionPageWizardContext.tsx
```diff
@@ -93,7 +95,8 @@ const useCreateCollectionPageWizardContext = ({
       },
       {
         onSuccess: ({ pageId }) => {
-          void router.push(`/sites/${siteId}/pages/${pageId}`)
+          const nextType = type === "page" ? "pages" : "links"
```
dcshzj, line 98: nit: I think this routes might need to be defined as a constant somewhere, especially since we have `folders`, `collections`, `pages` and now `links`.

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
```diff
@@ -0,0 +1,80 @@
...
+    title: "An Isomer Site",
+    summary: "",
+    children: [
+      {
+        id: "collection",
+        permalink: parentPermalink,
+        lastModified: "02 May 2023",
+        layout: "collection",
+        title: parentTitle,
+        summary: "Learn more about PREPARE here.",
```
dcshzj, line 50: nit: can make this an empty string

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
```diff
@@ -0,0 +1,80 @@
...
+        children: children.map((props) => ({
+          ...props,
+          date: "12 May 2023",
```
dcshzj, line 53: issue(non-blocking): the date should be today's date, otherwise might be confusing to the editor

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -255,10 +261,19 @@ export function JsonFormsLinkControl({
   required,
 }: ControlProps) {
   const dataString = data && typeof data === "string" ? data : ""
+  // NOTE: We need to pass in `siteId` but this component is automaticall used by JsonForms
```
dcshzj, line 264:
```suggestion
  // NOTE: We need to pass in `siteId` but this component is automatically used by JsonForms
```

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsRefControl.tsx
```diff
@@ -0,0 +1,127 @@
...
+                variant="clear"
+                colorScheme="critical"
+                aria-label="Remove file"
+                icon={<BiTrash />}
+                onClick={() => handleChange(path, "")}
+              />
+            </>
+          ) : (
+            <>
+              <Text>Choose a page or file to link to this Collection item</Text>
```
dcshzj, line 104: issue(design): I think the copy here is a little weird

```suggestion
              <Text>Choose a page or file to link this Collection item to</Text>
```

Paging @sehyunidaaa

apps/studio/src/features/editing-experience/components/utils.ts
```diff
@@ -74,3 +78,16 @@ export const uploadModifiedAssets = async ({
...
+  sitemap: typeof collectionSitemap,
+  title = "Your filename",
+) => {
+  return {
+    ...sitemap,
+    children: sitemap.children.map(({ children, ...rest }) => ({
+      ...rest,
+      children: children.map((props) => ({ ...props, title })),
+    })),
+  } as IsomerGeneratedSiteProps["siteMap"]
```
dcshzj, line 92: nit: can this be defined as part of the function return type?

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useAtomValue } from "jotai"
+
+import { useQueryParse } from "~/hooks/useQueryParse"
+import { editLinkSchema } from "~/pages/sites/[siteId]/links/[linkId]"
+import { trpc } from "~/utils/trpc"
+import { linkAtom } from "../atoms"
+import PreviewWithoutSitemap from "./PreviewWithoutSitemap"
+import { ViewportContainer } from "./ViewportContainer"
+
+export const EditLinkPreview = (): JSX.Element => {
```
dcshzj, line 11: nit: can rename this to `EditCollectionLinkPreview`, otherwise can be confusing with normal link editing

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/DetailsScreen.tsx
```diff
@@ -142,7 +142,10 @@ export const CreateCollectionPageDetailsScreen = () => {
               </FormControl>
 
               {/* Section 2: Page URL */}
-              <FormControl isInvalid={!!errors.permalink}>
+              <FormControl
+                isInvalid={!!errors.permalink}
+                display={type === "link" ? "None" : "auto"}
```
dcshzj, line 147: nit: should be lowercase?

```suggestion
                display={type === "link" ? "none" : "auto"}
```

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
```diff
@@ -255,10 +261,19 @@ export function JsonFormsLinkControl({
...
+  // so we are unable to pass props down
+  const { siteId } = useQueryParse(siteSchema)
+  const dummyFile = new File(
+    [],
+    // NOTE: Technically guaranteed since our s3 filepath has a format of `/<site>/.../<filename>`
+    dataString.split("/").at(-1) ?? "Uploaded file",
```
dcshzj, line 270: nit: hmm we can't guarantee this right, because this covers normal links and files, so the link may not necessarily be a File.

But this is covered under the LinkHrefEditor so I'm guessing it should be fine?

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -97,7 +97,12 @@ export const resourceRouter = router({
           .selectFrom("Resource")
           .where("siteId", "=", Number(siteId))
           .where("id", "=", String(resourceId))
-          .where("Resource.type", "=", "Folder")
+          .where((eb) =>
+            eb.or([
+              eb("Resource.type", "=", "Folder"),
+              eb("Resource.type", "=", "Collection"),
```
dcshzj, line 103: nit: should use `ResourceType`

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -139,6 +139,7 @@ export const getPageById = (
         eb("type", "=", "CollectionPage"),
         eb("type", "=", "RootPage"),
         eb("type", "=", "IndexPage"),
+        eb("type", "=", "CollectionLink"),
```
dcshzj, line 142: nit: should use `ResourceType`.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -299,6 +288,7 @@ export const getLocalisedSitemap = async (
               "Page",
               "CollectionPage",
               "RootPage",
+              "Collection",
```
dcshzj, line 291: nit: should use `ResourceType`.

apps/studio/src/features/editing-experience/atoms.ts
seaerchin: oh yes, updated

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/CreateCollectionPageWizardContext.tsx
seaerchin: actually this is already defined as `ResourceType`

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
seaerchin: done

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
seaerchin: actually this didn't quite work - oddly enough, only may worked. i'm not too sure why because this format `dd MMM yyyy` seems to be valid? lmk if you have any idea what could be the issue

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
seaerchin: thx for spotting

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsRefControl.tsx
seaerchin: fixedc

apps/studio/src/features/editing-experience/components/utils.ts
seaerchin: it needs the cast because the json tyhpe cannot be infered properly - it's inferred as `string` rather than the most specific but becaus eit's a readonly constant, we know it's correct (hence the cast)

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
seaerchin: sounds good

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
seaerchin: yeap we can't, but this is wrapped under the radio button as you've rightly pointed out so this case is technically impossible.

we could also use a regex this issue as the link itself adhere to strict formatting

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: not doing - these have the types automatically inferred so very little value add. ie, we can't do `Collections` (for eg) here, because it's a type error

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const [{ parent }] = trpc.resource.getParentOf.useSuspenseQuery({
+    resourceId: String(linkId),
+    siteId,
+  })
+
+  const [children] = trpc.collection.getSiblingsOf.useSuspenseQuery({
+    siteId,
+    resourceId: linkId,
+  })
```
dcshzj, line 30: issue: do we need to actually show all the siblings of the collection? I thought we were only intending to show only one, which is the exact item that the user is editing.

My concern is that due to the way we are sorting (type then title, from the trpc procedure), we cannot guarantee that the actual item that the user is editing will be visible in the preview.

My suggestion is to just skip showing of the siblings here when editing a specific link, and only show that single item within the context of the collection.

packages/components/src/utils/validation.ts
```diff
@@ -0,0 +1,11 @@
+export const ALLOWED_URL_REGEXES = {
+  external: "^https:\\/\\/",
+  mail: "^mailto:",
+  internal: "^\\[.*\\]$",
+  files: ".*\\.pdf",
```
dcshzj, line 5: issue: hmm I think we cannot enforce that it must be PDFs here, because we can potentially support other formats like docx, xlsx, etc.

I suggest perhaps we should consider the following format:
- External: `https://`
- Mail: `mailto:`
- Internal: `[resource:(d+):(d+)]` (I think I wrote this down in `getReferenceLinkHref`)
- Files: `/(d+)/...` (because files, like images, will be in the format `/${siteId}/${uuid}/${fileName}`)

Alternatively, since this is inside the components package, we can be less prescriptive about it and let Studio handle the pattern validation for us. The components will just assume that the link is valid and pass it through to the `href` attribute.

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/CreateCollectionPageWizardContext.tsx
dcshzj: Ah found it, this should be using `getResourceSubpath` so we can be consistent.

packages/components/src/utils/validation.ts
seaerchin: quite hard to do this in studio cos we use our existing validator for the blob **and not zod**

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
seaerchin: removed!

packages/components/src/types/page.ts
seaerchin: fixed

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/PreviewLayout.tsx
```diff
@@ -57,7 +58,7 @@ export const PreviewLayout = (): JSX.Element => {
             justify="center"
             whiteSpace="pre"
           >
-            {`You're previewing a ${currentType === "link" ? "PDF " : ""}collection page`}
+            {`You're previewing a ${currentType === ResourceType.CollectionLink ? "collection link" : "collection page"}`}
```
dcshzj, line 61: nit: can move the `collection` part out
```suggestion
            {`You're previewing a collection ${currentType === ResourceType.CollectionLink ? "link" : "page"}`}
```

packages/components/src/utils/validation.ts
dcshzj: Sure hmm but for files need to allow other kinds of files beyond PDF

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/PreviewLayout.tsx
seaerchin: done

PR #784, opened 2024-10-16
chore: update seed script to insert theme

seaerchin:
## Problem
as per title
## Solution
copy over theme from db

### DIFF
--- a/apps/studio/prisma/seed.ts
+++ b/apps/studio/prisma/seed.ts
@@ -162,10 +162,6 @@ async function main() {
               pressed: "#00283b",
             },
           },
-          "site-primary-100": "#fef4e8",
-          "site-primary-200": "#ffeec2",
-          "site-primary-default": "#f78f1e",
-          "site-secondary-default": "#877664",
         },
       }),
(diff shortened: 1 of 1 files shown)


apps/studio/prisma/seed.ts
```diff
@@ -146,7 +146,28 @@ async function main() {
...
+            interaction: {
+              hover: "#002e44",
+              default: "#00405f",
+              pressed: "#00283b",
+            },
+          },
+          "site-primary-100": "#fef4e8",
+          "site-primary-200": "#ffeec2",
+          "site-primary-default": "#f78f1e",
+          "site-secondary-default": "#877664",
```
dcshzj, line 168: nit: I think these are no longer needed now since we have shifted all our components over to the colour tokens

dcshzj (approved):

adriangohjw (approved):

apps/studio/prisma/seed.ts
seaerchin: removed

PR #786, opened 2024-10-17
Origin/adriangohjw/isom 1586 establish our standards of support

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1586/establish-our-standards-of-support

## Solution

- generate supportedBrowsers that achieved good audience coverage
  - image
  - Ref: https://browsersl.ist/#q=defaults%0Achrome+%3E%3D+64%2C+edge+%3E%3D+79%2C+firefox+%3E%3D+67%2C+opera+%3E%3D+51%2C+safari+%3E%3D+12%2C%0Anot+dead%0A

- check userAgent via `navigator.userAgent` and check it against supportedBrowser.js - if unsupported, shows the banner

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

image

## Tests

- Check the new storybook `?path=/docs/next-internal-components-unsupportedbrowserbanner--docs`

**New scripts**:

- `script` : `"supportedBrowsers": "echo \"export default $(browserslist-useragent-regexp --allowHigherVersions);\" > supportedBrowsers.js"`
  - generate `supportedBrowsers.js` which will be referenced when deciding if we should show the banner 

**New dependencies**:

- `dependency` : `browserslist-useragent-regexp`


seaerchin (commented):
due to inexperience on my end, i went to research abit on this topic and found some interesting stuff. according to the mdn docs [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent), there are cases where the user agent outright lies. due to gotchas like this, it's hard to determine if the regex is correct.

we could consider using the official browserlist utility available [here](https://github.com/browserslist/browserslist-useragent) to check the agent .

if you feel confident in this change, i'm ok with approving it. wanted to just clarify some stuff here! i do, however, acknowledge that the blast radius of this change is pretty small if the browser agent is wrongly recognised (assuming false positive, ie i don't display the banner when i'm supposed to) because such browsers are already super rare owing to age but given that there's an official utiilty which seems to ease maintainability and is able to use our `package.json` for the actual check, it seems to be worth the exttra effort to investigate

adriangohjw:
@seaerchin 

> there are cases where the user agent outright lies.

yup i think i have some experience with this. But IMO it's not an area of concern since 1) they are usually some forms of bots, 2) people who we want to show the banner with are usually less tech savvy and probably uses default browsers in their phones

> we could consider using the official browserlist utility available [here](https://github.com/browserslist/browserslist-useragent) to check the agent .

sounds good, i didn't see this. Lemme look into it, thanks!

seaerchin:
Think we're talking about slightly different things! Mdn said that chrome, for example, reports itself as being safari so we might get a false positive where we don't show banner when we need to. Quoting from mdn:

> Most browsers set the name and version in the format BrowserName/VersionNumber. But as the name is not the only information in a user agent string that is in that format, you can not discover the name of the browser, you can only check if the name you are looking for exists. But note that some browsers are lying: Chrome for example reports both as Chrome and Safari. So to detect Safari you have to check for the Safari string and the absence of the Chrome string, Chromium often reports itself as Chrome too or Seamonkey sometimes reports itself as Firefox.

adriangohjw:
@seaerchin 

I looked into browserslist-useragent and found it was originally designed for server-side environments so it relies on process, which is a Node.js global, and that won't be available client-side in Next.js. browserslist-useragent-regex (current implementation) is recommended as a workaround if it's for client-side

You are right that the user agent isn't the most reliable. ideally we check if a particular technology is present using something like Modernizr but I haven't got the chance to look into it, plus the risk is quite small, so I think current implementation is alright given the resource constraints 

I will also wait for my other PR regarding adding vitest to the Components folder to work - we can then write some tests against some user agents to verify it

seaerchin (approved):
sounds good, lgtm! didn't realise that it was server only (maybe too hasty on my end), thanks for catcching!

adriangohjw:
> sounds good, lgtm! didn't realise that it was server only (maybe too hasty on my end), thanks for catcching!

haha yeah don't worry about it, i also only realized after i faced a bug when trying to implement that

PR #787, opened 2024-10-18
isom-1286 add indexes for tables

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1286/add-indexes-for-tables

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- Removed 2 composite indexes from `Resource` as they are already covered and overlapping with another index
- Added index on `type` for `Resource`

### DIFF
--- a/apps/studio/prisma/migrations/20241018013545_resource_remove_overlapping_indexes/migration.sql
+++ b/apps/studio/prisma/migrations/20241018013545_resource_remove_overlapping_indexes/migration.sql
@@ -1,0 +1,5 @@
+-- DropIndex
+DROP INDEX "Resource_siteId_id_idx";
+
+-- DropIndex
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
lgtm. prisma indexes also have the `sort` [option](https://www.prisma.io/docs/orm/prisma-schema/data-model/indexes#configuring-the-index-sort-order-with-sort) which might help for guaranteeing a stable retrieval order of items. 

i was thinking we could index `type, id` then add the `sort` in but as a first pass this already looks good to me.

adriangohjw:
@seaerchin thanks for sharing!

yeah i think at current stage no need too much optimization - from experience, better to look at usage data. I have looked into AWS RDS and saw that we already have Performance Insights set up so lets monitor along

PR #789, opened 2024-10-18
fix(links): add validation to components that are links or refs

seaerchin:
## Problem
We previously did not sanitise our URLs at all. this meant that we were allowign insecure URLs on our frontend/backend. in order to alleviate this **as a first pass**, i added regex validation

## Solution
1. add URL validation for ext links (https at start of URL), internal links (verify that the format is `[...]`), files (end in `.pdf`) and mails (begins with `mailto:`)
2. for our frontend, i added the https left addon and appended if it didn't exist. also soe minor fixes regarding otggle and state reset.
    - for some reason, using the `switch` statement still renders the component (and triggers validation). using the `if` statement doesn't render and has the desired behaviour

## Screenshots

https://github.com/user-attachments/assets/0c63b76b-9838-4a47-ae12-e233833b3628


adriangohjw (approved):
as discussed in person, there are some ways to improve this but will merge first since it's blocking

will fix/deconflict/improve in https://github.com/opengovsg/isomer/pull/779 after PDF collections feat merged

PR #790, opened 2024-10-18
feat: add Datadog RUM to template and components

dcshzj:
## Problem

We don't have RUM set up on our template, so we are not able to monitor the user behaviour on the actual Isomer Next sites.

## Solution

**Breaking Changes**

**Features**:

- Set up Datadog RUM monitoring as a component, and use the component inside the Skeleton layout.
- The sampling rate is configured using the environment variable `NEXT_PUBLIC_ISOMER_NEXT_RUM_SAMPLING_RATE`, which is injected at build time to feature flag the level of sampling.


packages/components/src/templates/next/components/internal/DatadogRum/DatadogRum.tsx
```diff
@@ -0,0 +1,30 @@
...
+    process.env.NEXT_PUBLIC_ISOMER_NEXT_RUM_SAMPLING_RATE ?? 0,
+  ),
+  sessionReplaySampleRate: Number(
+    process.env.NEXT_PUBLIC_ISOMER_NEXT_RUM_SAMPLING_RATE ?? 0,
+  ),
+  trackUserInteractions: true,
+  trackResources: true,
+  trackLongTasks: true,
+  defaultPrivacyLevel: "allow",
+})
```
adriangohjw, line 24: should we pass in `production` as ENV too? or check that its production/staging before rendering. so as to not dirty the data + not have it run on dev and incur unncessary cost

apps/studio/public/assets/css/preview-tw.css
adriangohjw, line 1: ooo why this file will change ah?

adriangohjw (approved):

packages/components/src/templates/next/components/internal/DatadogRum/DatadogRum.tsx
dcshzj: Yup, the env is provided earlier on line 13, but I also added a check to only render this component if the site is not on staging

apps/studio/public/assets/css/preview-tw.css
dcshzj: This one is an automatic update on build

PR #794, opened 2024-10-20
isom-1619 keystatistics should have a link out to a page

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1619/keystatistics-should-have-a-link-out-to-a-page

## Solution

- update the schema
- update Component to display (+ stories)
  - @sehyunidaaa - need your design input on this, but my current implementation for mobile is to display a button (similar to Hero Banner) at the end of all the statistics 
- note:
  - current CTA is hardcoded to be "Learn more"
  - current implementation does not limit users the option to only select internal page for linking. That will require updating how link options are implemented, which is pending changes on other PRs at the moment

**Breaking Changes**

  - Details ...

**Features**

## Screenshots

image

image


sehyunidaaa:
@adriangohjw Thanks for taking a stab at this first! For the designs I've added some more details on the ticket https://linear.app/ogp/issue/ISOM-1619/keystatistics-should-have-a-link-out-to-a-page

Direct link to Figma if you need it: https://www.figma.com/design/rsKQdmtyoOWawyAv1LkJJK/(New%2C-WIP)-Isomer-Next-Component-Library?node-id=5937-458&t=MA0QhpNyIsYxlbvl-1

adriangohjw:
@sehyunidaaa currently the gap is `48px` but Figma states `40px (no change)` - can i clarify which should we follow?

image

sehyunidaaa:
@adriangohjw Ohh no okie we can keep it to 48px! It's minor. Will update Figma instead

adriangohjw:
@sehyunidaaa done 👍

sehyunidaaa:
@adriangohjw there were some discrepancies between figma vs. storybook. No need to change any of the gaps, I've updated[ the redline on figma](https://www.figma.com/design/rsKQdmtyoOWawyAv1LkJJK/(New%2C-WIP)-Isomer-Next-Component-Library?node-id=5974-1145&t=MA0QhpNyIsYxlbvl-1)

packages/components/src/interfaces/complex/KeyStatistics.ts
```diff
@@ -28,6 +28,13 @@ export const KeyStatisticsSchema = Type.Object(
         maxItems: 4,
       },
     ),
+    url: Type.Optional(
+      Type.String({
+        title: "Link destination",
+        description: "When this is clicked, open:",
+        format: "link",
```
seaerchin, line 35: note: the `link` format here implies that this can accept **any one of**: `internal link/ext link/image/file`. 

from what this component sounds liek it's meant to be used for, we should probably dissuade the last 2? but it might be very annoying to fix tbh - will let sehyun make the call. imo, opening files/images are fine in my book becauase they could convey more meaningful info (ie, this `url` is a see more for the key stats) and is also a pattern that we find elsewhere (eg: see more on current financials downloads a pdf report for companies)

packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
```diff
@@ -78,6 +87,19 @@ const KeyStatistics = ({ title, statistics, layout }: KeyStatisticsProps) => {
           </div>
         ))}
       </div>
+
+      {url !== undefined && url !== "" && (
```
seaerchin, line 91: nit: `!!url && ...`

packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
```diff
@@ -78,6 +87,19 @@ const KeyStatistics = ({ title, statistics, layout }: KeyStatisticsProps) => {
...
+        <div className={compoundStyles.urlButtonContainer()}>
+          <LinkButton
+            href={getReferenceLinkHref(url, site.siteMap)}
```
seaerchin, line 94: **note:** i didn't look into `getReferenceLinkHref` but the name here sounds like it shuold only allow internal links, which have a format of `[ref,$siteId,$resourceId]`. this is in direct contrast to the link format, which also allows normal urls + `mailto`

seaerchin (commented):
minor concerns but i skimmed

packages/components/src/interfaces/complex/KeyStatistics.ts
sehyunidaaa: Oh Isn't it `mailto` instead of `image`? 

`Internal link` is what MOH needs now so by right we could start with just that and expand later, but then we'd need to change the link selector on studio right. Doesn't seem like worth it

Technically there isn't a huge risk to supporting all 4 imo, like you said you could link a PDF report or email someone about the stats

If it is indeed `image` instead of `mailto` err that one I'd be a bit worried, strange to link an image to a block

adriangohjw: > note: the `link` format here implies that this can accept **any one of**: `internal link/ext link/image/file`. 

Yup! Noted in the PR that it was intentional (for now) since there're quite a number of pending PR and changes made to this link component so I don't want to edit the component to have the disabled option as it will have many conflicting changes anyway

packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
adriangohjw: Yeah I haven't looked very deep into this too but copied this from infobar component

packages/components/src/interfaces/complex/KeyStatistics.ts
seaerchin: update - we can use the `linkType` prop in my PR to do this

adriangohjw: re: https://github.com/opengovsg/isomer/pull/782

packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
seaerchin: issue: i went to look at this function definition and this function doesn't quite work here **unless we strictly allow internal links**. this is because this function does a regex for `[resource, <number>, < number>]` then does a lookup for this number to generate the permalink.

seaerchin (commented):
i think the `getReferenceLink` doesn't quite work if we waznt to allow external links too. we'd need to check for a protocol then conditionally do `getReferenceLink` or just the ext url

packages/components/src/templates/next/components/complex/KeyStatistics/KeyStatistics.tsx
adriangohjw: thanks! will fix together once the other PR is merged and we can pass in linktypes

seaerchin (approved):

PR #796, opened 2024-10-21
isom-1637 bug html tags from tiptap editor not stripped

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1637/bug-html-tags-from-tiptap-editor-not-stripped

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add a check to sanitize and remove html tags before displaying for prose

### Before

image

### After

image


packages/components/src/utils/getTextAsHtml.ts
```diff
@@ -22,13 +22,26 @@ interface GetTextAsHtmlArgs {
...
+
+// Strips HTML tags using the DOM
+// Using regexp works, but just in case we are dealing with complex HTML
+// or need to handle specific edge cases (like script or style tags),
+// we want to use a more robust solution like using the DOM
+// We want to prevent user-injected HTML tags from breaking the formatting
+function stripHtmlTagsUsingDOM(input: string): string {
+  const div = document.createElement("div")
+  div.innerHTML = input
+  return div.textContent || div.innerText || ""
```
seaerchin, line 36: honestly this feels very iffy AHHAHAHA but i guess we can accept this first? the only issue is that text that users input as html tags knowingly might get stripped but we can live with that for now ba

seaerchin (approved):

packages/components/src/utils/getTextAsHtml.ts
adriangohjw: ya i know haha but this is the most straightforward solution i found that doesn't require complex regex-ish approach or 3rd party library

https://chatgpt.com/share/6717caab-e52c-800f-b39b-986f5decbf65

PR #798, opened 2024-10-21
Fix unsupported browser flashing bug

adriangohjw:
## Problem

https://github.com/user-attachments/assets/a7a2d336-921a-4664-8be0-775b58a0f3f4

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- use `useState` and `useEffect` to retrieve useragent again when `window` loads after the first render
- add more explicit checks for supportedBrowser for empty string and undefined value


adriangohjw:
if this one doesn't work, we should temporarily roll this back for launch

adriangohjw:
re: https://github.com/opengovsg/isomer/pull/786

packages/components/src/templates/next/components/internal/UnsupportedBrowserBanner/UnsupportedBrowserBanner.tsx
```diff
@@ -19,13 +20,32 @@ const getUserAgent = (): string | undefined => {
 const isSupportedBrowser = ({
   userAgent,
 }: UnsupportedBrowserBannerProps): boolean => {
+  if (userAgent === undefined) return true
+  if (userAgent === "") return true
```
seaerchin, line 24: why would both of these be `true`? i'm assumign the reason here is that these are bots and we just ignore? additionally, should we also ignore `null`?

packages/components/src/templates/next/components/internal/UnsupportedBrowserBanner/UnsupportedBrowserBanner.tsx
```diff
@@ -19,13 +20,32 @@ const getUserAgent = (): string | undefined => {
...
+    // The check for typeof window and navigator ensures this only runs in browser environments, not during server-side rendering
+    // We use setNavigatorUserAgent to update the state, which will trigger a re-render with the correct user agent
+    if (
+      !initialUserAgent &&
+      typeof window !== "undefined" &&
+      typeof navigator !== "undefined"
+    ) {
+      setNavigatorUserAgent(navigator.userAgent)
+    }
+  }, [initialUserAgent])
```
seaerchin, line 46: i think since `navigator.userAgent` is a derived value, we might want to call `useMemo` rather htan `useEffect` here?

adriangohjw: if im not wrong, `useMemo` is for memoizing computed values to improve performance, not for running side effects or updating state (which we are doing here in this case). not sure if im missing out anything?

seaerchin: just from looking, it looks like [this](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state) might be what we're doing!

i'm not entirely sure here because of `window` and the fact that it might live outside of react but if we pass it to `useMemo` as a dep, it should function like a computed value

packages/components/src/templates/next/components/internal/UnsupportedBrowserBanner/UnsupportedBrowserBanner.tsx
adriangohjw: anyway, don;t think we need `null` but added in just in case haha

packages/components/src/templates/next/components/internal/UnsupportedBrowserBanner/UnsupportedBrowserBanner.tsx
adriangohjw: @seaerchin `navigator.userAgent` is a derived value but the component wouldn't know it has changed since it's not dependent on that. 

This current approach is to indirectly force a re-render after the first render so that we can ideally retrieve the window.navigator.userAgent (if its available)

adriangohjw: unless you mean removing **both** `useState` and `useEffect` altogether and doing something like this?

image

seaerchin: yea i was thinking something like 
```ts
const navigatorUserAgent = useMemo(() => 
    initialUserAgent || 
    typeof window !== undefined && typeof navigator !== undefined && navigator.userAgent ||"", 
    [window, navigator, initialUserAgent]
)}
```

cos this would have the same rendering/re-rendering logic as the `useEffect + setState` above wtihout the extra logical hops

seaerchin (approved):
lgtm to unblock first, we can tidy up the code later

PR #799, opened 2024-10-22
fix - datadog RUM check for process first

adriangohjw:
## Problem

Storybook Layout section breaks when datadog is loaded (re: https://github.com/opengovsg/isomer/pull/790)

Datadog is using `process.env` which is unavailable for client side

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- check for `process` is defined first (not ideal but Datadog RUM is a launch thing which will be removed soon)

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-10-22 at 11 45 54 AM


seaerchin (commented):
can we do a conditional init? `env === "production" && !!process && datadogRum.init`

adriangohjw:
> can we do a conditional init? `env === "production" && !!process && datadogRum.init`

`env` can also only be retrieved via `process.env` too which require `process`
but i think you have a point so im doing a `typeof process !== "undefined" && process.env.NODE_ENV === "production"` as a conditional check for init DD

seaerchin (approved):

PR #800, opened 2024-10-22
isom-1624 remove year filter for non date variant of collections

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1624/remove-year-filter-for-non-date-variant-of-collections

## Solution

**Breaking Changes**

  - Details ...

### Changes

- add more stories for Collection and CollectionCard (+ refactor to be leaner and more readable)
- add an option to pass in `shouldShowDate` for collection results
- update `getAvailableFilters` to inject "Not specified" year option
- update `getFilteredItems` to handle "Not specified" year option
- refactor utils to use const instead for single source of truth
- update copywriting of search results from "XXX search results" to "XXX articles"

_(refer to storybook changes for more UI)_

image


sehyunidaaa (changes requested):
@adriangohjw Left some comments on Chromatic! :)

packages/components/src/templates/next/layouts/Collection/utils.ts
```diff
@@ -24,12 +30,20 @@ export const getAvailableFilters = (
...
+    label,
+    count,
+  }))
+
   const availableFilters: FilterType[] = [
     {
-      id: "category",
+      id: FILTER_ID_CATEGORY,
       label: "Category",
       items: Object.entries(categories).map(([label, count]) => ({
```
dcshzj, line 48: nit: btw I think we should sort the labels by ascending alphabetical order

packages/components/src/templates/next/layouts/Collection/utils.ts
```diff
@@ -38,13 +52,22 @@ export const getAvailableFilters = (
...
-        count,
-      })),
+      items:
+        // do not show "not specified" option if all items have undefined dates
+        yearFilterItems.length === 0
+          ? []
+          : numberOfUndefinedDates === 0
+            ? yearFilterItems
+            : [
+                ...yearFilterItems,
```
dcshzj, line 64: nit: same here regarding the ordering of the years, should be in ascending order

packages/components/src/templates/next/layouts/Collection/utils.ts
adriangohjw: or should it be based on "count" descreasing? since it's likely more relevant?

packages/components/src/templates/next/layouts/Collection/utils.ts
adriangohjw: okay!

anyway, will put this in backlog since its unrelated to this PR

packages/components/src/templates/next/layouts/Collection/CollectionClient.tsx
```diff
@@ -119,6 +120,7 @@ const CollectionClient = ({
               paginatedItems={paginatedItems}
               searchValue={searchValue}
               totalCount={totalCount}
+              shouldShowDate={useMemo(() => shouldShowDate(items), [items])}
```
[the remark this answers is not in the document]
adriangohjw: sounds good, will do

dcshzj (approved):

sehyunidaaa (approved):
l g t m !

PR #811, opened 2024-10-24
refactor(settings): edit settings flow

seaerchin:
## Problem
We want to change the edit page flow so that users get to update the page settings in a modal and the meta settings within the page itself

## Solution
1. shift out code for updating page settings into a modal
2. add the trigger for the modal to be an `atom` so that the modal can fully own its own state and be shared easily via the atom
3. created a new endpoint for updating the meta because we were previously doing logical checks and overriding. 
    a. this was done so that the editing of page settings did not override meta due to the checks

## Video

https://github.com/user-attachments/assets/557b00c8-724b-417d-886d-eb33cb1790cd


adriangohjw (approved):

sehyunidaaa:
reviewed chromatic!

nit: on the video there seems to be a short flash before the Page Settings modal appears. is this a local environment behaviour?

PR #813, opened 2024-10-24
chore(migrations): fix migrations from prisma

seaerchin:
## Problem
1. we didn't squash our migrations so what happened was that we get migrations that were erroneous and being ran in prod

## Solution
1. squash the link migration as it was previously added, removed and then re-added which led to conflictss

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -40,7 +40,8 @@
     "build-storybook": "storybook build",
     "clean": "git clean -xdf .next .turbo node_modules build",
     "jump:vapt": "source .ssh/.env.vapt && ssh -L 5433:$DB_HOST:5432 $SSH_USER@$SSH_HOST -i .ssh/studio-vapt-bastion.pem",
-    "migrate:vapt": "source .ssh/.env.vapt && npx prisma migrate deploy"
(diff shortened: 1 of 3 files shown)


apps/studio/package.json
```diff
@@ -40,7 +40,8 @@
     "build-storybook": "storybook build",
     "clean": "git clean -xdf .next .turbo node_modules build",
     "jump:vapt": "source .ssh/.env.vapt && ssh -L 5433:$DB_HOST:5432 $SSH_USER@$SSH_HOST -i .ssh/studio-vapt-bastion.pem",
-    "migrate:vapt": "source .ssh/.env.vapt && npx prisma migrate deploy"
+    "migrate:vapt": "source .ssh/.env.vapt && npx prisma migrate deploy",
+    "jump": "source .ssh/.env.prod && ssh -L 5433:$DB_HOST:5432 $SSH_USER@$SSH_HOST -i .ssh/studio-prod-bastion.pem"
```
dcshzj, line 44: nit: I think calling it `jump:prod` might be safer and more explicit

dcshzj (approved):

PR #821, opened 2024-10-25
add NEXT_PUBLIC_APP_ENV to .env.example

adriangohjw:
`NEXT_PUBLIC_APP_ENV` not added to .env which lead to error since it's a required env


seaerchin (approved):
TYVM!!!

PR #822, opened 2024-10-25
fix: update dd env and service

seaerchin:
## Problem
our task definition has the wrong values set for dd keys

## Solution
update tsak def


adriangohjw (commented):
im assuming <ENV> is a valid syntax?

seaerchin:
@adriangohjw we do a `sed` replacement in our workflow!

adriangohjw (approved):

PR #829, opened 2024-10-26
Keystats limit linktypes option to internal page

adriangohjw:
## Problem

KeyStats should not allow for any other links except for internal page link

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- pass in `linkTypes` to KeyStats schema on component
- add custom `IsomerExtendedLinkJsonSchema` and also check for presence of `linkTypes`
- add `filterLinkTypes` for reusability down the road - fallback to default all options if undefined or not specified

## Screenshots

image

### DIFF
--- a/packages/components/src/interfaces/complex/KeyStatistics.ts
+++ b/packages/components/src/interfaces/complex/KeyStatistics.ts
@@ -33,6 +33,7 @@ export const KeyStatisticsSchema = Type.Object(
         title: "Link destination",
         description: "When this is clicked, open:",
         format: "link",
+        linkTypes: ["page"],
       }),
     ),
   },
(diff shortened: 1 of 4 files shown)


adriangohjw:
Note: I'm not fully confident this is the best approach for the current codebase, but based on the existing structure, I think this is the cleanest and most consistent way, though I could be wrong.

packages/components/src/interfaces/complex/KeyStatistics.ts
```diff
@@ -33,6 +33,7 @@ export const KeyStatisticsSchema = Type.Object(
         title: "Link destination",
         description: "When this is clicked, open:",
         format: "link",
+        linkTypes: ["page"],
```
adriangohjw, line 36: wondering if we should make this code a type on its own because `linkTypes` is a custom prop that's unusual of that of String

adriangohjw:
too stale - will update in another PR separaetly

PR #831, opened 2024-10-26
isom-1514 - add highlighted text for link insertion in tiptap

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1626/[bug]-links-over-highlighted-text-causes-a-replacement

## Solution

**Breaking Changes**

  - Details ...

https://github.com/user-attachments/assets/19d5f9bf-58e6-44ac-9683-48f8a8af12e7


seaerchin:
1 thing i tried but found interesting was that using this across **multi line** text converts them into a single line link. i do think this is a limitation of how we're doing things now (we get the `textContent` because we're forced to show as single line input).

we could consider an alternate flow, where **we disable** editing of link text (see attached image) in similar fashion to slack, which solves this quite nicely. Principally speaking, i think allowing them to edit the text **in tiptap** doesn't have much value because tiptap is already a text editor - this is just double work being done leading to bad ux. we should just disable the input if it's called from tiptap but this might not be an easy change to make. looping in @sehyunidaaa here for ideas - i think i'm willing to accept this because it fixes known issues and replaces the problem with a less glaring one, but just bouncing ideas off you so if you're amenable to disabling input, we can file a ticket


image

https://github.com/user-attachments/assets/51d88585-2c82-4744-9455-014875b9abbe

sehyunidaaa:
@seaerchin can you clarify what you mean by this?
> Principally speaking, i think allowing them to edit the text in tiptap doesn't have much value because tiptap is already a text editor

Also, when I edit links on Slack, I can still edit both the link text and the URL. Not entirely sure if I understand your proposal here! 
image

seaerchin:
> @seaerchin can you clarify what you mean by this?
> 
> > Principally speaking, i think allowing them to edit the text in tiptap doesn't have much value because tiptap is already a text editor

typo, i meant that allowing htem to edit hte text inside the modal doesn't have much value. re: editing - that's only true for **single line links**! i tried with multi line links and you can't edit them

adriangohjw:
I do agree that the better solution will be to disable add links if it spans across multiple lines e.g. how Notion does it

IMO this make up a very small subset of the editing experience and not worth the extra effort to look into

sehyunidaaa:
Tbh I don't have strong opinions on what's shown on Chin's screen recording, agreed it's not a common operation anyway. If we disable link on multiple links highlight, wouldn't that be more complex? Not sure if it's a case we have to specifically engineer for

seaerchin (approved):
Lgtm since sehyun no strong opinion HAHA I was hoping we'd be able to check for the presence of new lines but I think feeding it into the component now to disable might be hard so let's go with this first

PR #832, opened 2024-10-26
refactor - use const for linktype

adriangohjw:
## Problem

Currently values are hardcoded
however, this means it's easy to miss out something when changing the value - leading to more time needed to make changes and higher likelihood of bugs

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- declare const
- single source of truth by importing them everywhere instead of hardcoding values


packages/components/src/types/components.ts
```diff
@@ -16,3 +16,8 @@ export type IsomerComponent = Static<typeof IsomerComponentsSchemas>
 export type IsomerComponentTypes =
   | keyof typeof IsomerComplexComponentsMap
   | "prose"
+
+export const LINK_TYPE_PAGE = "page"
+export const LINK_TYPE_EXTERNAL = "external"
+export const LINK_TYPE_FILE = "file"
+export const LINK_TYPE_EMAIL = "email"
```
dcshzj, line 23: issue: this is a Studio-specific construct and should be added there. Components are flexible in the sense that it can take in any string (since they are forwarded verbatim to the `href` attribute), subject to the pattern defined for these 4 types of links.

apps/studio/src/features/editing-experience/components/LinkEditor/utils.ts
```diff
@@ -1,23 +1,32 @@
-export const getLinkHrefType = (href: string | undefined) => {
+import {
+  LINK_TYPE_EMAIL,
+  LINK_TYPE_EXTERNAL,
+  LINK_TYPE_FILE,
+  LINK_TYPE_PAGE,
```
dcshzj, line 5: issue: we should make use of a JavaScript object instead to define the link types. For example:

```ts
export const LINK_TYPE = {
    Email: "Email",
    External: "External",
    File: "File",
    Page: "Page"
}
```

That way, we can use things like `LINK_TYPE.Email` to represent an email link, and to perform exhaustiveness checks.

packages/components/src/types/components.ts
adriangohjw: sounds good, thanks for pointing it out!

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsRefControl.tsx
```diff
@@ -30,9 +30,9 @@ export const jsonFormsRefControlTester: RankedTester = rankWith(
 
 const parseHref = (
   href: string,
-  pageType: Omit<LinkTypes, "email" | "page">,
+  pageType: Omit<LinkTypes, typeof LINK_TYPES.Email | typeof LINK_TYPES.Page>,
```
dcshzj, line 33: issue: this should be Exclude

```suggestion
  pageType: Exclude<LinkTypes, typeof LINK_TYPES.Email | typeof LINK_TYPES.Page>,
```

seaerchin (approved):

PR #840, opened 2024-10-27
replace with RoleType for SSOT

adriangohjw:
## Problem

Currently values are hardcoded

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- single source of truth by importing them everywhere instead of hardcoding values


seaerchin (approved):
i think the `buildPermissions` method has a type restriction anyways so the two are equivalent but lgtm

PR #845, opened 2024-10-29
feat: use latest release tag instead of main for publish

dcshzj:
## Problem

We are always using the latest main branch to build our components and templates, which can be dangerous.

## Solution

**Breaking Changes**

**Features**:

- Fetch the latest release tag from the repository and use that release tag for the site builds.
    - NOTE: We still retain the ability to set a custom branch to build from using the `ISOMER_BUILD_REPO_BRANCH` env var.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

### DIFF
--- a/tooling/build/scripts/publisher.sh
+++ b/tooling/build/scripts/publisher.sh
@@ -11,9 +11,9 @@ calculate_duration() {
   echo "Time taken: $duration seconds"
 }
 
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
The initial entrypoint is still to curl this and exec so there might be some inconsistencies there but I don't think it's urgent to fix

PR #846, opened 2024-10-29
isom-1651 - add acceptable file type messages

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1651/inform-users-that-only-pdfs-are-supported

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- add a copywriting text

**AFTER**:

image

### DIFF
--- a/apps/studio/src/components/PageEditor/FileAttachment.tsx
+++ b/apps/studio/src/components/PageEditor/FileAttachment.tsx
@@ -15,6 +15,12 @@ interface FileAttachmentProps {
 
 type FileRejections = AttachmentProps<false>["rejections"]
 
+const ACCEPTED_FILE_TYPES: string[] = ["application/pdf"]
+const ACCEPTED_FILE_TYPES_MESSAGE = ACCEPTED_FILE_TYPES.map((filetype) => {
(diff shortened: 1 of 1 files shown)


sehyunidaaa:
@adriangohjw I'm okay with this, thanks for taking a shot at it! My only concern is that the text at the bottom sometimes gets replaced by the error message. Example screenshot attached!

Currently the error message for unspported file extension reads: 
"Your file's extension ending in {extension} is not allowed" 
And it's not too clear what I did wrong.

Should we change the error message for unsupported file extensions to be more instructive:
"For security reasons, only .pdf files are accepted. Please upload a .pdf file." 

image

sehyunidaaa (changes requested):
Left a comment re: error message

adriangohjw:
@sehyunidaaa unfortunately that's not possible with OGPDS too 😢 
https://design.open.gov.sg/?path=/docs/components-attachment--docs

sehyunidaaa:
@adriangohjw gooosh okay :( i'm good with this workaround!

sehyunidaaa (approved):

harishv7 (approved):

PR #850, opened 2024-10-31
feat: support storing page ordering of folders in DB

dcshzj:
## Problem

We need to support page ordering within folders, but there is no way to represent this inside Studio.

## Solution

**Breaking Changes**

**Features**:

- Add support for storing the page ordering of folders inside the `_pages` permalink with a new `PageOrder` resource type.
- Prevent this new `PageOrder` resource from appearing in the list of pages when viewing a folder inside Studio.
- Updated the publisher script to take the page ordering inside this `PageOrder` resource into account.

**Notes**:

- We are not adding support for editing this page ordering yet, this has to be manually done via editing the database.


apps/studio/prisma/generated/generatedEnums.ts
```diff
@@ -11,6 +11,7 @@ export const ResourceType = {
   CollectionLink: "CollectionLink",
   CollectionPage: "CollectionPage",
   IndexPage: "IndexPage",
+  PageOrder: "PageOrder",
```
harishv7, line 14: Could _pages be used for potentially storing other page related metadata? If so, using a more generic name like PageMeta or PageSettings would be good

tooling/build/scripts/publishing/index.ts
```diff
@@ -139,6 +146,8 @@ async function main() {
       }
     }
 
+    logDebug("Sitemap entries:", sitemapEntries)
```
harishv7, line 149: is this to be removed?

tooling/build/scripts/publishing/index.ts
```diff
@@ -159,7 +168,7 @@ async function main() {
       ),
     }
 
-    logDebug("Intermediate sitemap:", sitemap)
+    logDebug("Intermediate sitemap:", JSON.stringify(sitemap, null, 2))
```
harishv7, line 171: same here

apps/studio/src/utils/sitemap.ts
```diff
@@ -23,14 +23,19 @@ const getSitemapTreeFromArray = (
...
+        resource.type !== "PageOrder"
+      )
     }
     return (
       resource.parentId === parentId &&
       resource.permalink !== INDEX_PAGE_PERMALINK
     )
   })
 
+  // TODO: Sort the children by the page ordering if the PageOrder resource exists
```
harishv7, line 38: does the order in the sitemap matter since we are already sorting this during publishing stage?

apps/studio/public/assets/css/preview-tw.css
```diff
@@ -4133,11 +4129,29 @@ video {
   border-color: rgb(247 143 30 / var(--tw-border-opacity));
 }
 
+.focus\:bg-utility-highlight:where([data-rac])[data-focused] {
```
harishv7, line 4132: are these css changes related to this page order?

apps/studio/prisma/generated/generatedEnums.ts
dcshzj: Hmm no, page-related metadata are stored within those individual pages. This is a page ordering that is tied to the folder and not to individual pages.

apps/studio/public/assets/css/preview-tw.css
dcshzj: No, I think might have been previous changes that went in without doing the build step.

apps/studio/src/utils/sitemap.ts
dcshzj: This is used for the preview only, so we are maintaining the same behaviour as much as possible with what the user will see on the actual site.

tooling/build/scripts/publishing/index.ts
dcshzj: This is `logDebug`, so it only appears if `DEBUG` is set to `true`. Keeping this so we can have a way to understand how the sitemap looks like during the build process.

tooling/build/scripts/publishing/index.ts
dcshzj: Same as above.

apps/studio/prisma/generated/generatedEnums.ts
harishv7: Hmm understand, I just meant if there's potentially other use cases for _pages in future in which using a generic name now will help prevent migrations etc

harishv7 (approved):
lgtm, with one main comment on the naming of the resource type to be more generic. leave it to @dcshzj to take a call on this

apps/studio/prisma/generated/generatedEnums.ts
dcshzj: Renamed PageOrder to FolderMeta and `_pages` to `_meta`. I think we can have the option to have a CollectionMeta type in the future to control the ordering of items within the collection, but will need to be careful about the difference between this folder ordering meta vs normal folder metadata.

PR #851, opened 2024-11-01
fix: set cache control for only _next folder

dcshzj:
## Problem

We currently do a complete invalidation on CloudFront, which has some race condition issues and causes visitors to encounter an irresponsive website.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Update the publisher script such that when we upload the files to S3, we set all page content to have a low cache of 10 minutes, whereas all JS and CSS files inside the `_next` folder has a much longer cache duration of 24 hours.
- Remove the need to perform a distribution invalidation as the cache control setting now takes precedence.

### DIFF
--- a/tooling/build/scripts/publisher.sh
+++ b/tooling/build/scripts/publisher.sh
@@ -106,7 +106,12 @@ ls -al
 # Publish to S3
 echo "Publishing to S3..."
 start_time=$(date +%s)
-aws s3 sync . s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest --delete --no-progress
+
+# Set all files to have 10 minutes of cache, except for those in the _next folder
+aws s3 sync . s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest --delete --no-progress --cache-control "max-age=600" --exclude "_next/*"
+
+# Set all files in the _next folder to have 1 day of cache
+aws s3 sync _next s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest/_next --delete --no-progress --cache-control "max-age=86400"
 
 # Update CloudFront origin path
 echo "Updating CloudFront origin path..."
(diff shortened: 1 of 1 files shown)


tooling/build/scripts/publisher.sh
```diff
@@ -106,7 +106,12 @@ ls -al
 # Publish to S3
 echo "Publishing to S3..."
 start_time=$(date +%s)
-aws s3 sync . s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest --delete --no-progress
+
+# Set all files to have 10 minutes of cache, except for those in the _next folder
+aws s3 sync . s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest --delete --no-progress --cache-control "max-age=600" --exclude "_next/*"
```
harishv7, line 111: technically each time this is run a new build folder is create, hence there is no impact of --delete flag right

tooling/build/scripts/publisher.sh
```diff
@@ -119,6 +124,3 @@ echo "ETag: $ETag"
 jq '.Distribution.DistributionConfig' distribution.json > distribution-new.json
 jq ".Origins.Items[0].OriginPath = \"/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest\"" distribution-new.json > distribution-config.json
 aws cloudfront update-distribution --id $CLOUDFRONT_DISTRIBUTION_ID --distribution-config file://distribution-config.json --if-match $ETag
-
-# Invalidate CloudFront cache
```
harishv7, line 123: if you don't invalidate, the max time for a change to be live is 30 mins of build + 10 mins of previous cache = 40 mins right

I think we should still invalidate the cache, but having the TTL longer for _next helps to solve for those weird edge cases where the old versions of the HTMLs are being loaded and is unable to find the right _next chunk

tooling/build/scripts/publisher.sh
dcshzj: Yeah no impact, but this is a more defensive approach in case there is existing data.

tooling/build/scripts/publisher.sh
dcshzj: Hmm no, we can't invalidate because it is the CloudFront invalidation process that is opaque to us and caused us some issues. If we perform the invalidation, then the cached file in CloudFront gets removed, which then causes the 404 errors for the chunks that we were seeing earlier. We still want the old cached file in CloudFront to be retained so that in case users are still seeing the old index.html file which references these old chunks, they don't receive a 404.

harishv7: okay so what is the maximum time for a change to be live to MOPs now? is 10 mins + build time right?

If builds are still around 20-30 mins + 10mins, I think we still need to optimise this to be faster, but can be a future improvement

harishv7 (approved):
lgtm with minor comments for future

tooling/build/scripts/publisher.sh
dcshzj: Yeah in the worst case would be 10 minutes + build times but I think we should accept this because otherwise we lose the benefits of having CloudFront provide the caching for us, which can affect our time to first byte.

PR #856, opened 2024-11-05
fix: preserve reference links when copy-pasting from preview

dcshzj:
## Problem

We had an issue where users can be overwhelmed with multiple text components in a page, so in an attempt to consolidate them into a single text component, they copy-pasted from the preview into the Tiptap editor.

This caused issues with rendering links, as all links in the preview are transformed to the `#` value, which gets published to the actual site.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Remove the transformation to the `#` value in the FakeLink, as the `e.preventDefault()` is sufficient to prevent links from being clickable in the Preview.
- Add a ProseMirror plugin to the link extension to handle reference links being copied.
    - This is needed as links inside the HTML copied from the preview is transformed to the absolute link (e.g. `[resource:1:2]` becomes `https://studio.isomer.gov.sg/sites/1/pages/[resource:1:2]`).
    - Hence, we will need to transform these absolute links back into reference links during the paste.
    - The `addPasteRules` function is insufficient, as it does not have access to the full HTML pasted. Only the ProseMirror state has this full information.

## Video demo

https://github.com/user-attachments/assets/7a815d54-6907-4d39-a872-8effd232a576

### DIFF
--- a/apps/studio/src/features/editing-experience/hooks/useTextEditor.ts
+++ b/apps/studio/src/features/editing-experience/hooks/useTextEditor.ts
@@ -27,6 +27,8 @@ import { Underline } from "@tiptap/extension-underline"
 import { Plugin, PluginKey } from "@tiptap/pm/state"
 import { textblockTypeInputRule, useEditor } from "@tiptap/react"
 
+import { getHtmlWithRelativeReferenceLinks } from "../utils"
+
 const HEADING_LEVELS: Level[] = [2, 3, 4, 5]
 
 export interface BaseEditorProps {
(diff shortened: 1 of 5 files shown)


seaerchin (approved):

apps/studio/src/features/editing-experience/hooks/useTextEditor.ts
```diff
@@ -34,7 +35,33 @@ export interface BaseEditorProps {
...
+        // links, so that the original link is preserved in the editor.
+        new Plugin({
+          key: new PluginKey("transformReferenceLinks"),
+          props: {
+            transformPastedHTML(html, _) {
+              return html.replaceAll(
+                /href="(?:http|https):\/\/[^[]*\/\[resource:(\d+):(\d+)\]"/g,
+                (_, siteId, resourceId) => {
+                  return `href="[resource:${siteId}:${resourceId}]"`
+                },
```
adriangohjw, line 56: should we 
1. put this into a util so we can write basic unit test for it?
2. extract `[resource:(\d+):(\d+)\]` into a const and import it instead? seems to be used in a number of places

adriangohjw (approved):

apps/studio/src/features/editing-experience/hooks/useTextEditor.ts
dcshzj: Updated in ced21bd362465da4508444f738b092a364f9d766. Didn't create the tests yet as I think we haven't decided how we want to do testing for the editing experience feature.

PR #857, opened 2024-11-05
feat: alternate list marker type for nested lists

dcshzj:
## Problem

Nested lists are currently using the same list marker type, which can be a little confusing and annoying to users.

Fixes ISOM-1662.

## Solution

**Breaking Changes**

**Features**:

- Make nested lists alternate between different list types:
    - For ordered lists: Use decimal, then lower-alpha, then lower-roman
    - For unordered lists: Use disc, then circle, then square

**Bug Fixes**:

- Sometimes when the parent list item is "empty" (i.e. just spaces but not entirely empty), the child bullet gets collapsed together with it. We now force a line break if it is empty.

## Before & After Screenshots

**BEFORE**:

image
image

**AFTER**:

image
image

### DIFF
--- a/packages/components/src/templates/next/components/native/ListItem/ListItem.tsx
+++ b/packages/components/src/templates/next/components/native/ListItem/ListItem.tsx
@@ -3,7 +3,7 @@ import OrderedList from "../OrderedList"
 import Paragraph from "../Paragraph"
 import UnorderedList from "../UnorderedList"
 
-const ListItem = ({ content, LinkComponent, site }: ListItemProps) => {
+const ListItem = ({ content, level, LinkComponent, site }: ListItemProps) => {
   return (
     <li className="my-[15px] pl-2 sm:my-5 [&_>_p]:inline">
       {content.map((item, index) => {
@@ -21,6 +21,7 @@ const ListItem = ({ content, LinkComponent, site }: ListItemProps) => {
             <OrderedList
               key={index}
               {...item}
+              level={!!level ? level + 1 : 1}
               LinkComponent={LinkComponent}
               site={site}
             />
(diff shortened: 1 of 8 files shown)


packages/components/src/templates/next/components/native/ListItem/ListItem.tsx
```diff
@@ -21,6 +21,7 @@ const ListItem = ({ content, LinkComponent, site }: ListItemProps) => {
             <OrderedList
               key={index}
               {...item}
+              level={!!level ? level + 1 : 1}
```
seaerchin, line 24: question: why is this `level + 1` and `1` and not `level` and `0`? does this have something to do with what tiptap sends us?

seaerchin (approved):
lgtm with 1 minor qn

packages/components/src/templates/next/components/native/ListItem/ListItem.tsx
dcshzj: Hmm nothing to do with Tiptap here. Cos `level` is optional so the burden of setting the level is not placed on users of a list component, so for the OrderedList and UnorderedList components, they will assume that they are at level 0 if it is not provided. This increment in level is because we are starting on the next nested list, so the first nested list will have a level of 1.

This has no schema change either, purely just a components thing to track which level the list is being called at.

adriangohjw (approved):

packages/components/src/templates/next/components/native/ListItem/ListItem.tsx
seaerchin: ah ok wait let me double check - this is an ordered list inside another ordered list so the level has to be minimally 1?

missed this on initial read

dcshzj: Oops missed this, the first list is always at level 0 (zero-based indexing), then we increment the level every time we start a new nested list.

PR #858, opened 2024-11-05
cache `node_modules` to reduce build time

adriangohjw:
## Problem

Reduce build time and cost by caching

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

TL;DR - cache `node_modules` and save ~2+ mins per build

approach
- storing cached files (based on release tag version) in S3 as there's no guarantee local cache will persist between builds given the low frequency of builds and the on-demand setup of our codebuild instances
- since releases are not very often (~1-2 times weekly), this reduces build time for every build across sites
- e.g. MOH-corp now build at ~14 mins when testing with PROD data

how:
- cache `node_modules` (isomer)
  - before: between 80 to 120 seconds to install
  - after: ~6 seconds to fetch from S3
  - if cache miss, takes ~6 seconds (up to 40 seconds) to upload to S3 too after installing
- cache all the `node_modules` (tooling/template)
  - before: ~50 (9 + 33 + 8) seconds to install
  - after: ~0 seconds to fetch from S3
  - if cache miss, takes ~1 seconds (up to 17seconds)to upload to S3 too after installing

note
- did not cache `.next/cache` because it is too big and does not really speed up `next build` much. In fact, it's actually slower if we have to download the cache from S3 than actually building it on the spot
- while ideally we use codebuild in-built cache mechanism in buildspec, but:
  - it's easier to maintain everything in the publisher script 
  - cache are tagged to dynamic keys like `ISOMER_BUILD_REPO_BRANCH` which are fetch on build-time, which can't be done easily since we are using pulumi to define the stack


adriangohjw:
note - will need to update infra repo

tooling/build/scripts/publisher.sh
```diff
@@ -37,21 +37,34 @@ echo $(git branch)
...
-cd packages/components
-npm run build
-mv opengovsg-isomer-components-0.0.13.tgz ../../tooling/template/
-cd ../.. # back to root
-echo $(pwd)
-calculate_duration $start_time
+# Try to fetch cached node_modules from S3
+echo "Fetching cached node_modules..."
+NODE_MODULES_CACHE_PATH="s3://$S3_CACHE_BUCKET_NAME/$ISOMER_BUILD_REPO_BRANCH/isomer/node_modules.tar.gz"
+aws s3 cp $NODE_MODULES_CACHE_PATH node_modules.tar.gz || true
```
dcshzj, line 43: issue: I think this will break our assumption that it takes at least 1 minute to reach the "fetching from database" step. Should we move that step further down after the "prebuilding" step and before the "building" step instead?

adriangohjw: @dcshzj thanks for catching this!

But moving it between prebuilding and building wouldn't help too since prebuilding is cached just like node_modules

With this PR, if cache is found (_which should be all builds except the first one after we publish a newer package version_), then the time to reach "fetching from database" right after the 1) queuing of build + 2) provisioning of instances. A safe min. time needed from observing past build is 20s (provisioning of instances)

adriangohjw: @dcshzj should we update it from 1 min to 30s?

dcshzj: Ah oops I missed this comment, yeah if based on tests + moving the steps around it is still 30s then we should update that threshold too.

dcshzj (approved):

adriangohjw:
@dcshzj can help review https://github.com/opengovsg/isomer-next-infra/pull/27 too? this can't be merged before that 😬

PR #860, opened 2024-11-06
upgrade to next15

adriangohjw:
## Problem

Hoping to leverage the performance improvement stated in https://nextjs.org/blog/next-15#faster-static-generation-for-the-app-router to speed up SSG which is the remaining main bottleneck of slow build

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- improved codebuild `next build` by 3 minutes (540s -> 360s) when tested on `moh-corp` PROD

note:
- only upgrading for `tooling/template` because next 15 is not supported by storybook used in studio at the moment. Thus we have to tell the lint to exclude it

## Tests

**New dependencies**:

- `dependency` : `next` to v15


adriangohjw:
need to put on HOLD because storybook is not compatible with next15 yet https://github.com/storybookjs/storybook/issues/29421

package.json
```diff
@@ -18,7 +18,7 @@
     "format:fix": "turbo format --continue -- --write --cache --cache-location node_modules/.cache/.prettiercache  --ignore-path ./.prettierignore",
     "lint": "turbo lint --continue -- --cache --cache-location node_modules/.cache/.eslintcache",
     "lint:fix": "turbo lint --continue -- --fix --cache --cache-location node_modules/.cache/.eslintcache",
-    "lint:ws": "npx sherif@0.10.0",
+    "lint:ws": "npx sherif@0.10.0 -i next",
```
harishv7, line 21: can i clarify what -i next does?

tooling/template/app/[[...permalink]]/page.tsx
```diff
@@ -12,11 +12,13 @@ import {
   RenderEngine,
 } from "@opengovsg/isomer-components"
 
+export const dynamic = "force-static"
```
Comment (harishv7) [inline tooling/template/app/[[...permalink]]/page.tsx:15]: do we need this statement across the files? could you summarise what this implies?

Comment (harishv7) [reply tooling/template/app/[[...permalink]]/page.tsx:15]: I would think that having the following in the next.config will auto-force all to be static

```
const nextConfig = {
  output: "export",
```

package.json
adriangohjw: it tells `sherif` to ignore the different `next` versions and not raise it as error

this is needed because `postinstall` runs `lint:ws` and block `npm i` from proceeding

tooling/template/app/[[...permalink]]/page.tsx
Comment (adriangohjw) [reply tooling/template/app/[[...permalink]]/page.tsx:15]: @harishv7 yeah i would think so too but somehow not including this will cause an issue during `next build`

harishv7 (approved):
@adriangohjw  approving, please test on our repos before merging to ensure builds succeed

PR #862, opened 2024-11-06
speed up codebuild - aws s3 sync higher concurrency

adriangohjw:
## Problem

Hoping to leverage the concurrent uploads stated in https://docs.aws.amazon.com/cli/latest/topic/s3-config.html to speed up `aws s3 sync`

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- set a higher S3 `max_concurrent_requests` based on the number of cores (arbitrary 4x multiplier since file size is small)
  - We set a floor (20, which is higher than the default 10) and a ceiling (re: https://jasonralph.org/?p=791)
  - experimented with 10x multiplier with no cap (which means 320 concurrency with 32 CPU core, but value is the same). Likely due to resource bottleneck beyond a certain point
- benchmarking
  - before: 140s to 240s
  - after: 80s to 90s 
  - improvement: 50s to 160s reduction

### DIFF
--- a/tooling/build/scripts/publisher.sh
+++ b/tooling/build/scripts/publisher.sh
@@ -107,6 +107,16 @@ ls -al
 echo "Publishing to S3..."
 start_time=$(date +%s)
 
+
+NUMBER_OF_CORES=$(nproc)
+echo "Number of cores: $NUMBER_OF_CORES"
+
+# Set the number of concurrent S3 sync operations
+S3_SYNC_CONCURRENCY=$(( 4 * NUMBER_OF_CORES )) # 4x is an arbitrary number that should work well for most cases
+S3_SYNC_CONCURRENCY=$(( S3_SYNC_CONCURRENCY < 20 ? 10 : S3_SYNC_CONCURRENCY )) # Minimum of 20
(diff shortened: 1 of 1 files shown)


tooling/build/scripts/publisher.sh
```diff
@@ -107,12 +107,25 @@ ls -al
...
 start_time=$(date +%s)
 
+
+NUMBER_OF_CORES=$(nproc)
+echo "Number of cores: $NUMBER_OF_CORES"
+
+# Set the number of concurrent S3 sync operations
+S3_SYNC_CONCURRENCY=$(( 4 * NUMBER_OF_CORES )) # 4x is an arbitrary number that should work well for most cases
+S3_SYNC_CONCURRENCY=$(( S3_SYNC_CONCURRENCY < 20 ? 10 : S3_SYNC_CONCURRENCY )) # Minimum of 20
+S3_SYNC_CONCURRENCY=$(( S3_SYNC_CONCURRENCY > 100 ? 100 : S3_SYNC_CONCURRENCY )) # Maximum of 100 (to prevent AWS from throttling us)
```
harishv7, line 117: Each site can have at most 2 builds at a time. Also, there are multiple sites and builds. Do we know whats the throttling limit?

harishv7 (approved):

PR #863, opened 2024-11-07
fix(link-editor): update behaviour so that we can save on link collection items

seaerchin:
## Problem
This PR fixes an issue with ISOM-1654 (and more deeply, the link editor)

Closes ISOM-1654

## Solution
1. add a `LinkEditorContext` that all the child components read from. we do this so that the state can be unified for what the href currently is + the currently active link item 
2. pass `onChange` to the context and have the children reference it. this isn't quite ideal yet because page/file elements live outside of the context but we're manually passing the same `onChange` + callsite is limited, this shuold be ok
3. fix overflow for ref control so that overflowing links display as truncated elllipsis
4. bug-fix for tab behaviour so that doing any action on any other radio tab resets the state but changing the radio itself doesn't reset the state

## Videos + Screenshots

https://github.com/user-attachments/assets/69d00479-c840-43f0-82f2-f9f39a73b013

image

overflow now


adriangohjw:
> for tab behaviour so that doing any action on any other radio tab resets the state but changing the radio itself doesn't reset the state

will this be confusing UX? because it's not obvious to the end user why and when it resets

seaerchin:
> > for tab behaviour so that doing any action on any other radio tab resets the state but changing the radio itself doesn't reset the state
> 
> will this be confusing UX? because it's not obvious to the end user why and when it resets

yeap, getting @sehyunidaaa to double check this but i thought this was intuitive because:
1. swapping tabs doesn't do anything
2. taking an action resets

sehyunidaaa:
Just to confirm - we had to use this logic of resetting only when there's a change because it's difficult to maintain the input across different tabs right (E.g., if file has an attachment and link also has a link, we can't keep the two)

But I agree with Chin on this, I think once the user inputs something in a different tab then it's a signal that they want to link something else

adriangohjw:
@sehyunidaaa @seaerchin i think we can use my previous method of keeping the states as history so it persists between tabs

seaerchin:
> @sehyunidaaa @seaerchin i think we can use my previous method of keeping the states as history so it persists between tabs

sounds good but i would be wary of adding more states because i think this might be pre-mature and unnecessary complexity. lets stick with this first and see if useres are confused by the behaviour since sehyun is ok w/ it?

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -1,84 +1,95 @@
...
+const HTTPS_PREFIX = "https://"
+type HttpsLink = `https://${string}`
+
+const generateHttpsLink = (data: string): HttpsLink => {
+  if (data.startsWith(HTTPS_PREFIX)) {
+    return data as HttpsLink
+  }
+
+  return `https://${data}`
+}
```
adriangohjw, line 23: nit - should we use `HTTPS_PREFIX` for the `https://${data}` too?

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -206,7 +199,13 @@ interface LinkEditorModalProps {
...
   isOpen: boolean
   onClose: () => void
-  linkTypes: LinkTypeMapping
+  linkTypes: Record<
+    string,
+    {
+      icon: IconType
+      label: Capitalize<LinkTypes>
+    }
+  >
```
adriangohjw, line 208: use `LinkTypeMapping` instead?

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -72,7 +80,7 @@ interface LinkEditorModalContentProps {
   linkText?: string
   linkHref?: string
   onSave: (linkText: string, linkHref: string) => void
-  linkTypes: LinkTypeMapping
+  linkTypes: LinkEditorModalProps["linkTypes"]
```
adriangohjw, line 83: why the change?

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+  error?: string
+  onChange: (value: string) => void
+}
+const useLinkEditorContext = ({
+  linkHref,
+  linkTypes,
+  error,
+  onChange,
+}: UseLinkEditorContextProps) => {
+  const [curType, setCurType] = useState<LinkTypes>("page")
```
adriangohjw, line 26: use const `LINK_TYPES.Page` instead?

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorRadio.tsx
```diff
@@ -0,0 +1,89 @@
...
+      </Box>
+    </Box>
+  )
+}
+
+export const LinkEditorRadio = () => {
+  const { linkTypes, setCurType } = useLinkEditor()
+  const { getRootProps, getRadioProps } = useRadioGroup({
+    name: "link-type",
+    defaultValue: "page",
```
adriangohjw, line 65: use const instead

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorRadio.tsx
```diff
@@ -0,0 +1,89 @@
...
+    // NOTE: This is a safe cast because we map over the `linkTypes` below
+    // so each time we are using the `linkType`
+    onChange: (value) => setCurType(value as LinkTypes),
```
adriangohjw, line 68: why do we need to cast it though?

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorRadio.tsx
adriangohjw, line 1: nit - should this be plural `LinkEditorRadios` instead? since it returns a few radios and will be less misleading

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsRefControl.tsx
```diff
@@ -93,8 +98,14 @@ export function JsonFormsRefControl({
...
-              <Button onClick={onOpen} variant="link">
-                <Text textStyle="subhead-2">Link something...</Text>
+              <Button
+                onClick={onOpen}
+                variant="link"
+                aria-labelledby="button label"
+              >
+                <Text id="button label" textStyle="subhead-2">
+                  Link something...
+                </Text>
```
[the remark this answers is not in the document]
seaerchin: easy fix so i make

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
```diff
@@ -1,84 +1,95 @@
...
-        <LinkTypeRadioContent
-          selectedLinkType={selectedLinkType}
-          data={value}
-          handleChange={onChange}
-          pageLinkElement={pageLinkElement}
-          fileLinkElement={fileLinkElement}
-        />
+        {curType === "page" && pageLinkElement}
+        {curType === "file" && fileLinkElement}
+        {curType === "external" && (
```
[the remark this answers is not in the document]
seaerchin: actually this isn't hardcoding because `curType === LinkTypes` here, so we:
1. have intellisense
2. cannot specify the wrong type ("Page" instead of "page")

regardless, i think since we have the `const` might as well use it

apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx
seaerchin: fixed!

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
seaerchin: the child component should depend on the parent rather than specifying an external dependency.

actually i'l just use `Extract` here rather than respecificying oops

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
seaerchin: i actually tried this and it didn't _quite_ work. this is because when we say `Record<LinkTypes, string>`, actually what we mean is a 1:1 mapping (ie, every type in `LinkType` **must** have a corresponding string) which isn't quite what we want here. what we want is "some subset of LinkTypes to string" - i considered using `Partial<Record<LinkTypes, string>>`, but unfortunately that means that **every type** is now optional. hence, i just went iwth string for now

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorContext.tsx
seaerchin: donezxc

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorRadio.tsx
seaerchin: donezxc

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorRadio.tsx
seaerchin: value is inferred as `string` which doesn't satisfy the type

adriangohjw: will it be better to ensure that the value passed into and handled by `onChange` enforce the LinkTypes

seaerchin: this value is given by an external api (`useRadioGroup`) and it doesn't accept generics as far as i can tell, so we couldn't enforce it :sadge:

apps/studio/src/features/editing-experience/components/LinkEditor/LinkEditorRadio.tsx
seaerchin: i actually think the `Radios` is more confusing sia - wdyt about `RadioGroup` or `Header`?

adriangohjw: true, `LinkEditorRadioGroup` sounds good!

adriangohjw (approved):

PR #864, opened 2024-11-07
feat: move to whitelist database table from growthbook

dcshzj:
## Problem

Our email whitelist lives in Growthbook, which makes adding new users a little difficult as it involves both updating the database and updating the Growthbook feature flag.

Fixes ISOM-1668.

## Solution

**Breaking Changes**

    - [ ] Add the existing whitelist emails on Growthbook into the database.

**Features**:

- Create a new Whitelist table that can take the email and expiry date.
- Enhance the whitelisting logic to allow for domain whitelists and domain suffix whitelisting.
- Added tests for the whitelist check logic.

### DIFF
--- a/apps/studio/src/server/modules/whitelist/whitelist.service.ts
+++ b/apps/studio/src/server/modules/whitelist/whitelist.service.ts
@@ -1,9 +1,18 @@
 import { TRPCError } from "@trpc/server"
 
+import { isValidEmail } from "~/utils/email"
 import { db } from "../database"
 
 export const isEmailWhitelisted = async (email: string) => {
   const lowercaseEmail = email.toLowerCase()
+
+  // Extra guard even if Zod validation has already checked
+  if (!isValidEmail(lowercaseEmail)) {
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: "Please sign in with a valid email address.",
+    })
+  }
 
   // Step 1: Check if the exact email address is whitelisted
   const exactMatch = await db
@@ -24,7 +33,7 @@ export const isEmailWhitelisted = async (email: string) => {
   if (emailParts.length !== 2 && !emailParts[1]) {
     throw new TRPCError({
       code: "BAD_REQUEST",
-      message: "An invalid email was provided",
+      message: "Please sign in with a valid email address.",
     })
   }
 
--- a/apps/studio/src/server/modules/whitelist/whitelist.service.ts
+++ b/apps/studio/src/server/modules/whitelist/whitelist.service.ts
@@ -1,0 +1,66 @@
+import { TRPCError } from "@trpc/server"
+
+import { db } from "../database"
+
+export const isEmailWhitelisted = async (email: string) => {
+  const lowercaseEmail = email.toLowerCase()
+
+  // Step 1: Check if the exact email address is whitelisted
+  const exactMatch = await db
+    .selectFrom("Whitelist")
+    .where("email", "=", lowercaseEmail)
+    .where(({ eb }) =>
+      eb.or([eb("expiry", "is", null), eb("expiry", ">", new Date())]),
+    )
+    .select(["id"])
+    .executeTakeFirst()
+
+  if (exactMatch) {
+    return true
+  }
+
+  // Step 2: Check if the exact email domain is whitelisted
+  const emailParts = lowercaseEmail.split("@")
+  if (emailParts.length !== 2 && !emailParts[1]) {
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: "An invalid email was provided",
+    })
+  }
+
+  const emailDomain = `@${emailParts[1]}`
+  const domainMatch = await db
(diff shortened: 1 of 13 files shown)


apps/studio/src/server/modules/whitelist/whitelist.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    )
+    .select(["id"])
+    .executeTakeFirst()
+
+  if (exactMatch) {
+    return true
+  }
+
+  // Step 2: Check if the exact email domain is whitelisted
+  const emailParts = lowercaseEmail.split("@")
```
harishv7, line 23: do we need whitelisting by domain? I thought we decided to whitelist only by exact emails going forward

dcshzj: This is for government agencies so they can self-service use Studio without having to send in a request for explicit whitelisting, otherwise the ops load will be quite high if we have to explicitly whitelist each individual emails.

apps/studio/src/server/modules/whitelist/whitelist.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { db } from "../database"
+
+export const isEmailWhitelisted = async (email: string) => {
+  const lowercaseEmail = email.toLowerCase()
+
+  // Step 1: Check if the exact email address is whitelisted
+  const exactMatch = await db
+    .selectFrom("Whitelist")
+    .where("email", "=", lowercaseEmail)
```
harishv7, line 11: is this userinput sanitised?

apps/studio/src/server/modules/whitelist/whitelist.service.ts
```diff
@@ -0,0 +1,66 @@
...
+      eb.or([eb("expiry", "is", null), eb("expiry", ">", new Date())]),
+    )
+    .select(["id"])
+    .executeTakeFirst()
+
+  if (domainMatch) {
+    return true
+  }
+
+  // Step 3: Check if the suffix of the email domain is whitelisted
```
harishv7, line 45: what is the use case for suffix match?

apps/studio/src/server/modules/whitelist/whitelist.service.ts
dcshzj: Yes it is sanitised via the zod schema of the caller procedure, but sure added an additional guard here in 805eebf4aa60387f9df7f8d26c09d0c42e876531.

apps/studio/src/server/modules/whitelist/whitelist.service.ts
dcshzj: This is to whitelist `.gov.sg`, so we can whitelist all domains that end with a `.gov.sg` in the domain name.

harishv7 (approved):
lgtm, pending approval from @spaceraccoon on security side of things

apps/studio/src/server/modules/whitelist/whitelist.service.ts
spaceraccoon: Bypassed with this POC:
```
var validator = require('validator');

const email = '"john..d@open.gov.sg@"@example.org'
const lowercaseEmail = email.toLowerCase()
console.log(validator.isEmail(email))

const emailParts = lowercaseEmail.split("@")
console.log(emailParts)

console.log(emailParts.length)
if (emailParts.length !== 2 && !emailParts[1]) {
  console.log('fail')
} else {
  const emailDomain = `@${emailParts[1]}`
  console.log(emailDomain)
}
```

```
node index.js
true
[ '"john..d', 'open.gov.sg', '"', 'example.org' ]
4
@open.gov.sg
```

This is because RFC-compliant emails that pass `validators` [allow a lot of sus stuff](https://portswigger.net/research/splitting-the-email-atom).

Don't merge pls; this is a common issue and we wrote a specific validator in `starter-kitty` for this, but I'd like to stress-test this a bit more: https://github.com/opengovsg/starter-kitty/tree/develop/packages/validators/src/email.

apps/studio/src/server/modules/whitelist/whitelist.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    .select(["id"])
+    .executeTakeFirst()
+
+  if (exactMatch) {
+    return true
+  }
+
+  // Step 2: Check if the exact email domain is whitelisted
+  const emailParts = lowercaseEmail.split("@")
+  if (emailParts.length !== 2 && !emailParts[1]) {
```
spaceraccoon, line 33:
```suggestion
  if (emailParts.length !== 2 || !emailParts[1]) {
```

This is the mistake that allowed my bypass, but overall is a bit sus

spaceraccoon (changes requested):

spaceraccoon (approved):
Noted that email whitelisting is not a long-term solution for authentication due to multiple downstream gotchas, and that SingPass SSO is one of the better solutions.

PR #865, opened 2024-11-07
isom-1659 internal links for folders

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1659/internal-links-for-folders
Closes https://linear.app/ogp/issue/ISOM-1656/linking-internal-pages-easily

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

I'm not proud of the size of this PR but most of the parts are interlocking and because of the feature (introducing search), it requires quite a bit of rework of how we are doing things - please refer to https://github.com/opengovsg/isomer/pull/865#pullrequestreview-2475146367 for a quick summary of the context and key changes in some files 😬 

reuse the logic from 
- https://github.com/opengovsg/isomer/pull/751
- https://github.com/opengovsg/isomer/pull/879

More:
- update `getAncestry` to `getAncestryWithSelf` to fit current logic for `resourceStack`
- pass in `resourceTypes` into `search` endpoint to make it more flexible
- write a custom hook `useResourceStack` to encapsulate everything inside it
- add checks to ensure folders cannot be moved into their own nested children
- refactor such that all components use this to reduce to keep it DRY'
- refactor to use `getIcon` 
- refactor search-related hooks into `useSearchQuery.ts` (used by global search too)

| scenario | recording |
| - | - |
| move modal<br>can double click | https://github.com/user-attachments/assets/f15af544-61c2-4990-b55a-f35078f6b287 |
| move modal<br>can search | https://github.com/user-attachments/assets/968a4c2f-7c64-45a3-9753-52c29bfeca8b |
| edit link<br>can search | https://github.com/user-attachments/assets/ba2cb0cc-1dc2-4df9-a0bc-b89c557b8845 |

# Tests

Sorry this is a big lengthy as there are many edge cases to test for! 😩 

## Moving pages

**Note: folder page cannot be moved into collection, and vice-versa.**
- However, both folders and collections are displayed
- This was already the case before this PR and this PR does not aim to fix that, to prevent further bloating. We should fix that in a separate PR
- Nevertheless, there's mechanism in the backend to prevent that and it throws error toast

### Can move successfully without entering the folder

1. Create a new page and click "move to..." in the resource menu dropdown
   - [ ] it should load the folders and collections from the root
2. Click on a folder once
   - [ ] it should highlight the folder
   - [ ] both text in the blue box below should update
3. Click "Move here"
   - [ ] go to the new folder and the page should be there 

### Can move successfully while in the folder

1. Create a new page and click "move to..." in the resource menu dropdown
   - [ ] it should load the folders and collections from the root
2. Click on a folder once
   - [ ] it should highlight the folder
   - [ ] both text in the blue box below should update
3. Click on the folder again
   - [ ] you should enter the folder
   - [ ] both text in the blue box below should **not** update
6. Click "Move here"
   - [ ] go to the new folder and the page should be there

## Moving folders

_(the same as "Moving pages" but you should not see non-folders, but here's additional things to take note of)_

- [ ] the current folder should be disabled and not possible to click on

### Can go back to parent folder while child folder is highlighted

1. Create a new page and click "move to..." in the resource menu dropdown
2. Click on a folder twice to enter it (make sure it has a child folder)
3. Click on the child folder **once** to select it
4. Click "Back to parent folder"
   - [ ] You should move back to the root folder
   - [ ] Both text in the blue box should be updated to original

### Can go back to parent folder while child folder is not highlighted

1. Create a new page and click "move to..." in the resource menu dropdown
2. Click on a folder twice to enter it
4. Click "Back to parent folder"
   - [ ] You should move back to the root folder
   - [ ] Both text in the blue box should be updated to original

## Can create/updat hyperlink for text

1. Go to a page and create a new text block component. Type something, highlight it and click on the "link" icon in tiptap
   - [ ] it should load the pages, folders and collections from the root
   - [ ] you should NOT see `_index` pages
2. Click on an item
   - [ ] the text in the blue box should update
3. click "add link"
   - [ ] it should add the hyperlink successfully
4. click on the hyperlinked text and click "link" icon in tiptap
   - [ ] it should load the folder content of the current page
   - [ ] current linked page should be already selected (highlighted)

### Searching in "hyperlink" modal

1. Go to a page and create a new text block component. Type something, highlight it and click on the "link" icon in tiptap
2. type something into the searchbar
   - [ ] it should load the pages, folders and collections from the root
   - [ ] you should NOT see `_index` pages
3. Click on something
   - [ ] you should be teleported into the parent of that item
   - [ ] that item should be highlighted (NOTE: the item MAY NOT be the first item and seen, but should be something that can be scrolled down and found)
   - [ ] the text in the blue box below is correctly updated 

### Searching in "move resource" modal

_(same as above, except you should only see folder showing up and NOT pages)_

### Others

- [ ] clicking on "X" in searchbar should clear the search field and load your last view
- [ ] type something gibberish and click "clear search" when no results found. it should clear the search field and load your last view


apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -43,28 +42,14 @@ const PageLinkElement = ({ value, onChange }: PageLinkElementProps) => {
...
-      {!!resource && (
-        <Box bg="utility.feedback.info-subtle" p="0.75rem" w="full" mt="0.5rem">
-          <Text textStyle="caption-1">
-            You selected /{resource.fullPermalink}
-          </Text>
-        </Box>
-      )}
-    </>
+    <ResourceSelector
+      queryFn={trpc.resource.getChildrenOf.useInfiniteQuery}
```
seaerchin, line 47: nit: to be honest, this feels like a code-smell. this works for now because they have relatively similar interfaces but we tie it very closely to the actual function too (the type of `queryFn` is specified as those 2 queries). i worry that if change is needed to either of the 2 functions, this component might not be as easily reusable down the line.

something we could consider is shifting the subcomponents out and favouring composition so that, for example, 1 component purely displays another and another triggers the refetch. another alternative we could do is specify the type ourselves + specify the refetch etc as extra props so we don't depend so strictly on teh trpc type!

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
```diff
@@ -1,86 +1,95 @@
...
-        return BiFile
-      case ResourceType.Collection:
-        return BiData
-    }
-  }, [type])
+  const { type, ...restWithoutType } = rest
+
+  const buttonProps = getButtonProps({
+    isHighlighted: !!isHighlighted,
+  })
```
seaerchin, line 69: this looks abit off to me - does the `active/focus` state not already account for the `highlighted` state? i thought the change would be to specify a base style and override on `_active`, for example

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
```diff
@@ -92,13 +101,13 @@ export const ResourceItem = (props: ResourceItemProps) => {
         <ErrorBoundary
           onReset={reset}
           fallbackRender={({ resetErrorBoundary }) => (
-            <Box>
+            <div>
```
seaerchin, line 104: no action required, just fyi: i think chakra box is actually just `chakra(div)`, which essentialliy means that it's a div that allows us to use props shorthand. this change shouldn't have any impact (and no need to change back) but was wondering why it was made

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,157 +1,220 @@
...
+import type { PendingMoveResource } from "~/features/editing-experience/types"
+import { useQueryParse } from "~/hooks/useQueryParse"
+import { sitePageSchema } from "~/pages/sites/[siteId]"
 import { trpc } from "~/utils/trpc"
 import { ResourceItem } from "./ResourceItem"
 
+interface ResourceSelectorProps {
+  queryFn:
+    | typeof trpc.resource.getChildrenOf.useInfiniteQuery
+    | typeof trpc.resource.getFolderChildrenOf.useInfiniteQuery
```
seaerchin, line 16: comment: here, we depend very strictly on the return type of these 2 specific functions and nexte time, to render, for example, collection pages only, we'd need to keep adding the type here. what happens if the types don't quite conform (eg: 1 query fn needs x prop but another needs y) - i think we should definitely specify a function type here as `queryFn: (input: string) => Data` rather than depending on the trpc type

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,157 +1,220 @@
...
+  // NOTE: We should always start the stack from `/` (root)
+  // so that the user will see a full overview of their site structure
+  const [resourceStack, setResourceStack] = useState<PendingMoveResource[]>([])
+  const [isResourceHighlighted, setIsResourceHighlighted] =
+    useState<boolean>(true)
+  const { siteId } = useQueryParse(sitePageSchema)
+
+  const moveDest = resourceStack[resourceStack.length - 1]
+  const parentDest = resourceStack[resourceStack.length - 2]
+  const curResourceId = moveDest?.resourceId
```
seaerchin, line 42: no action needed - some repeated code here which we could separate into a custom hook (`useMoveState` for example) but no urgency because

1. this likely hasn't been finalised yet, so behaviour can still differ
2. only 2 call sites (here + move modal) so the need for a custom hook is still low.

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,157 +1,220 @@
...
-      },
-      {
-        getNextPageParam: (lastPage) => lastPage.nextOffset,
-      },
-    )
+  useEffect(() => {
+    if (curResourceId) {
+      onChange(curResourceId)
+    }
+  }, [curResourceId])
```
seaerchin, line 77: clarification needed - i don't quite understand this section here, i'd expect to see teh `onChange` on a button as that's the trigger for the change rather than on `useEffect`, as that implies we're syncing state somewhere

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -512,6 +512,13 @@ export const resourceRouter = router({
             ])
             .where("Resource.siteId", "=", Number(siteId))
             .where("Resource.id", "=", resourceId)
+            .where((eb) =>
+              // to exclude root page
+              eb.and([
+                eb("Resource.permalink", "is not", null),
+                eb("Resource.permalink", "!=", ""),
+              ]),
+            )
```
seaerchin, line 521: nit: we should use the Resource.type!

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -493,14 +493,14 @@ export const resourceRouter = router({
       return query.select(["role"]).execute()
     }),
 
-  getAncestryOf: protectedProcedure
+  getAncestryWithSelf: protectedProcedure
```
seaerchin, line 496: stupid question - but why change this ah? is it to avoid another query so we can generate the full permalink from this procedure alone?

seaerchin (commented):
i like the chagne to standardise on using resource selector but i think tying the trpc.procedure's type directly to the selector itself might be abit sussy. i think we can be abit more verbose and just specify types + functions etc so we avoid this lock in

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
adriangohjw: @seaerchin have refactored such that this is also being used by "move modal" so it's only happening once

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
adriangohjw: not much reason haha, just thought that since we aren't using the feature of `Box` then a bit pointless to import it

apps/studio/src/server/modules/resource/resource.router.ts
adriangohjw: Previously, the `resourceStack` was designed to store only ancestor resources, as there was no need to "double click" into folders to view contained resources. We've updated this implementation to also include the current "self" resource within the resourceStack.

Adding "self" to the query enables the "link modal" to accurately display any already linked resources.

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
adriangohjw: OGPDS does but our design is slightly different so there's a number of overwriting here to fit it to our use case

seaerchin: sounds good!

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
adriangohjw: in some cases, the parent component that uses `ResourceSelector` will need `onChange` (props passed in) to trigger state change

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
adriangohjw: i thought about it for some time and decided to update it to pass in a prop `onlyShowFolders` which will be used to decide which query to used

```ts
  const queryFn = onlyShowFolders
    ? trpc.resource.getFolderChildrenOf.useInfiniteQuery
    : trpc.resource.getChildrenOf.useInfiniteQuery
```

Why this is an improvement:
- **Centralized Updates**: If there are changes to the `getFolderChildrenOf` or `getChildrenOf` props, we only need to update them in a single location (this file). Most of the changes will be confined to updating the props, which simplifies maintenance and enhances extensibility without over-engineering the solution.
- **Flexibility in Handling Query Changes**: If the interfaces of the queries change significantly, we can simply create a copy and preserve the original query. This approach avoids forcing them to reuse the same query, providing more flexibility as the system evolves.
- **Simplified Parent Component Props**: The parent component no longer needs to be aware of which specific query to pass in. This reduces complexity and decouples the parent from needing to handle the specifics of the query logic.

adriangohjw:
converting to draft first - need to update it to show full permalink of the resource instead of the permalink only

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,157 +1,228 @@
...
+
 const SuspensableResourceSelector = ({
-  siteId,
-  selectedResourceId,
   onChange,
-  isDisabledFn,
+  selectedResourceId,
+  existingResource,
+  onlyShowFolders = false,
 }: ResourceSelectorProps) => {
```
dcshzj, line 30: thought: I think I relooked at this piece of code a couple of times for this review and it is quite a bit of things that is going on within this single component. Putting this as a thought because it is non-blocking in nature, but would be good to refactor this component (either in this PR directly or in a subsequent one).

Main points:
1. The `resourceStack` state is the data structure that is holding most of the information needed to get the selector working as intended, but we are currently manipulating this state directly, rather than exposing it through more understandable APIs like sink and lift (or push and pop, etc).
    1. Recommendation: To use a custom hook that manages this state and exposes only the APIs from the hook. From what I see, the APIs exposed should be the sink and lift calls, moveDest, parentDest and curResourceId.
2. The second `useEffect` to call the `onChange` still feels weird to me, as it still seems to be synchronising states.
    1. Recommendation: We can probably pass the `onChange` function to the custom hook suggested in 1, then let the custom hook handle calling this `onChange` whenever the resource stack changes, so we don't need to rely on the `useEffect` to synchronise the states.

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,157 +1,228 @@
...
-    siteId,
-    resourceId: selectedResourceId,
-  })
-  const [parentIdStack, setParentIdStack] = useState<string[]>(
-    ancestryStack.map((item) => item.id),
+  // NOTE: This is the stack of user's navigation through the resource tree
+  // NOTE: We should always start the stack from `/` (root)
+  // so that the user will see a full overview of their site structure
+  const [resourceStack, setResourceStack] = useState<PendingMoveResource[]>([])
+  const [isResourceHighlighted, setIsResourceHighlighted] =
```
dcshzj, line 35: question(non-blocking): this `isResourceHighlighted` isn't very clear to me, is this referring to whether there is a resource that has been selected?

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
```diff
@@ -1,86 +1,95 @@
...
-      case ResourceType.Folder:
-        return BiFolder
-      case ResourceType.CollectionPage:
-      case ResourceType.Page:
-        return BiFile
-      case ResourceType.Collection:
-        return BiData
-    }
-  }, [type])
+  const { type, ...restWithoutType } = rest
```
dcshzj, line 65: nit: we should destructure the `type` directly in the function signature

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
```diff
@@ -1,86 +1,95 @@
...
         _hover: {
-          bgColor: "unset",
+          color: buttonProps._hover.color,
+          bg: buttonProps._hover.bg,
         },
-      }}
-      _hover={{
-        bgColor: "interaction.muted.main.hover",
-      }}
+      })}
```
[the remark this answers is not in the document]
adriangohjw: IIRC cannot as it will overwrite the default hover state incorrectly

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
adriangohjw: we cannot pass `type` into `Button` thus we need to exclude it

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
adriangohjw: yup!

I was thinking about this but `isResourceSelected` might be misleading too because it's unclear too selecting it (single click) and going into it (double click) means its selected either way

`isInFolder` (or equivalent) won't work for non-folder/collection pages

adriangohjw: lemme know if you have a better suggestion! i can't think of anything else 🥲

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
adriangohjw: yup actually this sounds good!

apps/studio/src/components/ResourceSelector/ResourceItem.tsx
```diff
@@ -1,86 +1,95 @@
...
+      return BiLink
+    case ResourceType.Folder:
+      return BiFolder
+    case ResourceType.CollectionPage:
+    case ResourceType.Page:
+      return BiFile
+    case ResourceType.Collection:
+      return BiData
+    default:
+      return BiData // Default to data icon
```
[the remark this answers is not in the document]
adriangohjw: i will refactor to use the existing `ICON_MAPPINGS` with some tweaks for single source of truth and reusability 

image

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
adriangohjw: attempted a refactor into `useResourceStack` custom hooks!

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
dcshzj: > it's unclear too selecting it (single click) and going into it (double click) means its selected either way

Hmm I'm a bit confused. I thought this meant that there was a selection made and if the user were to insert the link at this point, we will insert the selection that the user has made. So if the user selected a normal page, this variable will be set to true. If the user selected a folder/collection, this variable will be set to true until the user double clicks, which then resets it to false.

Maybe I'm misunderstanding what this variable does? Might need a bit more clarification

adriangohjw:
@dcshzj @seaerchin 
Actually no need review this yet - Im working on https://github.com/opengovsg/isomer/pull/903 and will be changing up the logic on how the results are fetched

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
adriangohjw, line 1: key changes
- refactoring logic into custom hooks `useSearchQuery` and `useResourceStack` to abstract away the inner works
- supporting rendering of different UI states

apps/studio/src/components/Searchbar/SearchModal.tsx
adriangohjw, line 1: key changes
- abstracting into custom hook `useSearchQuery` to hide the inner works + make it DRY with `ResourceSelector`

apps/studio/src/features/dashboard/components/DirectorySidebar/constants.ts
adriangohjw, line 1: no longer required after refactoring to all use `getIcon`

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
adriangohjw, line 1: logic mostly moved into `ResourceItem`

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
adriangohjw, line 1: logic mostly updated and moved into `useResourceStack` custom hook in `ResourceSelector`

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsLinkControl.tsx
adriangohjw, line 1: context
- no longer require the query and text `You selected /{resource.fullPermalink}` as it's handled by `ResourceSelector`

apps/studio/src/features/editing-experience/types.ts
adriangohjw, line 1: no longer needed by simplifying to only use `ResourceItemContent`

adriangohjw (commented):
tried my best to add some context and key changes for some files, hopefully will help code reviewers with their code review 😬

adriangohjw:
@sehyunidaaa updated to 12rem for file explorer in link editor modal (instead of 13rem as there's still double scrolling with 13rem)

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -1,160 +1,190 @@
...
+import {
+  BackButtonHeader,
+  HomeHeader,
+  LoadingHeader,
+  SearchResultsHeader,
+} from "./ResourceSelectorHeader"
+import { SearchBar } from "./SearchBar"
+import { useResourceStack } from "./useResourceStack"
+
+const FILE_EXPLORER_DEFAULT_HEIGHT_IN_REM = 17.5
```
seaerchin, line 24: i think this  might not be the way to do it! instead, we should extend a chakra prop, which allows us to override props like this naturally.

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+    resourceItemsWithAncestryStack,
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    fetchNextPage: useResourceIdsFromSearch ? () => {} : fetchNextPage,
+    hasNextPage: useResourceIdsFromSearch ? false : hasNextPage,
+    isFetchingNextPage: useResourceIdsFromSearch ? false : isFetchingNextPage,
+    isResourceIdHighlighted,
+    isResourceItemDisabled,
+    hasParentInStack,
+    handleClickBackButton,
+    handleClickResourceItem,
```
seaerchin, line 207: i think this return api feels abit bloated and is indicative of this hook trying to do abit too much. should we just keep this hook scoped to managing the stack and offer stack-like APIs (push, pop, peek) only?

apps/studio/src/utils/resources.ts
```diff
@@ -19,3 +50,14 @@ export const isAllowedToHaveLastEditedText = (
     resourceType === ResourceType.CollectionPage
   )
 }
+
+// only show user-viewable resources (excluding root page, folder meta etc.)
+export const getUserViewableResourceTypes = (): ResourceType[] => {
```
seaerchin, line 55: nit: we probably should make a custom type like `UserViewableResourceType = Extract...`

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+      (resourceStack.length === 1 && !isResourceHighlighted) ||
+      resourceStack.length > 1,
+    [resourceStack.length, isResourceHighlighted],
+  )
+
+  const fullPermalink: string = useMemo(() => {
+    return resourceStack.map((resource) => resource.permalink).join("/")
+  }, [resourceStack])
+
+  const handleClickResourceItem = (
```
karrui, line 130: nit: wrap in `useCallback`

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const pendingMovedItemAncestryStack: ResourceItemContent[] =
+    trpc.resource.getAncestryWithSelf
+      .useSuspenseQuery({
+        siteId: String(siteId),
+        resourceId: selectedResourceId,
+      })[0]
+      .map((resource) => ({ ...resource, resourceId: resource.id }))
+
+  useEffect(() => {
```
karrui, line 170: this effect seems redundant. Since this is a suspenseQuery, you can probably init the `useState` directly instead of using the effect (since this effect seems to be run only once)

```ts
const pendingMovedItemAncestryStack = trpc.resource.getAncestryWithSelf
      .useSuspenseQuery(...)

const [resourceStack, setResourceStack] = useState<ResourceItemContent[]>(pendingMovedItemAncestryStack) // probably good to rename this too
```

can move this up if need be.

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }, [isResourceHighlighted, removeFromStack])
+
+  const pendingMovedItemAncestryStack: ResourceItemContent[] =
+    trpc.resource.getAncestryWithSelf
+      .useSuspenseQuery({
+        siteId: String(siteId),
+        resourceId: selectedResourceId,
+      })[0]
+      .map((resource) => ({ ...resource, resourceId: resource.id }))
```
karrui, line 168: calling a map function right after a useSuspenseQuery feels like an antipattern. This seems to be a new procedure. You might want to rewrite the procedure return the exact structure you need instead of modifying it on the frontend. i.e. `getAncestryWithSelf` returns an array with the `resourceId` too?

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    setResourceStack(pendingMovedItemAncestryStack)
+  }, [])
+
+  useEffect(() => {
+    if (curResourceId) {
+      onChange(curResourceId)
+    }
+  }, [curResourceId])
```
karrui, line 192: [you might not need an effect](https://react.dev/learn/you-might-not-need-an-effect)

you update the resource stack only when you click on the resource item. You might want to trigger `onChange` there (inside `handleClickResourceItem`).

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  // currently do not support fetching next page for search
+  return {
+    fullPermalink,
+    moveDest,
+    resourceItemsWithAncestryStack,
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    fetchNextPage: useResourceIdsFromSearch ? () => {} : fetchNextPage,
+    hasNextPage: useResourceIdsFromSearch ? false : hasNextPage,
+    isFetchingNextPage: useResourceIdsFromSearch ? false : isFetchingNextPage,
```
karrui, line 202: comment: this feels a little wrong, though i cannot put my finger on it. might want to add a comment on why this is done this way, since the purpose of these conditionals is not very obvious to me.

apps/studio/src/hooks/useSearchQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const resources = useCallback((): SearchResultResource[] => {
+    return data?.pages.flatMap((page) => page.resources) ?? []
+  }, [data])()
+
+  const totalResultsCount = useCallback(() => {
+    return (
+      data?.pages.reduce((acc, page) => acc + (page.totalCount ?? 0), 0) ?? 0
+    )
+  }, [data])()
```
karrui, line 42: this is an antipattern -- if you are using `useCallback` as an IIFE, you should be directly assigning this to a constant (and optionally using `useMemo` instead)

```tsx
const totalResultsCount = useMemo(() => data?.pages.reduce((acc, page) => acc + (page.totalCount ?? 0), 0) ?? 0, [data])
```

apps/studio/src/hooks/useSearchQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+        query: debouncedSearchTerm,
+        resourceTypes,
+      },
+      {
+        onSuccess: onSearchSuccess,
+      },
+    )
+  }, [siteId, debouncedSearchTerm, resourceTypes, onSearchSuccess])()
+
+  const resources = useCallback((): SearchResultResource[] => {
```
karrui, line 34: same here. use `useMemo` instead.

apps/studio/src/hooks/useSearchQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+export const useSearchQuery = ({
+  siteId,
+  resourceTypes,
+  onSearchSuccess,
+}: UseSearchQueryProps) => {
+  const [searchValue, setSearchValue] = useState("")
+  const debouncedSearchTerm = useDebounce(searchValue, 300)
+
+  const { data, isLoading } = useCallback(() => {
```
karrui, line 21: This extra `useCallback` IIFE is redundant (and an antipattern). use the `useInfiniteQuery` hook directly. The hook already has memoization built in.

apps/studio/src/hooks/useSearchQuery.ts
```diff
@@ -0,0 +1,62 @@
...
+    return data?.pages.flatMap((page) => page.resources) ?? []
+  }, [data])()
+
+  const totalResultsCount = useCallback(() => {
+    return (
+      data?.pages.reduce((acc, page) => acc + (page.totalCount ?? 0), 0) ?? 0
+    )
+  }, [data])()
+
+  const recentlyEditedResources = useCallback((): SearchResultResource[] => {
```
karrui, line 44: same here. use `useMemo` instead.

apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
```diff
@@ -735,6 +736,143 @@ describe("resource.router", async () => {
...
+      await expect(result).rejects.toThrowError(
+        new TRPCError({
+          code: "FORBIDDEN",
+          message:
+            "You do not have sufficient permissions to perform this action",
+        }),
+      )
+    })
+
+    it.skip("should throw 403 if user does not have read access to resource", async () => {})
```
karrui, line 826: should this test be written instead of skipped? now that you have rbac

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -141,21 +149,14 @@ export const resourceRouter = router({
...
-        }
-      }
-      return {
-        items: result,
-        nextOffset: null,
-      }
+      return result.length > limit
+        ? // Dont' return the last element, it's just for checking if there are more
+          { items: result.slice(0, limit), nextOffset: offset + limit }
+        : { items: result, nextOffset: null }
```
karrui, line 155: comment/nit: this seems like unnecessary optimisation (actually it's less performant so not really an optimisation too (`slice` recreates the array, `pop` mutates)) + less readable. Suggest sticking with previous implementation.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -200,19 +202,40 @@ export const resourceRouter = router({
       } else {
         query = query.where("Resource.parentId", "=", String(resourceId))
       }
-
       const result = await query.execute()
-      if (result.length > limit) {
+      return {
```
karrui, line 206: comment/nit: this seems like unnecessary optimisation (actually it's less performant so not really an optimisation too (`slice` recreates the array, `pop` mutates)) + less readable. Suggest sticking with previous implementation.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -474,54 +497,44 @@ export const resourceRouter = router({
...
     .input(searchSchema)
     .output(searchOutputSchema)
     .query(
-      async ({ ctx, input: { siteId, query = "", cursor: offset, limit } }) => {
+      async ({
+        ctx,
+        input: {
+          siteId,
+          query = "",
+          resourceTypes = Object.values(ResourceType),
```
karrui, line 533: suggest not setting defaults on the params itself, but inside the schema for a single source of truth. you can use zod schema's [`.default`](https://zod.dev/?id=default) function.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -411,6 +412,55 @@ export const publishResource = async (
...
+          "Resource.title",
+          "Resource.permalink",
+          "Resource.parentId",
+          "Resource.type",
+        ])
+        .where("Resource.siteId", "=", Number(siteId))
+        .where("Resource.id", "=", String(resourceId))
+        .where((eb) =>
+          eb.and([eb("Resource.type", "!=", ResourceType.RootPage)]),
+        )
```
karrui, line 437: this `AND` clause seems redundant since there is only one conditional. Can be simplified?

```suggestion
        .where("Resource.type", "!=", ResourceType.RootPage),
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -597,6 +642,34 @@ export const getSearchRecentlyEdited = async ({
...
 }
 
+export const getNestedFolderChildren = async ({
+  siteId,
+  resourceId,
+}: {
+  siteId: number
+  resourceId: number
+}): Promise<ResourceItemContent[]> => {
+  return (async function getNestedChildren(
```
karrui, line 652: you might want to rewrite this function as a recursive CTE instead of doing it in code. This current implementation suffers from the [N+1 query problem](https://docs.sentry.io/product/issues/issue-details/performance-issues/n-one-queries/) and may cause some performance issues in the db.

also may not be consistent (since using a db connection instead of doing it in a transaction)

this [stackoverflow question](https://stackoverflow.com/questions/54907495/postgresql-recursive-parent-child-query) might have some answers (though i do also remember we already have something similar already in this repo)

karrui (dismissed):

karrui:
oops, i never read chin's review comments, so might have duplicate

apps/studio/src/utils/resources.ts
adriangohjw: right now we are using it more of an array of values (const) rather than a type, so should be okay to leave it as it is

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/865/commits/1e649ec81355584c5372a80b0463f9cd7799c767 and https://github.com/opengovsg/isomer/pull/865/commits/f522727ecd1becd9bd8ea8770cd63f04d7780e37 !

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+  )
+  const parentDest = useMemo(
+    () => resourceStack[resourceStack.length - 2], // second last item in stack
+    [resourceStack],
+  )
+  const curResourceId = useMemo(() => moveDest?.id, [moveDest])
+
+  const queryFn = onlyShowFolders
+    ? trpc.resource.getFolderChildrenOf.useInfiniteQuery
+    : trpc.resource.getChildrenOf.useInfiniteQuery
```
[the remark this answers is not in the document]
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/865/commits/1e649ec81355584c5372a80b0463f9cd7799c767

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -44,31 +42,16 @@ interface PageLinkElementProps {
...
-      )}
-    </>
+    <ResourceSelector
+      interactionType="link"
+      siteId={siteId}
+      onChange={(resourceId) =>
+        onChange(getReferenceLink({ siteId: String(siteId), resourceId }))
+      }
+      selectedResourceId={getResourceIdFromReferenceLink(value)}
+      fileExplorerHeight={12}
```
seaerchin, line 53: nit: the props for `ResourceSelector` might benefit from extending `BoxProps` from chakra so this `fileExplorerHeight` + `hasAdditionalLeftPadding` props can be written naturally

apps/studio/src/components/ResourceSelector/ResourceSelectorContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+  hasAdditionalLeftPadding,
+  handleClickResourceItem,
+}: {
+  resourceItemsWithAncestryStack: ResourceItemContent[][]
+  isResourceIdHighlighted: (resourceId: string) => boolean
+  isResourceItemDisabled: (resourceItem: ResourceItemContent) => boolean
+  hasAdditionalLeftPadding: boolean
+  handleClickResourceItem: (
+    resourceItemWithAncestryStack: ResourceItemContent[],
+  ) => void
```
seaerchin, line 37: style: usually if it gets so big the props shud be typed outside but not really a big issue

apps/studio/src/components/ResourceSelector/ResourceSelectorContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+  return resourceItemsWithAncestryStack.map((resourceItemWithAncestryStack) => {
+    const lastChild = lastResourceItemInAncestryStack(
+      resourceItemWithAncestryStack,
+    )
+
+    if (!lastChild) {
+      throw new Error(
+        "Unexpected undefined lastChild from lastResourceItemInAncestryStack",
+      )
+    }
```
seaerchin, line 48: double check - usually when we throw errors, we should have a `ErrorBoundary` to recover. this allows us to 
1. isolate the error cleanly 
2. allow the users to self-recover, if possible. 

i think not a big deal if htis error is indeed never expected (and hence crashworthy) as we have top level error boundaries for precisely this situation (unrecoverable errors)

apps/studio/src/components/ResourceSelector/ResourceSelectorContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+export const SuspendableContent = ({
+  resourceItemsWithAncestryStack,
+  isResourceIdHighlighted,
+  isResourceItemDisabled,
+  hasAdditionalLeftPadding,
+  handleClickResourceItem,
+  isSearchQueryEmpty,
+  searchQuery,
+  clearSearchValue,
+  isLoading,
```
seaerchin, line 114: this component is getting very bloated already - we used to have this issue in Classic, where components had 10+ props and took on a load of responsibilities, which in turn limited dev speed as we were afraid of breaking something accidentally. 

not going to block this PR because i think the functionality should come first but it would be good if we could separate out these props (if possible) in a future PR

apps/studio/src/components/ResourceSelector/ResourceSelectorContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+  if (resourceItemsWithAncestryStack.length === 0) {
+    return isSearchQueryEmpty ? (
+      <NoItemsInFolderResult />
+    ) : (
+      <ZeroResult
+        searchQuery={searchQuery}
+        handleClickClearSearch={clearSearchValue}
+      />
+    )
+  }
```
seaerchin, line 140: style (not actionable): should we prefer doing early returns rather than having the nested ternary for readability?

```tsx
const hasNoItems = resourceItemsWithAncestryStack.length === 0
if (hasNoItems && isSearchQueryEmpty) return ...
if (hasNoItems) return ...
```

apps/studio/src/components/ResourceSelector/useResourceQuery.tsx
```diff
@@ -0,0 +1,55 @@
...
+        ? resourceIds
+        : pages.flatMap(({ items }) => items).map((item) => item.id),
+    })
+
+  return {
+    resourceItemsWithAncestryStack,
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    fetchNextPage: useResourceIdsFromSearch ? () => {} : fetchNextPage,
+    hasNextPage: useResourceIdsFromSearch ? false : hasNextPage,
+    isFetchingNextPage: useResourceIdsFromSearch ? false : isFetchingNextPage,
```
seaerchin, line 53: why would we ever make this determination? the caller gets these functions/props back and is in charge of using them as-is. 

the caller also knows if we should `useResourceIdsFromSearch` as it's derived from something they pass (`!!resourceIds`) 

all of this implies that we should just eliminate the ternaries here and let the caller make this judgement themselves, no?

apps/studio/src/components/ResourceSelector/useResourceSelector.tsx
```diff
@@ -0,0 +1,144 @@
...
+}: {
+  interactionType: "move" | "link"
+  siteId: number
+  moveDest: ResourceItemContent | undefined
+  resourceStack: ResourceItemContent[]
+  isResourceHighlighted: boolean
+  setIsResourceHighlighted: (isResourceHighlighted: boolean) => void
+  existingResource: ResourceItemContent | undefined
+  setResourceStack: (resourceStack: ResourceItemContent[]) => void
+  removeFromStack: (numberOfResources: number) => void
```
seaerchin, line 29: not sure if i'm missing context - why not let the client handle this since they can choose to call `useResourceStack`?

alternatively, why not just call `useResourceStack` inside this function? do we foresee these props changing (ie, not from `useResourceStack`?)

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  // NOTE: This is the stack of user's navigation through the resource tree
+  // NOTE: We should always start the stack from `/` (root)
+  // so that the user will see a full overview of their site structure
+  const [resourceStack, setResourceStack] = useState<ResourceItemContent[]>(
+    pendingMovedItemAncestryStack,
+  )
+
+  const [isResourceHighlighted, setIsResourceHighlighted] =
+    useState<boolean>(!!selectedResourceId)
```
seaerchin, line 29: nit: not entirely sure this should be here rather than letting the caller handle it

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -413,6 +414,107 @@ export const publishResource = async (
   return addedVersionResult
 }
 
+interface ResourceItemContentWithPath extends ResourceItemContent {
+  path: string[]
+}
+export const getBatchAncestryWithSelfQuery = async ({
```
seaerchin, line 420: not super sure i understand this!

based off what i can tell, we first get all the resources inside this site and the list of resource Ids. thereafter, we find each resource and all connected resources in its path and push them to groups then return it? 

can just confirm if my understanding si correct?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -506,14 +610,7 @@ export const getSearchResults = async ({
...
   })
-    .where("Resource.type", "in", [
-      // only show user-viewable resources (excluding root page, folder meta etc.)
-      ResourceType.Page,
-      ResourceType.Folder,
-      ResourceType.Collection,
-      ResourceType.CollectionLink,
-      ResourceType.CollectionPage,
-    ])
+    .where("Resource.type", "in", resourceTypes)
```
seaerchin, line 613: didn't check, but a friendly reminder to update the callsites to pass in `resourceTypes`

adriangohjw:
@adriangohjw 
- add comments for CTE to help improve understanding
- add backend tests for batch queries

apps/studio/src/components/ResourceSelector/ResourceSelectorContent.tsx
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/865/commits/97de6fef8fd11fba128e8928d71cea07df2df357

adriangohjw:
@harishv7 @seaerchin updates:
- https://github.com/opengovsg/isomer/pull/865/commits/463d54ee55f7444c17342bb21d8b2c99a47af5c6 : simplified the query code by shifting the application code into DB code using postgres native `jsonb_build_array` and `jsonb_build_object`
- https://github.com/opengovsg/isomer/pull/865/commits/3f257b95b2bf55c1e7eb6c6117ee3ef20948abac : added unit test for query

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: replaced with native PG code. re: https://github.com/opengovsg/isomer/pull/865#issuecomment-2628133528

apps/studio/src/components/ResourceSelector/ResourceSelectorContent.tsx
adriangohjw: addressed in https://github.com/opengovsg/isomer/pull/865/commits/0f3333eda8216a3d3ef229592e13cf04dc31aef3

seaerchin (changes requested):
i think we need to handle the case of someone wanting to move a resource to root - either we need to disallow the move button or to show a toast and block the move!

right now, what happens is that it fails because the resource id is `null` on `getNestedFolderChildrenOf` (see video) 

Not a hard blocker on my end - but having 2 loading animations is kinda weird; as far as i can tell, it's only blocking on the the title for initial load, so i'm not sure why we need it; 

cc @sehyunidaaa for other ideas re loading and whether we want to disable the save button at root or allow but show toast (i think disable cleanest + cos we maintain stack, we can just disable when stack is empty)


https://github.com/user-attachments/assets/41c1dbfe-495b-4b20-af23-b5d2a6a5b6d4

adriangohjw:
> i think we need to handle the case of someone wanting to move a resource to root - either we need to disallow the move button or to show a toast and block the move!
> 
> right now, what happens is that it fails because the resource id is `null` on `getNestedFolderChildrenOf` (see video)

fixed in https://github.com/opengovsg/isomer/pull/865/commits/ea428d73c32cb3430a0f1ffc1a3decb36456483a !
 
> Not a hard blocker on my end - but having 2 loading animations is kinda weird; as far as i can tell, it's only blocking on the the title for initial load, so i'm not sure why we need it;

fixed in https://github.com/opengovsg/isomer/pull/865/commits/bdd3b2eed98f0ee3ab430bb6a540ecbc941dc69c !

> cc sehyunidaaa for other ideas re loading and whether we want to disable the save button at root or allow but show toast (i think disable cleanest + cos we maintain stack, we can just disable when stack is empty)

actually why wanna disable save button at root ah? that should be an option?

seaerchin:
> actually why wanna disable save button at root ah? that should be an option?

we don't necessarily need to disable the save button, but we need to block the save if the user isn't an admin - we don't allow creation of folders/pages at root because of our permissions. this is because sites like MOH have sub-teams, which we use folders to partition the stuff belonging to each team and the "root" space is for the overall MOH itself. 

hence, if we allow moving of pages in folders -> root, we'll be circumventing this measure for preventing cross team edits. lmk if this makes sense

adriangohjw:
@seaerchin okay thanks for explaining the context! Is this a MOH only thing? Anyway i will proceed to add fix this with a comment to this convo, to give more context for subsequent maintainers (+ reminder to fix this once we fix the permission issues)

seaerchin:
> @seaerchin okay thanks for explaining the context! Is this a MOH only thing? Anyway i will proceed to add fix this with a comment to this convo, to give more context for subsequent maintainers (+ reminder to fix this once we fix the permission issues)

not a MOH only thing, but MOH is the biggest case. alot of ministry sites have this issue, where different sectinos of the url are owned by different teams. 

eg: last time, biosafety etc used to be on www.moh.gov.sg/biosafety

apps/studio/src/components/ResourceSelector/useResourceStack.tsx
```diff
@@ -38,9 +40,17 @@ export const useResourceStack = ({
...
+    (numberOfResources: number): ResourceItemContent[] => {
+      let updatedStack: ResourceItemContent[] = []
+      setResourceStack((prev) => {
+        updatedStack = prev.slice(0, -numberOfResources)
+        return updatedStack
+      })
+      return updatedStack
+    },
+    [],
+  )
```
seaerchin, line 53: [question] is this just to return the new state of the stack? curious why we don't just return `prev` directly so it's easier to understand (but not a huge issue since this part of the codebase is unlikely to change)

seaerchin (approved):
lgtm code-wise, but could i just trouble you to add a few more error messages in our call out to `move`?


https://github.com/user-attachments/assets/dae9c14a-fdfa-4bb1-8b09-1c57bbf6fe5f

right now, we show `FORBIDDEN` in 2 cases: 
1. move to original location - i'm not opiniated here, can either allow (less jarring, no consequence) or show an informative error message
2. move to root - i think showing `FORBIDDEN` isn't quite useful enough for the user to self-resolve and this might benefit from a slightly clearer error message. 

approving first but please add a new error message for case #2!

adriangohjw:
> move to original location - i'm not opiniated here, can either allow (less jarring, no consequence) or show an informative error message

fixed in https://github.com/opengovsg/isomer/pull/865/commits/0efba8f1bb79672ca39c0605ae0d20ce030ca318

PR #867, opened 2024-11-07
isom-1655 display block contents for identification

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1655/display-block-contents-for-identification

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- added logic to generate label based on block content

## Screenshots

(refer to storybook for full list of examples)

image


packages/components/src/templates/next/render/renderComponentPreviewText.ts
seaerchin, line 1: just abit confused - why are we adding this here when the dependency is in studio?

packages/components/src/templates/next/render/renderComponentPreviewText.ts
```diff
@@ -0,0 +1,69 @@
+import Ajv from "ajv"
+
+import type { ProseContent, TextProps } from "~/interfaces"
+import type { IsomerSchema } from "~/types"
+import { TextSchema } from "~/interfaces"
+
+function getTextContentOfProse(content: ProseContent): string {
+  const values: string[] = []
+
+  function recursiveSearch(obj: Record<string, unknown>) {
```
seaerchin, line 10: i haven't tried it yet, but `editor.getText` (see [here](https://tiptap.dev/docs/editor/api/editor#gettext)) seems ideal for this - should we do it on studio since this appears to be studio specific?

seaerchin (commented):
alot of the code here is complicated because it appears to be dealing with prose - if we use tiptap on studio (**not on components**), this would be alot easier

packages/components/src/templates/next/render/renderComponentPreviewText.ts
adriangohjw: we aren't using the tiptap editor for this part yet

seaerchin: nono, i mean since this part comes from tiptap, we can shift this over to tiptap

packages/components/src/templates/next/render/renderComponentPreviewText.ts
adriangohjw: yeah i think it's a tough greyline as the initial implementation is to have it in studio

but had a discussion with @dcshzj and decided that studio shouldn't have to know what's going on in the Component library. Plus when we add new component on the Component library, we will also be less likely to forget about it

packages/components/src/templates/next/render/renderComponentPreviewText.ts
adriangohjw: oh as in, current implementation is to put it in Component library which don't have tiptap installed

seaerchin: eh bad phrasing on my end, i meant since this part comes from tiptap, we can shift over to studio!

seaerchin (approved):
lgtm

PR #869, opened 2024-11-11
fix(404): add button for searches

seaerchin:
## Problem
on the 404 page, users are not able to take action to self-resolve the issue

## Solution
1. add a button (for now) that redirects to the searchsg page so that users can self-resolve the issue


packages/components/src/templates/next/layouts/NotFound/SearchButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const lastUrlSegment = permalink.split("/").at(-1) ?? ""
+  // NOTE: Replace all non-alphanumeric characters with spaces
+  // then remove all spaces and join by `+`.
+  // This is because we might have run-on spaces from sequences of symbols
+  // like: `+=`, which would lead to 2 spaces
+  const missingPath = lastUrlSegment
+    .replaceAll(/[\W_]/gi, " ")
+    .split(" ")
+    .filter((v) => !!v)
+    .join("+")
```
adriangohjw, line 33: perhaps move this into a utils to add unit test?

seaerchin: sounds good!

seaerchin: donezxc

packages/components/src/utils/__tests__/getWordsFromPermalink.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Arrange
+    const singleLevelPermalink = "/this-._single=level|"
+    const expected = "this+single+level"
+
+    // Act
+    const actual = getWordsFromPermalink(singleLevelPermalink)
+
+    // Assert
+    expect(actual).toBe(expected)
+  })
```
adriangohjw, line 40: unsure if i understand the test description correctly but `=` isn't preserve but converted?

packages/components/src/utils/getWordsFromPermalink.ts
```diff
@@ -0,0 +1,15 @@
+export const getWordsFromPermalink = (permalink: string): string => {
```
harishv7, line 1: checking understanding:

so https://example.com/some-page_name+ will yield `some+page+name` right

seaerchin: yeap excatly

packages/components/src/templates/next/layouts/NotFound/SearchButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { LinkButton } from "../../components/internal/LinkButton"
+
+type NotFoundSearchButtonProps = Pick<NotFoundPageSchemaType, "LinkComponent">
+export const NotFoundSearchButton = ({
+  LinkComponent,
+}: NotFoundSearchButtonProps) => {
+  const [permalink, setPermalink] = useState("")
+
+  useLayoutEffect(() => {
+    // The check for typeof window and navigator ensures this only runs in browser environments, not during server-side rendering
```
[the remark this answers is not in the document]
seaerchin: it's not server-side rendered but nextjs will transpile then serve the js at runtime. this led to compile errors as `window` is `undefined` when transpiling

packages/components/src/utils/__tests__/getWordsFromPermalink.test.ts
seaerchin: this is my bad, i was flipflopping on whether we should keep this due to query params but i decided teh `=` wasn't going to impact the search results much - i'll just remove this test and add a `=` character ot the other permlainks

packages/components/src/templates/next/layouts/NotFound/SearchButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { getWordsFromPermalink } from "~/utils"
+import { LinkButton } from "../../components/internal/LinkButton"
+
+type NotFoundSearchButtonProps = Pick<NotFoundPageSchemaType, "LinkComponent">
+export const NotFoundSearchButton = ({
+  LinkComponent,
+}: NotFoundSearchButtonProps) => {
+  const [permalink, setPermalink] = useState("")
+
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
seaerchin: ah gotcha, i thuoght we only want to do this post-paint but i'll just update to `useEffect` for effciency

adriangohjw (approved):

dcshzj:
> When deploying to a test site, im unsure why does the URL render an incorrect link `/search/?q=...` instead of `/search?q=...` when the react props is correct, will look into it

We know that this comes from the `trailingSlash` config inside next.config.js. The CloudFront redirect function will need to be updated to handle this.

adriangohjw:
need to update CF function from 

```js
async function handler(event) {
  const request = event.request;
  const originalUri = request.uri;
  const sanitizedUri = request.uri.split("/").filter((part) => part !== "").join("/");
  request.uri = "/" + sanitizedUri;

  // rest of the code
}
```

to 

```js
async function handler(event) {
  const request = event.request;
  const originalUri = request.uri;

  // Separate path and query string
  const [path, queryString] = originalUri.split('?')
  const sanitizedPath = path.split('/').filter((part) => part !== "").join('/')
  request.uri = "/" + sanitizedPath;

  // Append query string back if it exists
  if (queryString) {
    request.uri += `?{queryString}`
  }
  
  // rest of the code
}
```

adriangohjw (approved):
after digging more into nextjs, it seems like entering `/search/?q=abc` doesn't work (incorrectly redirected by our CF to `/search`) but clicking on the button rendered by nextjs does not go through our CF redirect so this still works.

i think we don't have to update our CF since we don't have to support people typing `/search/?q=abc` for now

nevertheless, have logged them into https://linear.app/ogp/issue/ISOM-1680/aws-cloudfront-function-does-not-handle-edge-case for future work

PR #873, opened 2024-11-12
Fix dependabot yml

adriangohjw:
## Problem

Currently dependabot isn't working. for example, there's a new version in `@opengovsg/starter-kitty-validators` but it's not automatically bumped

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- increase `open-pull-requests-limit` to 10 (default 5)

**Bug Fixes**:

- add `versioning-strategy: increase` according to https://github.com/dependabot/dependabot-core/issues/5226 as this seems to be due to us using monorepo npm workspaces

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+    versioning-strategy: increase
     groups:
       aws-sdk:
         applies-to: version-updates
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,7 @@ updates:
     schedule:
       interval: daily
     versioning-strategy: increase
+    open-pull-requests-limit: 10
     groups:
       aws-sdk:
         applies-to: version-updates


.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+    versioning-strategy: increase
```
harishv7, line 7: actually both of these keys are not required according to the [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file). Just curious why adding this 2 fixes the issue?

adriangohjw: im not exactly sure but i was wondering why the bot isn't triggering and seems like someone else had the same issue and was fixed by adding `versioning-strategy` (https://github.com/dependabot/dependabot-core/issues/5226) so I thought I'd give it a try

the other one `open-pull-requests-limit` had nothing to do with the issue but a good to have, to increase the limit from the default 5 to 10. Once the limit is hit, it stopped opening a new PR and from experience it's very easy to hit the default 5 limit

harishv7: alright

harishv7 (approved):

PR #875, opened 2024-11-13
global search backend

adriangohjw:
## Problem

Backend API needed for global search

## Solution

**Breaking Changes**

  - Details ...

### API Endpoint structure

```ts
{
  totalCount: number : null,
  resources: ResourceResultType[],
  suggestions: {
    recentlyEdited: ResourceResultType[]
  }
}
```

where 

ResourceResultType is `{ id, title, type, parentId, lastUpdatedAt, fullPermalink }`
where as all are `String` types

for example

```json
{
    "id": "21326",
    "title": "Update on Local COVID-19 Situation (8 Sep 2021)",
    "type": "CollectionPage",
    "parentId": "14258",
    "lastUpdatedAt": "2024-10-28T18:27:42.687Z",
    "fullPermalink": "newsroom/update-on-local-covid-19-situation-(8-sep-2021)"
},
```

### Description of matching conditions

- only return resources that's viewable by users (exclude meta resources etc.)
- ordered by the number of matching terms in the query with the title of the resources
  - for each word in the query (split by white spaces), we do a `ILIKE` lookup against Resource.title and return if:
    - the word is at the start of the title e.g. search term `hello` in `hello world`
    - the word is in the middle of the title e.g. search term `world` in `hello world`
  -  we currently do not return the following to keep the query simple and fast:
     - the word is prefix by non-spaces e.g. search term `world` in `hello-world`, `hello "world"`
- `lastUpdatedAt` is generated by getting the latest value between the Resource `updatedAt` and it's draft blob `updatedAt` (if applicable)
 
### Suggestions (recently edited etc.)

currently, `recentlyEdited` are an array of resources placed in `suggestions` object. The reason for this data structure is for extensibility where we want to give other form of suggestions to the users

### How

added gin trigram index on `Resource.title`

### Benchmarking

benchmarked by duplicating MOH prod data to local DB (>10k resources)
- if there's 5 words in the query (e.g. `minister covid parliament 2024 daily`), it takes 120-140ms
- if there's only 1 word in the query (e.g. `covid`), it takes 60-90ms
- note: very basic benchmarking done here + time shown is time for TRPC response, not DB query 

## Tests

go to `apps/studio` and run `npm run test:unit`


apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -535,4 +542,158 @@ export const resourceRouter = router({
...
+    .input(searchSchema)
+    .query(async ({ input: { siteId, query = "", cursor: offset, limit } }) => {
+      interface ResourceInterface {
+        id: string
+        title: string
+        type: string
+        parentId: string | null
+        lastUpdatedAt: Date | null
+        fullPermalink: string
+      }
```
adriangohjw, line 556: note - this has been extracted out into another file in https://github.com/opengovsg/isomer/pull/882 - will not duplicate the effort here

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,183 @@ export const publishResource = async (
 
   return addedVersionResult
 }
+
+export const getWithFullPermalink = async ({
```
adriangohjw, line 414: this is refactored and extracted out from an existing code in resource.router so that its reusable

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,183 @@ export const publishResource = async (
...
+  resources,
+}: {
+  resources: Omit<SearchResultResource, "fullPermalink">[]
+}): Promise<SearchResultResource[]> => {
+  return await Promise.all(
+    resources.map(async (resource) => ({
+      ...resource,
+      fullPermalink: await getWithFullPermalink({
+        resourceId: resource.id,
+      }).then((r) => r?.fullPermalink ?? ""),
```
dcshzj, line 481: question: seems like this is the same steps as using the `getResourceFullPermalink` function above?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,183 @@ export const publishResource = async (
...
+  query,
+  offset,
+  limit,
+}: {
+  siteId: number
+  query: string
+  offset: number
+  limit: number
+}): Promise<{
+  totalCount: number | null
```
dcshzj, line 497: issue: having `null` here seems weird at first glance, should we set `null` to 0 instead?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,183 @@ export const publishResource = async (
...
+    orderedResources = orderedResources.orderBy(
+      sql`(
+        ${sql.join(
+          searchTerms.map(
+            (searchTerm) =>
+              // 1. Match if the search term is at the start of the title
+              // 2. Match if the search term is in the middle of the title (after a space)
+              sql`
+                CASE WHEN "Resource"."title" ILIKE ${searchTerm + "%"} THEN 1 ELSE 0 END +
+                CASE WHEN "Resource"."title" ILIKE ${"% " + searchTerm + "%"} THEN 1 ELSE 0 END
```
dcshzj, line 540: nit: the interpolation I think can put the `%` outside (easier to read but not sure if it works, please check)

```suggestion
                CASE WHEN "Resource"."title" ILIKE ${searchTerm}% THEN 1 ELSE 0 END +
                CASE WHEN "Resource"."title" ILIKE %${searchTerm}% THEN 1 ELSE 0 END
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,183 @@ export const publishResource = async (
...
+    )
+
+  // Currently ordered by number of words matched
+  // followed by `lastUpdatedAt` if there's a tie-break
+  let orderedResources = queriedResources
+  if (searchTerms.length > 1) {
+    orderedResources = orderedResources.orderBy(
+      sql`(
+        ${sql.join(
+          searchTerms.map(
```
dcshzj, line 534: question(non-blocking): I'm not too sure _why_ this works because it seems like we don't sum the number of matches then use that sum for ordering, so the max that each resource will have is 1. If it has been tested and it works then ok!

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,183 @@ export const publishResource = async (
...
+    .where((eb) =>
+      eb.or([
+        ...searchTerms.map((searchTerm) =>
+          // Match if the search term is at the start of the title
+          eb("Resource.title", "ilike", `${searchTerm}%`).or(
+            // Match if the search term is in the middle of the title (after a space)
+            eb("Resource.title", "ilike", `% ${searchTerm}%`),
+          ),
+        ),
+      ]),
```
dcshzj, line 524: minor nit: I think we don't need to spread the array? (editing on GitHub so the indentation might be incorrect)

```suggestion
      eb.or(searchTerms.map((searchTerm) =>
          // Match if the search term is at the start of the title
          eb("Resource.title", "ilike", `${searchTerm}%`).or(
            // Match if the search term is in the middle of the title (after a space)
            eb("Resource.title", "ilike", `% ${searchTerm}%`),
          ),
        )),
```

apps/studio/tests/integration/helpers/seed/index.ts
```diff
@@ -211,11 +211,32 @@ export const setupPageResource = async ({
...
+      break
+    case "IndexPage":
+      fallbackTitle = "test index page"
+      fallbackPermalink = "test-index-page"
+      break
+    default:
+      fallbackTitle = "test page"
+      fallbackPermalink = "test-page"
+      break
+  }
```
dcshzj, line 233: minor nit: probably can abstract as a function then return both fallback title and permalink as an object

apps/studio/src/schemas/resource.ts
```diff
@@ -70,3 +71,18 @@ export const getAncestrySchema = z.object({
...
+    query: z.string().optional(),
+  })
+  .merge(infiniteOffsetPaginationSchema)
+
+export const searchOutputSchema = z.object({
+  totalCount: z.number().nullable(),
+  resources: z.array(z.custom<SearchResultResource>()),
+  suggestions: z.object({
+    recentlyEdited: z.array(z.custom<SearchResultResource>()),
+  }),
```
dcshzj, line 87: question: will there be other kinds of suggestions in the future?

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -535,4 +512,36 @@ export const resourceRouter = router({
 
       return ancestors.reverse().slice(0, -1)
     }),
+
+  search: protectedProcedure
```
dcshzj, line 516: question: can't seem to find it but it seems that we are not validating the user's permissions to view the resources? Would it mean that it is possible for a user with insufficient permissions to be able to view a certain resource that they don't have access to?

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: i was wondering what's the best way to return this `totalCount` because if a user types in empty spaces, it's not exactly necessarily `0` since it would be misleading someone to think "no results found" as its more of "results not returned"

apps/studio/tests/integration/helpers/seed/index.ts
adriangohjw: its only used within this method so i think it's fine, might be more complicated to extract out and refer to it elsewhere

dcshzj: Hmm it's more of safety as `let` can be dangerous to use especially when there are more stuff to be done after this part. Using `const` here gives the assurance that it wouldn't accidentally be changed down the line, and the function does one thing by itself.

This is more of a best practices thing but given this is a helper function I leave it up to you.

apps/studio/src/schemas/resource.ts
adriangohjw: no such explicit product or design requirement

but i remember during designcrit there seemed to be a common theme and questions around supporting other form of suggestions e.g. non-resource pages like settings, or isomer guide

was also taking inspiration from other search systems e.g. Notion, Slack, and they seem to support other form of suggestions. Or even filters

thus decided to make the call to pre-empt that and structure it this way

apps/studio/src/server/modules/resource/resource.router.ts
adriangohjw: yup you are right!

decided there's no need for this now as
- users would still be able to see the resources on the left sidebar either way
- the current way of setting resource permission could potentially lead to very nested query to check if a user has permission for each resource

will leave this for later when we want to really to implement a studio-wide resource-based permission

apps/studio/src/schemas/resource.ts
dcshzj: Hmm I feel that we shouldn't be doing this pre-emptively especially when this is more of an internal API that is relatively easy to change. Having this means that we will definitely support providing other kinds of suggestions (which may or may not require a backend call to retrieve), but we don't have a foreseeable timeline yet regarding whether we will actually add such a support.

apps/studio/src/server/modules/resource/resource.router.ts
dcshzj: Hmm but I thought the permissions is on a site level? It seems like it is possible for someone without any role to access the site to be able to use this search endpoint to get the list of resources for the site, even though by right they don't have access.

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: oh, this actually does because for every match for the searchterm we give it a score of `1` then we sum them based on the number of terms
```ts
    orderedResources = orderedResources.orderBy(
      sql`(
        ${sql.join(
          searchTerms.map(
            (searchTerm) =>
              // 1. Match if the search term is at the start of the title
              // 2. Match if the search term is in the middle of the title (after a space)
              sql`
                CASE WHEN "Resource"."title" ILIKE ${searchTerm + "%"} THEN 1 ELSE 0 END +
                CASE WHEN "Resource"."title" ILIKE ${"% " + searchTerm + "%"} THEN 1 ELSE 0 END
              `,
          ),
          sql` + `,
        )}
      ) DESC`,
```

an equivalent SQL query will be
```sql
order by 
  CASE WHEN "Resource"."title" ILIKE 'word1%' THEN 1 ELSE 0 END +
  CASE WHEN "Resource"."title" ILIKE '% word1%' THEN 1 ELSE 0 END +
  CASE WHEN "Resource"."title" ILIKE 'word2%' THEN 1 ELSE 0 END +
  CASE WHEN "Resource"."title" ILIKE '% word2%' THEN 1 ELSE 0 END +
  ...
```

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: for some reason on how kysely construct the query, this will not work. suspect it has something to do with how `sql` is parsed in kysely 🥲

apps/studio/src/schemas/resource.ts
adriangohjw: okay makes, sense, shall do!

apps/studio/src/server/modules/resource/resource.router.ts
adriangohjw: ah okay i misunderstood what you meant. Will add the check for site-level permission!

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: since the backend do something params processing e.g. `query.trim() === ""`, it might be a duplicate of logic - unless we move that to frontend too

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: yup. don't think there's much difference between either other than data structure and `getWithFullPermalink` seem more used by resource side of things

didn't wanna touch `getResourceFullPermalink` - perhaps can add this as a note to refactor to be the same in subsequent PRs?

dcshzj (approved):

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -409,3 +410,189 @@ export const publishResource = async (
...
+    .limit(limit)
+    .execute()) as SearchResultResource[]
+
+  const totalCount: number = (
+    await db
+      .with("queriedResources", () => queriedResources)
+      .selectFrom("queriedResources")
+      .select(db.fn.countAll().as("total_count"))
+      .executeTakeFirstOrThrow()
+  ).total_count as number // needed to cast as the type can be `bigint`
```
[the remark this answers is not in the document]
adriangohjw: yup it doesn't

PR #878, opened 2024-11-14
feat: allow dividers in Tiptap editors

dcshzj:
## Problem

We previously removed the ability to create dividers inside the Tiptap editor due to schema incompatibility, so it is not possible to add dividers in Tiptap currently.

## Solution

**Breaking Changes**

**Features**:

- Add dividers to all Tiptap editors.
    - The components schema also matches the schema defined within the Tiptap editor and hence, it is now safe to add this back.
- Fixed the Table schema to disallow dividers within table cells (which was the original intent from design).


adriangohjw (commented):
> Fixed the Table schema to disallow dividers within table cells (which was the original intent from design).

does this currently break any of the existing schema in DB?

dcshzj:
> does this currently break any of the existing schema in DB?

Checked through all the blobs in the GitHub repos and in the DB, no dividers are currently being used inside table cells.

packages/components/src/interfaces/native/Table.ts
```diff
@@ -50,7 +50,6 @@ const TableCellSchema = Type.Unsafe<TableCellProps>(
     attrs: Type.Optional(TableBaseCellSchema),
     content: Type.Array(
       Type.Union([
-        Type.Ref(DividerSchema),
```
adriangohjw, line 53: remove unused import `DividerSchema`

dcshzj: Updated in ff787dfb48ef98b6ed626620874ccab632937dd8.

adriangohjw (approved):

PR #880, opened 2024-11-15
Studio - only show intercom post-login

adriangohjw:
## Problem

Currently, Intercom script rendered on every page on studio.

However, we only want to show it on pages that can be seen by login-ed users to reduce unnecessary intercom spam.

Our workabout now is to disable it on Intercom's settings side 

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- make a `WithIntercomWrapper` that takes in children and then conditionally render the Intercom component (from the JS library) if the environment variable found + wrap it around pages that we want to show it


seaerchin (commented):
could we shift the code inside `Intercom` to `EnforceLoginStateWrapper`?  should be automatic then

PR #885, opened 2024-11-16
feat: support editing map and video embeds

dcshzj:
## Problem

We have the map and video embed components, but no way to create/edit them on Studio.

Closes ISOM-1676.

## Solution

**Breaking Changes**

**Features**:

- Add the new map and video embed components as supported components that can be added for article and content pages.
- Introduced a new embed form control in JSONForms to allow users to paste in the embed code, which is then sanitised and parsed to extract the actual URL of the embed using DOMPurify.
- Expanded the Content Security Policy to allow the various map and video embed providers.

## Screenshots

image
image
image


sehyunidaaa:
Hey just curious! is there a reason why there's an 'update' button instead of having the textarea directly on the form?

dcshzj:
> Hey just curious! is there a reason why there's an 'update' button instead of having the textarea directly on the form?

Clarified offline! Mainly is because we are constructing the iframe ourselves (for security) rather than letting the user input any iframe that they want. The validation for iframes would also be harder to perform compared to just checking the URL which we are currently doing.

sehyunidaaa:
Thank you! Added some minor design suggestions here after our discussion! @dcshzj 

https://www.figma.com/design/g7LnLn4IUcUC9dYxgsHEJM/IsomerCMS-V2-(working-title%3A-%22Isomer-Studio%22)?node-id=10835-42086

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsEmbedControl.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+              <FormErrorMessage>{errors.embedCode?.message}</FormErrorMessage>
+            </FormControl>
+
+            <Infobox size="sm" variant="info" mt="1.25rem">
+              <Box>
+                <Text>You can embed content from:</Text>
+                <UnorderedList ml="1.5rem" mt="0.25rem">
+                  <ListItem>Video: {SUPPORTED_VIDEOS.join(", ")}</ListItem>
+                  <ListItem>Map: {SUPPORTED_MAPS.join(", ")}</ListItem>
```
adriangohjw, line 138: This component is shared between both maps and videos, so both lines will always appear

However, this can be confusing on the UI side because `Map: Google Map, OneMap` shows up when user is trying to add Video instead

Perhaps we can detect and only show the relevant ones?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsEmbedControl.tsx
```diff
@@ -0,0 +1,242 @@
...
+)
+
+interface EmbedCodeModalProps {
+  isOpen: boolean
+  onClose: () => void
+  onSave: (embedCode: string) => void
+  urlPattern?: string
+}
+
+function EmbedCodeModal({
```
adriangohjw, line 67: curious - why did we choose to open this in a modal instead of editing with the "label" field on the left-hand sidebar instead? seems like an additional step

apps/studio/src/features/editing-experience/components/utils.ts
```diff
@@ -91,3 +107,35 @@ export const generatePreviewSitemap = (
...
+      return acc
+    }
+
+    const [embedName, regex] = curr
+    if (new RegExp(regex).test(url)) {
+      return EMBED_NAME_MAPPING[embedName as keyof typeof EMBED_NAME_MAPPING]
+    }
+
+    return undefined
+  }, undefined)
```
adriangohjw, line 141: do we need a `reduce` here? since we aren't really needing the accumulation aspect of 

what about a simple loop with early return?

```suggestion
export const getEmbedNameFromUrl = (url: string): string | undefined => {
  const entries = Object.entries({
    ...MAPS_EMBED_URL_REGEXES,
    ...VIDEO_EMBED_URL_REGEXES,
  })

  for (const [embedName, regex] of entries) {
    if (new RegExp(regex).test(url)) {
      return EMBED_NAME_MAPPING[embedName as keyof typeof EMBED_NAME_MAPPING]
    }
  }

  return undefined
}
```

apps/studio/src/features/editing-experience/components/utils.ts
```diff
@@ -91,3 +107,35 @@ export const generatePreviewSitemap = (
...
+    RETURN_DOM_FRAGMENT: true,
+  })
+  const sanitizedUrl = elem.firstElementChild?.getAttribute("src")
+
+  if (sanitizedUrl === null) {
+    return undefined
+  }
+
+  return sanitizedUrl
+}
```
adriangohjw, line 123: - explicitly state the return types so its upfront to the future user of this method 
- simplify the return statements

```suggestion
export const getIframeSrc = (embedCode: string): string | undefined => {
  const elem = DOMPurify.sanitize(embedCode, {
    ALLOWED_TAGS: ["iframe"],
    RETURN_DOM_FRAGMENT: true,
  })
  const sanitizedUrl = elem.firstElementChild?.getAttribute("src")
  return sanitizedUrl ?? undefined
}
```

adriangohjw (commented):
can we also update `readPageAndBlob` for mocking so that it shows up in the stories for page editing?

apps/studio/next.config.mjs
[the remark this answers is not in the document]
dcshzj: Hmm the list seems to be for embedding Google Maps using the JavaScript API, but we are using iframe currently so it's a bit different? Embedding maps locally on my machine doesn't show any CSP errors

apps/studio/next.config.mjs
```diff
@@ -63,7 +68,6 @@ const ContentSecurityPolicy = `
     ;
   connect-src
     'self'
-    https://schema.isomer.gov.sg
```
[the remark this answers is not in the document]
dcshzj: This one was legacy when we hot load the schema from this URL, removing it since we now bundle the schema together with Studio.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsEmbedControl.tsx
dcshzj: Hmm it's cos we are only storing the URL in the schema rather than the entire iframe. If we had this directly inside the page editor, then we will either need to reconstruct how the iframe code looks like, or when users paste into the input field, it gets automatically translated into just the URL which might be unexpected.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsEmbedControl.tsx
dcshzj: Hmm yeah, but mainly wanted to keep things simple first because we are not sure the universe of the embeds yet, and this is still initial designs pending improvements. I think once confirmed, we can improve this logic.

apps/studio/src/features/editing-experience/components/utils.ts
dcshzj: Hmm this one was because I wanted to keep it the same as the one from `getParsedDate`:

https://github.com/opengovsg/isomer/blob/78368f5a50f98f4e62528ccc057c64fb688b2837/packages/components/src/utils/getParsedDate.ts#L13-L40

This was a comment from @seaerchin regarding using reducer pattern.

apps/studio/next.config.mjs
```diff
@@ -36,10 +36,15 @@ const ContentSecurityPolicy = `
...
     https://intercom-sheets.com
-    https://www.intercom-reporting.com 
-    https://www.youtube.com
+    https://www.intercom-reporting.com
     https://player.vimeo.com
     https://fast.wistia.net
+    https://www.google.com
+    https://www.youtube.com
+    https://www.youtube-nocookie.com
+    https://player.vimeo.com
```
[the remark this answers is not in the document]
dcshzj: Oops, removed in bac88de4fe3d5e3260e22751ad78b8a3c56a9213

apps/studio/src/features/editing-experience/components/utils.ts
dcshzj: Updated in bac88de4fe3d5e3260e22751ad78b8a3c56a9213

apps/studio/src/features/editing-experience/components/utils.ts
adriangohjw: ooo i can see the code but not the comment

dcshzj: Ah, comment is here: https://github.com/opengovsg/isomer/pull/732#discussion_r1794060218

dcshzj:
> can we also update `readPageAndBlob` for mocking so that it shows up in the stories for page editing? easier to review the design also

Updated in 8e1889e7a0b54033368546d1d0d4c8c3cf2829ab

adriangohjw (approved):

PR #889, opened 2024-11-19
chore: make growthbook client ID env-specific

dcshzj:
## Problem

The Growthbook client ID is currently using the `production` environment, regardless of the actual deployment environment.

## Solution

**Breaking Changes**

**Improvements**:

- Make the Growthbook client ID be an input for the AWS deploy workflow, so that each environment can define their own client IDs.
    - Note: Both staging and vapt are using the staging client, hence they are the same.
- Also made Intercom take inputs the same way as standardisation.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -39,6 +39,8 @@ jobs:
       app-s3-region: "ap-southeast-1"
       app-s3-assets-bucket-name: "isomer-next-infra-prod-assets-private-a319984"
       app-s3-assets-domain-name: "isomer-user-content.by.gov.sg"
+      app-growthbook-client-key: "sdk-r07MHTLLgfdVDThi"
+      app-intercom-app-id: "jv2tjc3g"
 
     secrets:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy_production.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       app-s3-region: "ap-southeast-1"
       app-s3-assets-bucket-name: "isomer-next-infra-prod-assets-private-a319984"
       app-s3-assets-domain-name: "isomer-user-content.by.gov.sg"
+      app-growthbook-client-key: "sdk-r07MHTLLgfdVDThi"
+      app-intercom-app-id: "jv2tjc3g"
```
adriangohjw, line 43: will the leading and trailing double quotes cause breaking changes?

dcshzj: Hmm I don't think so, because this is a string input into the workflow, and the interpolation inside aws_deploy.yml will remove the quotes.

adriangohjw (approved):

PR #890, opened 2024-11-19
Intercom - Add user info

adriangohjw:
## Problem

Currently intercom is implemented but does not take in user's information. However Intercom stores information in browser cookie (might be wiped out), thus conversations not preserved across devices and sessions

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- pass in information into IntercomSDK after retrieving from `getMe()`
- extract into a standalone component to abstract away logics

### DIFF
--- a/apps/studio/src/components/Intercom.tsx
+++ b/apps/studio/src/components/Intercom.tsx
@@ -1,0 +1,12 @@
+import Intercom from "@intercom/messenger-js-sdk"
+
+import { env } from "~/env.mjs"
+
+export const initIntercom = () => {
+  // Initialize Intercom
+  if (env.NEXT_PUBLIC_INTERCOM_APP_ID) {
+    Intercom({
+      app_id: env.NEXT_PUBLIC_INTERCOM_APP_ID,
+    })
+  }
+}
--- a/apps/studio/src/components/Intercom.tsx
+++ b/apps/studio/src/components/Intercom.tsx
@@ -1,12 +1,24 @@
-import Intercom from "@intercom/messenger-js-sdk"
+import { Intercom as IntercomSDK } from "@intercom/messenger-js-sdk"
 
 import { env } from "~/env.mjs"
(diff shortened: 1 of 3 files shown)


apps/studio/src/components/Intercom.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const Intercom = () => {
+  const { me } = useMe()
+
+  if (env.NEXT_PUBLIC_INTERCOM_APP_ID) {
+    IntercomSDK({
+      app_id: env.NEXT_PUBLIC_INTERCOM_APP_ID,
+      user_id: me.id,
+      name: me.name,
```
dcshzj, line 17: issue: I think we should be using `preferredName` here instead? Also, the `name` column can be empty, so we should fallback to the front part of the email address with replacements to spaces if it is not present.

adriangohjw: actually, on this, why is there a need for both `preferredName` and `name` columns?

Also, i've looked at the DB, `name` are all filled but not `preferredName`, so it's safer to use `name` instead

Noted on the fallback on email, will do!

dcshzj: Hmm the `preferredName` column was created specifically for Intercom actually, iirc it is to make the widget more personal to the user. I think it's just very low on our priority list to allow users to define their preferred names in Studio.

adriangohjw: so intended use case was for name to be `Adrian Goh Jun Wei` and preferred name be `Adrian`?

dcshzj: Yup, so the Intercom widget will say "Hi, Adrian" rather than the full name

adriangohjw: okay lemme fallback from `preferredName` > `name` > `email` username

dcshzj (approved):

PR #896, opened 2024-11-26
feat: add meta image for social media sharing

dcshzj:
## Problem

Sharing our site links on social media doesn't show much, and there is no way to show an image.

## Solution

**Breaking Changes**

**Features**:

- Add the meta image as part of the OpenGraph protocol as an optional field in the meta settings field.

**Improvements**:

- Extract out the functionality to get the meta description as a standalone function for easier readability.

## Tests

- [ ] Go to the "Meta settings" tab
- [ ] Verify that the image attachment component is visible and the label is "Meta image".

NOTE: The image will NOT be uploaded even if you try to attach. This is a bug that will be tackled separately.


packages/components/src/engine/metadata.ts
```diff
@@ -19,34 +82,25 @@ export const getMetadata = (props: IsomerPageSchemaType) => {
...
+      siteName: props.site.siteName,
+      type: props.layout === "article" ? "article" : "website",
+      images: !!metaImageUrl
+        ? [
+            {
+              url: metaImageUrl,
+            },
+          ]
+        : undefined,
+    },
```
seaerchin, line 98: question: what's `openGraph` here and what does it do?

dcshzj: Opengraph is the protocol for sharing links on social media (docs: https://ogp.me/). The object's shape is taken from here: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#opengraph

.github/workflows/chromatic.yml
```diff
@@ -40,6 +40,8 @@ jobs:
               - 'packages/components/**'
             studio:
               - 'apps/studio/**'
+              - 'packages/components/src/interfaces/**'
+              - 'packages/components/src/types/**'
```
adriangohjw, line 44: curious - why is there a need to add this?

packages/components/src/engine/metadata.ts
adriangohjw, line 1: import and use `ISOMER_PAGE_LAYOUTS` instead of hardcoding e.g. `ISOMER_PAGE_LAYOUTS.Database`

packages/components/src/engine/metadata.ts
```diff
@@ -1,12 +1,75 @@
...
+    default:
+      const _: never = props
+      return undefined
+  }
+}
+
+const getMetaImage = (props: IsomerPageSchemaType) => {
+  switch (props.layout) {
+    case "article":
+      return props.page.image?.src || props.meta?.image
```
adriangohjw, line 36: if `meta.image` is found, which means it's intentionally uploaded by user, should we prioritize that?

similar to `getMetaDescription` where we do
```ts
if (props.meta?.description) {
    return props.meta.description
  }
```

packages/components/src/engine/metadata.ts
```diff
@@ -1,12 +1,75 @@
...
+
 export const getMetadata = (props: IsomerPageSchemaType) => {
   const faviconUrl = `${props.site.assetsBaseUrl ?? ""}${props.site.favicon || "/favicon.ico"}`
+  const fullUrl = props.site.url
+    ? new URL(props.page.permalink, props.site.url).toString()
+    : props.page.permalink
+  const metaImage = getMetaImage(props)
+  const metaImageUrl = metaImage
+    ? `${props.site.assetsBaseUrl ?? ""}${metaImage}`
+    : undefined
```
adriangohjw, line 68: suggestion - fallback to `site.logoUrl` instead of undefined?

packages/components/src/engine/metadata.ts
```diff
@@ -1,12 +1,75 @@
...
+      const _: never = props
+      return undefined
+  }
+}
+
 export const getMetadata = (props: IsomerPageSchemaType) => {
   const faviconUrl = `${props.site.assetsBaseUrl ?? ""}${props.site.favicon || "/favicon.ico"}`
+  const fullUrl = props.site.url
+    ? new URL(props.page.permalink, props.site.url).toString()
+    : props.page.permalink
```
adriangohjw, line 64: understand that this is defensive but will `props.site.url` be undefined?
plus can be misleading to call this `fullUrl` and potentially return `props.page.permalink`

adriangohjw (changes requested):

.github/workflows/chromatic.yml
dcshzj: Context: https://opengovproducts.slack.com/archives/C06R4DX966P/p1732607260729969?thread_ts=1732597220.921359&cid=C06R4DX966P

packages/components/src/engine/metadata.ts
dcshzj: Updated in 252ce172e3223a985e958c0b12cc5f2fda117008.

packages/components/src/engine/metadata.ts
dcshzj: Updated in 252ce172e3223a985e958c0b12cc5f2fda117008.

packages/components/src/engine/metadata.ts
dcshzj: Updated in 252ce172e3223a985e958c0b12cc5f2fda117008 as `canonicalUrl` instead

packages/components/src/engine/metadata.ts
dcshzj: Updated in 252ce172e3223a985e958c0b12cc5f2fda117008.

packages/components/src/engine/metadata.ts
```diff
@@ -1,52 +1,106 @@
...
+    case ISOMER_PAGE_LAYOUTS.NotFound:
+      // NOTE: These pages do not appear in search results, so we don't need to
+      // provide a meta description
+      return undefined
+    default:
+      const _: never = props
+      return undefined
+  }
+}
+
```
adriangohjw, line 60:
```suggestion
const getMetaImage = (props: IsomerPageSchemaType) => {
  if (props.meta?.image) {
    return props.meta.image
  }

  switch (props.layout) {
    case ISOMER_PAGE_LAYOUTS.Article:
      return props.page.image?.src
    case ISOMER_PAGE_LAYOUTS.Content:
    case ISOMER_PAGE_LAYOUTS.Database:
    case ISOMER_PAGE_LAYOUTS.Index:
    case ISOMER_PAGE_LAYOUTS.Collection:
      return undefined
    case ISOMER_PAGE_LAYOUTS.Homepage:
      return props.content.find((item) => item.type === "hero")?.backgroundUrl
    case ISOMER_PAGE_LAYOUTS.File:
    case ISOMER_PAGE_LAYOUTS.Link:
    case ISOMER_PAGE_LAYOUTS.Search:
    case ISOMER_PAGE_LAYOUTS.NotFound:
      // NOTE: These pages do not appear in search results, so we don't need to
      // provide a meta description
      return undefined
    default:
      const _: never = props
      return undefined
  }
}

```

dcshzj: Hmm no, I see where you're coming from but we can't do this. The file and link layouts do not have support for meta images, so the attribute will not be present.

Actually in fact I think even for meta descriptions they should be removed.

adriangohjw: if the attribute isn't present, won't `props.meta?.image` be false and skip through that check?

dcshzj: It's typescript that will not allow it, because the attribute doesn't exist on the link and file layouts so there will be a type error.

adriangohjw (approved):

dcshzj:
Thanks @adriangohjw @seaerchin for reviewing. Will put this PR on hold first while I do a refactor of the image uploading. @sehyunidaaa brought up a good point about the uploading not working as intended, so will do that first in a separate PR before this gets merged

dcshzj:
Going to merge this first, the image uploading aspect **will not work** on Studio but we can deprioritise this temporarily as it is not user-facing.

PR #897, opened 2024-11-26
fix: add search to settings

seaerchin:
## Problem
We want to add search to the settigns page but it was previously missing
## Solution
1. rename the layout (since it now has a header)
2. extract header into component since it is now shared
3. add header to settigns layout

## Screenshots
image


apps/studio/src/components/CmsSidebar/CmsSidebarContainer.tsx
adriangohjw, line 1: sorry a bit confused - is this still being used? saw that in the other file `CmsSidebarOnlyContainer` has alos been renamed to `CmsSidebarContainer` and this `CmsSidebarContainer` doesn't seem to be exported/imported anywhere

seaerchin: i think my editor wonked when renaming file/variable, lemme fix it

adriangohjw (approved):

PR #901, opened 2024-11-27
chore: add max length for alt text

seaerchin:
## Problem
previously, our image alt text had no max length. 

## Solution
1. add a max length of 120 - same as our other components and close enough to wcag recommendation of 125


adriangohjw (commented):
are these existing blobs that contains infocards and infopic with length over 120? if so, we would need to give Ops a heads-up as agencies won't be able to update those resources until it's fixed

seaerchin:
merging, no offending instances found (refer to [slack](https://opengovproducts.slack.com/archives/C07CWUNUL68/p1732765586491049?thread_ts=1732692669.634269&cid=C07CWUNUL68))

PR #902, opened 2024-11-27
isom-1678 dynamic blocks for MUIS

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1678/dynamic-blocks-for-muis

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Add `DynamicStatistics` component
  - While this is a MUIS specific component, tried my best to make it a middle ground and make it sufficiently reusable in the future if we need to do date-based API fetching e.g. weather, temperature, PSI, stock tickets. If we want to modify to be a more generic dynamic component, it can be achievable by tweaking it slightly
- use `@tanstack/react-query` to fetch the query from an API endpoint and handle the different non-success states

Note  
- This isn't supported in Studio yet and we have to set a yearly reminder to update this
- The prayers timetable has to be added to `isomer-user-content` S3 bucket. Here's a sample file following the required format for a year that's 73kB - [muis_prayer_timetable_sample.json](https://github.com/user-attachments/files/17933616/muis_prayer_timetable_sample.json) **(THIS ISN'T ACCURATE AND SHOULD NOT BE USED IN PRODUCTION)**

## Screenshots

image

_(refer to storybook for tablet and mobile)_

**New dependencies**:

- Component `@tanstack/react-query` : for making API call and handling the different states

**New dev dependencies**:

- Component `storybook-addon-mock` : needed to mock API calls for storybook without making external network call


packages/components/src/interfaces/complex/DynamicStatistics.ts
```diff
@@ -0,0 +1,80 @@
+import type { Static } from "@sinclair/typebox"
+import { Type } from "@sinclair/typebox"
+
+export const DYNAMIC_STATISTICS_TYPE = "dynamicstatistics"
```
adriangohjw, line 4: Note: im proposing for us to define the type as a const.

Right now, the existing components hardcode it as a string literal which can be less intuitive, especially when they are used in keys.

packages/components/src/templates/next/render/renderComponentPreviewText.ts
```diff
@@ -98,6 +99,8 @@ export function renderComponentPreviewText({
       return getTextContentOfProse(component.content)
     case "video":
       return component.title || "Video embed"
+    case DYNAMIC_STATISTICS_TYPE:
+      return component.apiEndpoint
```
adriangohjw, line 103: don't know what's the best preview text to display - will use `apiEndpoint` now since it's a required field + we aren't supporting this on studio yet

adriangohjw:
Ah marking as draft, will need to change it to use Link component

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -175,6 +176,40 @@ export const DEFAULT_BLOCKS: Record<
     title: "Rick Astley - Never Gonna Give You Up",
     url: "https://www.youtube.com/embed/dQw4w9WgXcQ?si=ggGGn4uvFWAIelWD",
   },
+  // TODO: Replace with actual working API endpoint
+  [DYNAMIC_STATISTICS_TYPE]: {
+    type: `${DYNAMIC_STATISTICS_TYPE}`,
+    apiEndpoint: "https://jsonplaceholder.com/muis_prayers_time",
```
harishv7, line 182: in actual case for this would be a link to a csv for MUIS right?

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -175,6 +176,40 @@ export const DEFAULT_BLOCKS: Record<
...
+        label: "Asar",
+        key: "asar",
+      },
+      {
+        label: "Maghrib",
+        key: "maghrib",
+      },
+      {
+        label: "Ishak",
+        key: "isyak",
```
harishv7, line 207: in your sample json i saw each gregorian date has a corresponding arabic date. how is that obtained?

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -175,6 +176,40 @@ export const DEFAULT_BLOCKS: Record<
...
+      },
+    ],
+    url: "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
```
harishv7, line 210: what's this url for?

apps/studio/src/components/PageEditor/constants.ts
adriangohjw: it will be a link to a URL that returns a JSON, which means we have to convert the CSV to JSON before uploading to S3

reason being 
1. unsure of how CSV parsing is supported on older browsers
2. using JSON allow us to reuse this block in the future for other API endpoints

apps/studio/src/components/PageEditor/constants.ts
adriangohjw: there will be a "view all dates" CTA under the actual date
image

reason being people might want to know prayers time of other dates
(similar to https://www.muis.gov.sg/Media/Islamic-Calendar)

discussed with @sehyunidaaa and we will be doing a similar approach where it's a link to a page with the PDFs

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -175,6 +176,40 @@ export const DEFAULT_BLOCKS: Record<
     title: "Rick Astley - Never Gonna Give You Up",
     url: "https://www.youtube.com/embed/dQw4w9WgXcQ?si=ggGGn4uvFWAIelWD",
   },
+  // TODO: Replace with actual working API endpoint
+  [DYNAMIC_STATISTICS_TYPE]: {
+    type: `${DYNAMIC_STATISTICS_TYPE}`,
+    apiEndpoint: "https://jsonplaceholder.com/muis_prayers_time",
+    title: "hijri_date",
+    statistics: [
```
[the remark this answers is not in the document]
adriangohjw: Thought a lot about this too on whether we want to call it "PrayersBanner"

My take is that while this is a MUIS specific component, we can make it sufficiently reusable in the future if we need to do date-based API fetching e.g. weather, temperature, PSI, stock tickers - all of these stuff being some form of "statistics"/"data" 

Open to other namings too!

apps/studio/src/components/PageEditor/constants.ts
adriangohjw: thinking of it being referenced by the `title` field
 https://github.com/opengovsg/isomer/blob/709c2511a485b12cc24fbc2cfd09b6b9f7b48b50/apps/studio/src/components/PageEditor/constants.ts#L183

apps/studio/src/components/PageEditor/constants.ts
adriangohjw: the label for the CTA is described here:
https://github.com/opengovsg/isomer/blob/709c2511a485b12cc24fbc2cfd09b6b9f7b48b50/apps/studio/src/components/PageEditor/constants.ts#L211

seaerchin:
not a review, just wanted to say that i got rickrolled in my emails frgom this PR LOL
image

packages/components/src/templates/next/layouts/Homepage/Homepage.stories.tsx
```diff
@@ -30,6 +31,24 @@ const meta: Meta<typeof Homepage> = {
...
       themeOverride: "Isomer Next",
     },
+    mockData: [
+      {
+        url: "https://jsonplaceholder.com/muis_prayers_time",
+        method: "GET",
+        status: 200,
+        response: {
+          [getSingaporeDateYYYYMMDD()]: {
+            hijri_date: "17 Jamadilawal 1442H",
```
harishv7, line 41: nit: can standardise to hijriDate to use camelCase as default

apps/studio/src/components/PageEditor/constants.ts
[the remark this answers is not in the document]
harishv7: in spirit of keeping it generic, maybe data would be better

packages/components/src/interfaces/complex/DynamicStatistics.ts
harishv7: then should these const definitions be in a separate file where it defines all the blocks?

packages/components/src/interfaces/complex/DynamicStatistics.ts
```diff
@@ -0,0 +1,85 @@
...
+import { Type } from "@sinclair/typebox"
+
+import type { IsomerSiteProps, LinkComponentType } from "~/types"
+
+export const DYNAMIC_STATISTICS_TYPE = "dynamicstatistics"
+
+// Hardcoded for now because
+// 1. MUIS is the only use case and there's always 6 prayer timeslots
+// 2. No other known use cases have been identified
+export const NUMBER_OF_STATISTICS = 6
```
harishv7, line 11: technically we can infer it from the received response of the json. but agree we don't have any use case for now and our UI is not generic enough as well. can be a future consideration.

packages/components/src/templates/next/components/complex/DynamicStatistics/DynamicStatistics.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export default meta
+type Story = StoryObj<typeof DynamicStatisticsUI>
+
+export const Default: Story = {
+  args: {
+    title: "1 Rejab 1446H",
+    statistics: [
+      { label: "Subuh", value: "5:43am" },
```
harishv7, line 23: nit: i don't think they add the "am" or "pm" at the back looking at muis website

apps/studio/src/components/PageEditor/constants.ts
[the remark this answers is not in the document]
adriangohjw: ok, will make the change!

packages/components/src/templates/next/components/complex/DynamicStatistics/DynamicStatistics.stories.tsx
adriangohjw: yeah tho i do agree it's clearer + otherwise the UI will be too too empty 🥲

packages/components/src/templates/next/layouts/Homepage/Homepage.stories.tsx
```diff
@@ -208,6 +227,39 @@ export const Default: Story = {
         secondaryButtonLabel: "Sub CTA",
         secondaryButtonUrl: "/",
       },
+      {
+        type: "dynamicstatistics",
+        apiEndpoint: "https://jsonplaceholder.com/muis_prayers_time",
+        title: "hijri_date",
```
harishv7, line 233: the date itself is also a value right?

adriangohjw: yup

packages/components/src/templates/next/components/complex/DynamicStatistics/DynamicStatistics.stories.tsx
sehyunidaaa: I added this in design because dates are generally more readable with am/pm, and 24-hour format (3rd alternative) is not super laymen friendly!

sehyunidaaa:
reviewed updated chromatic, lgtm! thanks for making the styling updates so quickly 🙏

packages/components/src/templates/next/layouts/Homepage/Homepage.stories.tsx
adriangohjw: addressed in https://github.com/opengovsg/isomer/pull/902/commits/55b01825d5ccef052f2b1bdce47f8f3dbf0aae88

apps/studio/src/components/PageEditor/constants.ts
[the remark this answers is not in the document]
adriangohjw: addressed in https://github.com/opengovsg/isomer/pull/902/commits/3ec4a13ea8ffcdbe3883a900a5e40279891bf68b

packages/components/src/templates/next/components/complex/DynamicStatistics/DynamicStatistics.stories.tsx
harishv7: okay sounds good

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -175,6 +176,40 @@ export const DEFAULT_BLOCKS: Record<
     title: "Rick Astley - Never Gonna Give You Up",
     url: "https://www.youtube.com/embed/dQw4w9WgXcQ?si=ggGGn4uvFWAIelWD",
   },
+  // TODO: Replace with actual working API endpoint
+  [DYNAMIC_DATA_BANNER_TYPE]: {
+    type: `${DYNAMIC_DATA_BANNER_TYPE}`,
+    apiEndpoint: "https://jsonplaceholder.com/muis_prayers_time",
+    title: "hijriDate",
+    statistics: [
```
harishv7, line 184: are we renaming this key to data?

adriangohjw: addressed in https://github.com/opengovsg/isomer/pull/902/commits/83f4204b5d70bbca484a443a22ff543f03c70eb3

harishv7 (approved):

harishv7 (commented):
Can we add a section on verification steps in PR description

adriangohjw:
> Can we add a section on verification steps in PR description

added in the PR description under "Verification"!

PR #904, opened 2024-11-28
feat: add complex blocks titles to table of contents

dcshzj:
## Problem

The titles of complex components do not show up in the table of contents for content pages.

## Solution

**Breaking Changes**

**Features**:

- Make complex component titles appear in the table of contents.
- Also added a new hidden form control to specifically hide fields from appearing in the form builder.

**Improvements**:

- Removed unused `getRandomNumberBetIntervals`. It was earlier replaced with the component index.

## Before & After Screenshots

image


sehyunidaaa:
Was a little conflicted with infobar (the CTA block) at first. No super confident answer here from me as both has clear pros and cons. I'm siding more with one of our principles which is scannability, and how we've built components that enable that (chunks/sections of content, headings, and ToC). Given the principle I think it makes sense to include infobar heading on the ToC.

Word vomit / thought process:

- One concern was explaining it to users on studio. But, thinking if we frame these as "sections" on the Studio in the future, then the whole ToC concept will be much clearer. Section headings text styles -> shows up on ToC, so a 'section' block also shows up, seems like an intuitive way to frame it
   - expecting users to differentiate it using the actual heading sizes feels way more unintuitive
- Another concern was the cta block is quite narrow and lacks content, do we want to have an anchor link to it
   - but you might have a subsidy page where you have 3 different schemes, and you want users to zoom straight to one of the application buttons
- This will affect live sites like togetherforbetter, will work with ops to give heads up

packages/components/src/utils/getTableOfContents.ts
```diff
@@ -8,21 +8,36 @@ export const getTableOfContents = (
...
+      return [
+        {
+          content: block.title,
+          anchorLink: `#${block.id}`,
+        },
+      ]
     }
 
-    const result = []
+    if (block.type === "prose" && block.content) {
```
seaerchin, line 26: should we still have a defensive check here? previously we allowed **only** `prose` with `content` - right now, we allow 4 blocks (assuming they have title), then we make the assumption that **the remaining blocks are prose with content**. 

what happens if we add a new block?

packages/components/src/utils/getTransformedPageContent.ts
```diff
@@ -31,11 +31,18 @@ export const getTransformedPageContent = (content: IsomerComponent[]) => {
           }
         }),
       }
-
-      transformedContent.push(transformedBlock)
+    } else if (
+      block.type === "infocards" ||
+      block.type === "infocols" ||
+      block.type === "infopic" ||
+      block.type === "keystatistics"
```
seaerchin, line 38: feels like this should already be a type or type guard

seaerchin (commented):
very minor issues, mostly with regards to forward compat and some assumptions we're making

packages/components/src/utils/getTableOfContents.ts
dcshzj: Hmm we are only selectively adding support for certain blocks, not all blocks will be eligible to be part of the table of contents. If there is a new block, it will fallback to the last return which is the empty array `[]`.

packages/components/src/utils/getTransformedPageContent.ts
dcshzj: Wait checking here, this one is for these specific blocks where we want to add the ID but the function will be able to take any block. We don't necessarily want to add the ID for blocks that does not support having anchors in them (e.g. infobar).

seaerchin: yea, but this checks for blocks that can have toc - so we can write a typeguard that accepts a string and narrows it to these 4 specifically after. 

only noting it here because this check is also repeated in `getTableOfContents`

seaerchin (approved):
code looks fine. some concerns re: repeated back - wonder if we can extract it out?

dcshzj:
I think we can do a refactor next time regarding this. We separated out the logic to get the table of contents earlier for some other reason but I think there is no need to have them separate. Will do so in a separate PR when got the time.

PR #907, opened 2024-11-29
chore(scripts): add db scripts for common tasks

seaerchin:
## Problem
Right now, we often have to whitelist/create site/update user permissions. This PR adds these 3 common functionality in via db scripts

## Solution
1. use kysely and write common scripts 
2. add simple readme for usage

### DIFF
--- a/apps/studio/prisma/scripts/addUsersToSite.ts
+++ b/apps/studio/prisma/scripts/addUsersToSite.ts
@@ -6,6 +6,7 @@ interface User {
   email: string
   name?: string
   phone?: string
+  role?: RoleType
 }
 interface AddUsersToSiteProps {
   siteId: number
@@ -16,14 +17,15 @@ export const addUsersToSite = async ({
   siteId,
   users,
 }: AddUsersToSiteProps) => {
-  const processedUsers = users.map(({ email, phone, name }) => ({
+  const processedUsers = users.map(({ email, phone, name, role }) => ({
     email: email.toLowerCase(),
     name: name ?? "",
     phone: phone ?? "",
+    role: role ?? RoleType.Editor,
   }))
 
   await Promise.all(
-    processedUsers.map(async (props) => {
+    processedUsers.map(async ({ role, ...props }) => {
       await db.transaction().execute(async (tx) => {
         const user = await tx
           .insertInto("User")
@@ -44,7 +46,7 @@ export const addUsersToSite = async ({
           .values({
             userId: user.id,
             siteId,
-            role: RoleType.Editor,
+            role,
           })
           .execute()
       })
--- a/apps/studio/prisma/scripts/addUsersToSite.ts
+++ b/apps/studio/prisma/scripts/addUsersToSite.ts
@@ -3,7 +3,7 @@ import cuid2 from "@paralleldrive/cuid2"
 import { db, RoleType } from "~/server/modules/database"
 
 // NOTE: add them as editor first as that is the one with the least permissions
-export const addUsersToSite = async (siteId: number, emails: string[]) => {
+const addUsersToSite = async (siteId: number, emails: string[]) => {
   const lowercasedEmails = emails.map((email) => email.toLowerCase())
 
   await Promise.all(
--- a/apps/studio/prisma/scripts/addUsersToSite.ts
+++ b/apps/studio/prisma/scripts/addUsersToSite.ts
@@ -2,8 +2,15 @@ import cuid2 from "@paralleldrive/cuid2"
 
 import { db, RoleType } from "~/server/modules/database"
 
+interface AddUsersToSiteProps {
+  siteId: number
+  emails: string[]
+}
 // NOTE: add them as editor first as that is the one with the least permissions
-const addUsersToSite = async (siteId: number, emails: string[]) => {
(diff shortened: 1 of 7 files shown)


apps/studio/prisma/scripts/addUsersToSite.ts
```diff
@@ -0,0 +1,45 @@
...
+            siteId,
+            role: RoleType.Editor,
+          })
+          .execute()
+      })
+    }),
+  )
+}
+
+const emails = ["your email here"]
```
dcshzj, line 41: nit: not sure if useful but mainly in prod would be good to capture their actual names and phone numbers. I see the migrators will provide us so would be good to have it inside our DB too

apps/studio/prisma/scripts/createSite.ts
```diff
@@ -0,0 +1,236 @@
...
+      .insertInto("Footer")
+      .values({
+        siteId,
+        content: jsonb({
+          contactUsLink: "/contact-us",
+          feedbackFormLink: "https://www.form.gov.sg",
+          privacyStatementLink: "/privacy",
+          termsOfUseLink: "/terms-of-use",
+          siteNavItems: FOOTER_ITEMS,
+        }),
```
dcshzj, line 176: nit: maybe can put this together with the `FOOTER_ITEMS`? Found that there were a couple of instances I had to override this

apps/studio/prisma/scripts/addUsersToSite.ts
```diff
@@ -0,0 +1,45 @@
...
+          )
+          .returning(["id", "name"])
+          .executeTakeFirstOrThrow()
+
+        await tx
+          .insertInto("ResourcePermission")
+          .values({
+            userId: user.id,
+            siteId,
+            role: RoleType.Editor,
```
dcshzj, line 33: nit: might want to customise as part of the `emails` variable too?

dcshzj (approved):
The script not really set in stone anyway so ok to approve with caveat for eng that the script will always need to customise before running

apps/studio/prisma/scripts/updateUserPermissions.ts
```diff
@@ -0,0 +1,25 @@
+import { db, RoleType } from "~/server/modules/database"
+
+export const updateUserPermissions = async (siteId: number, role: RoleType) => {
```
adriangohjw, line 3: IMO this function name is potentially misleading as it doesn't require us to pass in any `user.id` but rather what it's doing is update the user permissions of **ALL** users of the site

apps/studio/prisma/scripts/createSite.ts
```diff
@@ -0,0 +1,236 @@
...
+  })
+
+  await addUsersToSite(
+    siteId,
+    [...ISOMER_ADMINS, ...ISOMER_MIGRATORS].map(
+      (email) => `${email}@open.gov.sg`,
+    ),
+  )
+
+  await updateUserPermissions(siteId, RoleType.Admin)
```
adriangohjw, line 232: would it be a better idea to not have these functions be included here?

apps/studio/prisma/scripts/createSite.ts
```diff
@@ -0,0 +1,236 @@
...
+    title: "Something else",
+    url: "/something-else",
+  },
+  {
+    title: "Resources",
+    url: "/resources",
+  },
+]
+
+export const createSite = async (siteName: string) => {
```
adriangohjw, line 127: nit - might want to remove `export` since it's not being used anywhere?

adriangohjw (commented):
might also want to consider updating these to be keyword arguments instead of positional arguments? to reduce chance of passing in the wrong arguments

apps/studio/prisma/scripts/createSite.ts
seaerchin: actually i want to have them here just because it makes it easier for us to run the script! i put fake values so that it'll fail and provide feedback, which minimises the risk of running a "false" migration (+ it's wrapped in `tx` also)

lmk if you don't think this is sufficient!

apps/studio/prisma/scripts/createSite.ts
seaerchin: sounds good

apps/studio/prisma/scripts/updateUserPermissions.ts
seaerchin: sure, lemme update the name to `updateAllUserPermissionsForSite`

apps/studio/prisma/scripts/addUsersToSite.ts
seaerchin: done!

apps/studio/prisma/scripts/addUsersToSite.ts
seaerchin: honestly, can - i just did this first because the requests we were getting were all to add users as `Admin`, which would be easy to do with the followup script. we didn't really see instances w mixed perms but better to have and set as default arg i g

apps/studio/prisma/scripts/createSite.ts
seaerchin: done

PR #912, opened 2024-12-02
Global search improvement (text hint + display personalized recently viewed)

adriangohjw:
## Problem

Improve global search based on https://opengovproducts.slack.com/archives/C06R4DX966P/p1732866658371749

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- added `useResourceLocalViewHistory` to store user recently viewed history (based on site) on their localstorage
  - add to view history when opening and viewing a resource (only page, not folder or links)
- added a hint component that will show during
  - no search result
  - searching on the 3rd search query everytime the modal open (resets on close) 
- added stories 

### showing recently viewed

https://github.com/user-attachments/assets/370dd08c-4601-4cdd-8da8-300a06c92b01

### showing hint after 3rd query

note: adjusted to be after the 3rd one since 2nd one might be a bit annoying IMO during development

https://github.com/user-attachments/assets/fad2972e-4b4c-4611-9f64-dff54570d4bb

### DIFF
--- a/apps/studio/src/components/Searchbar/SearchModalBodyContentStates.tsx
+++ b/apps/studio/src/components/Searchbar/SearchModalBodyContentStates.tsx
@@ -61,10 +61,12 @@ const ModalBody = ({ children, ...props }: PropsWithChildren & ChakraProps) => {
 
 const HeaderTextAndContent = ({
   headerText,
+  showSearchResultHint = false,
   content,
   ...props
 }: {
   headerText?: string
+  showSearchResultHint?: boolean
   content: React.ReactNode
 } & ChakraProps) => {
   return (
@@ -74,6 +76,7 @@ const HeaderTextAndContent = ({
           {headerText}
         </Text>
       )}
+      {showSearchResultHint && <SearchResultHint />}
       {content}
     </VStack>
   )
@@ -154,6 +157,7 @@ export const SearchResultsState = ({
     <ModalBody>
       <HeaderTextAndContent
         headerText={`${totalResultsCount} search result${totalResultsCount === 1 ? "" : "s"} with "${searchTerm}" in title`}
+        showSearchResultHint
         content={
           <SearchResults
             siteId={siteId}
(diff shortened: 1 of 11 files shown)


apps/studio/src/components/Searchbar/SearchModalBodyContentStates.tsx
```diff
@@ -70,39 +91,67 @@ export const InitialState = ({
   siteId: string
   items: SearchResultResource[]
 }) => {
+  const { get } = useResourceLocalViewHistory({ siteId })
+  const { data: localViewHistorySearchResults = [] } =
+    trpc.resource.searchWithResourceIds.useQuery({
+      siteId,
+      resourceIds: get().map((history) => history.resourceId),
```
harishv7, line 98: if a resource was deleted, will it still return a result

adriangohjw: nope, because the underlying query is using a `IN`

https://github.com/opengovsg/isomer/blob/e64857553ada4ff98601d0fee13ef7445e0b065f/apps/studio/src/server/modules/resource/resource.service.ts#L607-L617

apps/studio/src/hooks/useResourceLocalViewHistory.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const entryToUpdate: LocalViewHistory | undefined =
+      localViewHistory[existingEntryIndex]
+
+    if (entryToUpdate) {
+      // Remove the existing entry
+      localViewHistory.splice(existingEntryIndex, 1)
+    }
+    // Add the new entry to the beginning of the array
+    localViewHistory.unshift({ resourceId, dateTime: new Date() })
```
harishv7, line 40: unshift is potentially expensive, but not really a concern for small number of items. a better way would be to use push and store in reverse order. Then can map in reverse index

adriangohjw: noted, thanks for sharing 😄 

the current approach `splice` it to only have 10 items (despite us only showing 3) so i wouldn't proceed to make this change. Furthermore, we prob wouldn't want to store uncapped number of items in localstorage too

harishv7: yea don't need to change, more fyi esp is only 10 items

harishv7 (approved):

PR #913, opened 2024-12-03
Fix - incorrect `getFullPermalink` for RootPage with _index permalink

adriangohjw:
## Problem

Closes https://opengovproducts.slack.com/archives/C06R4DX966P/p1733189421170829

if a `RootPage` type resource has the permalink `_index`, it will cause the API endpoint to throw an error

image

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- since `RootPage` is a special type, wrote a check to ensure it's not incorrectly filtered out if it's permalink is `_index`
- add unit tests

## Tests

running `npm run test:unit` should pass

### DIFF
--- a/apps/studio/src/server/modules/resource/resource.service.ts
+++ b/apps/studio/src/server/modules/resource/resource.service.ts
@@ -353,7 +353,7 @@ export const getResourcePermalinkTree = async (
       return []
     }
 
-    const resourcePermalinks = await tx
+    const resources = await tx
       .withRecursive("Ancestors", (eb) =>
         eb
           // Base case: Get the actual resource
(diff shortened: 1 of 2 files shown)


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -371,13 +371,21 @@ export const getResourcePermalinkTree = async (
...
+      resources[0]?.permalink === INDEX_PAGE_PERMALINK
+    ) {
+      return [""] // empty string to represent the root page's permalink
+    }
+
+    return resources
       .reverse()
-      .filter((v) => v !== INDEX_PAGE_PERMALINK)
+      .filter((r) => r.type !== ResourceType.IndexPage)
+      .map((r) => r.permalink)
```
seaerchin, line 388: actually just double checking here - our publishing script uses `permalink === "/"` to identify a root page. it also explicitly adds this terminating `_index` in `writeContentToFile`, so i think some assumptions might be violated here. 

the `_index` is required for run-time, where we look it up if it is a folder, but not required for `sitemap` (and must explicitly be removed ). as we just take the `permalink` from db, this code change seems to allow `_index` and violate this at run-time. 

i think we should instead do a record level change tbh - i have no idea why dc has the `permalink` set as `_index` though, pinging @dcshzj

dcshzj: Yeah I think might have been some data error somewhere but not too sure. For now root pages should be empty `""` for the permalink.

The risk seems to be quite high for us to switch to using `_index` for RootPage to standardise with the one used by IndexPage compared to the rewards that come with the standardisation, so I think we should just make the data correction and leave it as status quo? There are quite some places that depend on the empty string as the permalink for RootPage and we need to check through the codebase to find out such instances.

adriangohjw: okay agree, will update the record and close off this PR, thanks!

PR #914, opened 2024-12-03
Fix multi accordion top margin

adriangohjw:
## Problem

Closes https://opengovproducts.slack.com/archives/C07CWUNUL68/p1733212670307009

> I think it's cos we applied a margin-top only when an accordion is :first-of-type. So when you have groups of accordions, only the first groups gets a top margin.

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- update to use `mt-7` by default and only `mt-0` if preceeded by another `<details>` (accordion)

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- added a story in Content layout

### DIFF
--- a/apps/studio/public/assets/css/preview-tw.css
+++ b/apps/studio/public/assets/css/preview-tw.css
@@ -5328,6 +5328,10 @@ video {
   }
 }
 
+.\[\&\+details\]\:mt-0 + details {
+  margin-top: 0px;
+}
+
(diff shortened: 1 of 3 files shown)


adriangohjw:
fixed according to chromatic comment
image

seaerchin (approved):

PR #911, opened 2024-12-02
feat(collection tags): frontend changes for tags in collection

seaerchin:
## Problem
We need to support tagging in our collections for the CLS use case. this PR adds the initial collection tags in 

## Solution
1. Add an internal `Tag` component for use
2. Add this `Tag` to the `CollectionCard`

PR #916, opened 2024-12-03
feat(multi-tags): add filtering to collections

seaerchin:
## Problem
This PR adds filtering to collections through the use of tags. 

## Solution
1. add `Tags` to the schema on `CollectionPagePageSchema`
    - note that in our publishing script, we explictly check for whether it's a `PAGE_RESOURCE_TYPE`, which means that even though the schema can only expose it on page items. we might want to expose it on files, which would necessitate a slight change inhow we generate hte `SitemapEntries`. not too sure if this is what we want, so i omitted it for now. need to think abit more here
2. add tags to `publishing/index.ts` so that it gets written to the sitemap entry (and hence, to disk)
3. in `Collection`, extract the `tags` from the `Sitemap`. add filtering logic onto our collection so that it will display the correct cards on filter click. take note that the logic here can be (and should be) reused for collection but not for years due to how the comparison is made
    - i didn't do it for collections because the id and the check is a lower-cased check via a hard-coded id + the property is different (takes from `item.category` and not `item.tags`  

## Videos

https://github.com/user-attachments/assets/db0c6179-e18d-497a-ac2f-38f683a4286c

### DIFF
--- a/packages/components/src/interfaces/internal/CollectionCard.ts
+++ b/packages/components/src/interfaces/internal/CollectionCard.ts
@@ -4,7 +4,6 @@ import type { IsomerSiteProps, LinkComponentType } from "~/types"
 export interface Tag {
   selected: string[]
   category: string
-  values: string[]
 }
 
 export interface FileDetails {
--- a/packages/components/src/interfaces/internal/CollectionCard.ts
+++ b/packages/components/src/interfaces/internal/CollectionCard.ts
@@ -2,7 +2,7 @@ import type { ImageProps } from "~/interfaces"
 import type { IsomerSiteProps, LinkComponentType } from "~/types"
 
 export interface Tag {
-  labels: string[]
+  selected: string[]
   category: string
   values: string[]
 }
--- a/packages/components/src/interfaces/internal/CollectionCard.ts
+++ b/packages/components/src/interfaces/internal/CollectionCard.ts
@@ -2,7 +2,7 @@ import type { ImageProps } from "~/interfaces"
 import type { IsomerSiteProps, LinkComponentType } from "~/types"
 
 export interface Tag {
-  label: string
+  labels: string[]
   category: string
   values: string[]
 }
--- a/packages/components/src/interfaces/internal/CollectionCard.ts
+++ b/packages/components/src/interfaces/internal/CollectionCard.ts
@@ -3,6 +3,8 @@ import type { IsomerSiteProps, LinkComponentType } from "~/types"
 
 export interface Tag {
   label: string
+  category: string
+  values: string[]
 }
 
 export interface FileDetails {
--- a/packages/components/src/templates/next/layouts/Collection/utils.ts
+++ b/packages/components/src/templates/next/layouts/Collection/utils.ts
@@ -19,7 +19,7 @@ const getCategories = (
       ([label, count]) => ({
         label,
         count,
-        id: label.toLowerCase(),
+        id: label,
       }),
     )
 
@@ -27,7 +27,7 @@ const getCategories = (
       ...acc,
       {
         items,
-        id: category.toLowerCase(),
+        id: category,
         label: category,
       },
     ]
@@ -72,17 +72,15 @@ export const getAvailableFilters = (
     // Step 3: Get all category tags
     if (tags) {
       tags.forEach(({ label, category }) => {
-        const lowercasedLabel = label.toLowerCase()
-
         if (!tagCategories[category]) {
           tagCategories[category] = {}
         }
 
-        if (!tagCategories[category][lowercasedLabel]) {
-          tagCategories[category][lowercasedLabel] = 0
+        if (!tagCategories[category][label]) {
+          tagCategories[category][label] = 0
         }
 
-        tagCategories[category][lowercasedLabel] += 1
+        tagCategories[category][label] += 1
       })
     }
   })
@@ -137,6 +135,7 @@ export const getFilteredItems = (
   appliedFilters: AppliedFilter[],
   searchValue: string,
 ): ProcessedCollectionCardProps[] => {
+  console.log(appliedFilters)
   return items.filter((item) => {
     // Step 1: Filter based on search value
     if (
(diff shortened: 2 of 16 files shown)


packages/components/src/interfaces/internal/CollectionCard.ts
```diff
@@ -3,6 +3,8 @@ import type { IsomerSiteProps, LinkComponentType } from "~/types"
 
 export interface Tag {
   label: string
```
harishv7, line 5: what does label represent here? is it the current value that the tag holds?

harishv7: Would it make more sense to call values and currentValues than "label"?

tooling/build/scripts/publisher.sh
```diff
@@ -70,7 +70,6 @@ start_time=$(date +%s)
 cd tooling/build/scripts/publishing
 echo $(pwd)
 npm ci
-npm install ts-node -g
```
harishv7, line 73: we don't need ts-node at build time?

tooling/build/scripts/publishing/tsconfig.json
```diff
@@ -13,6 +13,9 @@
 
     /* Completeness */
     // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
-    "skipLibCheck": true /* Skip type checking all .d.ts files. */
+    "skipLibCheck": true /* Skip type checking all .d.ts files. */,
+    "paths": {
```
harishv7, line 17: what's this change for?

packages/components/src/interfaces/internal/CollectionCard.ts
```diff
@@ -3,6 +3,8 @@ import type { IsomerSiteProps, LinkComponentType } from "~/types"
...
+  category: string
+  values: string[]
```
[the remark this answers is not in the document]
seaerchin: yea, this is taken at build time. you can see this at `publishing/index.ts` where we write the `tag` to the sitemap

tooling/build/scripts/publisher.sh
seaerchin: we shouldn't specify dependencies like this - we have a `package.json`. doing this makes it confusing when trying to test locally (as i found out). 

what happened is that:
1. i specified `tsx` as a dependency (faster, respects `tsconfig.paths` so we can share types from db -> build, which is important to make sure our build step works properly)
2. updated the `npm run start` command to use `tsx` - which is now a dependency

packages/components/src/interfaces/internal/CollectionCard.ts
seaerchin: > yea, i thought this was a distinct feature of multi-tags to make it useful for things like TOSP. Currently, we are duplicating entries based on the body parts.  

ah ok, can! i'll do it in this pr - thought having singly tagged values w support for multiple topics was enough for the cls use case.

seaerchin: updated the filtering behaviour

seaerchin: i'll update `label` -> `selected`! i think the `curValues` might be abit misleading, especialy in conjunction with our `setState`

packages/components/src/interfaces/internal/CollectionCard.ts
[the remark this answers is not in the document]
harishv7: Hmm yea, similar thoughts. I didn't think we will need the `values` array here. In the long term, this will be moved to the collection level when it will be required.

However, for current implementation, values will not be known till build time

tooling/build/scripts/publisher.sh
harishv7: oops missed the addition of tsx in package. Okay lgtm

packages/components/src/interfaces/internal/CollectionCard.ts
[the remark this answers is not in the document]
adriangohjw: correct me if im wrong, but will not knowing the values till build time be an issue? 

seems like we only require all values for the filters and options, which is fine to only be known during build time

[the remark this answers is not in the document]
seaerchin: yea, you're right - i thought we were gonna ask the users for the set of values then display them in the filters (eg: a: 0) but actually we'll hide those wtihout any matches so i'll remove this

packages/components/src/templates/next/layouts/Collection/utils.ts
```diff
@@ -33,6 +68,22 @@ export const getAvailableFilters = (
...
+
+    // Step 3: Get all category tags
+    if (tags) {
+      tags.forEach(({ selected: selectedLabels, category }) => {
+        selectedLabels.forEach((label) => {
+          if (!tagCategories[category]) {
+            tagCategories[category] = {}
+          }
+          if (!tagCategories[category][label]) {
+            tagCategories[category][label] = 0
```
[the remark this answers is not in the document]
seaerchin: no! the sequence of events here is:
1. i see a category label 
2. if don't have the count of this stored, create an empty object
3. add 1 to the label. 

i did this so i don't have to do the `if/else` down the line and it's easier to comprehend

packages/components/src/interfaces/internal/CollectionCard.ts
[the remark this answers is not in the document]
seaerchin: @adriangohjw - not sure i understand your question! i think we don't require the values at all - i think, for now, we can just fetch all values from db (collections are 1 level deep), which shuold still be manageable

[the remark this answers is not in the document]
seaerchin: updated!

packages/components/src/templates/next/layouts/Collection/utils.ts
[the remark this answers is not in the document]
adriangohjw: ah okay 👍

adriangohjw (approved):

seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/916?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#916** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#911** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/911?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

seaerchin:
### Merge activity

* **Dec 19, 10:18 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/916).

PR #919, opened 2024-12-06
feat: add visited colour for links

dcshzj:
## Problem

We lack a visited colour for links, which is an IM8 requirement.

## Solution

**Breaking Changes**

**Improvements**:

- Introduce a new visited pseudo-class for links.
- Also fixed a typo for link hovers.

## Tests

- [ ] Create a new text block with a link inside it
- [ ] Verify that the preview shows the link with a light blue colour
- [ ] Hover over the link and verify that the link changes colour
- [ ] Click on the link and verify that it becomes a violet colour


dcshzj:
> can i check - is the footer link also going to be this colour?

For now nope, by right need but cos the text is white so not immediately clear if it is possible to show a visited colour.

adriangohjw (commented):
should we use tailwind `@layer` and `@apply` to apply the hover and visited state to every link text component instead? current implementation seems like it might be forgotten for links in new components in the future

adriangohjw:
> should we use tailwind `@layer` and `@apply` to apply the hover and visited state to every link text component instead? current implementation seems like it might be forgotten for links in new components in the future

sorry just saw https://github.com/opengovsg/isomer/pull/919#issuecomment-2522392127 - i think it's alright to proceed then

PR #922, opened 2024-12-09
isom-1678 dynamic blocks for muis fix

adriangohjw:
## Problem

https://github.com/opengovsg/isomer/pull/902
- does not have CSP in studio thus demo fails
- `fetch` structured in a way that cause nextjs static export to fail

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- update studio's CSP to allow connect-src based on S3 asset URL instead of hardcoding
- replace `react-query` - `fetch` isn't available during build time since it's a browser API, and suggested solution is to wrap it in `useEffect`. however, `useQuery` itself is a hook and should not be placed inside `useEffect` due to potential hydration error. To keep things simple, i have replace `useQuery` altogether with the basic `useState` and `useEffect`

### Studio

note: first part on CSP only required for local development

https://www.loom.com/share/acc94c03b3bf4486882c23c0c93c154b?sid=7ef0be6f-d0b8-4d7b-8623-4db9e210393e

Sample component block
```json
    {
      "type": "dynamicdatabanner",
      "apiEndpoint": "https://isomer-user-content-stg.by.gov.sg/staging_muis_prayer_timetable_sample_2024.json",
      "title": "hijri_date",
      "data": [
        {
          "label": "Subuh",
          "key": "subuh"
        },
        {
          "label": "Syuruk",
          "key": "syuruk"
        },
        {
          "label": "Zohor",
          "key": "zohor"
        },
        {
          "label": "Asar",
          "key": "asar"
        },
        {
          "label": "Maghrib",
          "key": "maghrib"
        },
        {
          "label": "Ishak",
          "key": "isyak"
        }
      ],
      "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ",
      "label": "View all dates",
      "errorMessage": [
        {
          "text": "Not seeing the prayer times? ",
          "type": "text"
        },
        {
          "text": "Report an issue",
          "type": "text",
          "marks": [
            { "type": "bold" },
            {
              "type": "link",
              "attrs": {
                "href": "https://www.form.gov.sg/some-link"
              }
            }
          ]
        }
      ]
    },
```

### Site build

1. After the above step is done, click "Publish" and wait for codebuild to be completed.
2. Go to the site URL and have a look - the component should be present (you can request for a manual invalidation in cloudfront to reduce waiting time)

### DIFF
--- a/apps/studio/next.config.mjs
+++ b/apps/studio/next.config.mjs
@@ -74,7 +74,11 @@ const ContentSecurityPolicy = `
     https://*.wogaa.sg
     https://placehold.co
     https://cdn.growthbook.io
-    ${env.NODE_ENV === "production" ? "https://isomer-user-content.by.gov.sg" : "https://*.by.gov.sg"}
+    ${
+      !!env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME
+        ? `https://${env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME}`
+        : "https://*.by.gov.sg"
+    }
     https://via.intercom.io
     https://api.intercom.io
     https://api.au.intercom.io
--- a/packages/components/src/templates/next/components/complex/DynamicDataBanner/DynamicDataBannerClient.tsx
+++ b/packages/components/src/templates/next/components/complex/DynamicDataBanner/DynamicDataBannerClient.tsx
@@ -11,7 +11,8 @@ import { getSingaporeDateLong, getSingaporeDateYYYYMMDD } from "./utils"
 
 const createDynamicDataBannerStyles = tv({
   slots: {
-    outerContainer: `${ComponentContent} grid grid-cols-1 gap-5 bg-brand-canvas px-6 pb-4 pt-6 md:gap-4 md:px-10 md:py-2 lg:grid-cols-12 lg:justify-between lg:justify-items-stretch lg:gap-0`,
+    screenWideOuterContainer: "bg-brand-canvas",
+    outerContainer: `${ComponentContent} grid grid-cols-1 gap-5 px-6 pb-4 pt-6 md:gap-4 md:px-10 md:py-2 lg:grid-cols-12 lg:justify-between lg:justify-items-stretch lg:gap-0`,
     basicInfoContainer:
       "flex flex-row items-center justify-between md:gap-1 md:py-3 lg:col-span-3 lg:flex-col lg:items-start lg:justify-start",
     basicInfoInnerContainer: "flex flex-col items-start justify-start gap-1",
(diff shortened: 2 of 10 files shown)


apps/studio/next.config.mjs
```diff
@@ -74,7 +74,11 @@ const ContentSecurityPolicy = `
     https://*.wogaa.sg
     https://placehold.co
     https://cdn.growthbook.io
-    ${env.NODE_ENV === "production" ? "https://isomer-user-content.by.gov.sg" : "https://*.by.gov.sg"}
+    ${
```
harishv7, line 77: Let's add verification steps section in PR's description

packages/components/src/templates/next/components/complex/DynamicDataBanner/DynamicDataBannerClient.tsx
```diff
@@ -69,32 +65,37 @@ export const DynamicDataBannerUI = ({
...
-    queryFn: async () => {
-      const response = await fetch(apiEndpoint)
-      return (await response.json()) as Record<string, object | string>
-    },
-  })
+}: Omit<DynamicDataBannerProps, "type" | "site">) => {
+  const [isLoading, setLoading] = useState(true)
+  const [apiData, setApiData] = useState<Record<string, object>>({})
+
+  // This is to ensure that the component is mounted before the query is executed
```
harishv7, line 79: did we manage to test the build?

adriangohjw: yup! https://test-isomer-next-staging.isomer.gov.sg/

image

apps/studio/next.config.mjs
adriangohjw: oh i was thinking this is a follow-up/hotfix of https://github.com/opengovsg/isomer/pull/902 so don't need, but will do!

packages/components/src/templates/next/components/complex/DynamicDataBanner/DynamicDataBannerClient.tsx
harishv7: Curious, what is the fallback state if the API returns with an error? 

Can we choose to not render this component in this case? Is that the current behaviour

adriangohjw: fallback state is unchanged, which means it will be the same as loading - banner is still showing but without the values) - "view all dates" button will still be available

we can don't render this whole banner if that's the case, but wondering if that's better since user unable to click "view all dates"?

harishv7: I think the current behaviour is fine for now
Cc @jandonyada

jandonyada: If there's an error, can we show some text with a link to a formsg or something? Like "Not seeing the prayer times? [Report an issue](www.form.gov.sg/some-link)."

adriangohjw: > If there's an error, can we show some text with a link to a formsg or something? Like "Not seeing the prayer times? [Report an issue](www.form.gov.sg/some-link)."

Do we already have an existing link I can use?

adriangohjw: @jandonyada @harishv7 done!

https://www.chromatic.com/build?appId=665439911775ce21bf8b8fb4&number=1365

harishv7 (approved):

PR #928, opened 2024-12-11
feat: logo cloud

seaerchin:
## Problem
we want to add a logo cloud for agencies - this will mostly be a collection of images 

## Solution
1. add logo cloud with support for images

### DIFF
--- a/packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.stories.tsx
+++ b/packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.stories.tsx
@@ -1,0 +1,51 @@
+import type { Meta, StoryObj } from "@storybook/react"
+
+import { LogoCloud } from "./LogoCloud"
+
+const meta: Meta<typeof LogoCloud> = {
+  title: "Next/Components/Logo Cloud",
+  component: LogoCloud,
+  argTypes: {},
+  parameters: {
+    themes: {
+      themeOverride: "Isomer Next",
+    },
+  },
+}
+export default meta
+type Story = StoryObj<typeof LogoCloud>
+
+const IMAGE = { src: "https://via.placeholder.com/150", alt: "placeholder" }
+
+// Default scenario
+export const Default: Story = {
+  args: {
+    images: [IMAGE],
+  },
+}
+
+export const ManyImages: Story = {
+  args: {
+    images: Array(10).fill(IMAGE),
+  },
+}
+// https://www.ncss.gov.sg/images/default-source/asset/fwdsg_logo.png?sfvrsn=91b23ea8_1
+
+export const Agency: Story = {
+  args: {
+    images: [
+      {
+        alt: "Some image",
+        src: "https://www.ncss.gov.sg/images/default-source/asset/fwdsg_logo.png?sfvrsn=91b23ea8_1",
+      },
+      {
+        alt: "Some image",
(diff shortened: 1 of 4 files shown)


packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.stories.tsx
adriangohjw, line 1: perhaps we can add a story for many images of different width and height

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.tsx
```diff
@@ -0,0 +1,45 @@
+import { IsomerSiteProps } from "~/types"
```
adriangohjw, line 1: nit
```suggestion
import type { IsomerSiteProps } from "~/types"
```

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { isExternalUrl } from "~/utils"
+import { ImageClient } from "../Image"
+
+interface Image {
+  src?: string
```
adriangohjw, line 6: why is it optional?

adriangohjw (commented):
curious - why aren't we defining the schema `packages/components/src/interfaces/complex/` ?

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.tsx
seaerchin: updated downstream to be required!

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.tsx
seaerchin: good call

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.stories.tsx
seaerchin: sure! will do 2 cases - one where most images are similar width but there's 1 with huge width and the opposite case, where the heights are similar

seaerchin: added, but i think the changes for using `maxh` is done in the follow up so minor caveat

adriangohjw (approved):

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.tsx
```diff
@@ -0,0 +1,45 @@
...
+        ? src
+        : `${assetsBaseUrl}${src}`
+
+    return { src: transformedSrc, alt }
+  })
+  return (
+    <div className="px-10 py-12">
+      <div className="flex flex-col gap-4">
+        <p className="prose-headline-base-medium self-center text-base-content-light">
+          {title ?? "With support from these agencies"}
```
adriangohjw, line 30: btw `title` is required and won't be falsy? just checking if this is just a preventive measures?

seaerchin: yes defensive! reason why is because this is written to disk and read back (+ done manually)

PR #930, opened 2024-12-11
feat(logo cloud): add schema for rendering 

seaerchin:
## Problem
logo cloud requires an update in `renderComponent` so that it can be rendered successfully from the `content` of the page

Closes [insert issue #]

## Solution
1. add the associated schema over to schemas

### DIFF
--- a/apps/studio/src/components/PageEditor/constants.ts
+++ b/apps/studio/src/components/PageEditor/constants.ts
@@ -325,7 +325,7 @@ export const BLOCK_TO_META: Record<
     // TODO: Add image source
     label: "Logo cloud",
     description: "Display logos of other agencies here",
-    usageText: "Add logos here to easily show logos from other agencies",
+    usageText: "Show an overview of related agencies",
   },
 }
 
--- a/apps/studio/src/components/PageEditor/constants.ts
+++ b/apps/studio/src/components/PageEditor/constants.ts
@@ -216,6 +216,11 @@ export const DEFAULT_BLOCKS: Record<
       },
     ],
   },
+  logocloud: {
+    type: "logocloud",
+    images: [],
+    title: "With support from these agencies",
+  },
 }
 
 export const BLOCK_TO_META: Record<
@@ -315,6 +320,12 @@ export const BLOCK_TO_META: Record<
     label: "Dynamic Data Banner",
     description: "Display dynamic data banner",
     usageText: "This block supports fetching data from an API endpoint.",
+  },
+  logocloud: {
+    // TODO: Add image source
+    label: "Logo cloud",
+    description: "Display logos of other agencies here",
+    usageText: "Add logos here to easily show logos from other agencies",
   },
 }
 
--- a/packages/components/src/interfaces/complex/LogoCloud.ts
+++ b/packages/components/src/interfaces/complex/LogoCloud.ts
@@ -1,9 +1,10 @@
 import type { Static } from "@sinclair/typebox"
 import { Type } from "@sinclair/typebox"
 
+export const LOGO_CLOUD_TYPE = "logocloud"
 export const LogoCloudSchema = Type.Object(
   {
-    type: Type.Literal("logocloud", { default: "logocloud" }),
+    type: Type.Literal(LOGO_CLOUD_TYPE, { default: LOGO_CLOUD_TYPE }),
     images: Type.Array(
       Type.Object({
         src: Type.String({
--- a/packages/components/src/interfaces/complex/LogoCloud.ts
+++ b/packages/components/src/interfaces/complex/LogoCloud.ts
@@ -1,0 +1,39 @@
+import type { Static } from "@sinclair/typebox"
+import { Type } from "@sinclair/typebox"
+
+export const LogoCloudSchema = Type.Object(
+  {
+    type: Type.Literal("logocloud", { default: "logocloud" }),
+    images: Type.Array(
+      Type.Object({
+        src: Type.String({
+          title: "Upload image",
+          format: "image",
+        }),
+        alt: Type.String({
+          title: "Alternate text",
+          maxLength: 120,
+          description:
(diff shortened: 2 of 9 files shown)


seaerchin:
marking as draft first for studio side

packages/components/src/interfaces/complex/LogoCloud.ts
```diff
@@ -0,0 +1,39 @@
...
+          title: "Upload image",
+          format: "image",
+        }),
+        alt: Type.String({
+          title: "Alternate text",
+          maxLength: 120,
+          description:
+            "Add a descriptive alternative text for this image. This helps visually impaired users to understand your image.",
+        }),
+      }),
```
adriangohjw, line 19: note to self @adriangohjw - to update this to use the common imagealt and imagesrc schema after https://github.com/opengovsg/isomer/pull/920 is merged

packages/components/src/interfaces/complex/LogoCloud.ts
```diff
@@ -0,0 +1,39 @@
...
+      {
+        title: "Images for logo cloud",
+        minItems: 1,
+        maxItems: 10,
+      },
+    ),
+    title: Type.String({
+      title: "Title for the logo cloud",
+      description: "Upload the images for the logo cloud here or provide a url",
```
adriangohjw, line 28: is this correct? seems like this `title` is more for what the content will display?

packages/components/src/schemas/components.ts
```diff
@@ -26,6 +26,7 @@ import {
   UnorderedListSchema,
   VideoSchema,
 } from "~/interfaces"
+import { LogoCloudSchema } from "~/interfaces/complex/LogoCloud"
```
adriangohjw, line 29: nit - can be `"~/interfaces` and merge with the above instead of `~/interfaces/complex/LogoCloud`

packages/components/src/templates/next/render/renderComponentPreviewText.ts
```diff
@@ -95,6 +95,8 @@ export function renderComponentPreviewText({
       return component.title
     case "map":
       return component.title || "Map embed"
+    case "logocloud":
+      return component.title || "Logo cloud"
```
adriangohjw, line 99: nit - since title is required, shouldn't need to fallback to "Logo cloud"

packages/components/src/interfaces/complex/LogoCloud.ts
```diff
@@ -0,0 +1,39 @@
+import type { Static } from "@sinclair/typebox"
+import { Type } from "@sinclair/typebox"
+
+export const LogoCloudSchema = Type.Object(
+  {
+    type: Type.Literal("logocloud", { default: "logocloud" }),
```
adriangohjw, line 6: perhaps we can make `logocloud` a const to be passed around instead (esp. to studio) instead of using string literal. this will ensure single source of truth and help with debugging/development down the road 

similar to `DYNAMIC_DATA_BANNER_TYPE`

packages/components/src/templates/next/render/renderComponentPreviewText.ts
seaerchin: kek i was copying HAHA thakns for catch, fixed!

packages/components/src/schemas/components.ts
seaerchin: thanks for the catch!

packages/components/src/interfaces/complex/LogoCloud.ts
seaerchin: sure

packages/components/src/interfaces/complex/LogoCloud.ts
seaerchin: this matches the header in our edit page! to give you an example, this is the `label` for statistics! i didn't put too much thought into this cos we're not allowing studio editing for logo cloud 
image
image

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -316,6 +321,12 @@ export const BLOCK_TO_META: Record<
     description: "Display dynamic data banner",
     usageText: "This block supports fetching data from an API endpoint.",
   },
+  logocloud: {
+    // TODO: Add image source
```
harishv7, line 325: when do we plan to add the image?

packages/components/src/interfaces/complex/LogoCloud.ts
adriangohjw: in this case, perhaps we can remove it so we won't have an issue of not remembering to fix this in the future when we support on studio?

adriangohjw (approved):

packages/components/src/templates/next/components/complex/LogoCloud/LogoCloud.tsx
```diff
@@ -34,7 +26,7 @@ export const LogoCloud = ({
             <ImageClient
               {...props}
               width="100%"
-              className="inset-0 w-fit max-w-52 object-contain p-2"
+              className="inset-0 max-h-24 w-fit object-contain p-2"
```
harishv7, line 29: do we need to verify chromatic for this pr?

seaerchin: image

approved by our stand in designer

apps/studio/src/components/PageEditor/constants.ts
seaerchin: when we need to add studio support! this is the image that shows on hover when adding the component on studio

packages/components/src/interfaces/complex/LogoCloud.ts
seaerchin: i added this here so it's less work next time when we want to implement it in studio ;--;

PR #931, opened 2024-12-11
chore: object spreads for meta

seaerchin:
## Problem
our metadata stuff broke with next15. 

## Solution
using an object spread for metadata seemed to work for the homepage. i have no clue - discovered this by accident


adriangohjw:
fixed in https://github.com/opengovsg/isomer/pull/933, thank you! 😄

PR #936, opened 2024-12-14
isom-1689 polish for component editing panels (tiptap editor)

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1689/polish-for-component-editing-panels (the tiptap editor ones)

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- allow prose schema to be optional or required
- add error messages if it doesn't fulfil the requirement
- make required prose for 1) accordion, 2) callout, 3) contentpic

**Bug Fixes**:

- props like `required` aren't passed correctly to `JsonFormsProseControl.tsx` as it has been intercepted by `JsonFormsObjectControl`. Updated the tester to match correctly with higher ranking too

**Refactoring**:
- create a single tester to remove the need to have multiple duplicated testers for different prose e.g. callout, accordion, contentpic
- simplified passing of props for the editors e.g. `useCalloutEditor`, `useAccordionEditor`

## Tests

https://www.loom.com/share/8969a576620d4ca09888cd8f70e4b22e?sid=bf74b0b9-c5d6-4343-8690-5c57b1027420

please check against 1) callout, 2) contentpic, 3) accordion, 4) other prose blocks


apps/studio/src/constants/formBuilder.ts
```diff
@@ -11,7 +11,7 @@ export const JSON_FORMS_RANKING = {
   ObjectControl: 2,
   AllOfControl: 3,
   AnyOfControl: 3,
-  ProseControl: 2,
+  ProseControl: 3,
```
seaerchin, line 14: question - can i check what this fixes?

apps/studio/src/features/editing-experience/components/TipTapProseComponent.tsx
```diff
@@ -57,7 +57,7 @@ function TipTapProseComponent({ content }: TipTapComponentProps) {
     },
   })
 
-  const updatePageState = (editorContent: JSONContent) => {
+  const updatePageState = (editorContent: JSONContent | undefined) => {
```
seaerchin, line 60: question - this is required so that it becomes `undefined` and then we prevent the users from saving rgiht?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
seaerchin, line 1: style (no action required) - tbh i preferred the old form cos it's more obvious but i guess this is ok too

seaerchin (approved):
lgtm w 1 question on why we change render priority

apps/studio/src/constants/formBuilder.ts
adriangohjw: it might be matched by `jsonFormsObjectControlTester` otherwise since both are `2`, so setting `3` will ensure it's checked first before that

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
adriangohjw: oh which part is more obvious? the UI is actually the same

apps/studio/src/features/editing-experience/components/TipTapProseComponent.tsx
adriangohjw: yup

PR #938, opened 2024-12-17
fix - remove "default" values in infocards

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1670/infocards-title-and-description-default-to-placeholder-when-empty

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- removed the `default` fields

note: there's a underlying issue with how we are rendering fields such that the "required" error message is not showing up. Should fix in a separate PR

## Tests

https://www.loom.com/share/847ca1fac61b427aa0e3c68907605f59?sid=5908c1bc-892e-4893-9717-7961b79f3851

### DIFF
--- a/packages/components/src/interfaces/complex/InfoCards.ts
+++ b/packages/components/src/interfaces/complex/InfoCards.ts
@@ -11,13 +11,11 @@ import { LINK_HREF_PATTERN } from "~/utils/validation"
 const SingleCardNoImageSchema = Type.Object({
   title: Type.String({
     title: "Title",
-    default: "This is the title of the card",
     maxLength: 100,
   }),
   description: Type.Optional(
     Type.String({
       title: "Description",
-      default: "This is an optional description for the card",
       maxLength: 150,
     }),
   ),


packages/components/src/interfaces/complex/InfoCards.ts
```diff
@@ -11,13 +11,11 @@ import { LINK_HREF_PATTERN } from "~/utils/validation"
 const SingleCardNoImageSchema = Type.Object({
   title: Type.String({
     title: "Title",
-    default: "This is the title of the card",
```
harishv7, line 14: The default field was meant to be there as a placeholder. Do we know why this occurs in first place? Because I remember other components do not have issues with default values

adriangohjw: the other components don't use `default` and the placeholders are derived from `apps/studio/src/components/PageEditor/constants.ts`

Not sure why this occur in the first place

harishv7 (approved):
@adriangohjw  Should also fix the required error msg which is currently missing - can file a ticket for this

adriangohjw:
https://linear.app/ogp/issue/ISOM-1705/bug-infocards-children-not-showing-error-message

PR #940, opened 2024-12-17
chore: fix env

seaerchin:
## Problem
the default starter kit logger uses `NODE_ENV` for env and automatically injects it, which causes wrong parsing on dd logs as it should always be set to `production` - see [here](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production)

Closes [insert issue #]

## Solution
use our `NEXT_PUBLIC_APP_ENV` instead

### DIFF
--- a/apps/studio/src/lib/logger.ts
+++ b/apps/studio/src/lib/logger.ts
@@ -54,7 +54,7 @@ export class PinoLogger {
         formatters: {
           bindings: () => {
             return {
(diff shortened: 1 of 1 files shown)


harishv7 (approved):
approving, test on staging before merging

adriangohjw:
oops sorry didn't see the comment 😅

PR #941, opened 2024-12-19
fix: collections active state

seaerchin:
## Problem
we were previously using the wrong property, leading to collections not having the correct state in side bar when we are already in the collection

## Solution
update to correct property 

## Checks
1. Click into a collection
2. the sidebar should be highlighted for the collection

### DIFF
--- a/apps/studio/src/features/dashboard/components/DirectorySidebar/useIsActive.ts
+++ b/apps/studio/src/features/dashboard/components/DirectorySidebar/useIsActive.ts
@@ -28,8 +28,9 @@ export const useIsActive = (
     case ResourceType.IndexPage:
       return siteProps.resourceId === currentResourceId
     case ResourceType.Folder:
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -33842,7 +33842,7 @@
         "eslint-plugin-import": "^2.30.0",
         "eslint-plugin-jsx-a11y": "^6.10.0",
         "eslint-plugin-react": "^7.36.1",
-        "eslint-plugin-react-hooks": "*",
+        "eslint-plugin-react-hooks": "beta",
```
harishv7, line 33845: is this change intended?

harishv7 (approved):
Approved with 1 minor comment

package-lock.json
seaerchin: i did an `npm i` and this got updated - i'll change it back in the interest of safety if you're concerned!

PR #942, opened 2024-12-19
fix(url): show full urls everywhere in UI

seaerchin:
## Problem
Some places don't show the full UI but we should always show the full UI everywhere. this also fixes a bug with the page settings modal where we were calling the wrong function for the close button

## Solution
1. update the creation modal to fetch the full permalink from the backend and show it
2. update the **pagecreationmodal** so that it uses a grid. this is because removing the left adornment makes the left section squished and shorter than required. 
3. update the folder settings modal to use the full url also (similar fashion as ^) 
4. update the box to use a `Suspense` so that we can load in the folder setings modal quickly

## Checks 
1. create a page in both folder and collection 
2. the page should show the full url on the details screen
3. type an extremely long url 
4. click on settings for a nested folder
5. the folder settings modal should show the full url
6. open the page settings modal 
7. click on the Close button 
8. the modal should close

## Video

https://github.com/user-attachments/assets/675c16cf-6c4d-4e95-80a5-97de87c404ee

### DIFF
--- a/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
@@ -42,6 +42,29 @@ import {
   folderSettingsModalAtom,
 } from "../../atoms"
 
+interface SuspendablePermalinkProps {
+  title: string
+  folderId: string
+}
+const SuspendablePermalink = ({
+  folderId,
+  title,
+}: SuspendablePermalinkProps) => {
+  const [{ fullPermalink }] =
+    trpc.resource.getWithFullPermalink.useSuspenseQuery({
+      resourceId: folderId ? String(folderId) : "",
+    })
+
+  return (
+    <Text textStyle="subhead-2" overflow="hidden">
+      <chakra.span color="base.content.medium">
+        {fullPermalink.split("/").slice(0, -1).join("/")}
+      </chakra.span>
+      /{title}
+    </Text>
+  )
+}
+
 export const FolderSettingsModal = () => {
   const { folderId } = useAtomValue(folderSettingsModalAtom)
   const { siteId } = useQueryParse(sitePageSchema)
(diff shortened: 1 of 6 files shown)


apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
```diff
@@ -1,6 +1,7 @@
 import { Suspense } from "react"
```
harishv7, line 1: More on the UI - should we keep the existing look but add the parent path prefix as part of the gray box?

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/DetailsScreen.tsx
```diff
@@ -179,34 +184,39 @@ export const CreateCollectionPageDetailsScreen = () => {
                     URL should be short and simple
                   </FormHelperText>
                 </FormLabel>
-                <InputGroup>
```
harishv7, line 182: Can we keep this? I thought it made it more intuitive for new users on Studio

apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
adriangohjw: had a brief discussion with Chin and it seems like it will be too long, especially when considering the parent permalink

seaerchin: actually the figma design is to have the infobox and not this, refer to the figma [here](https://www.figma.com/design/g7LnLn4IUcUC9dYxgsHEJM/Isomer-Studio-working-file?node-id=7037-20229&m=dev)
image

apps/studio/src/features/editing-experience/components/CreateCollectionPageModal/DetailsScreen.tsx
seaerchin: see above

apps/studio/src/features/dashboard/components/FolderSettingsModal/FolderSettingsModal.tsx
harishv7: Yea that's fine. What I felt helped was having the <example.com>/ or the site domain in front

For now a simple fix would be to keep the `your-site.gov.sg/` in front of the permalink

seaerchin: we shuold change everything at once but idt that should be a blocker here

harishv7 (approved):
Approving first, should follow up with future PR to add the site domain prefix

PR #945, opened 2024-12-19
fix - add tooltip for navbar logo

adriangohjw:
## Problem

isomer logo in studio navbar is missing tooltip on hover

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- add tooltip
- change copywriting from "back to dashboard" to "back to sites" as dashboard can be vague

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image


apps/studio/src/components/AppNavbar.tsx
```diff
@@ -33,22 +33,24 @@ export function AppNavbar(): JSX.Element {
...
-              <Image
-                src="/assets/isomer-logo-color.svg"
-                height={24}
-                width={22}
-                alt="Back to dashboard"
-                priority
-              />
-            }
-          />
+          <Tooltip label={"Back to sites"} placement="right">
```
harishv7, line 36: Will this also show on the "All sites" page? That might be confusing

adriangohjw: It does. Currently there's already an existing link to do that without the tooltip, so IMO not adding can be more confusing because users click and "nothing happens". Perhaps a better improvement is to remove the link altogether?

adriangohjw: updated for homepage to not be a button and not be clickable

dcshzj (approved):

PR #952, opened 2024-12-20
add soft delete for users and permissions

harishv7:
## Problem

Closes ISOM-1701

## Solution

**Breaking Changes**

**Features**:
- Added soft delete functionality for Users and ResourcePermissions via `deletedAt` column
- Added user deletion check during email authentication flow

**Improvements**:
- Enhanced permission queries to exclude soft-deleted records
- Added filtering for deleted users in ongoing sessions

**Bug Fixes**:
- Fixed potential security issue where deleted users could still access the system with active sessions

## Tests

- Verify soft delete functionality for Users, by setting deletedAt to any valid timestamp
- Users with deletedAt in the User table should not be able to login
- Users who are already logged in with an active session should be logged out when the deletedAt is set in User table
- Users can have Resource Permissions to several sites. If a Resource Permission is set, then the User should not be able to see that specific site in the All Sites view. Visitng the deleted resource Site's URL directly should show that they are unauthorised to view the page

**New scripts**:
- `20241220130729_add_deletedat_user_perms` : Adds deletedAt columns to User and ResourcePermission tables

### DIFF
--- a/apps/studio/src/server/modules/user/user.service.ts
+++ b/apps/studio/src/server/modules/user/user.service.ts
@@ -1,0 +1,18 @@
+import { db } from "../database"
+
+export const isUserDeleted = async (email: string) => {
+  const lowercaseEmail = email.toLowerCase()
+  const user = await db
+    .selectFrom("User")
+    .where("email", "=", lowercaseEmail)
+    .select(["deletedAt"])
+    // Email is a unique field in User table
+    .executeTakeFirst()
+
+  console.log(`isUserDeleted: ${JSON.stringify(user)}`)
+
+  if (user?.deletedAt) {
+    return true
+  }
+  return false
+}
--- a/apps/studio/src/server/modules/user/user.service.ts
+++ b/apps/studio/src/server/modules/user/user.service.ts
@@ -9,7 +9,5 @@ export const isUserDeleted = async (email: string) => {
     // Email is a unique field in User table
     .executeTakeFirst()
 
-  console.log(`isUserDeleted: ${JSON.stringify(user)}`)
-
-  return user?.deletedAt
+  return !!user?.deletedAt
 }
--- a/apps/studio/src/server/modules/user/user.service.ts
+++ b/apps/studio/src/server/modules/user/user.service.ts
@@ -9,5 +9,5 @@ export const isUserDeleted = async (email: string) => {
     // Email is a unique field in User table
     .executeTakeFirst()
 
-  return !!user?.deletedAt
+  return user?.deletedAt ? true : false
 }
--- a/apps/studio/src/server/modules/user/user.service.ts
+++ b/apps/studio/src/server/modules/user/user.service.ts
@@ -11,8 +11,5 @@ export const isUserDeleted = async (email: string) => {
 
   console.log(`isUserDeleted: ${JSON.stringify(user)}`)
 
-  if (user?.deletedAt) {
-    return true
-  }
-  return false
+  return user?.deletedAt
 }
--- a/apps/studio/prisma/generated/generatedEnums.ts
+++ b/apps/studio/prisma/generated/generatedEnums.ts
@@ -1,23 +1,23 @@
 export const ResourceState = {
-  Draft: "Draft",
-  Published: "Published",
-} as const
-export type ResourceState = (typeof ResourceState)[keyof typeof ResourceState]
+    Draft: "Draft",
+    Published: "Published"
+} as const;
+export type ResourceState = (typeof ResourceState)[keyof typeof ResourceState];
 export const ResourceType = {
-  RootPage: "RootPage",
-  Page: "Page",
-  Folder: "Folder",
-  Collection: "Collection",
-  CollectionMeta: "CollectionMeta",
-  CollectionLink: "CollectionLink",
-  CollectionPage: "CollectionPage",
-  IndexPage: "IndexPage",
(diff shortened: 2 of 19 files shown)


harishv7:
* **#952** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -0,0 +1,18 @@
...
+export const isUserDeleted = async (email: string) => {
+  const lowercaseEmail = email.toLowerCase()
+  const user = await db
+    .selectFrom("User")
+    .where("email", "=", lowercaseEmail)
+    .select(["deletedAt"])
+    // Email is a unique field in User table
+    .executeTakeFirst()
+
+  console.log(`isUserDeleted: ${JSON.stringify(user)}`)
```
adriangohjw, line 12: nit - remove?

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  if (user?.deletedAt) {
+    return true
+  }
+  return false
```
adriangohjw, line 17: nit
```suggestion
  return user?.deletedAt
```

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: okay nvm, saw it has been removed in subsequent PR

adriangohjw (commented):
curious, it seems like this solution isn't "audit complete".

For example, how do we handle situation where agency wants to restore a user's permission for a site's resource? we cannot add another record due to ResourcePermission's `@@unique([userId, siteId, resourceId, role])`. We can remove the `deletedAt` but it will mean remove the audit trail aspect of it

It's alright if this is a stopgap measure, but just want to clarify that this is intentional

Not urgent but we might also want to add a backlog ticket to write a script to automate this.

harishv7:
@adriangohjw this measure is to support soft deletes for now. In terms of audit completeness, we will have to re-look into that in [ISOM-1072](https://linear.app/ogp/issue/ISOM-1702/improve-audit-logging)

To re-store a user's permissions, we will need to remove the `deletedAt` in `User` table. For `ResourcePermission` we will add a new entry. To find out the audit trail, we can check based on the `userId` when were the resource permissions removed fully and re-added.

Couple the above with Ops email/Slack request, there is still some kind of full audit picture. Lmk if that makes sense!

adriangohjw (approved):

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -0,0 +1,13 @@
...
+export const isUserDeleted = async (email: string) => {
+  const lowercaseEmail = email.toLowerCase()
+  const user = await db
+    .selectFrom("User")
+    .where("email", "=", lowercaseEmail)
+    .select(["deletedAt"])
+    // Email is a unique field in User table
+    .executeTakeFirst()
+
+  return !!user?.deletedAt
```
spaceraccoon, line 12:
```suggestion
  return user !== undefined && user.deleted !== null
```

Wondering if a more precise comparison might be better; since `deletedAt` is a `Timestamp` type which includes `string`, if `user.deletedAt === ''`, this would still pass the initial check even if it's not intended.

Original is still fine though, just a nit

spaceraccoon (commented):
Minor nits but looks ok since this is just introducing the field! Main concern is that now you have to ensure `deletedAt` is checked/included in all queries / updates are users. Avoid mass assignments like:

```
db.users.update({
  ..user
})
```

As this could overwrite the `deletedAt` field when it's not intended / authorised. I suggest a custom linting rule or work with Copilot to write a CodeQL rule.

You may want to update your seed script as well to ensure the provided whitelisted emails are lowercased.

I noticed `agnosticProcedure`/`nonStrictAuthMiddleware` exists and doesn't do the deletedAt check; should this be added as well now that it's a soft delete? In fact, this procedure isn't currently in use, does it still need to be around?

apps/studio/src/server/modules/user/user.service.ts
harishv7: The deletedAt field which translated back can only be of type Date, null or undefined so wouldn't be an issue

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -0,0 +1,13 @@
+import { db } from "../database"
+
+export const isUserDeleted = async (email: string) => {
+  const lowercaseEmail = email.toLowerCase()
```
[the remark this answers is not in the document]
harishv7: sounds good, added a normalise function and updated the test/seed functions

harishv7:
Thanks @spaceraccoon have made the fixes and good catch on the unused middleware/procedure. We kept it there for future use initially, but let me go ahead and remove it a subsequent PR.

harishv7:
### Merge activity

* **Jan 23, 8:37 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/952).

PR #953, opened 2024-12-20
add tests and handling for deleted users

harishv7:
## Problem

Deleted users were still able to access the system and view sites they previously had access to.

Closes ISOM-1701

## Solution

Added checks to prevent deleted users from logging in and viewing sites they previously had access to.

**Breaking Changes**:

**Features**:
- Added user deletion status check during login
- Filtered out sites where user permissions have been deleted

**Improvements**:
- Removed debug console log from user deletion check

## Tests

- Added test for login attempt with deleted user
- Added test for site listing with deleted permissions
- Added user service tests for checking deletion status
- Added helper function for setting up test users with deletion status

**New scripts**:
None

**New dependencies**:
None

**New dev dependencies**:
None

### DIFF
--- a/apps/studio/src/server/modules/user/__tests__/user.service.test.ts
+++ b/apps/studio/src/server/modules/user/__tests__/user.service.test.ts
@@ -1,0 +1,47 @@
+import { resetTables } from "tests/integration/helpers/db"
+import { setupUser, setUpWhitelist } from "tests/integration/helpers/seed"
+
+import { isUserDeleted } from "../user.service"
+
+describe("user.service", () => {
+  beforeAll(async () => {
+    await resetTables("User")
+    await setUpWhitelist({ email: "@example.com" })
+
+    // Setup active user
+    await setupUser({
+      name: "Active User",
+      userId: "active123",
+      email: "active@example.com",
+      phone: "12345678",
+      isDeleted: false,
+    })
+
+    // Setup deleted user
+    await setupUser({
+      name: "Deleted User",
+      userId: "deleted123",
+      email: "deleted@example.com",
+      phone: "12345678",
+      isDeleted: true,
+    })
+  })
+
+  it("should return false if user is not deleted", async () => {
+    // Arrange
+    const email = "active@example.com"
+    // Act
+    const result = await isUserDeleted(email)
+    // Assert
+    expect(result).toBe(false)
+  })
+
+  it("should return true if user is deleted", async () => {
+    // Arrange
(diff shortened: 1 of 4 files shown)


apps/studio/src/server/modules/user/__tests__/user.service.test.ts
```diff
@@ -0,0 +1,47 @@
+import { resetTables } from "tests/integration/helpers/db"
+import { setupUser, setUpWhitelist } from "tests/integration/helpers/seed"
+
+import { isUserDeleted } from "../user.service"
+
+describe("user.service", () => {
+  beforeAll(async () => {
+    await resetTables("User")
+    await setUpWhitelist({ email: "@example.com" })
```
adriangohjw, line 9: do we need this? don't see any logic within `isUserDeleted` that's using the whitelisting table + believe this is handled by `isEmailWhitelisted`?

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+
+    // Setup deleted user
+    await setupUser({
+      name: "Deleted User",
+      userId: "deleted123",
+      email: "deleted@example.com",
+      phone: "12345678",
+      isDeleted: true,
+    })
```
adriangohjw, line 27: nit - might be better to move each of this into the respective test case in the "arrange" stage?

adriangohjw: nit - do we want to make the non-essential fields (name, userId, phone) to fallback to a default value in `setupUser` instead of having to specify them? doing so will make this setup cleaner and more readable since the "attention" will be on the `isDeleted` arg

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
harishv7: good catch! removed

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
harishv7: good points on both, have updated

adriangohjw (approved):

harishv7:
### Merge activity

* **Jan 23, 8:37 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/953).

PR #961, opened 2024-12-26
refactor: logocloud

seaerchin:
## Problem
1. logocloud requires `assetsBaseUrl` for relative links but `site` already has this - we should account for this and just derive it from `site` 
2. studio uses `title` to render so now for logocloud it's showing `Title` (see SS)

## Solution
1. update logocloud from using `assetsBaseUrl` to using `site.assetsBaseUrl`
2. update `title` in the schema to `Logocloud component`

## How to test
1. open up studio
2. add a logocloud component
3. add a relative url
4. the image shuold render without issue
5. for the title, it shuold display logocloud component

## Migration plan
- none -> because now it's no longer picking up the property post release and it's isntead deriving from sites.

## screenshots
image

https://github.com/user-attachments/assets/1101fb41-e89e-48dc-89f6-cc997230502d

### DIFF
--- a/packages/components/src/interfaces/complex/LogoCloud.ts
+++ b/packages/components/src/interfaces/complex/LogoCloud.ts
@@ -1,6 +1,7 @@
 import type { Static } from "@sinclair/typebox"
 import { Type } from "@sinclair/typebox"
 
+import { IsomerSiteProps } from "~/types"
 import { AltTextSchema, ImageSrcSchema } from "./Image"
 
 export const LOGO_CLOUD_TYPE = "logocloud"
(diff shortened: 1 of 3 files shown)


packages/components/src/interfaces/complex/LogoCloud.ts
```diff
@@ -1,6 +1,7 @@
 import type { Static } from "@sinclair/typebox"
 import { Type } from "@sinclair/typebox"
 
+import { IsomerSiteProps } from "~/types"
```
adriangohjw, line 4: nit
```suggestion
import type { IsomerSiteProps } from "~/types"
```

adriangohjw (approved):

adriangohjw (changes requested):
need fix Logocloud storybook to pass in the right args

adriangohjw (approved):

PR #963, opened 2024-12-27
fix getParsedDate - add 8h to account for timezone

adriangohjw:
## Problem

date in collection preview is one day behind on studio

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- add a 8h addition to date to account for SG timezone

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-12-27 at 9 28 07 PM

**AFTER**:

Screenshot 2024-12-27 at 10 32 13 PM

## Tests

1. create a link in a collection
2. select a date
3. the date in the date selector and that in the preview should be the same

### DIFF
--- a/packages/components/src/utils/getParsedDate.ts
+++ b/packages/components/src/utils/getParsedDate.ts
@@ -1,5 +1,6 @@
 import { isMatch, parse } from "date-fns"
 
+const TIMEZONE_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
 const SUPPORTED_DATE_FORMATS = [
   "dd/MM/yyyy",
   "d MMM yyyy",
@@ -7,7 +8,7 @@ const SUPPORTED_DATE_FORMATS = [
   "dd MMM yyyy",
   "dd MMMM yyyy",
   "yyyy-MM-dd",
-  "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
+  TIMEZONE_DATE_FORMAT,
 ]
 
 export const getParsedDate = (dateString: string) => {
@@ -20,7 +21,14 @@ export const getParsedDate = (dateString: string) => {
 
       try {
         if (isMatch(dateString, format)) {
-          return parse(dateString, format, new Date())
+          let offsetDate = dateString
+          if (format === TIMEZONE_DATE_FORMAT) {
+            offsetDate = new Date(
+              // Add 8 hours to account for the Singapore timezone offset
+              new Date(dateString).getTime() + 8 * 60 * 60 * 1000,
+            ).toISOString()
+          }
+          return parse(offsetDate, format, new Date())
         }
       } catch (e) {
         return new Date()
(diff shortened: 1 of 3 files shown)


packages/components/src/utils/getParsedDate.ts
```diff
@@ -20,7 +21,14 @@ export const getParsedDate = (dateString: string) => {
 
       try {
         if (isMatch(dateString, format)) {
-          return parse(dateString, format, new Date())
+          let offsetDate = dateString
+          if (format === TIMEZONE_DATE_FORMAT) {
+            offsetDate = new Date(
+              // Add 8 hours to account for the Singapore timezone offset
+              new Date(dateString).getTime() + 8 * 60 * 60 * 1000,
```
harishv7, line 28: Doesn't `new Date()` by default get the date in local timezone?

adriangohjw: from what i understand, `new Date()` is the reference date that provides default values for any missing components in `dateString`. e.g. if the `dateString` includes only the month and day, the year is taken from the reference date

packages/components/src/utils/getParsedDate.ts
```diff
@@ -20,7 +21,14 @@ export const getParsedDate = (dateString: string) => {
 
       try {
         if (isMatch(dateString, format)) {
-          return parse(dateString, format, new Date())
+          let offsetDate = dateString
+          if (format === TIMEZONE_DATE_FORMAT) {
+            offsetDate = new Date(
+              // Add 8 hours to account for the Singapore timezone offset
```
dcshzj, line 27: issue: I think hardcoding the 8 hours here might not be an ideal solution (although the vast majority of our users are in this timezone). Instead, we should be parsing the date in the machine's local time zone, as we can see that the date changes when you try to change your computer's timezone.

https://github.com/user-attachments/assets/6cd10435-3b1c-4ba9-b4e3-44647011a82a

This doesn't affect actual live sites since the date formats have already been statically generated at build time.

https://github.com/user-attachments/assets/944f341c-781d-469d-a628-669ef672e3fa

Hence, instead of `8`, we should determine the timezone offset that the user's machine is in, then inputing that as part of the calculation here. Alternatively, we can discard the time part of this date string and only work with the date part (i.e. `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` becomes just `yyyy-MM-dd`) before performing the parsing, so we are guaranteed to be working without consideration for the timezone.

adriangohjw: address in https://github.com/opengovsg/isomer/pull/963/commits/3c4c605bb68401f7ccacd7cfbdcaaa4cb31f6d83, thanks!

dcshzj (approved):

packages/components/src/utils/getParsedDate.ts
adriangohjw: btw have also updated GH actions to use `Asia/Singapore` during CI - https://github.com/opengovsg/isomer/pull/963/commits/c7177d5621cfbfd826d1f71dad460d1cca087ab5

PR #964, opened 2024-12-27
Fix inconsistent studio dashboard layout

adriangohjw:
## Problem

currently the dashboard layout for "root page", "folder" and "collection" are all incorrect, which leads to a mild inconsistencies when navigating throughtout the pages

it also means logic are duplicated and fixes have to be checked against every part

this also fixes the double scrollbar issue introduced in https://github.com/opengovsg/isomer/pull/959

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- refactor to use common `Dashboard` component

## Before & After Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/a378f0a4-01da-4d54-b519-eee27a2fc9df

**AFTER**:

https://github.com/user-attachments/assets/e44e8e07-8aec-4bed-949c-5e914e6c668c

## Tests

- [ ] navigate between the root page, folder and collection -> the layout should be consistent


apps/studio/src/features/dashboard/components/Dashboard.tsx
```diff
@@ -0,0 +1,109 @@
...
+              wordBreak="break-all"
+            >
+              {title}
+            </Text>
+          </HStack>
+          <HStack>{buttons}</HStack>
+        </Flex>
+      </VStack>
+      {preTableContent && preTableContent}
+      <Box width="100%">{table}</Box>
```
seaerchin, line 106: hmm, i feel like this should be a `children` prop because this is a wrapper component because this is a layout component that determines how the page itself is laid out.

see, for example, `AdminSidebarOnlyLayout`!

seaerchin (commented):
minor nit but i don't want this to snowball!

apps/studio/src/components/CmsSidebar/CmsContainer.tsx
```diff
@@ -75,7 +75,7 @@ export function CmsContainer({
           {sidenav}
         </Box>
       </GridItem>
-      <GridItem as="main" area="main" overflow="auto">
+      <GridItem as="main" area="main" overflow="hidden">
```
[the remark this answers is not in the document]
adriangohjw: while changing to `auto` solves the issue, it causes a minor issue of having multiple scrollbar

by using this new layout (plucked from folder pages), changing to `hidden` will still allow the navigation to be viewable while also solving the multiple scrollbar

apps/studio/src/features/dashboard/components/Dashboard.tsx
adriangohjw: i think `table` can be a children, but if we also want to pass in `preTableContent`, then user will have to remember to wrap table in <Box width="100%">

seaerchin: we can still use `children` prop but the caller needs to wrap it in a fragment `<></>`, which seems to be better fitting as the caller is the one that makes the judgement of whether there is `preTableContent`

adriangohjw: oh my bad, i typed `<Box width="100%">` without the quotes and it disappeared when github renders the comment - updated my initial reply

seaerchin: if we always need this `Box w = 100%` on all the callsites, then i don't think it shuold be here. instead, the `Table` itself should specify it

seaerchin (approved):
lgtm but could we shift the `w = 100%` to be on the table instead?

PR #965, opened 2024-12-29
isom-1709 - chore: do not wrap link text in span

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1709/[lighthouse-seo]-non-dropdown-link-in-navbar-does-not-have-descriptive

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- do not wrap non-dropdown navbar item in `<span>`

### DIFF
--- a/apps/studio/public/assets/css/preview-tw.css
+++ b/apps/studio/public/assets/css/preview-tw.css
@@ -1418,10 +1418,6 @@ video {
   -webkit-line-clamp: 3;
 }
 
-.\!block {
-  display: block !important;
-}
-
 .block {
   display: block;
 }
@@ -3154,11 +3150,6 @@ video {
 .text-link {
   --tw-text-opacity: 1;
   color: rgb(26 86 229 / var(--tw-text-opacity));
-}
-
-.text-link-hover {
-  --tw-text-opacity: 1;
-  color: rgb(21 71 190 / var(--tw-text-opacity));
 }
 
 .text-site-secondary {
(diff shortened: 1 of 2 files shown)


apps/studio/public/assets/css/preview-tw.css
adriangohjw, line 1: this isn't from this PR but thought to commit it in together

packages/components/src/templates/next/components/internal/Navbar/NavItem.tsx
```diff
@@ -64,9 +64,9 @@ export const NavItem = forwardRef<HTMLButtonElement, NavbarItemProps>(
             isExternal={isExternalUrl(url)}
             showExternalIcon={isExternalUrl(url)}
             href={referenceLinkHref}
-            className={item({ isOpen })}
+            className={`${item({ isOpen })} ${groupFocusVisibleHighlight()}`}
```
dcshzj, line 67: issue: I think the group focus visible highlight might need to be replaced with the non-group version (`focusVisibleHighlight`), as it doesn't work currently (but please verify also):

https://github.com/user-attachments/assets/4a8cfafc-9fd9-4b8d-b58c-0e0c50c171ad

adriangohjw: @dcshzj address in latest commits, thanks!

dcshzj (approved):

PR #968, opened 2024-12-30
Fix index page not rendering link

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1712/bug-linking-to-an-index-page-cause-incorrect-rendering-of

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

**this has been tested on STAGING and verified to be working**

- add `indexPageId` attribute to SitemapEntry
- use that value for matching in `getReferenceLinkHref.ts`

## Tests

1. go to studio
2. create a page, add a Text block and link to a IndexPage resource
3. trigger a codebuild
4. go to the published page and check that the URL is valid and not incorrect `.../[resource:...:.....]`

### DIFF
--- a/tooling/build/scripts/publishing/index.ts
+++ b/tooling/build/scripts/publishing/index.ts
@@ -111,17 +111,19 @@ async function main() {
           resource.parentId,
         )
 
+        const isIndexPage =
+          resource.fullPermalink.endsWith(INDEX_PAGE_PERMALINK) &&
+          resource.type !== "RootPage"
+
         const resourceConvertedPermalink = getConvertedPermalink(resource.fullPermalink)
 
         const idOfFolder = resources.find(
-          (item) =>
-            resource.fullPermalink.endsWith(INDEX_PAGE_PERMALINK) &&
-            resource.type !== "RootPage" &&
-            item.fullPermalink === resourceConvertedPermalink,
+          (item) => isIndexPage && item.fullPermalink === resourceConvertedPermalink,
         )?.id
 
         const sitemapEntry: SitemapEntry = {
           id: idOfFolder ?? resource.id,
+          indexPageId: isIndexPage ? resource.id : undefined,
           type: resource.type,
           title: resource.title,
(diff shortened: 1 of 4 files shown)


adriangohjw:
not sure why build fail though - re-running the storybook on chromatic works

tooling/build/scripts/publishing/index.ts
```diff
@@ -111,19 +111,25 @@ async function main() {
...
-            resource.fullPermalink.endsWith(INDEX_PAGE_PERMALINK) &&
-            resource.type !== "RootPage" &&
-            item.fullPermalink ===
-              getConvertedPermalink(resource.fullPermalink),
+            isIndexPage && item.fullPermalink === resourceConvertedPermalink,
         )?.id
 
         const sitemapEntry: SitemapEntry = {
           id: idOfFolder ?? resource.id,
+          indexPageId: isIndexPage ? resource.id : undefined,
```
dcshzj, line 129: issue: I think this logic is not quite right. The `id` field is present here because we are attempting to provide a mapping of the ID stored inside the database to the actual permalink that the user uses to access the specific page on the site. A folder does not physically exist in the "real world", as all permalinks must be actual pages (whether manually created or automatically generated).

In this part, we store the ID of the folder because we cannot guarantee that the `_index` custom page will always exist. Hence, we use a mapping of the folder resource ID to the permalink instead, rather than the actual index page's ID.

When linking to a folder/index page, the reference link should be linking to the folder ID rather than the custom index page, so that it is consistent with the case where the custom index page does not exist. This might be a Studio problem where we are missing the ability to link to a folder (and also missing the functionality to prevent users from being able to link directly to the custom index page).

Looking at the original issue, it might have been that the original reference link was pointing to the custom index page rather than the folder ID, so we should update the reference link to point to the folder ID rather than keep the reference to the custom index page.

PR #970, opened 2025-01-02
enhancement/add copyright year component

adriangohjw:
## Problem

Currently the year is generated on build time

But ideally, we want it to be generated on runtime since we don't know when will agencies trigger a new build + it's easier than having to remember to trigger a new build for all sites every new year

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- `use client` for the copyright and year

### DIFF
--- a/packages/components/src/templates/next/components/internal/Footer/ClientCopyrightYear.tsx
+++ b/packages/components/src/templates/next/components/internal/Footer/ClientCopyrightYear.tsx
@@ -1,5 +1,5 @@
 "use client"
 
 export const ClientCopyrightYear = () => {
-  return <>&copy; {new Date().getFullYear()} </>
+  return <>&copy; {new Date().getFullYear()}</>
 }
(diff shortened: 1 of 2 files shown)


packages/components/src/templates/next/components/internal/Footer/ClientCopyrightYear.tsx
```diff
@@ -0,0 +1,5 @@
+"use client"
+
+export const ClientCopyrightYear = () => {
+  return <>&copy; {new Date().getFullYear()} </>
```
dcshzj, line 4: nit: I think the space should be part of Footer.tsx rather than here? At first glance it wasn't clear why there was an additional space here.

dcshzj (approved):

packages/components/src/templates/next/components/internal/Footer/ClientCopyrightYear.tsx
adriangohjw: thanks for catching this! changed it in https://github.com/opengovsg/isomer/pull/970/commits/077edc12143eaf837f7a1ed5357659317d281b44

PR #972, opened 2025-01-02
Enhancement - collection do not render empty filter

adriangohjw:
## Problem

Screenshot 2025-01-02 at 9 38 47 PM

We are returning filters with only one filter option. This doesn't make sense since clicking on it has no impact - it will only clutter the UI

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- Only render a filter if it contains at least 2 items
- I refactored a large function into smaller ones to improve readability and maintainability, accepting a slight performance cost from iterating three times instead of once. Since this runs only during build time, it doesn’t affect the end-user experience.

## Tests

1. go to studio and create a collection with only two items (same date, different category)
2. publish it
3. verify that "year" filter should not appear
4. repeat with another collection with "same category, different year"


packages/components/src/templates/next/components/internal/BlogCard/BlogCard.tsx
```diff
@@ -84,9 +86,11 @@ export const BlogCard = ({
           </Text>
         )}
         {/* TODO: Feature enhancement? Filter by category when clicked */}
-        <Text className="prose-label-sm-medium text-base-content-subtle">
-          {category}
-        </Text>
+        {shouldShowCategory && (
```
dcshzj, line 89: issue: I think we should still show the category even if it is the only one? When clicking the link to open the article, the category will still be shown on the article page, so initial thoughts is that not showing the category here makes the category on the article page a little out of place.

packages/components/src/templates/next/layouts/Collection/filterUtils/index.ts
dcshzj, line 1: nit: can name the folder as just `utils`

packages/components/src/templates/next/layouts/Collection/utils.ts
```diff
@@ -249,6 +124,11 @@ export const updateAppliedFilters = (
   }
 }
 
+// use filters generated as the single source of truth
+export const shouldShowCategory = (filters: Filter[]): boolean => {
+  return Boolean(filters.find((filter) => filter.id === FILTER_ID_CATEGORY))
```
dcshzj, line 129:
```suggestion
  return filters.some((filter) => filter.id === FILTER_ID_CATEGORY)
```

packages/components/src/templates/next/layouts/Collection/filterUtils/getTagFilters.ts
```diff
@@ -0,0 +1,51 @@
...
+    const filters: Filter[] = [
+      ...acc,
+      {
+        items,
+        id: category,
+        label: category,
+      },
+    ]
+
+    return filters
```
dcshzj, line 49: issue: I think this is missing a sort function (both on the category and on the individual tags)

packages/components/src/templates/next/layouts/Collection/filterUtils/index.ts
adriangohjw: there's already an existing `utils` file that contains non-filter related utils so don't want to cause any confusion

packages/components/src/templates/next/layouts/Collection/filterUtils/getTagFilters.ts
adriangohjw: this was copied over from `getCategories` from `utils` as-is so didn't touch it

but good idea to sort it, i will just add it in nevertheless

packages/components/src/templates/next/layouts/Collection/filterUtils/index.ts
dcshzj: Hmm didn't notice that file existed, then maybe we should split the functions inside the utils file into its individual files and put inside this `utils` folder?

packages/components/src/templates/next/components/internal/BlogCard/BlogCard.tsx
adriangohjw: we only not show category when there's no category filter on the lefthand side (since that's where `shouldShowCategory` is derived from), which means there's either no/one category. IMO will clutter the UI on the collection results since it then serves no purpose in this scenario

re: comment from Sehyun https://opengovproducts.slack.com/archives/C07CWUNUL68/p1735803802912019?thread_ts=1735799448.795359&cid=C07CWUNUL68

packages/components/src/templates/next/layouts/Collection/filterUtils/index.ts
adriangohjw: addressed in https://github.com/opengovsg/isomer/pull/972/commits/c5ab110c006981444f8e20fe9a57159671f7f3f3

packages/components/src/templates/next/layouts/Collection/filterUtils/getTagFilters.ts
adriangohjw: addressed in https://github.com/opengovsg/isomer/pull/972/commits/6611f6e18e55931741f27e37b8f6c2b7df93e536

packages/components/src/templates/next/layouts/Collection/utils/__tests__/getCategoryFilter.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      items: [],
+    })
+  })
+
+  it("should count and format categories correctly", () => {
+    // Arrange
+    const items: ProcessedCollectionCardProps[] = [
+      {
+        category: "guides",
+      } as ProcessedCollectionCardProps,
```
dcshzj, line 27: issue: this type-cast is a little weird. I understand that we just want to test specific parts of the collection card props but this might hide problems if the implementation were to be changed, or if the props has changed upstream.

We can probably create a few complete test objects to be used across the different tests, or have some builder function that will automatically populate the remaining fields so that we can avoid this type casting.

packages/components/src/templates/next/layouts/Collection/utils/getAvailableFilters.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getYearFilter } from "./getYearFilter"
+
+export const getAvailableFilters = (
+  items: ProcessedCollectionCardProps[],
+): Filter[] => {
+  return [
+    getCategoryFilter(items),
+    getYearFilter(items),
+    ...getTagFilters(items),
+  ].filter((filter) => filter.items.length >= 1)
```
dcshzj, line 14: nit: I think we might be able to remove this filter? There shouldn't be a case where a filter exists but there are 0 items.

adriangohjw: currently the `get*Filter` functions returned a `Filter` object regardless, and they could be empty `items`

for example, if all the `lastUpdated` are undefined (it's an optional field, though it auto set to Date.now() in Studio UI on initialization) 

thought that we can keep this as a more defensive approach

packages/components/src/templates/next/layouts/Collection/utils/__tests__/getCategoryFilter.test.ts
adriangohjw: > might hide problems if the implementation were to be changed, or if the props has changed upstream.

hmm not quite understanding this, as this test should fail if the implementation changes (e.g. using more props), which should be the intended outcome of a test?

dcshzj (approved):
Approving first due to urgency, let's address the type-casting issue in a separate PR

PR #976, opened 2025-01-08
fix: adjust design for collection blog variant

dcshzj:
## Problem

The collection blog variant has a couple of design issues that were raised by CSA.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Removed the `capitalize` function to respect the original capitalisation of the tag category provided by the user.
- Added a small gap between different tag categories.
- Added `whitespace: pre-wrap` so that newlines can be handled properly in the collection card description. Also switched Text back to the `p` tag for simplicity.

## Before & After Screenshots

image

## Tests

- [ ] On Studio, create a collection with an article that contains multiple tags and tag categories with random capitalisation in the tag category, then publish the collection and article.
- [ ] Create another article with a different tag, but with the same tag categories. In the article page summary, write a random description and use Shift+Enter to create new lines, then publish this article.
- [ ] Verify on the published site that the capitalisation of the tag category is respected, there is a small gap between the tag categories and the newlines in the description are rendered correctly rather than in one line.

### DIFF
--- a/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
+++ b/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
@@ -3,7 +3,7 @@ import type { Meta, StoryObj } from "@storybook/react"
 import { withChromaticModes } from "@isomer/storybook-config"
 
 import type { CollectionCardProps } from "~/interfaces"
-import { Tag } from "~/interfaces/internal/CollectionCard"
+import type { Tag } from "~/interfaces/internal/CollectionCard"
 import { BlogCard } from "./BlogCard"
 
 const meta: Meta<typeof BlogCard> = {
@@ -100,7 +100,7 @@ export const TagsWithoutImage: Story = {
   args: generateArgs({
     title: "Collection card without tags",
     withoutImage: true,
-    description: "This is a random description that will be on the card",
+    description: "This is a random description\nthat will be on the card",
     tags: [
       {
         category: "very long",
--- a/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.tsx
+++ b/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.tsx
@@ -65,7 +65,7 @@ export const BlogCard = ({
           </Link>
         </h3>
         {tags && tags.length > 0 && (
-          <div className="-mt-1">
+          <div className="-mt-1 flex flex-col gap-1">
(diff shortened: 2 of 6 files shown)


packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
```diff
@@ -100,7 +100,7 @@ export const TagsWithoutImage: Story = {
   args: generateArgs({
     title: "Collection card without tags",
     withoutImage: true,
-    description: "This is a random description that will be on the card",
+    description: "This is a random description\nthat will be on the card",
```
adriangohjw, line 103: nit - should we make a new story for this? since we are testing for different things (one is tag without image, another is newline)

adriangohjw (approved):

packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
dcshzj: Hmm I didn't create a new story cos it isn't really a standalone feature, plus it doesn't conflict with the other functionality of the card.

adriangohjw: oh i was thinking it might be misleading to lump them together under `TagsWithoutImage` because we should be treating stories like actual test cases and thus following the principle of isolating every test case to testing only one thing

dcshzj: Yup agreed if it is for a specific feature, but in this case it is like any other description which I thought was not sufficient to have its own stories.

Also I'm not sure when did we start to treat stories like test cases, but I think visual tests should be kept separate from Storybook. These stories are more for simulating instances which are difficult to get to (e.g. loading states) to facilitate review by design, whereas visual tests say using Playwright are more appropriate to perform the testing that you mentioned.

adriangohjw: Thanks for sharing - i see where you are coming from and agree!

Though since we’re not quite using Playwright for visual testing (at the moment) + this approach is using Storybook to some extent as a testing proxy, it might be clearer to split them up so that it's easier to understand and maintain in the future.

That said, i’m not blocking this PR - just thought to share that this approach might feel a bit conflated and might introduce some potential for confusion. Let's proceed to merge this and we can maybe discuss this again later on

PR #977, opened 2025-01-08
isom-1714: Fix missing autoscrolling for next/link

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1714/bug-scroll-position-doesnt-reset-for-internal-links

related issue: https://github.com/vercel/next.js/issues/45187 (still happening with nextjs in next 14)

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:
- adding a `ScrollToTop` component to `Skeleton`
- update KeyStats and Infocards to use `LinkComponent` (currently left out)

## Tests

**Tested on staging to work**

1. trigger a new build on codebuild
2. scroll down and click on any internal link. It should scroll user to the top of the page instead of retaining their existing scroll position

### DIFF
--- a/packages/components/src/templates/next/components/internal/ScrollToTop.tsx
+++ b/packages/components/src/templates/next/components/internal/ScrollToTop.tsx
@@ -1,22 +1,19 @@
 "use client"
 
 import { useEffect } from "react"
-import { usePathname } from "next/navigation"
 
 /*
+  NOTE: This is a temporary fix until we have a better solution.
+
   Fix with issue w/ "next/Link" not scrolling to the top when clicking on <Link />
   if we have a header with "sticky" that's not in its sticky position.
   Ref: https://github.com/vercel/next.js/issues/45187#issuecomment-1639518030
 */
 export const ScrollToTop = () => {
-  const pathname = usePathname()
-
   useEffect(() => {
-    // Required because "window" is a browser-only API that's not available
-    // during static site generation (SSG)
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    if (window) window.scroll(0, 0)
-  }, [pathname])
+    if (window) window.scrollTo(0, 0)
+  }, [])
 
   return <></>
 }
--- a/packages/components/src/templates/next/components/internal/ScrollToTop.tsx
+++ b/packages/components/src/templates/next/components/internal/ScrollToTop.tsx
@@ -1,0 +1,22 @@
+"use client"
+
+import { useEffect } from "react"
+import { usePathname } from "next/navigation"
+
+/*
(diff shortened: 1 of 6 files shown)


packages/components/src/templates/next/components/internal/ScrollToTop.tsx
```diff
@@ -0,0 +1,22 @@
...
+*/
+export const ScrollToTop = () => {
+  const pathname = usePathname()
+
+  useEffect(() => {
+    // Required because "window" is a browser-only API that's not available
+    // during static site generation (SSG)
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (window) window.scroll(0, 0)
+  }, [pathname])
```
seaerchin, line 19: actually i'm abit confused here - why is `pathname` the dependency in `useEffect` and not `window`?

packages/components/src/templates/next/components/internal/ScrollToTop.tsx
```diff
@@ -0,0 +1,22 @@
+"use client"
+
+import { useEffect } from "react"
+import { usePathname } from "next/navigation"
```
dcshzj, line 4: not a review but saw this, `next/navigation` is not available in components so we will need to depend on `window` if we need the path name

packages/components/src/templates/next/components/internal/ScrollToTop.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  Fix with issue w/ "next/Link" not scrolling to the top when clicking on <Link />
+  if we have a header with "sticky" that's not in its sticky position.
+  Ref: https://github.com/vercel/next.js/issues/45187#issuecomment-1639518030
+*/
+export const ScrollToTop = () => {
+  useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (window) window.scrollTo(0, 0)
+  }, [])
```
seaerchin, line 16: not too sure if this works? won't the empty dependency array check once at initialisation and not check any further? 

as a matter of style, i think we should specify `window` here even though at run-time `window` will be available since it's on the browser

seaerchin (approved):
approving because i saw the comment re: testing on `staging`, but could we add the `window` dependency!

packages/components/src/templates/next/components/internal/ScrollToTop.tsx
adriangohjw: yup it will still work and should be empty dependency as this is only to prevent this from running at build-time

According to [nextjs doc](https://nextjs.org/docs/app/building-your-application/deploying/static-exports#browser-apis)
image

actually after re-reading this, i realized we can do away with `if (window)` - will make the change

PR #979, opened 2025-01-09
fix - set selection on update for tiptap

adriangohjw:
## Problem

When users type in tiptap, the cursor will jump. This is caused by the `useEffect` we have which was necessary to ensure data is repopulated back when user dismisess the "are you sure you want to drop changes" modal

RE: https://opengovproducts.slack.com/archives/C07CWUNUL68/p1736426201036129

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- to note the cursor position and add it back

## Tests

1. on studio, create a Callout component with some text. Save the component
2. open the component again. Put your cursor in the middle of the text and start typing - your cursor should not jump to the end of the text

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
@@ -94,7 +94,7 @@ export function JsonFormsProseControl({
       const selection = editor?.state.selection
       if (!selection) return
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-      editor.commands.setContent(data, false)
+      editor.commands.setContent(data)
       editor.commands.setTextSelection(selection)
     }
   }, [data])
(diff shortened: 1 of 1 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
```diff
@@ -91,8 +91,11 @@ export function JsonFormsProseControl({
   // Needed to force the value to be set when user clicks on "Go back to editing" in the exit modal
   useEffect(() => {
     if (data !== undefined) {
+      const selection = editor?.state.selection
+      if (!selection) return
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-      editor?.commands.setContent(data)
+      editor.commands.setContent(data, false)
```
seaerchin, line 97: image

not sure if we need this second argument?

seaerchin (commented):
i'm abit hesistant on this because it seems like we're adding alot of overhead that we must remember as `useEffect` is not very tightly coupled to the call site - should we try to find another solution?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsProseControl.tsx
adriangohjw: resolved in https://github.com/opengovsg/isomer/pull/979/commits/cb8ab4f586a060493951334d2614f65cefa5266d

adriangohjw:
> i'm abit hesistant on this because it seems like we're adding alot of overhead that we must remember as useEffect is not very tightly coupled to the call site - should we try to find another solution?

@seaerchin agree. perhaps i can add a ticket for that since this issue is quite a blocker to the site users since they cannot edit certain components

seaerchin (approved):

PR #981, opened 2025-01-10
fix: duplicate image thumbnail logic for GitHub sites

dcshzj:
## Problem

The image thumbnail for collection is not applying for GitHub sites.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Duplicate the image thumbnail logic to generate-sitemap.js which is used for GitHub sites.


adriangohjw (approved):
nit - can refactor `tooling/build/scripts/publishing/utils/getResourceImage.ts` to use this to reduce logic duplication but i think it's fine

PR #982, opened 2025-01-13
feat: vica integration

adriangohjw:
## Problem

Add integration for VICA

## Solution

**Breaking Changes**

  - Details ...

**Features**:

Refer to this slack message for installation guide: https://opengovproducts.slack.com/archives/C087MUEJAMA/p1736330772752669?thread_ts=1736245461.677389&cid=C087MUEJAMA

**_note: decided to follow vica's way of specifying the keys and values as will be easier to port over from agency_**

- Add Vica stylesheet and widget component
- Add vica props to siteprops
- todo - clarifying with VICA on usage of props to pass in colour config (installation guide wasn't clear). In the meantime, we can manually pass it in via the props

## Tests

note: slightly tricky to test as we will need to get the app-id and app-name from agencies

1. add the necessary vica config in `site` in the Site table 
2. rebuild and see if it shows up

### DIFF
--- a/packages/components/src/interfaces/internal/Vica.ts
+++ b/packages/components/src/interfaces/internal/Vica.ts
@@ -1,5 +1,8 @@
 import type { ScriptComponentType } from "~/types"
 
+// NOTE: not all props will be used even if we passed them in
+// as we will override some of them with Isomer's configuration e.g. font-family
+// Nevertheless, keeping them here for reference
 interface VicaWidgetProps {
   // UI Theme
   "app-id": string
--- a/packages/components/src/interfaces/internal/Vica.ts
+++ b/packages/components/src/interfaces/internal/Vica.ts
@@ -1,0 +1,48 @@
+import type { ScriptComponentType } from "~/types"
+
+interface VicaWidgetProps {
+  // UI Theme
+  "app-id": string
+  "app-name": string
+  "app-icon"?: string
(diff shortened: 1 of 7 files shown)


packages/components/src/interfaces/internal/Vica.ts
```diff
@@ -0,0 +1,51 @@
+import type { ScriptComponentType } from "~/types"
+
+// NOTE: not all props will be used even if we passed them in
+// as we will override some of them with Isomer's configuration e.g. font-family
+// Nevertheless, keeping them here for reference
+interface VicaWidgetProps {
```
seaerchin, line 6: abit confused - how will we override? i'm assuming the compiled css will have `font-family` but how will vica pick it up

seaerchin (approved):
mostly lgtm - i just took a (brief) look at the vica docs linked in slack

packages/components/src/interfaces/internal/Vica.ts
adriangohjw: can override directly in https://github.com/opengovsg/isomer/blob/25eee45d5b4748f7b2138bb3acd02f4f5c1c7caa/packages/components/src/templates/next/components/internal/Vica/Vica.tsx#L30-L36

once VICA got back to me, we also ideally want to use the site's `color` to prefill this. For now the only props we know we will override what's passed in is font family

PR #984, opened 2025-01-13
improve lighthouse performance score - preload GTM 

adriangohjw:
## Problem

Similar to https://github.com/opengovsg/isomer/pull/966, we can preload GTM to improve loading speed

image

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- preload GTM
- refactor into `PreloadHelper`


seaerchin (approved):
lgtm, dk if we want to do this for wogaa/vica too

adriangohjw:
@seaerchin haven't seen the warning show up for wogaa yet so i think no need for now. VICA doesn't have to be preloaded since it only loads the widget after page load so don't think it will benefit from this!

PR #986, opened 2025-01-13
improve lighthouse performance - do not lazyload above-fold image

adriangohjw:
_**Note: not urgent, good-to-have improvement**_

## Problem

Follow up enhancement of https://github.com/opengovsg/isomer/pull/967. 

We followed next.js convention and lazy loaded all images. While the overall sites' lighthouse performance score has shown improvement, there are further rooms for improvement for certain pages.

For example, some pages have images above the fold being lazy loaded, thus having poor lighthouse score

image

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

**Tested to be working on STG**

- use `doesComponentHaveImage` to determine which component has images
- in `renderPage`, pass in `shouldLazyLoad` into the components (true if they are the first component, and false otherwise)

## Tests

1. go to a page in Studio and add multiple different components with images
2. publish the page and go to the published page. inspect the dev console and you should see only image(s) in the first component is eager loaded, while image(s) in the subsequent components are lazy loaded

### DIFF
--- a/packages/components/src/interfaces/complex/Contentpic.ts
+++ b/packages/components/src/interfaces/complex/Contentpic.ts
@@ -22,4 +22,5 @@ export const ContentpicSchema = Type.Object(
 export type ContentpicProps = Static<typeof ContentpicSchema> & {
   LinkComponent?: LinkComponentType
   site: IsomerSiteProps
+  shouldLazyLoad?: boolean
 }
--- a/packages/components/src/interfaces/complex/Image.ts
+++ b/packages/components/src/interfaces/complex/Image.ts
@@ -61,4 +61,5 @@ export const ImageSchema = Type.Object(
 
 export type ImageProps = Static<typeof ImageSchema> & {
   site: IsomerSiteProps
+  shouldLazyLoad?: boolean
 }
--- a/packages/components/src/interfaces/complex/InfoCards.ts
+++ b/packages/components/src/interfaces/complex/InfoCards.ts
@@ -175,10 +175,12 @@ export type SingleCardWithImageProps = Static<
   layout: IsomerPageLayoutType
   isExternalLink?: boolean
   LinkComponent?: LinkComponentType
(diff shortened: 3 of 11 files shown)


seaerchin (commented):
what happens if there's a single small image?

seaerchin:
based off what i've found [here](https://webperf-snippets.nucliweb.net/Loading/Find-Above-The-Fold-Lazy-Loaded-Images), we should try to find viewport height then greedily load as much as we can

adriangohjw:
> what happens if there's a few small images above the fold?

based on current component design, the subsequent images are very likely below the fold. Even if they can above the fold, they likely won't be as large as the first image, which is the LCP (Largest Contentful Paint) and what lighthouse uses

adriangohjw:
> based off what i've found [here](https://webperf-snippets.nucliweb.net/Loading/Find-Above-The-Fold-Lazy-Loaded-Images), we should try to find viewport height then greedily load as much as we can

i did a brief look and it seems like this only tell us which is lazily loaded but does not change the loading behaviour (since the content is already loaded)

if im not wrong, this will be useful for cases where we know how the page will look like e.g. our own marketing pages, but in our context as a CMS where the content is dynamically generated, i don't think we will be able to determine that beforehand? 

correct me if im wrong!

seaerchin (approved):
lgtm, i was thinking of a flow where:
1. we lazy load all images
2. at run time, `useEffect` to check viewport and change to eager load everything above fold **but**

i think this is v minor because our current designs stack images vertically (eg: Homepage) so this improvement is marginal

adriangohjw:
@seaerchin don't think that's possible because `useEffect` happens after the initial render after the components have rendered. Any changes to image loading strategies would occur post-render, which is too late

However, we can potentially explore doing this in the future during build time, tho don't think its worth the bandwidth to explore now IMO

PR #987, opened 2025-01-13
isom-1719 non-prod sites can be indexed by search engine

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1719/test-sites-can-be-indexed-by-search-engine

while staging sites have `disallow` in `robots.txt`, according to [google](https://developers.google.com/search/docs/crawling-indexing/robots/intro)
> A page that's disallowed in robots.txt can still be indexed if linked to from other sites.
While Google won't crawl or index the content blocked by a robots.txt file, we might still find and index a disallowed URL if it is linked from other places on the web. As a result, the URL address and, potentially, other publicly available information such as anchor text in links to the page can still appear in Google search results. To properly prevent your URL from appearing in Google search results, [password-protect the files on your server](https://developers.google.com/search/docs/crawling-indexing/control-what-you-share), [use the noindex meta tag or response header](https://developers.google.com/search/docs/crawling-indexing/block-indexing), or remove the page entirely.

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- add `shouldBlockIndexing` based on site's environment
- pass in true/false into meta.noIndex for `/page.tsx` based on `shouldBlockIndexing`

## Tests

trigger a new codebuild, and inspect the HTML. It should/should not have `<meta name="robots" content="noindex">` based on the table below

| type of site | should have noindex in HTML |
| - | - |
| PROD site | no |
| PROD staging site | yes |
| STG/UAT site | yes |

### DIFF
--- a/packages/components/src/engine/metadata.ts
+++ b/packages/components/src/engine/metadata.ts
@@ -111,6 +111,12 @@ export const getMetadata = (props: IsomerPageSchemaType) => {
   }
 }
 
+export const shouldBlockIndexing = (
+  environment: IsomerPageSchemaType["site"]["environment"],
+): boolean => {
+  return environment === "staging"
+}
+
 export const getRobotsTxt = (props: IsomerPageSchemaType) => {
   const rules = [
     {
--- a/packages/components/src/engine/metadata.ts
+++ b/packages/components/src/engine/metadata.ts
@@ -114,7 +114,7 @@ export const getMetadata = (props: IsomerPageSchemaType) => {
 export const shouldBlockIndexing = (
   environment: IsomerPageSchemaType["site"]["environment"],
 ): boolean => {
-  return environment === "staging"
+  return environment !== "production"
 }
 
 export const getRobotsTxt = (props: IsomerPageSchemaType) => {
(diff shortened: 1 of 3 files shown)


packages/components/src/engine/metadata.ts
```diff
@@ -111,6 +111,12 @@ export const getMetadata = (props: IsomerPageSchemaType) => {
   }
 }
 
+export const shouldBlockIndexing = (
+  environment: IsomerPageSchemaType["site"]["environment"],
+): boolean => {
+  return environment === "staging"
+}
+
```
seaerchin, line 119: shuold we make this more restrictive? we might hvae environments that are non-production which shouldn't be indexed also like `vapt` or `uat`

seaerchin (approved):

packages/components/src/engine/metadata.ts
adriangohjw: right now we only support `production` / `staging` (generated from our infra)

image

we can make it more generic to return true if it **DOES NOT** equate to `production` since we don't know how many non-production sites are there (could have more than staging, vapt, uat in the future too)

WDYT?

seaerchin: yea sgtm!

adriangohjw: updated in https://github.com/opengovsg/isomer/pull/987/commits/e810697509aba6b7547bf72e7408d3f126de1102, thanks!

PR #988, opened 2025-01-14
chore: make publish toast message dismissable

dcshzj:
## Problem

We have a migrator that complained that the publish toast message is blocking the search bar, and valuable productive time is lost waiting for the toast message to disappear.

## Solution

**Breaking Changes**

**Improvements**:

- Make the publish toast message dismissable.

## Screenshots

https://github.com/user-attachments/assets/8d4860e7-0690-47a6-956e-2b6a04374489


adriangohjw (approved):
should we make all toast closable then?

PR #989, opened 2025-01-14
fix: empty results message is at a wrong position in blog layout 

sehyunidaaa:
## Problem

image

Visual jank, the no results message should be center-aligned to the entire results area. Currently it's on the left column.

After:

Storybook screen capture for blog layout story:
image


adriangohjw (approved):
thank you!

PR #990, opened 2025-01-15
feat: enable raw JSON editor mode for collection links

dcshzj:
## Problem

The raw JSON editor mode is not available for collection links, which adds ops load to engineers when migrators need to change tags (e.g. for REACH).

## Solution

**Breaking Changes**

**Features**:

- Extended the existing raw JSON editor mode to the collection links editor. This involved:
    - Small refactor of ActivateRawJsonEditorMode.tsx to allow the consumer to determine the action to take after a successful activation (removed the dependency on the EditorStateDrawer context).
    - Refactored RawJsonEditorModeStateDrawer.tsx by extracting out the UI from the business logic, so that the UI can be reused by the LinkEditorDrawer.
    - Extended LinkEditorDrawer to handle two states - root and rawJsonEditor states. Also split the existing page state into a saved and preview version, so that the raw JSON editor can discard changes to revert back to the original saved state.

**Bug fixes**:

- The ref attribute in the link preview was hard-coded, so it could be misleading when users hover over the link. Have updated it to use the value provided by the user.

**Notes**:

- It is **NOT** possible to change the title using the raw JSON editor mode. This is not something I'm tackling now, since the title change needs to be done in a separate flow.
- It is also not possible to edit something on the link editor first, then go into the raw JSON editor mode and see those changes reflected. I was thinking if we want this behaviour, then we may have to handle 3 page states which makes this feature overly complicated (since it is only used by migrators), so opted to keep things simple.

## Before & After Screenshots

https://github.com/user-attachments/assets/6265aec3-316b-42ce-80f6-cf04492b79f9

## Tests

- [ ] Navigate to any collection link in a collection inside Studio.
- [ ] Press the special sequence to activate the raw JSON editor mode.
- [ ] Make changes to the category and/or year of the collection link item and verify that the changes are reflected on the preview on the right (provided the schema is valid).
- [ ] Save the changes made, verify that you should be able to save.
- [ ] Refresh the page, verify that the changes you made persist.

### DIFF
--- a/apps/studio/src/features/editing-experience/components/RawJsonEditor.tsx
+++ b/apps/studio/src/features/editing-experience/components/RawJsonEditor.tsx
@@ -1,0 +1,121 @@
+import {
+  Box,
+  Button,
+  Heading,
+  HStack,
+  Icon,
+  IconButton,
+  Spacer,
+  Textarea,
+  useClipboard,
+  useDisclosure,
+} from "@chakra-ui/react"
+import { BiDollar, BiX } from "react-icons/bi"
+
+import { DiscardChangesModal } from "./DiscardChangesModal"
+
+interface RawJsonEditorProps {
+  pendingChanges: string
+  isModified: boolean
+  isLoading: boolean
+  isPendingChangesValid: boolean
+  handleChange: (value: string) => void
+  handleDiscardChanges: () => void
+  handleSaveChanges: () => void
+}
+
+export const RawJsonEditor = ({
+  pendingChanges,
+  isModified,
+  isLoading,
+  isPendingChangesValid,
+  handleChange,
+  handleDiscardChanges,
+  handleSaveChanges,
+}: RawJsonEditorProps) => {
+  const {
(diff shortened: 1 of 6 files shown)


apps/studio/src/features/editing-experience/components/RawJsonEditor.tsx
```diff
@@ -0,0 +1,121 @@
+import {
```
harishv7, line 1: Can help explain why is editing the title not possible?

harishv7 (approved):
looks fine, tested locally on collection links

apps/studio/src/features/editing-experience/components/RawJsonEditor.tsx
dcshzj: The title right now is being stored as a separate column in the Resources table, whereas the content is taken from the Blob table, so editing the title directly on the content right now does not propagate the change back to the Resources table.

If it's a more common request from the migrators then maybe we can add this functionality in the future.

PR #991, opened 2025-01-15
fix(create-site): update scripts for creating site + publishing

seaerchin:
**NOTE: do not merge till release as this will go into effect IMMEDIATELY**

## Problem
https://opengovproducts.slack.com/archives/C06R4DX966P/p1736928912803439

1. studio and components use different properties for `siteName` - which in turns impact the footer
2. `createSite.ts` has a default `siteName` set, even though the script already gives a `siteName`

## Solution
1. remove the spread in property - the `config` object already has `siteName` 
2. use the passed in `siteName` for `createSite`

## Tests
1. create a site locally using `createSite.ts` 
2. verify that the site has a correct `siteName` and not the default of MTI 
3. point a **staging** site to this branch
4. update teh **staging** db so that the site's `site.name` and `site.config.siteName` differ
5. publish a change 
6. verify that the end site uses `site.config.siteName`

### DIFF
--- a/apps/studio/src/server/modules/site/__tests__/site.router.test.ts
+++ b/apps/studio/src/server/modules/site/__tests__/site.router.test.ts
@@ -65,7 +65,6 @@ describe("site.router", async () => {
       expect(result).toEqual([
         {
           id: site.id,
-          name: site.name,
           config: site.config,
         },
       ])
@@ -87,7 +86,6 @@ describe("site.router", async () => {
       expect(result).toEqual([
         {
           id: site1.id,
-          name: site1.name,
           config: site1.config,
         },
       ])
@@ -123,7 +121,6 @@ describe("site.router", async () => {
       expect(result).toEqual([
         {
           id: site1.id,
-          name: site1.name,
           config: site1.config,
         },
       ])
--- a/apps/studio/src/server/modules/site/site.router.ts
+++ b/apps/studio/src/server/modules/site/site.router.ts
@@ -42,11 +42,13 @@ export const siteRouter = router({
         action: "read",
       })
 
-      return db
+      const { config } = await db
         .selectFrom("Site")
         .where("Site.id", "=", siteId)
-        .select("name")
+        .select("config")
         .executeTakeFirstOrThrow()
+
+      return { name: config.siteName }
     }),
   getConfig: protectedProcedure
     .input(getConfigSchema)
--- a/apps/studio/tests/integration/helpers/seed/index.ts
+++ b/apps/studio/tests/integration/helpers/seed/index.ts
@@ -49,16 +49,18 @@ export const setupSite = async (siteId?: number, fetch?: boolean) => {
       return { site, navbar, footer }
     })
   }
+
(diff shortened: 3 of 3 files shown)


tooling/build/scripts/publishing/index.ts
```diff
@@ -485,7 +485,6 @@ async function fetchAndWriteSiteData(client: Client) {
       const config = {
         site: {
           ...configResult.rows[0].config,
-          siteName: configResult.rows[0].name,
```
dcshzj, line 488: issue: I recall we did this so that we store the site name as a separate column inside the DB rather than inside the JSON column, as it is used under the sites dashboard. We couldn't remove the `siteName` prop inside the JSON blob because then the types would not match.

I think we should still stick to storing the site name in the separate column and have this set to say "Ignore this" (until we have a good solution for the typing mismatch).

seaerchin: > issue: I recall we did this so that we store the site name as a separate column inside the DB rather than inside the JSON column, as it is used under the sites dashboard. We couldn't remove the siteName prop inside the JSON blob because then the types would not match.

can clarify this? not sure what the issue here is apart from the types - if it's just about the name that presents on the dashboard, we can also chagne it to read from `site.config.siteName`, which i thought is better because that's a "user level view" whereas the db one (`site.name`) is not tied under `site.config`

seaerchin: ok then i'll just make the change to read directly from `site.config.siteName` HAHA

seaerchin: done, see video here. re: performance impact - it's close to 0 because we already read it at present lol

https://github.com/user-attachments/assets/0228f4af-a657-443d-bc06-27500bb41927

seaerchin: tried this! 

**on this branch**
image

the slow part is in fetching all the site's logos - the call itself is relatively fast (73kb though) and resolves in 177ms
image

**on `main`**
image

the call itself (same size - we still return `site.name` + `site.config`)
image

at 73kb, for this to take >1s, the connection is slower than 1mbps so i don't really think it's a concern. (3g download speed is 1.1mbps onwards)

but a good point to note is that we should probably have pagination way before this. just based off what we observe here, i think we should have it at around 100-200 sites onboarded (assuming it scales linearly)

dcshzj: Nice, I think the results looks promising, agreed we can add the pagination later (anyway it should only affect Isomer admins since we have access to more sites than usual, so low priority)

Should we go ahead and drop the Name column from the Sites table completely? Or are you intending to do that in a separate PR?

seaerchin: not intending to drop - i think we should preserve it just in case our internal names differ from agency's preferred name (eg: we want to do `<agency> - <site> - <version>`)

dcshzj: Sounds good yup

dcshzj (approved):

PR #993, opened 2025-01-15
fix(ISOM-1694): refactor image uploading

seaerchin:
## Problem
1. image uploading is lazy atm, which greatly complicates logic
2. additionally, we need to wait for malware scan to complete before we can display the image (see ticket for more details)

Closes ISOM-1694
Closes ISOM-1713
Closes ISOM-1711

## Solution
1. Keep refetching the url from s3 until it succeeds so that we know we can show the image. We `console.error` if it still fails after 3 tries so that it's easier to debug - technically this is production but i think this is ok to do for now since it's not user facing. Ideally, we should ping to a dd endpoint tgt w user id so we can browse rum replay

## Videos

https://github.com/user-attachments/assets/a491e4d5-6af1-4923-a549-298daded7f52

## Tests 
- upload an image 
- delete the image **immediately** without saving/exiting hte modal
- upload another image
- the newly uploaded image should display on preview

## Notes
1. this also fixes the image not available after upload bug


seaerchin:
this doesn't quite fix it i realised - it stil uploads a blob url so we'd have to go with the other way, which is to allow backend to read from s3 bucket directly

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/constants.ts
```diff
@@ -8,6 +8,16 @@ export const IMAGE_UPLOAD_ACCEPTED_MIME_TYPES = [
   "image/bmp",
   "image/webp",
 ]
```
dcshzj, line 10: nit: (outside of the diff) I think `IMAGE_UPLOAD_ACCEPTED_MIME_TYPES` is no longer used?

packages/components/src/hooks/useImage.ts
```diff
@@ -0,0 +1,91 @@
...
+  fallback,
+  retries = 3,
+  baseTimeoutMs = 500,
+}: UseImageProps) => {
+  const [isLoading, setIsLoading] = useState(false)
+  const [isError, setIsError] = useState(false)
+  const [isSuccess, setIsSuccess] = useState(false)
+
+  // NOTE: We require the `useEffect` because `fetch` is a browser API
+  // and it won't exist at compile time for NextJS
```
dcshzj, line 55: question(non-blocking): is this comment still true? Can't seem to find any usage of `useEffect` here.

packages/components/src/hooks/useImage.ts
```diff
@@ -0,0 +1,91 @@
...
+  promise: () => Promise<string>,
+  baseTimeoutMs: number,
+  maxRetries: number,
+) => {
+  const retryWithBackoff = async (retries: number) => {
+    try {
+      // Make sure we don't wait on the first attempt
+      if (retries > 0) {
+        const timeToWait = baseTimeoutMs * 2 ** retries
+        console.log(`waiting for ${timeToWait}ms...`)
```
dcshzj, line 19: issue(non-blocking): hmm this console.log will appear on the user's browser (since this is client-side), I think we should remove?

dcshzj (approved):

apps/studio/src/components/ImageUpload.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Button, FormLabel } from "@opengovsg/design-system-react"
+
+interface ImageUploadProps {
+  data?: string
+  required?: boolean
+  label: string
+  onClick: () => void
+  description?: string
+}
+export const ImageUpload = ({
```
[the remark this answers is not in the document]
seaerchin: hmm - `ImageUploadInfobox`?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/constants.ts
seaerchin: good catch, deleted!

packages/components/src/hooks/useImage.ts
seaerchin: ah ps, deleted

packages/components/src/hooks/useImage.ts
seaerchin: sounds good

PR #994, opened 2025-01-16
Inverse hero banner variant for sites with yellow as a colour theme

sehyunidaaa:
## Problem

The template can't support extremely vibrant colours like Yellow. Some default variants will fail contrast with a yellow. Unlike other colours, yellow loses its unique vibrance when you lower its lightness. 

## Solution

- The Hero banner (block variant) should have a variant that support inverse.
- This variant uses a different mapping of colour tokens.
- This variant should NOT be exposed to the user on Studio. It is set by us only for sites that have a high luminosity theme colour. 

[More details on Notion
](https://www.notion.so/opengov/Mini-Inverse-hero-banner-variant-for-LiveOn-1af77dbba788803ba2a7dcf4ea782b25?pvs=4)

image


sehyunidaaa:
FYI - this feature might not be needed anymore because LiveOn doesn't want yellow all of a sudden. Will close PR if this is the case (branch will still be alive for future use)

seaerchin (approved):

PR #997, opened 2025-01-17
fix: update title to be from folder title

seaerchin:
## Problem
we want to update index page title to be from the folder title and not auto gen

## Solution
use the folder title

## Tests
1. go to staging and update a site's build branch to be `fix/custom-index-pages`
2. update a title for folder to be different from permalink
3. start a new build
4. check the index page 
5. it should be different


dcshzj:
Replaced with #1002 as the solution is different.

PR #999, opened 2025-01-18
Enhancement/add user email in studio

adriangohjw:
## Problem

Feedback from agencies about not knowing which email they are using

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- add `me.email` to the dropdown
- refactor to be reusable between different navbars
- add stories

## Before & After Screenshots

image

## Tests

1. click on avatar menu in both Sites Management page `/` AND resource management page `/sites/XXX`
2. it should show your email - no behaviour on hover + non-clickable


apps/studio/src/components/AvatarMenu.tsx
```diff
@@ -0,0 +1,35 @@
+import { Divider, Flex, Text } from "@chakra-ui/react"
+import {
+  AvatarMenu as AvatarMenuComponent,
```
dcshzj, line 3: nit: Can name this as `OgpAvatarMenu` since technically this file is also a component.

apps/studio/src/components/AvatarMenu.tsx
```diff
@@ -0,0 +1,35 @@
...
+      >
+        <Flex alignItems="center" gap="0.75rem">
+          <BiUser size="1.25rem" color="#666c7a" />
+          <Text textStyle="subhead-1" textColor="base.content.medium">
+            {me.email}
+          </Text>
+        </Flex>
+      </Menu.Item>
+      <Divider orientation="horizontal" />
+      <Menu.Item onClick={() => logout()}>Sign out</Menu.Item>
```
dcshzj, line 32: suggestion(design): will leave to design to comment on this one but was thinking that since we have a logo for the email, I thought would be good to have a logout logo here as well, so that the text can be aligned. It would look something like this:

image

adriangohjw: good suggestion! will check with Sehyun

adriangohjw: added in https://github.com/opengovsg/isomer/pull/999/commits/60805b2fd7ca3280df91cd3ac86c72ca8d7ae7e0

apps/studio/src/components/AvatarMenu.tsx
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/999/commits/9182d9528711b32625c17ba909bb934a621161be

dcshzj (approved):

PR #1000, opened 2025-01-20
fix: allow editing of published collection links

dcshzj:
## Problem

Published collection links do not have a draftBlobId, which causes an error in saving published collection links.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Use the correct function to update the blob (updateBlobById), as that function handles creating the draftBlobId if it is not present.

## Tests

- [ ] Create a collection link item within a collection.
- [ ] Save changes to the collection link item and publish, make sure the publish button is greyed out and the publish is successful.
- [ ] Make another change to the collection link.
- [ ] Verify that the success toast message appears with the text "Link updated"


seaerchin (approved):

harishv7:
PR #1000 🎉

PR #1001, opened 2025-01-20
Chore/allow collections move

seaerchin:
## Problem
previously we disallowed users from shifting items inside collections. this pr allows shifting of collection items between collections. 
1. this means that collection items can only be moved between collections and **not** into folders 
2. similarly, folder items can only be moved between folders and not into collections

## Solution
1. fetch data using `getChildrenOf` and filter
2. show the menu tooltip on collection items
3. this is a **frontend only** change - the backend method will still allow users if they `post` directly and are verified 

## Tests
1. create 2 collections
2. create a folder
5. create pages inside the collections + the folder
6. create links inside the collections 
7. for the page inside the folder, verify that you cannot move it to another collection
8. for the pages inside the collection + the link, verify that you can move them between collections
9. verify that the pages + link **once moved**, doesnt show inside the original collectino both on the dashboard and the sidebar
10. verify that the pages + link **once moved**, shows inside the new collection on the sidebar + the dashboard


apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
```diff
@@ -59,7 +60,7 @@ const SuspendableMoveItem = ({
       pl="2.25rem"
       size="xs"
       onClick={handleOnClick}
-      leftIcon={<BiFolder />}
+      leftIcon={type === "Collection" ? <BiData /> : <BiFolder />}
```
dcshzj, line 63: nit: I think I've seen this if-else a couple of times, might be good to standardise to use a common function that determines the icon to use

For example, `ICON_MAPPINGS` inside `apps/studio/src/features/dashboard/components/DirectorySidebar/constants.ts` has this mapping for all the resources currently.

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -71,7 +72,7 @@ const MoveResourceContent = withSuspense(
     const parentDest = resourceStack[resourceStack.length - 2]
     const curResourceId = moveDest?.resourceId
     const { data, fetchNextPage, hasNextPage, isFetchingNextPage } =
-      trpc.resource.getFolderChildrenOf.useInfiniteQuery(
+      trpc.resource.getChildrenOf.useInfiniteQuery(
```
dcshzj, line 75: issue(perf): I think we should still continue using `getFolderChildrenOf`, otherwise it seems a bit heavy for the client browser to perform the filtering on the entire list of resources, rather than letting the backend perform an initial filtering to folder-like resources.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -265,7 +268,8 @@ export const resourceRouter = router({
               // NOTE: we only allow moves to folders/root.
               // for moves to root, we only allow this for admin
               (parent.type !== ResourceType.RootPage &&
-                parent.type !== ResourceType.Folder)
+                parent.type !== ResourceType.Folder &&
+                parent.type !== ResourceType.Collection)
```
dcshzj, line 272: nit: might be good to add a comment about the frontend-side filtering, which makes moving items from Folder to Collection (and vice versa) not possible to a normal user.

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
seaerchin: sounds good!

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveItem.tsx
seaerchin: udpated

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: frontend side filtering removed, now added a backend side rejection + toast

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
seaerchin: went back to using `getFolderChildrenOf`

dcshzj (approved):

PR #1004, opened 2025-01-21
fix: properly retrieve nested folder resources

dcshzj:
## Problem

Nested folder resources are not properly retrieved, which causes the casing in the index page to not be correct for those cases.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Correctly retrieve the folder resource using the full permalink instead.

## Tests

- Repeat #1000 but for nested folders.

### DIFF
--- a/tooling/build/scripts/publishing/index.ts
+++ b/tooling/build/scripts/publishing/index.ts
@@ -251,10 +251,16 @@ function generateSitemapTree(
       const pageName = danglingDirectory.replace(/-/g, " ")
       const generatedTitle =
         pageName.charAt(0).toUpperCase() + pageName.slice(1)
-      const folderResourceTitle = resources.find(
-        (resource) => resource.fullPermalink === danglingDirectory,
-      )?.title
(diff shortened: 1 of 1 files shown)


seaerchin (approved):

seaerchin (commented):
just for my understanding - previously this didn't work because we rely on `fullPermalink === danglingDirectory`, which meant that nested folders wouldn't work

now, since we use a full permalink comparison, the `folder` itself would match the permalink, even if nested?

dcshzj:
@seaerchin yup, `danglingDirectory` is only the last part of the permalink (i.e. the slug, which is `senior-management` if the full permalink is `about-us/senior-management`). We can immediately tell this is true because the previous code does a replacement of dashes to spaces and capitalise the first letter to be the title, so means `danglingDirectory` is only the last part.

Now, it's doing a full comparison using the permalink (which we actually already are doing a few lines down, which I now moved up). This guarantees that we are getting the actual folder resource that we want, which we can then extract the title from.

PR #1006, opened 2025-01-22
Chore tweak dynamicdatabanner UI (mobile view)

adriangohjw:
## Problem

MUIS prayers banner on mobile is a bit cut off due to large min. size

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- adding mobile grid

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

1. Publish new build for staging site `test-test`
2. Use mobile view on dev console (iphone SE) to ensure its not cut off. Should look similar to the "After" and not "Before" screenshots above


seaerchin:
went to check the story - it didn't quite reflect what was happening on mobile even in the before view, which seemed counter intuitive 

how do we check that this is working

adriangohjw:
> went to check the story - it didn't quite reflect what was happening on mobile even in the before view, which seemed counter intuitive
> 
> how do we check that this is working

@seaerchin ooo im able to replicate both on storybook manually

| before | after |
| - | - |
| image | Screenshot 2025-01-22 at 4 43 20 PM |

but if you are referring to chromatic, then yeah it missed this out because we are using `mobile` which is `sm` and not showing error. It only shows error for `xs`. Lemme add `xs` for this ticket so moving forward there will be a check for regression too!

harishv7 (approved):
lgtm, can merge with chromatic review

PR #1018, opened 2025-01-23
Fix vica hydration bug

adriangohjw:
## Problem

VICA widget sometimes loading sometimes not

Error log shows https://react.dev/errors/418?args%5B%5D= 
image

Suspected it's due to it being rendered SSG and causing incocrrect hydration error

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

tested on STG site and works

- explicitly declare `use client`
- explicitly declare for it not to be rendered during SSG

## Tests

1. rebuild a site with VICA (https://test-isomer-vica-staging.isomer.gov.sg/)
2. try spamming reloading multiple times (soft reload, hard reload, hard reload with empty cache etc.) - the VICA widget should show up in the bottom right consistently **EVERY** time

### DIFF
--- a/packages/components/src/templates/next/components/internal/Vica/VicaWidget.tsx
+++ b/packages/components/src/templates/next/components/internal/Vica/VicaWidget.tsx
@@ -1,17 +1,5 @@
 import type { VicaProps } from "~/interfaces"
 
-export const VicaStylesheet = () => {
-  return (
-    <>
-      <link
-        href="https://webchat.vica.gov.sg/static/css/chat.css"
-        referrerPolicy="origin"
-        rel="stylesheet"
-      />
-    </>
-  )
-}
-
 export const VicaWidget = ({
   ScriptComponent = "script",
   ...props
--- a/packages/components/src/templates/next/components/internal/Vica/VicaWidget.tsx
+++ b/packages/components/src/templates/next/components/internal/Vica/VicaWidget.tsx
@@ -1,9 +1,14 @@
+"use client"
+
 import type { VicaProps } from "~/interfaces"
 
 export const VicaWidget = ({
   ScriptComponent = "script",
   ...props
 }: VicaProps) => {
+  // to not render during static site generation on the server
+  if (typeof window === "undefined") return null
+
   return (
     <>
       <ScriptComponent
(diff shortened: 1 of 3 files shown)


packages/components/src/templates/next/components/internal/Vica/VicaWidget.tsx
```diff
@@ -1,21 +1,14 @@
...
-  )
-}
+import type { VicaProps } from "~/interfaces"
 
 export const VicaWidget = ({
   ScriptComponent = "script",
   ...props
 }: VicaProps) => {
+  // to not render during static site generation on the server
+  if (typeof window === "undefined") return null
```
seaerchin, line 10: nit: prefer `!window`

seaerchin (approved):
lgtm! jsut to double check why this bug occurs and how this fixes: 

1. previously, we rendered both `VicaStylesheet` + `VicaWidget` on the server
2. when we sent it to the client side, `chat.js` altered some portions of the html? (hence causing the bug) 
3. how, we mark `Widget` as `use client` so that on the server, it doesn't do anything
4. on client side now, it's initially `null` (so matches server) 
5. fetches `VicaWidget` since `window` is defined (technically this is the rsc script which then re-renders) 
6. `VicaWidget` loads in properly

if this isnt true, please dismiss the review!

adriangohjw:
> lgtm! jsut to double check why this bug occurs and how this fixes:
> 
> 1. previously, we rendered both `VicaStylesheet` + `VicaWidget` on the server
> 2. when we sent it to the client side, `chat.js` altered some portions of the html? (hence causing the bug)
> 3. how, we mark `Widget` as `use client` so that on the server, it doesn't do anything
> 4. on client side now, it's initially `null` (so matches server)
> 5. fetches `VicaWidget` since `window` is defined (technically this is the rsc script which then re-renders)
> 6. `VicaWidget` loads in properly
> 
> if this isnt true, please dismiss the review!

im _not exactly very sure_ why this is happening because the integration doc for VICA seems to differ for each agency 😢 

my observation is that on staging site, VICA widgets actually rendered before disappearing.

my suspicion is that during SSG, VicaWidget was indeed rendered during SSG and modifed the DOM after the initial render, leading to differences between the server-rendered HTML and the client-side DOM during hydration.

PR #1033, opened 2025-01-24
feat: introduce temporary admin mode for site config

dcshzj:
## Problem

Ops load on engineers is quite high for editing the site config, navbar and footer.

## Solution

**Breaking Changes**

**Features**:

- Add a new `/admin` route for Isomer Admins to directly modify the site's config, theme, navbar and footer items.
    - Access to this feature is gated for just Isomer Admins, all other users will not know of its existence and even if they do, they will not have the permission to access the form to edit.

## Before & After Screenshots

**View for Isomer Admins**:

image

**View for unauthorised users**:

image

## Tests

- [ ] Log in to Studio as a whitelisted Isomer Admin (your email should be whitelisted inside Growthbook).
- [ ] Go to any site.
- [ ] Verify that a star icon appears on the navigation bar on the left side.
- [ ] Click on the star icon, verify that you are able to see the form to change the site configurations.
- [ ] Change any configuration and save, verify that you are able to save successfully.
- [ ] Navigate to any page and verify that those changes have persisted (or change another configuration that is more visible, such as the navigation bar or footer).
- [ ] Log in to Studio as a non-whitelisted user.
- [ ] Go to any site, verify that there are only 2 icons on the navigation bar on the left side.
- [ ] Append `/admin` to the URL and navigate to the page.
- [ ] Verify that the page shows that you do not have permission to access the page.

### DIFF
--- a/apps/studio/src/pages/sites/[siteId]/admin.tsx
+++ b/apps/studio/src/pages/sites/[siteId]/admin.tsx
@@ -25,7 +25,7 @@ import { useIsUserIsomerAdmin } from "~/hooks/useIsUserIsomerAdmin"
 import { useQueryParse } from "~/hooks/useQueryParse"
 import { useZodForm } from "~/lib/form"
 import { type NextPageWithLayout } from "~/lib/types"
-import { setSiteConfigAdminSchema } from "~/schemas/site"
+import { setSiteConfigByAdminSchema } from "~/schemas/site"
 import { AdminSidebarOnlyLayout } from "~/templates/layouts/AdminSidebarOnlyLayout"
 import { trpc } from "~/utils/trpc"
 
@@ -56,31 +56,6 @@ const SiteAdminPage: NextPageWithLayout = () => {
     void router.push(`/sites/${siteId}`)
   }
 
-  const { mutate, isLoading } = trpc.site.setSiteConfigAdmin.useMutation({
-    onSuccess: async () => {
-      // reset({ config, theme, navbar, footer })
-      await trpcUtils.site.getConfig.invalidate({ id: siteId })
-      await trpcUtils.site.getTheme.invalidate({ id: siteId })
-      await trpcUtils.site.getNavbar.invalidate({ id: siteId })
-      await trpcUtils.site.getFooter.invalidate({ id: siteId })
-      toast({
-        title: "Saved site config!",
-        description: "Check your site in 5-10 minutes to view it live.",
-        status: "success",
-        ...BRIEF_TOAST_SETTINGS,
-      })
-    },
-    onError: () => {
-      toast({
-        title: "Error saving site config!",
-        description:
-          "If this persists, please report this issue at support@isomer.gov.sg",
-        status: "error",
-        ...BRIEF_TOAST_SETTINGS,
-      })
-    },
-  })
-
   const [previousConfig] = trpc.site.getConfig.useSuspenseQuery({
     id: siteId,
   })
@@ -94,13 +69,15 @@ const SiteAdminPage: NextPageWithLayout = () => {
     id: siteId,
   })
 
-  // NOTE: Refining the setNotificationSchema here instead of in site.ts since omit does not work after refine
+  // NOTE: Refining the setSiteConfigByAdminSchema here instead of in site.ts since omit does not work after refine
   const {
+    reset,
+    watch,
     register,
     handleSubmit,
     formState: { isDirty, errors },
   } = useZodForm({
-    schema: setSiteConfigAdminSchema
+    schema: setSiteConfigByAdminSchema
       .omit({ siteId: true })
       .refine((data) => !!data.config, {
         message: "Site config must be present",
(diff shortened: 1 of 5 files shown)


seaerchin (dismissed):

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -0,0 +1,228 @@
...
+      }
+    }
+
+    if (!isOpen) {
+      router.events.on("routeChangeStart", handleRouteChange)
+    }
+    return () => {
+      router.events.off("routeChangeStart", handleRouteChange)
+    }
+  }, [isOpen, router.events, isDirty])
```
Comment (seaerchin) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx:140]: i'm not quiet sure what this `useEffect` does? if the modal is closed + there are changes, we unsub to `routeChangeStart` (so this iwll only ever fire once) -> and we set the next url

why wouldn't we use the `nextUrl` of the unsaved settings modal

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -0,0 +1,228 @@
...
+        minH="100%"
+      >
+        <Center py="5.5rem" px="2rem">
+          <VStack w="48rem" alignItems="flex-start" spacing="1.5rem">
+            <Text w="full" textStyle="h3-semibold">
+              Manage site configurations
+            </Text>
+            <Infobox variant="warning" textStyle="body-2" size="sm">
+              No validation is done on the JSON input. Please ensure that they
+              are valid before saving.
```
Comment (seaerchin) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx:177]: can we warn the migrators that these changes are **immediately** published? since this will be on prod, it might not be very evident to the migrators but it'l impact the site hugely (eg: we add a link on footer before the page (which is linked) is created)

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -0,0 +1,228 @@
...
+                  >
+                    Site {type}
+                  </Text>
+
+                  <Textarea
+                    fontFamily="monospace"
+                    boxSizing="border-box"
+                    minH="18rem"
+                    {...register(type)}
+                  />
```
Comment (seaerchin) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx:196]: not actionable: see if it's easy to format - i found [this](https://github.com/microlinkhq/react-json-view) which seems to be the one used by our `play.isomer.gov.sg`

seaerchin (commented):
minor issue - not too sure what the `useEffect` is doing! i think because this affects end user sites (and the site wide config), we might want to sanitise the json object to make sure it's valid (cos we're just parsing to make sure it's valid json, but not that the jsonobject matches what we expect - eg: `{siteNavItems: [...]}` vs `{siteNavItem: [...]}`

but no actionable re: the second point of matching - onlyi want to make sure the `useEffect` isn't sussy

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -0,0 +1,228 @@
...
+    })
+  })
+
+  if (!isUserIsomerAdmin) {
+    return (
+      <Center h="full">
+        <Text>You do not have permission to access this page.</Text>
+      </Center>
+    )
+  }
```
Comment (seaerchin) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx:155]: minor super nit: we should redirect + toast unauth users!

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (dcshzj) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx:149]: Hmm this was taken directly from the settings.tsx file, mainly to trigger the unsaved settings modal if the form is dirty and the user is attempting to change to a different route (e.g. clicking on the pages icon to go back to the site dashboard). This `useEffect` will prevent that.

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (dcshzj) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx:178]: Yup the warning is shown at the bottom next to the save button, same design as the site settings page.

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (dcshzj) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx:197]: Hmm I didn't intend to do formatting here because main workflow from migrators is that they will edit on VSCode first before pasting in here.

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (dcshzj) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: Switched in d549351e72d5802c0e65a91969627e67cbb0d080.

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -0,0 +1,229 @@
...
+      theme: JSON.stringify(previousTheme, null, 2),
+      navbar: JSON.stringify(previousNavbar.content, null, 2),
+      footer: JSON.stringify(previousFooter.content, null, 2),
+    },
+  })
+
+  const [nextUrl, setNextUrl] = useState("")
+  const isOpen = !!nextUrl
+
+  useEffect(() => {
```
Comment (seaerchin) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx:132]: ok, sounds good! this honestly look abit sus but given that it's used on `settings.tsx` , i think it's ok.

i think it might be worth to double-check the function is cleaning up properly; just based off what i see here, we turn off the events if `isDirty` then we `setNextUrl`,  but we only turn on the event when the modal is closed again? wondering if it's possible for our users to bypass the `!isOpen` and for `routeChangeStart` to be off permanently

based off what i know i don't think so, because any closing of the modal will trigger the state change, but just wanted to be sure

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -136,4 +144,51 @@ export const siteRouter = router({
 
       return input
     }),
+  setSiteConfigAdmin: protectedProcedure
```
seaerchin, line 147: nit: i think this might be better named as `setSiteConfigByAdmin` or `setAdminSiteConfig` but very minor

seaerchin (approved):
mostly lgtm - i think blind spots for me are the `useEffect`, which is actually pretty confusing

the router method can be updated but not a real blocker, given that it's narrowly used and easy to infer from context

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx:149]: ok, sounds good! this honestly look abit sus but given that it's used on settings.tsx , i think it's ok.

i think it might be worth to double-check the function is cleaning up properly; just based off what i see here, we turn off the events if isDirty then we setNextUrl, but we only turn on the event when the modal is closed again? wondering if it's possible for our users to bypass the !isOpen and for routeChangeStart to be off permanently

based off what i know i don't think so, because any closing of the modal will trigger the state change, but just wanted to be sure

Comment (dcshzj) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx:152]: Yeah you're right mainly in that as long as the modal is closed, the state change will be triggered so it will ensure that the router events listener is enabled correctly.

apps/studio/src/server/modules/site/site.router.ts
dcshzj: Updated in 31dcf08f43695cdc78c70983cff0df8d9a24e9dd.

PR #1046, opened 2025-01-27
fix(meta): show image preview

seaerchin:
## Problem
Fixes ISOM-1729
Right now on studio, we don't show the meta image on upload. this PR implements a simple preview for meta image. This also solves the bug of not allowing `.` in image names

## Solution
1. add a `meta` format inside our components. this is required because we use json forms to render our meta settings but we don't have a preview 
2. in our renderers, add a `meta` renderer that uses the retry strategy in the base branch. we render the preview right below (pending design comments omegalul)

## Video

https://github.com/user-attachments/assets/149d1718-3af7-48e1-bab6-e902700e0a93

## Tests
1. create a page
2. go to the meta settings
4. upload an image
5. the image should render after a while
6. delete the image 
7. upload another image
8. the new image should render


apps/studio/src/components/PageEditor/FileAttachment.tsx
```diff
@@ -7,7 +7,7 @@ import { useUploadAssetMutation } from "~/hooks/useUploadAssetMutation"
 import { getPresignedPutUrlSchema } from "~/schemas/asset"
 
 interface FileAttachmentProps {
-  setHref: (href?: string) => void
+  setHref: (href?: string, original?: File) => void
```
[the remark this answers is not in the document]
seaerchin: this `original` prop is never used! i'll remove it. the flow now is:
1. call `upload` 
2. upload to s3 with presigned url 
3. on studio frontend, keep retrying until the URL we get 200s 
4. set the `src` to the (now working) url

package-lock.json
```diff
@@ -34354,6 +34354,21 @@
     "tooling/typescript": {
       "name": "@isomer/tsconfig",
       "version": "0.0.0"
+    },
+    "tooling/template/node_modules/@next/swc-win32-ia32-msvc": {
```
harishv7, line 34358: what's this change btw?

seaerchin: idk, this seems to be an autogenerated change every time...

apps/studio/src/components/PageEditor/FileAttachment.tsx
```diff
@@ -82,7 +79,7 @@ export const FileAttachment = ({
         />
       </Skeleton>
       <Text textStyle="body-2" textColor="base.content.medium" pt="0.5rem">
-        {`Maximum file size: ${MAX_FILE_SIZE_BYTES / 1000000} MB`}
+        {`Maximum file size: ${maxSizeInBytes / 1000000} MB`}
```
harishv7, line 82: nit: can just make the 1000000 a constant `ONE_MB_IN_BYTES` since we have a couple of areas of this use

seaerchin: i did this but after searching through the codebase, i think this is the only use!

apps/studio/src/features/editing-experience/components/form-builder/hooks/useS3Image.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect, useState } from "react"
```
harishv7, line 1: for my understanding can i check when is useImage used over useS3Image? I believe all image uploads will now be using useS3Image right?

harishv7 (approved):
minor non blocking comments

apps/studio/src/features/editing-experience/components/form-builder/hooks/useS3Image.ts
```diff
@@ -0,0 +1,33 @@
...
+const assetsBaseUrl = `https://${env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME}`
+export const useS3Image = (imagePath: string) => {
+  const [image, setImage] = useState<File | undefined>()
+
+  useEffect(() => {
+    if (!imagePath) {
+      setImage(undefined)
+      return
+    }
+    void getImageAsFile(`${assetsBaseUrl}${imagePath}`)
```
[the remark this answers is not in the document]
seaerchin: i think i mistakenly added it for the linter, i removed this!

apps/studio/src/features/editing-experience/components/form-builder/hooks/useS3Image.ts
seaerchin: `useImage` is used to handle uploading to s3 with the retries - this is required so that we don't run into the malware scan not being completed and get a 403.

`useS3Image` is when the image has already successfully uploaded and we want to download + show the image using the `FileAttachment` (as it requires a `file` but we only have a relative path `/<site_id>/<hash>/<image>`) 

i think i'll update `useImage` to `useImageUpload` for clarity in this case

PR #1048, opened 2025-01-27
feat: dropdown for collections

seaerchin:
## Problem
Add dropdown for collections, see ticket for deets

## Solution
1. use a db query to get all the category values - this takes the draft blob if it is there, otherwise it orders by the publishedAt and takes the latest one 
2. render this on the frontend as a dropdown of the values
3. Add a `Suspense` + `Skeleton` so there's a nice loading state

## Screenshots
image

## Tests 
- [ ] Go to site 7 on `staging`
- [ ] Create a collection 
- [ ] Create a collection page 
    - [ ] edit the db so that this page's category is updated
- [ ] Create another collection page 
- [ ] Check the dropdown 
- [ ] it should have `Feature Articles` (default value) + the previous pag e
- [ ] Publish this page 
- [ ] The dropdown should still have the same values
- [ ] Make some change so that there's a draft blob id 
- [ ] update the category for this draft blob to some other value (now, published blob + draft blob has diff values)
- [ ] Click on dropdown again 
- [ ] it should have the new values

### DIFF
--- a/apps/studio/src/server/modules/page/page.router.ts
+++ b/apps/studio/src/server/modules/page/page.router.ts
@@ -141,8 +141,9 @@ export const pageRouter = router({
                 "in",
                 eb
                   .selectFrom("Blob")
-                  .select("id")
-                  .where("Version.resourceId", "=", eb.ref("Resource.id"))
+                  .select("Blob.id")
+                  .leftJoin("Version", "Version.blobId", "Blob.id")
+                  .where("Version.blobId", "=", eb.ref("Blob.id"))
                   .orderBy("publishedAt desc")
                   .limit(1),
               ),
@@ -155,17 +156,20 @@ export const pageRouter = router({
         // because kysely doesn't know that certain properties only exist
         // for certain `Resources` (as we are selecting from the `Blob` table)
         .select((eb) => eb.ref("content", "->").key("page").as("page"))
-        .distinct()
         .execute()
 
+      const duplicatedCategories = pages
+        .filter(({ page }) => {
+          return !!(page as { category?: string }).category
+        })
+        .map(({ page }) => {
+          return (page as unknown as { category: string }).category
+        })
+
+      const categories = Array.from(new Set<string>(duplicatedCategories))
+
       return {
-        categories: pages
-          .filter(({ page }) => {
-            return !!(page as { category?: string }).category
-          })
-          .map(({ page }) => {
-            return (page as unknown as { category: string }).category
-          }),
+        categories,
       }
     }),
 
--- a/apps/studio/src/server/modules/page/page.router.ts
+++ b/apps/studio/src/server/modules/page/page.router.ts
@@ -119,57 +119,29 @@ export const pageRouter = router({
         .executeTakeFirstOrThrow()
 
       const blobs = await db
-        .with("draftBlobs", (db) => {
-          return db
-            .selectFrom("Resource")
-            .leftJoin("Blob", "Resource.draftBlobId", "Blob.id")
-            .select([
-              "Resource.id as resourceId",
-              "Resource.draftBlobId as draftBlobId",
-              sql<string>`"Blob"."content"->'page'->>'category'`.as(
-                "draft_category",
-              ),
-            ])
-            .where("Resource.parentId", "=", String(parentId))
-        })
-        .with("publishedBlobs", (db) => {
-          return db
-            .selectFrom("draftBlobs")
-            .where("draftBlobId", "is", null)
-            .innerJoin("Version", "Version.resourceId", "draftBlobs.resourceId")
-            .innerJoin("Blob", "Version.blobId", "Blob.id")
-            .select([
-              "draftBlobs.resourceId as resourceId",
-              sql<string>`"Blob"."content"->'page'->>'category'`.as(
-                "published_category",
-              ),
-            ])
-            .orderBy("Version.versionNum", "desc")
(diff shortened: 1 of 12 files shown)


apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -99,6 +99,80 @@ export const pageRouter = router({
         .execute()
     }),
 
+  getCategories: protectedProcedure
```
adriangohjw, line 102: to replace this nested inner join for better performance and more suitable for production use
https://github.com/opengovsg/isomer/pull/1056

adriangohjw (changes requested):

apps/studio/src/server/modules/page/page.router.ts
seaerchin: actually good call! benched this against local (10k rows) - took ~10s. lemme take a stab at refactoring this first before i use 1056!

seaerchin: updated the query - ran against MOH newsroom, about ~300ms consistently! local (10k rows) took around 80ms - i think the difference is on the initial select, lemme see what i can do

image

adriangohjw (changes requested):
Right now, this is only being used for the collection's category, but I feel like it could be generalized for more use cases since the only real difference is the query we're using. Plus, the query logic is in the form renderer, which ideally should be only UI-focused.

That said, I think this simple approach works fine for now - it’s not over-engineered, and we can always refactor it later, like we’re doing with `JsonFormsProseControl`

Not blocking though - just wanted to call out the trade-offs we’re making here

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -99,6 +99,55 @@ export const pageRouter = router({
...
+        .leftJoin("Version as v", "r.publishedVersionId", "v.id")
+        .leftJoin("Blob as vb", "v.blobId", "vb.id")
+        .where("r.type", "in", [
+          ResourceType.CollectionPage,
+          ResourceType.CollectionLink,
+        ])
+        .select((eb) => {
+          return eb.fn
+            .coalesce(
+              sql<string>`b.content->'page'->>'category'`,
```
harishv7, line 133: for future: might wanna add indexes on `content->'page'->>'category'`

package-lock.json
```diff
@@ -33932,6 +33932,21 @@
     "tooling/typescript": {
       "name": "@isomer/tsconfig",
       "version": "0.0.0"
+    },
+    "tooling/template/node_modules/@next/swc-win32-ia32-msvc": {
```
harishv7, line 33936: whats this change

harishv7 (approved):

package-lock.json
seaerchin: i'm not very sure - this was something auto installed/updated when i did `npm i`.

based off what i googled (nextjs issue [here](https://github.com/vercel/next.js/discussions/61407)), this appears to be a nextjs binary for win32 (specifically for swc, a js runtime). 

i can't be sure about this, but i suspect this (swc) is used under the hood by next and this got updated as part of deps

apps/studio/src/server/modules/page/page.router.ts
seaerchin: i can do this in a follow-up, no worries

package-lock.json
adriangohjw: yeah can confirm - this is what's autoinstalled during npm i, it's also stated in the console

apps/studio/src/server/modules/page/page.router.ts
adriangohjw: addon - potentially wanna create a autogenerated column on this and index this autogen column -> easier to reference and search for it in the codebase than calling `content->'page'->>'category'` + easier to identify it when its unused thus safer to remove

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -119,57 +119,29 @@ export const pageRouter = router({
...
-                .coalesce("draft_category", "published_category")
-                .as("category")
-            })
+            .as("category")
         })
-        .selectFrom("coalescedBlobs")
-        .select("coalescedBlobs.category")
         .distinct()
-        .where("coalescedBlobs.category", "is not", null)
+        .where("r.parentId", "=", String(parentId))
```
adriangohjw, line 139: nit - might wanna mover this up together with the other `where` statement to improve readability

package-lock.json
harishv7: sounds good thanks for confirming

harishv7:
Adding the sql here for posterity

```
SELECT DISTINCT 
    COALESCE(
        b.content->'page'->>'category', -- prefer draftBlob category if available
        vb.content->'page'->>'category' -- if not, we fallback to publishedVersion category
    ) AS category
FROM "Resource" r
LEFT JOIN "Blob" b ON r."draftBlobId" = b.id
LEFT JOIN "Version" v ON r."publishedVersionId" = v.id
LEFT JOIN "Blob" vb ON v."blobId" = vb.id
WHERE r.type = 'CollectionPage' AND r."siteId" = 36;

```

seaerchin:
### Merge activity

* **Feb 2, 11:59 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1048).

PR #1056, opened 2025-01-28
Feat/dropdown collections refactor queries

adriangohjw:
refactor to 

1. replace nested joins with CTE. simple benchmarking 
  - 60 resources -> ~50ms
  - 7.5k resources -> ~250ms 

2. put category filtering logic from application layer to database layer for speed and improve readability 

3. use CTE for improved readability too

### DIFF
--- a/apps/studio/src/server/modules/page/page.router.ts
+++ b/apps/studio/src/server/modules/page/page.router.ts
@@ -118,79 +118,70 @@ export const pageRouter = router({
         .select("parentId")
         .executeTakeFirstOrThrow()
 
-      const { categories } = await db
-        .selectFrom(
-          sql<{ categories: string[] }>`
-            WITH
-              -- Step 1: Get all of the latest published versions
-              _latest_published_versions AS (
-                SELECT
-                  "resourceId",
-                  "blobId"
-                FROM
-                  (
-                    SELECT
-                      "resourceId",
-                      "blobId",
-                      ROW_NUMBER() OVER (
-                        PARTITION BY
-                          "resourceId"
-                        ORDER BY
-                          "versionNum" DESC
-                      ) AS row_num
-                    FROM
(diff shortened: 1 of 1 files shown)


adriangohjw:
@seaerchin please test this first! it works okay for me when running as raw sql on Tableplus

adriangohjw:
updated to use a kysely-ish syntax

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -118,59 +118,70 @@ export const pageRouter = router({
...
-                "Blob.id",
-                "in",
-                eb
-                  .selectFrom("Blob")
-                  .select("Blob.id")
-                  .leftJoin("Version", "Version.blobId", "Blob.id")
-                  .where("Version.blobId", "=", eb.ref("Blob.id"))
-                  .orderBy("publishedAt desc")
-                  .limit(1),
+      const result = await db
```
harishv7, line 121: Is this to get the unique categories of a collection?

Would something like this suffice?

```
SELECT DISTINCT 
    COALESCE(
        b.content->'page'->>'category', -- prefer draftBlob category if available
        vb.content->'page'->>'category' -- if not, we fallback to publishedVersion category
    ) AS category
FROM "Resource" r
LEFT JOIN "Blob" b ON r."draftBlobId" = b.id
LEFT JOIN "Version" v ON r."publishedVersionId" = v.id
LEFT JOIN "Blob" vb ON v."blobId" = vb.id
WHERE r.type = 'CollectionPage' AND r."siteId" = 36;

```

adriangohjw:
closing as there are cleaner solutions too
- https://github.com/opengovsg/isomer/pull/1056#discussion_r1936716756
- https://github.com/opengovsg/isomer/pull/1048#discussion_r1936712969

PR #1065, opened 2025-01-31
Feat/admin create index page temp workabout

adriangohjw:
## Problem

Currently, migrators depend on isoengineers to create the index page, which consumes engineering bandwidth and leads to costly context switching.

Since the feature for agency users to customize the index page is unlikely to be released soon, this serves as a temporary workaround to reduce the engineering load.

**Note: This is an admin-only interim solution until the customizable index page is fully developed.**

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add a "Add index page" button for folders and collections for isomer admins

## Before & After Screenshots

https://github.com/user-attachments/assets/52917c57-3288-47c9-95d1-7ca310d8b57e

### Non Isomer admins should not see the button

1. login with a non-isomer admin account e.g. `adriangoh+1@open.gov.sg` 
2. go inside a folder -> you should not see the new button
3. go inside a collection -> you should not see the new button

### Isomer admins should see the button

1. login with a isomer-admin account e.g. `adriangoh@open.gov.sg`
2. go inside a folder without an index page
    - [ ] you should see the button
    - [ ] clicking on the button should create an index page
    - [ ] opening the index page in json editor mode should show correct values
    - [ ] button should be disabled after the index page is created
    - [ ] button should still be disabled after reloading the page
3. repeat step 2 for collection


apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -534,4 +541,97 @@ export const pageRouter = router({
       }
       return permalinkTree
     }),
+
+  createIndexPage: protectedProcedure
```
adriangohjw, line 545: note - this mimics how on-call isoengineer currently add an index page manually

apps/studio/src/features/permissions/hooks/useIsIsomerAdmin.ts
seaerchin, line 1: i think we have an existing hook for this already - `useIsUserIsomerAdmin`

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -572,4 +574,23 @@ export const resourceRouter = router({
         (a, b) => resourceIds.indexOf(a.id) - resourceIds.indexOf(b.id),
       )
     }),
+
+  getIndexPage: protectedProcedure
+    .input(getIndexPageSchema)
+    .output(getIndexPageOutputSchema)
```
seaerchin, line 580: clarification - this specifies the return type of this procedure and shouldn't be needed normally because trpc can infer the return type.

could i check why we need to type hint for `parent` here? i'd prefer that we just let `trpc` handle the typing tbh but this isn't a big deal

apps/studio/src/features/dashboard/components/AdminCreateIndexPageButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+        toast({
+          status: "error",
+          title: "Failed to create index page",
+          description: error.message,
+          ...BRIEF_TOAST_SETTINGS,
+        })
+      },
+    })
+
+  if (!isIsomerAdmin) return null
```
seaerchin, line 60: nit: should we be using `can` from `@casl/react` here? not a big deal but just for standardisation

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -534,4 +541,97 @@ export const pageRouter = router({
...
+                lastModified: new Date().toISOString(),
+                contentPageHeader: {
+                  // follow the same format as autogenerated index pages
+                  summary: `Pages in ${parent.title}`,
+                },
+              },
+              layout:
+                parent.type === ResourceType.Collection
+                  ? ISOMER_USABLE_PAGE_LAYOUTS.Collection
+                  : ISOMER_USABLE_PAGE_LAYOUTS.Index,
```
seaerchin, line 603: clarification - shouldn't this be `Index`?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -534,4 +541,97 @@ export const pageRouter = router({
...
+      }
+
+      return db.transaction().execute(async (tx) => {
+        const blob = await tx
+          .insertInto("Blob")
+          .values({
+            content: jsonb({
+              page: {
+                title: parent.title,
+                permalink: parentFullPermalink,
```
seaerchin, line 593: clarification: could you explain abit more why we need this? based off my understanding, we get the `r.permalink` from the `resource` and use that to construct the full permalink so i'm not sure if this is required/used

seaerchin (commented):
some clarifications but overall ok, no issues w approach. happy to approve later on

apps/studio/src/server/modules/resource/resource.router.ts
adriangohjw: @seaerchin not much reasons actually, was just thinking of using the schema as a self-documentation!

apps/studio/src/server/modules/page/page.router.ts
adriangohjw: i looked through existing custom index pages for collections and they are all Collection

apps/studio/src/server/modules/page/page.router.ts
adriangohjw: @seaerchin current implementation mimicked that of oncall, but just clarified with ZJ that this is a legacy thing that's no longer needed - removed in https://github.com/opengovsg/isomer/pull/1065/commits/8fd3184d08efba2f82d850174ae72028be02e5de . thanks!

apps/studio/src/server/modules/page/page.router.ts
adriangohjw: i doubled checked with existing collextion indexpages and have slightly tweaked it in https://github.com/opengovsg/isomer/pull/1065/commits/f8a14bb6f63a4141c85e5a345b8ac307c9aea30e - thanks for catching

apps/studio/src/features/permissions/hooks/useIsIsomerAdmin.ts
adriangohjw: ahh thanks for pointing out, didn't know it exists - have updated in https://github.com/opengovsg/isomer/pull/1065/commits/2f50412fb6e5c30f5cb88017285cf30b06bfc070

apps/studio/src/features/dashboard/components/AdminCreateIndexPageButton.tsx
adriangohjw: was considering that but feel like i didn't wanna overcomplicate it.
anyway have updated it to use growthbook feature flagging in https://github.com/opengovsg/isomer/pull/1065/commits/2f50412fb6e5c30f5cb88017285cf30b06bfc070 !

seaerchin (approved):

PR #1066, opened 2025-02-01
Enhancement/add collection in folders

adriangohjw:
## Problem

During [Isoranges workabouts sync](https://www.notion.so/opengov/Workarounds-in-Isomer-18277dbba7888070922ff637552ca467), it was identified that one of the limitations for studio is the inability to create non-root level collections.

This PR aims to enables that by replicating the logic for creating non-root level folders, to:
1) improve the information architecture of the sites _(for those that found workabouts)_
2) reduce support load from oncall isoengineers _(for those who reached out to isoengeers for help)_

**Note**: this PR was not set out to allow the moving of collection across folders - we can skip that for now IMO as it's not a commonly requested/used feature.
 
## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- replicating the logic for creating non-root level folders
- updating breadcrumbs for collections to remove previous assumption of hardcoding only the root page
- refactor to reduce code duplication amongst the two flows (creating non-root folders and non-root collections)

## Before & After Screenshots

https://github.com/user-attachments/assets/297d525a-5f7e-4c68-bf68-b1a29d688e76

## Tests

1. go to a folder
2. click on the dropdown button "Add new item" and select Collection
3. fill in the details and click "create" -> the collection should appear

### DIFF
--- a/apps/studio/src/features/dashboard/components/DashboardLayout.tsx
+++ b/apps/studio/src/features/dashboard/components/DashboardLayout.tsx
@@ -11,6 +11,51 @@ import {
 } from "@chakra-ui/react"
 import { Breadcrumb } from "@opengovsg/design-system-react"
 
+import type { RouterOutput } from "~/utils/trpc"
+import { getFolderHref } from "~/utils/resource"
+
+/**
+ * NOTE: This returns the path from root down to the parent of the element.
+ * The element at index 0 is always the root
+ * and the last element is always the parent of the current folder
+ */
+export const getBreadcrumbsFromRoot = (
+  resource: RouterOutput["resource"]["getParentOf"],
+  siteId: string,
+): { href: string; label: string }[] => {
+  // NOTE: We only consider the 3 cases below:
+  // Root -> Folder
+  // Root -> Parent -> Folder
+  // Root -> ... -> Parent -> Folder
+  const rootHref = `/sites/${siteId}`
+
+  if (resource.parent?.parentId) {
+    return [
+      { href: rootHref, label: "Home" },
+      {
+        href: getFolderHref(siteId, resource.parent.parentId),
+        label: "...",
+      },
+      {
+        href: getFolderHref(siteId, resource.parent.id),
+        label: resource.parent.title,
+      },
+    ]
+  }
+
+  if (resource.parent?.id) {
+    return [
+      { href: rootHref, label: "Home" },
+      {
+        href: getFolderHref(siteId, resource.parent.id),
+        label: resource.parent.title,
+      },
+    ]
+  }
+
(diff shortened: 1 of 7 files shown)


apps/studio/src/pages/sites/[siteId]/collections/[resourceId].tsx
```diff
@@ -42,16 +51,10 @@ const CollectionResourceListPage: NextPageWithLayout = () => {
...
-        breadcrumbs={[
-          {
-            href: `/sites/${siteId}`,
-            label: "Home",
-          },
-          {
-            href: `/sites/${siteId}/collections/${resourceId}`,
-            label: metadata.title,
-          },
-        ]}
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/collections/[resourceId].tsx:54]: previously it was hardcoded because we assumed it will be root level, which is no longer true with this update

apps/studio/src/features/dashboard/components/DashboardLayout.tsx
```diff
@@ -11,6 +11,51 @@ import {
...
 
+import type { RouterOutput } from "~/utils/trpc"
+import { getFolderHref } from "~/utils/resource"
+
+/**
+ * NOTE: This returns the path from root down to the parent of the element.
+ * The element at index 0 is always the root
+ * and the last element is always the parent of the current folder
+ */
+export const getBreadcrumbsFromRoot = (
```
adriangohjw, line 22: shifted from `apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx` to be reusable across both flows

apps/studio/src/utils/resource.ts
```diff
@@ -35,3 +35,11 @@ export const getLinkToResource = ({
 }) => {
   return `/sites/${siteId}/${getResourceSubpath(type)}/${resourceId}`
 }
+
+export const getFolderHref = (siteId: string, folderId: string) => {
```
adriangohjw, line 39: shifted from `apps/studio/src/pages/sites/[siteId]/folders/[folderId]/index.tsx` to be reused

seaerchin (commented):
mostly lgtm! i tested this by: 
1. creating collection at root (all good) 
2. creating collection inside another folder (root -> folder -> collection)
3. creating collection inside a nested folder (root -> folder -> folder -> collection)

functionality wise it looks fine becuase our pages exist apart from other pages (ie, the only dependency a page will ever have is only on `category` and that's only if they are a gb-ed site + it's an article page) 

~**however**, there is a tiny visual issue - the page itself will display a breadcrumb from the root down to the page and inside a nested collection, this is wrong~
image

cross-checked against folders - behaviour is consistent, so i think this is ok

seaerchin (approved):

PR #1068, opened 2025-02-03
fix - remove conditional check for resoucetype

adriangohjw:
## Problem

hotfix for https://github.com/opengovsg/isomer/pull/1048 - `getCategories` are used by folders too instead of only collections, and this query throw an error as it assumes theres only collections

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- remove conditional check for resource types


seaerchin (approved):

adriangohjw:
chromatic test failure unrelated, merging as other tests passed

PR #1070, opened 2025-02-03
fix: set collection item dates to be optional

dcshzj:
## Problem

The collection item dates are currently required, but they should be optional because we already support the dateless variant for collections.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Set the date field to be optional for collection items (links and articles).

## Screenshots

image

## Tests

- Go to Studio and enter any collection.
- Edit any article page in the collection, remove the article date and save the page. Verify that the page can be saved.
- Edit any collection link in the collection, remove the item date and save the item. Verify that the page can be saved.

### DIFF
--- a/apps/studio/src/features/editing-experience/components/LinkEditorDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/LinkEditorDrawer.tsx
@@ -200,7 +200,7 @@ export const LinkEditorDrawer = () => {
     title,
   }
   const handleChange = (data: IsomerLinkSchema) =>
-    setLinkAtom((oldData) => ({ ...oldData, ...data }))
+    setLinkAtom(data as CollectionLinkProps)
 
   return (
     <ErrorProvider>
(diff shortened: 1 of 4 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDateControl.tsx
```diff
@@ -33,7 +33,7 @@ export function JsonFormsDateControl({
         <FormLabel description={description}>{label}</FormLabel>
         <DatePicker
           inputValue={!!data ? String(data) : undefined}
-          allowManualInput={false}
+          allowManualInput={true}
```
dcshzj, line 36: note: this was set to true to allow users to remove the value from the date picker, as there is no "clear input" functionality in the OGP design system datepicker

apps/studio/src/features/editing-experience/components/LinkEditorDrawer.tsx
```diff
@@ -200,7 +200,7 @@ export const LinkEditorDrawer = () => {
     title,
   }
   const handleChange = (data: IsomerLinkSchema) =>
-    setLinkAtom((oldData) => ({ ...oldData, ...data }))
+    setLinkAtom(data as CollectionLinkProps)
```
dcshzj, line 203: note: the `oldData` causes the `date` field to be preserved if it was later removed, which causes the preview to be incorrect.

adriangohjw (approved):

sehyunidaaa:
Not approving the chromatic reviews because the way to delete off the date field is unintuitive.

dcshzj:
> Not approving the chromatic reviews because the way to delete off the date field is unintuitive.

Discussed offline, for now is okay, this will be reviewed when we overhaul how optional fields work (e.g. with a toggle)

PR #1086, opened 2025-02-03
vica: support asset URL + use site color + fix disappearing between navigation

adriangohjw:
## Problem

currently
- we don't support image asset url from isomer assets bucket (unless we explicitly include the base url)
- colors are hardcoded (because VICA didn't get back to us)
- vica disappears in between navigation

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- preprocess image url during build time and then pass into a client widget component (to prevent passing entire site to client side)
- pass in colors based on site config directly

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/packages/components/src/templates/next/components/internal/Vica/VicaWidgetClient.tsx
+++ b/packages/components/src/templates/next/components/internal/Vica/VicaWidgetClient.tsx
@@ -1,0 +1,32 @@
+"use client"
+
+import type { VicaProps } from "~/interfaces"
+
+export const VicaWidgetClient = ({
+  ScriptComponent = "script",
+  ...vica
+}: Omit<VicaProps, "site">) => {
+  // to not render during static site generation on the server
+  if (typeof window === "undefined") return null
+
+  return (
+    <>
+      <ScriptComponent
+        async
+        type="text/javascript"
+        src="https://webchat.vica.gov.sg/static/js/chat.js"
+        referrerPolicy="origin"
+        // 'lazyOnload' is recommended by Next.js for chat widgets
+        // Reference: https://nextjs.org/docs/pages/api-reference/components/script#lazyonload
+        strategy="lazyOnload"
+      />
+      <div
+        id="webchat"
+        {...vica}
+        app-font-family="Inter, system-ui, sans-serif"
+        // NOTE: Clarifying with VICA regarding color scheme
+        // Once confirmed, will override with site's color scheme for consistency
+      />
+    </>
+  )
+}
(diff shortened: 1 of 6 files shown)


adriangohjw:
tested to be working on staging

packages/components/src/templates/next/components/internal/Vica/VicaWidgetClient.tsx
```diff
@@ -0,0 +1,46 @@
...
+    // to not render during static site generation on the server
+    if (typeof window === "undefined") return
+
+    // Only inject the script after everything else has finished loading
+    // This is to replicate Next.js lazyOnload behaviour (as recommended for widgets)
+    if (document.readyState === "complete") {
+      loadVicaScript()
+    } else {
+      window.addEventListener("load", loadVicaScript)
+      return () => window.removeEventListener("load", loadVicaScript)
```
dcshzj, line 29: issue: I think we should give the script an ID and remove the script element if it is previously present, similar to the SearchSGInputBox. Otherwise, it can cause multiple scripts to be loaded in the DOM after a couple of route changes.

image

packages/components/src/templates/next/components/internal/Vica/VicaWidgetClient.tsx
```diff
@@ -0,0 +1,46 @@
...
+const loadVicaScript = () => {
+  const scriptTag = document.createElement("script")
+  scriptTag.async = true
+  scriptTag.type = "text/javascript"
+  scriptTag.src = "https://webchat.vica.gov.sg/static/js/chat.js"
+  scriptTag.referrerPolicy = "origin"
+  document.body.appendChild(scriptTag)
+}
+
+export const VicaWidgetClient = (vica: Omit<VicaProps, "site">) => {
```
dcshzj, line 18: nit: I think this should be `VicaWidgetProps` instead of the `Omit`

adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1086/commits/58e8143fd5e2a655a7ab62a826ca3d1b9aa1eeae

packages/components/src/templates/next/components/internal/Vica/VicaWidgetClient.tsx
adriangohjw: thanks for catching, updated in https://github.com/opengovsg/isomer/pull/1086/commits/34a185ffa87aaf5d6bfa8465a1f72b5623dbbe9d !

dcshzj (approved):

PR #1088, opened 2025-02-04
chore(copy): update copy + add toast

seaerchin:
## Problem
- toast copy on page settings save is misleading
- article page header lacks toast on save

## Solution
- update toast copy for page settings save
- add toast for article page header save

### DIFF
--- a/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
@@ -121,8 +121,8 @@ const PageSettingsModalContent = ({
       await utils.collection.invalidate()
 
       toast({
-        title: "Saved page settings",
-        description: "Publish this page for your changes to go live.",
+        title: "Saved and published page settings",
+        description: "Check your site later to see the changes",
         status: "success",
       })
     },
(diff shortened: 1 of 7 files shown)


apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
```diff
@@ -121,8 +121,8 @@ const PageSettingsModalContent = ({
       await utils.collection.invalidate()
 
       toast({
-        title: "Saved page settings",
-        description: "Publish this page for your changes to go live.",
+        title: "Saved and published page settings",
+        description: "Check your site later to see the changes",
```
harishv7, line 125: What would "later" mean to a user?

Might want to check the copy with @sehyunidaaa

sehyunidaaa:
Would be helpful to have a video or new story on PRs like this to verify the behaviour!

apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
harishv7: I think "few mins" works

Avg case scenario, it should not take more than 10 mins. 
Worst case scenario is 15min

seaerchin: Lets just use the original wording? 

`description: "Check your site in 5-10 minutes to view it live."`

seaerchin:
> Would be helpful to have a video or new story on PRs like this to verify the behaviour!

added for homepage/contentpage/articlepage

apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
harishv7: lgtm

harishv7 (approved):

PR #1097, opened 2025-02-06
Fix aspect ratios of images in contentPic and blogCard

sehyunidaaa:
## Problem

Components using images had varying aspect ratios, making it tricky for migrators and agency users to apply images. This PR deals with contentPic and blogCards in blog collection variant.

Closes [ISOM-1690]

## Solution

- Fix contentPic image ratio to 5:6 regardless of screen size / breakpoint
- Fix blogCard image ratio to 2:1 regardless of screen size / breakpoint

Please check Chromatic for before & after


packages/components/src/templates/next/components/complex/Contentpic/Contentpic.tsx
```diff
@@ -11,8 +11,7 @@ const contentpicStyles = tv({
     // margin used for margin collapse
     container:
       "mb-7 flex flex-col gap-7 md:flex-row [&:not(:first-child)]:mt-7",
-    image:
-      "max-h-[400px] w-full rounded object-cover md:h-[240px] md:max-h-full md:w-[200px]",
+    image: "aspect-[5/6] h-auto rounded object-cover md:h-[240px] md:w-[200px]",
```
dcshzj, line 14: issue: I think might need to run `npm run build`? I think these aspect ratio CSS classes are new and won't work (at least on Studio) until the tailwind CSS file is rebuilt

adriangohjw: fixed here https://github.com/opengovsg/isomer/pull/1097/commits/b568b7da5d3ccab94c1ee806d55f24579319a56a

sehyunidaaa: thank you!

adriangohjw (approved):

PR #1098, opened 2025-02-07
chore: fix publishing for empty links

seaerchin:
## Problem
1. users can publish empty links 

## Solution
1. add a `ref` atom that we use **ONLY** to validate to see if a user can publish
2. this is only updated `onSave` - so we know it's accurate
3. we cannot use the other `linkAtom` as that's required for updating our UI

## Videos
https://github.com/user-attachments/assets/6554f0fb-49da-47c3-8e05-27179477410d

## Tests
1. create a new collection link 
2. ensure that publishing is blocked
3. add a ref
4. publishing shuold still be blocked
5. click save
6. publishing should be allowed
7. go back to the collection
8. repeat steps 1-2

### DIFF
--- a/apps/studio/src/features/editing-experience/components/PublishButton.tsx
+++ b/apps/studio/src/features/editing-experience/components/PublishButton.tsx
@@ -20,6 +20,7 @@ const SuspendablePublishButton = ({
   pageId,
   siteId,
   isDisabled,
+  onClick,
   ...rest
 }: PublishButtonProps): JSX.Element => {
   const toast = useToast()
@@ -73,7 +74,10 @@ const SuspendablePublishButton = ({
               isDisabled={!isChangesPendingPublish || isDisabled}
               variant="solid"
               size="sm"
-              onClick={handlePublish}
+              onClick={(e) => {
+                handlePublish()
+                onClick?.(e)
+              }}
               isLoading={isLoading}
               {...rest}
             >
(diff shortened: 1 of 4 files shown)


apps/studio/src/features/editing-experience/components/PublishButton.tsx
```diff
@@ -10,14 +11,16 @@ import { Can } from "~/features/permissions"
 import { withSuspense } from "~/hocs/withSuspense"
 import { trpc } from "~/utils/trpc"
 
-interface PublishButtonProps {
+interface PublishButtonProps extends ButtonProps {
   pageId: number
   siteId: number
 }
```
dcshzj, line 17: thought(non-blocking): this seems a bit extensive that we allow the consumer to be able to customise the publish button, since this is quite a specific component and ideally should be standardised across the entire site?

Suggestion is that we only need to add the `isDisabled` prop as part of the props, but the rest of the button props should not be customisable by the consumer.

dcshzj (approved):
Approving first but do have some consideration for the comment I have earlier!

apps/studio/src/features/editing-experience/components/PublishButton.tsx
seaerchin: 2 things to note here:
1. we are the consumer here, so we dictate how this component is used 
2. this is on the frontend, so behaviour is still gated by backend

i think the principle i use to think about this kind of component is - what is the component _really_ trying to do? in this case, the value of the `PublishButton` (at least to me) was the fact that it fetched from backend cross checks the state to conditionally disable and has a mutate also.

i think the styling part of it, at least to me, wasn't really much of a concern because it's more a "i will only customise it when i need it" and the real business value is in the logic that this button itself has regarding UI states + onclick.

that being said, i'll add a case for `onClick` to fire alongisde our mutation function so that the caller's onclick doesn't override our mutate

PR #1103, opened 2025-02-12
chore: fix infinite login spinner

seaerchin:
## Problem
See ticket

Closes ISOM-1716 (hopefully)

## Solution
- removed the full screen spinner and instead, return null since we're redirected out to `/` 

## Notes
- i tested this via setting the redirect in `/:page/settings`; previously, it showed teh spinner **even after** redirection but now, it displays only the expected content
- i think this happens because we wrap our app in a default error boundary, so this spinner displays **even after** redirection out to the new page

## Tests
1. give yourself access to some site 
2. go to that site 
3. now, delete your access to the site
4. refresh teh page
5. you should get redirected out to `/`
6. there shuold not be a loading spinner

### DIFF
--- a/apps/studio/src/components/ErrorBoundary/DefaultTrpcError.tsx
+++ b/apps/studio/src/components/ErrorBoundary/DefaultTrpcError.tsx
@@ -3,12 +3,11 @@ import { useRouter } from "next/router"
 import { type TRPC_ERROR_CODE_KEY } from "@trpc/server/rpc"
 
 import { trpc } from "~/utils/trpc"
-import { FullscreenSpinner } from "../FullscreenSpinner"
 import { DefaultNotFound } from "./DefaultNotFound"
 import { DefaultServerError } from "./DefaultServerError"
 import { UnexpectedErrorCard } from "./UnexpectedErrorCard"
 
-function UnauthorizedError() {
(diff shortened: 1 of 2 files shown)


apps/studio/src/components/ErrorBoundary/DefaultTrpcError.tsx
```diff
@@ -3,20 +3,19 @@ import { useRouter } from "next/router"
 import { type TRPC_ERROR_CODE_KEY } from "@trpc/server/rpc"
 
 import { trpc } from "~/utils/trpc"
-import { FullscreenSpinner } from "../FullscreenSpinner"
 import { DefaultNotFound } from "./DefaultNotFound"
 import { DefaultServerError } from "./DefaultServerError"
 import { UnexpectedErrorCard } from "./UnexpectedErrorCard"
 
-function UnauthorizedError() {
+export const UnauthorizedError = () => {
```
adriangohjw, line 10: nit - remove `export` since it's not used outside of this file

adriangohjw (dismissed):
test in staging first? can't actually replicate the actual flow on dev (seeing spinner on post-login)

adriangohjw (approved):

PR #1105, opened 2025-02-12
chore: update collection links

seaerchin:
## Problem
Right now, collection links aren't the greatest - they show the permalink of the collection link itself but this is **not the actual permalink** of the item on the website (and in fact, the collection link isn't accessible via said permalink) 

Closes ISOM-1737

## Solution
1. allow `CollectionLink` to be updated (abit hidden cos we're using page router .-.) 
2. hide the `CollectionLink`'s permalink on studio

## Videos 

https://github.com/user-attachments/assets/24b29fd4-dc5a-42c1-bd68-ee31aab6d941

## Tests
1. go to a collection **with an existing collection link**
2. make sure that hte collection link's permalink is hidden
3. try and update teh permalink/title
4. it shuold succeed
5. create a new collection link
6. it should succeed
7. repeat steps 3-4


apps/studio/src/features/dashboard/components/CollectionTable/CollectionTable.tsx
```diff
@@ -26,7 +27,11 @@ const getColumns = ({ siteId }: CollectionTableProps) => [
         siteId={siteId}
         id={row.original.id}
         title={row.original.title}
-        permalink={`/${row.original.permalink}`}
+        permalink={
+          row.original.type === ResourceType.CollectionLink
+            ? ""
+            : `/${row.original.permalink}`
+        }
```
adriangohjw, line 34: nit - might wanna pass in `null` and then not rendered as the `VStack` it is in will still render this empty string permalink, causing a "0.25rem margin-bottom" look to the title and it being non-centered with the left icon

https://github.com/opengovsg/isomer/blob/1a66183a6b6cf411f9fc69880619fb56258555df/apps/studio/src/features/dashboard/components/ResourceTable/TitleCell.tsx#L65-L86

seaerchin: good catch, lemme update it and try!

apps/studio/src/features/dashboard/components/CollectionTable/CollectionTableMenu.tsx
```diff
@@ -33,19 +33,20 @@ export const CollectionTableMenu = ({
...
-                  pageId: resourceId,
-                  type: resourceType,
-                })
-              }
-            >
-              Edit page settings
-            </MenuItem>
-          )}
+          {resourceType === ResourceType.CollectionPage ||
+            (resourceType == ResourceType.CollectionLink && (
```
adriangohjw, line 37: should be `===` instead of `==`

apps/studio/src/features/dashboard/components/CollectionTable/CollectionTableMenu.tsx
```diff
@@ -33,19 +33,20 @@ export const CollectionTableMenu = ({
...
+                onClick={() =>
+                  setPageSettingsModalState({
+                    pageId: resourceId,
+                    type: resourceType,
+                  })
+                }
+              >
+                Edit page settings
+              </MenuItem>
+            ))}
```
adriangohjw, line 49: instead of
```ts
{resourceType === ResourceType.CollectionPage ||
  (resourceType == ResourceType.CollectionLink && (...)
 )}
```

should the bracket be around the entire condition instead?
```ts
{(resourceType === ResourceType.CollectionPage || resourceType === ResourceType.CollectionLink) && (...)}
```

seaerchin: ah yes, ps, thanks for the catch!

apps/studio/src/features/dashboard/components/CollectionTable/CollectionTableMenu.tsx
seaerchin: thanks for spotting!

adriangohjw (approved):

seaerchin:
not merging yet, checking if @sehyunidaaa ok

sehyunidaaa:
I will check later this week and comment; please don't merge yet. No bandwidth to review and suggest alternatives today

sehyunidaaa:
Hello! I'm a bit confused by the role of the permalink on this item! Primarily two concerns:
- It's not clear how the permalink is used; when MOPs hover on the link item on a site, will they see this permalink? or will they see the original URL?
- Does a collection link/file item _have_ to have a permalink associated with it? Even though its target is technically its URL? 
- It's also quite hidden and not communicated when the link is first created

Other comments / questions:
- I agree that the title editing modal should be available, but should we just have a title field?
- If this ^ is possible though, does that mean we can put the title field in the page editing view itself?
- On the dashboard level, if it's an external URL, should we display that in place of the permalink? Currently it looks a bit empty!

@seaerchin thank you!

PR #1107, opened 2025-02-13
fix(bug): link editor throws an error when the referenced item has been deleted

seaerchin:
## Problem
Currently, when we delete a link and go back to an item that is referencing it, we will see an error page with no way for the user to self resolve; this isn't ideal

## Solution
1. add error boundary and use it in `BaseLinkControl`
2. the error stems from `getWithFullPermalink` - at present, only this method throws an error for the children of the error boundary, so we won't display this error boundary for an unrelated error
3. for other call sites of `getWithFullPermalink`, the precondition (at least on frontend) is that hte resource exists 
    - link editor modal, where we click on an item 
    - folder settings modal (for full permalink of folder) 
    - create page (full permalink of parent) 
    hence, i didn't add on those places

## Video

https://github.com/user-attachments/assets/d35f38b1-4e1b-4ad7-a82e-13351dbc3271

## Tests 
1. Go to any Studio site and edit the home page.
2. Use the raw JSON editor mode to change the primary CTA button link to a reference link pointing to a non-existent resource ID, then save.
3. Click on the button to edit the hero banner.
4. You should see the infobox on the button link 
5. Click the trash icon
6. Relink an existing page and save
7. reload the page
8. the page should not crash

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/components/LinkErrorBoundary.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/components/LinkErrorBoundary.tsx
@@ -1,0 +1,50 @@
+import { PropsWithChildren } from "react"
+import { IconButton, Stack, Text } from "@chakra-ui/react"
+import { Infobox } from "@opengovsg/design-system-react"
+import { ErrorBoundary } from "react-error-boundary"
+import { BiTrash } from "react-icons/bi"
+
+interface LinkErrorBoundaryProps {
+  resetLink: () => void
+}
+export const LinkErrorBoundary = ({
+  resetLink,
+  children,
+}: PropsWithChildren<LinkErrorBoundaryProps>) => {
+  return (
+    <ErrorBoundary
+      fallbackRender={({ resetErrorBoundary }) => (
+        <Infobox
+          variant="error"
+          borderRadius="4px"
+          borderColor="utility.feedback.critical"
+          border="1px solid"
+          bg="utility.feedback.critical"
+          w="100%"
+          size="sm"
+        >
+          <Stack direction="column" w="full">
+            <Text textStyle="subhead-2">
+              The page you linked no longer exists
+            </Text>
+            <Text> Pick a different destination</Text>
+          </Stack>
+          <IconButton
+            size="xs"
+            variant="clear"
+            alignSelf="center"
+            colorScheme="critical"
+            aria-label="Remove file"
+            icon={<BiTrash />}
+            onClick={() => {
+              resetLink()
+              resetErrorBoundary()
+            }}
+          />
+        </Infobox>
+      )}
+    >
+      {children}
+    </ErrorBoundary>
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/editing-experience/components/form-builder/components/LinkErrorBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const LinkErrorBoundary = ({
+  resetLink,
+  children,
+}: PropsWithChildren<LinkErrorBoundaryProps>) => {
+  return (
+    <ErrorBoundary
+      fallbackRender={({ resetErrorBoundary }) => (
+        <Infobox
+          variant="error"
+          borderRadius="4px"
```
adriangohjw, line 19: nit - use 1rem instead since we have mostly been using rem over px

apps/studio/src/features/editing-experience/components/form-builder/components/LinkErrorBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+  resetLink: () => void
+}
+export const LinkErrorBoundary = ({
+  resetLink,
+  children,
+}: PropsWithChildren<LinkErrorBoundaryProps>) => {
+  return (
+    <ErrorBoundary
+      fallbackRender={({ resetErrorBoundary }) => (
+        <Infobox
```
adriangohjw, line 17: question - should we also standardize to use Flex like in `BaseLinkControl` instead to have a more consistent code and UI?

https://github.com/opengovsg/isomer/blob/fc2128e2070ed16e69a81d71c686501d966aa592/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/BaseLinkControl.tsx#L76-L83

apps/studio/src/features/editing-experience/components/form-builder/components/LinkErrorBoundary.tsx
seaerchin: oh yes! good catch! have to talk to @sehyunidaaa regarding this because there's a [difference](https://austingil.com/px-or-rem-in-css/) - but i think since we are already using `rem` for borders, might as well continue omegalull

apps/studio/src/features/editing-experience/components/form-builder/components/LinkErrorBoundary.tsx
seaerchin: most of the time, yes - however i didn't do it here because we want the `Infobox` component from ogpds. the reason why i didn't use it last time was because (sadly enough) it [always](https://github.com/opengovsg/camp-design-system/blob/b9bddc4299d1b8c87801bb3635598f001019b3fb/packages/camp/src/Infobox/Infobox.tsx#L37) provides an icon

adriangohjw (approved):

PR #1109, opened 2025-02-15
chore: break word and resolve overflow

seaerchin:
## Problem
1. right now we are using `break-all` in tiptap, which causes wordbreak to be annoying.

## Solution
1. update to `break-word` - however, the preview text can now overflow (very long word, no space) so i had to add `overflow: auto` 

## Video

https://github.com/user-attachments/assets/5db726c1-987a-4f73-bba1-b8a5555b6a14

## Tests
1. create a new page
2. add a new text block
3. go into the text block
4. type out a sentence with a long word (like: supercalifragilisticexplalidicious) at the end
5. the long word should not be randomly truncated in the middle but reflow to the next time
6. type out a huge chunk of words without space (eg: testtesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttest) 
7. click save
8. the preview text should be automatically hidden rather than overflowing

### DIFF
--- a/apps/studio/src/features/editing-experience/components/Block/BaseBlock.tsx
+++ b/apps/studio/src/features/editing-experience/components/Block/BaseBlock.tsx
@@ -59,7 +59,7 @@ export const BaseBlock = ({
           <Icon as={icon} fontSize="0.75rem" color="base.content.default" />
         </Flex>
       )}
-      <Stack align="start" gap="0.25rem">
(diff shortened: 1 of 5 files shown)


adriangohjw (approved):
nice fix! 

perhaps it's worth adding a storybook for this so there's no accidental degradation down the road?

seaerchin:
sounds good, lemme do that

PR #1117, opened 2025-02-18
isom-1740 backend user management service (tracking last login)

adriangohjw:
## Problem

Track user login timestamps by adding a `lastLoginAt` field to the User model.

This change is essential for improving user authentication tracking and analytics.

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible

**Features**:

- Added `lastLoginAt` field to the User model.
- Updated authentication logic to set `lastLoginAt` during user login.

## Tests

The following tests have been added to confirm functionality:

- Tests for setting `lastLoginAt` to null when creating a new user.
- Tests for updating `lastLoginAt` when a user logs in.

### Manual tests

1. Login into studio with a new email -> `lastLoginAt` should be current timestamp
2. Log out and login again -> `lastLoginAt` should be a new timestamp

PR #1118, opened 2025-02-18
isom-1740 USER MGMT (backend user management service)

adriangohjw:
### note - don't merge yet until security review

Closes https://linear.app/ogp/issue/ISOM-1740/backend-user-management-service

Refer to the following sub branches for more details:
- https://github.com/opengovsg/isomer/pull/1117
- https://github.com/opengovsg/isomer/pull/1121
- Adding DB column index on `lastLoginAt`
- Additional commits to beef up tests in this branch directly


apps/studio/src/server/modules/user/__tests__/user.router.test.ts
```diff
@@ -0,0 +1,1001 @@
...
+        users: [{ email: TEST_EMAIL }],
+      })
+
+      // Assert
+      await expect(result).rejects.toThrowError(
+        new TRPCError({ code: "UNAUTHORIZED" }),
+      )
+    })
+
+    it("should throw 403 if user is not admin of the site", async () => {
```
harishv7, line 62: i didn't get this part - we are setting up admin perms for the user in arrange but the assert expects it to throw forbidden for a non-admin user

adriangohjw: this checks that admin of SiteA cannot create user for SiteB

apps/studio/src/schemas/user.ts
```diff
@@ -0,0 +1,87 @@
+import { RoleType } from "~prisma/generated/generatedEnums"
+import { z } from "zod"
+
+import { offsetPaginationSchema } from "./pagination"
+
+export const createInputSchema = z.object({
```
seaerchin, line 6: [nit] actually why do we omit the `User` here (ie, why not `createUserInputSchema`)? i thought we'd err on teh side of verbosity but this isn't a huge issue because this is used in the router - just makes it easier next time to know that it's the correct schema (eg: if we have 2 `createInputSchema`)

apps/studio/src/schemas/user.ts
```diff
@@ -0,0 +1,87 @@
...
+  id: z.string(),
+  name: z.string(),
+  email: z.string().email(),
+  role: z.nativeEnum(RoleType),
+  lastLoginAt: z.date().nullable(),
+})
+
+export const listInputSchema = z.object({
+  siteId: z.number().min(1),
+  getIsomerAdmins: z.boolean().optional().default(false),
```
seaerchin, line 49: [nit] after reviewing the downstream PRs, this feels like it should be an array of enum of `RoleType` + `IsomerAdmin` and the backend will return all users whose roles are in this array

not a big deal but using booleans like this is a code smell

apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
seaerchin, line 1: i skipped this

apps/studio/src/server/modules/auth/sgid/__tests__/sgid.service.test.ts
seaerchin, line 1: skipped

apps/studio/src/server/modules/user/__tests__/testHelper.ts
seaerchin, line 1: [nit] can we rename this to `utils` hahahahaha

apps/studio/src/server/modules/user/__tests__/user.router.test.ts
seaerchin, line 1: skipped

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
seaerchin, line 1: skipped

apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -0,0 +1,282 @@
...
+              email: user.email.toLowerCase(),
+              siteId,
+              trx,
+            })
+            return {
+              id: createdUser.id,
+              email: createdUser.email,
+              role: resourcePermission.role,
+            }
+          }),
```
seaerchin, line 53: [issue] this feels like an issue because we're holding a lock and inserting row after row into db. wonder if we should use a bulk insert here? (see kysely's [docs](https://kysely.dev/docs/examples/insert/multiple-rows))

seaerchin (commented):
review abit first...

apps/studio/src/lib/logger.ts
adriangohjw, line 1: unrelated but threw issue midweek. something to do with dependencies

apps/studio/src/server/modules/permissions/permissions.service.ts
```diff
@@ -109,3 +110,50 @@ export const validateUserPermissionsForResource = async ({
...
+    .selectFrom("ResourcePermission")
+    .where("userId", "=", userId)
+    .where("siteId", "=", siteId)
+    .where("resourceId", "is", null)
+    .where("deletedAt", "is", null)
+    .select("role")
+    .execute()
+
+  // NOTE: Any role should be able to read the list of user permissions
+  if (roles.length > 0) {
```
harishv7, line 129: technically length of roles should be just 1 right? >1 would be an error state

apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -0,0 +1,282 @@
...
+        )
+      })
+
+      // TODO: Send welcome email to users
+      // Email service is not ready yet
```
harishv7, line 58: @adriangohjw will this be implemented in this sprint? It's labelled as a P1. Do heads up if we need to deprio this

apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -0,0 +1,282 @@
...
+
+      return createdUsers
+    }),
+
+  delete: protectedProcedure
+    .input(deleteInputSchema)
```
harishv7, line 64: can we rename as deleteUserInputSchema etc. for clarity

apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -0,0 +1,282 @@
...
+      }
+
+      return result
+    }),
+
+  list: protectedProcedure
+    .input(listInputSchema)
+    .output(listOutputSchema)
+    .query(
+      async ({ ctx, input: { siteId, getIsomerAdmins, offset, limit } }) => {
```
harishv7, line 151: do we have default values for the offset, limit?

apps/studio/src/server/modules/user/user.router.ts
adriangohjw: oops this comment was added before i worked on email service. Email service has since been done, i just need to pull the changes and integrate them (call the service endpoint)

apps/studio/src/server/modules/permissions/permissions.service.ts
adriangohjw: yes you are right, but currently we don't have constaints on the DB layer to check for uniqueness, because we are using `@@unique([userId, siteId, resourceId, role])`, and a `NULL` `resourceId` is not "unique" per se due to how DB compare and handles `NULL`

thus is more defensive to make sure any error-nous entries don't throw any error

apps/studio/src/server/modules/user/user.router.ts
adriangohjw: the input schema is importing from existing `offsetPaginationSchema` which uses default 10 as limit

apps/studio/src/server/modules/user/user.router.ts
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/1118/commits/146a8c2a92d5f105eebb40842ff061c01d1ec48a

apps/studio/src/schemas/user.ts
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/1118/commits/146a8c2a92d5f105eebb40842ff061c01d1ec48a

apps/studio/src/schemas/user.ts
adriangohjw: `getIsomerAdmins ` is for us to return only isomer admins, or non isomer admins, for this tab here

image

i was contemplating:
1. whats the best params for this (couldn't think of one) and
2. whether i should spin this off as a standalone endpoint

but given that this is confusing, i think i will go with number 2!

oh and no need to return by `RoleType` for now since we are returning all users without any filtering logic

apps/studio/src/server/modules/permissions/permissions.service.ts
```diff
@@ -109,3 +110,50 @@ export const validateUserPermissionsForResource = async ({
...
+  userId,
+  siteId,
+}: Omit<PermissionsProps, "resourceId">) => {
+  const builder = new AbilityBuilder<UserManagementAbility>(createMongoAbility)
+  const roles = await db
+    .selectFrom("ResourcePermission")
+    .where("userId", "=", userId)
+    .where("siteId", "=", siteId)
+    .where("resourceId", "is", null)
+    .where("deletedAt", "is", null)
```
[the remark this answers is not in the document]
adriangohjw: yup we soft-delete all `ResourcePermission` of the users for that site https://github.com/opengovsg/isomer/blob/146a8c2a92d5f105eebb40842ff061c01d1ec48a/apps/studio/src/server/modules/user/user.router.ts#L80-L85

apps/studio/src/server/modules/user/__tests__/testHelper.ts
adriangohjw: done in https://github.com/opengovsg/isomer/pull/1118/commits/af96b48a49c52a37586d02b5356b07f818263eac

njenwei:
Just popping by cause you shared on slack!

Some questions/comments:
- Consider having separate tables for Site and Resource permissions
  - When combined, you have to code around it so often - having to specify `resourceId === null` so often feels quite clunky, 
  - The DB cannot enforce the correct unique constraints with the null column
  - `deletedAt` doesn't work when resourceId is set, yet works when resourceId is null (actually how do you delete resource level permissions?)
- Consider if you need to soft delete the old permissions? Might be cleaner to keep the Permissions table "operational", only reflecting the current state
  - To soft delete during an update feels weird... might be confusing someday
  - You probably want an audit log table (eg, to capture who actually made the changes), no need to track history here
  - Because resourceId is null, you don't have proper indexes - might result in full table scans
- Consider moving `validateUserPermissionsForSite` into `permissions.service.ts` for consistency
- Consider extending `protectedProcedure` to take in your CASL checks, so that it's enforced by default on all routes that use the procedure `update: protectedProcedure('RESOURCE', 'ACTION')`

Apologies if you've already had some of these discussions with your team prior to this!

apps/studio/src/server/modules/user/user.router.ts
harishv7: got it, thanks!!

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +20,175 @@ export const isUserDeleted = async (email: string) => {
...
+  getIsomerAdmins: boolean
+}
+
+interface RankedResourcePermission extends ResourcePermission {
+  rn: number
+}
+
+export const getUsersQuery = ({
+  siteId,
+  getIsomerAdmins,
```
harishv7, line 153: does getIsomerAdmins **only** return admins or **also includes** admins in original result set?

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +20,175 @@ export const isUserDeleted = async (email: string) => {
...
+  return db
+    .with("ActiveResourcePermission", (qb) =>
+      qb
+        .selectFrom(
+          sql<RankedResourcePermission>`(
+          SELECT *,
+            ROW_NUMBER() OVER (
+              PARTITION BY "userId", "siteId", "resourceId"
+              ORDER BY CASE 
+                WHEN role = ${RoleType.Admin} THEN 1
```
harishv7, line 164: why do we need the role ranking here? each user should have exactly 1 role per site right?

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +20,175 @@ export const isUserDeleted = async (email: string) => {
...
+                WHEN role = ${RoleType.Publisher} THEN 2
+                WHEN role = ${RoleType.Editor} THEN 3
```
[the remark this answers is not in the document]
harishv7: brainstorming:

```
export const getUsersQuery = ({ siteId, getIsomerAdmins }: GetUsersQueryProps) => {
  return db
    .with("ActiveUser", (qb) => {
      let query = qb.selectFrom("User").selectAll().where("deletedAt", "is", null);
      
      if (!getIsomerAdmins) {
        query = query.where("email", "not in", ISOMER_ADMINS_AND_MIGRATORS_EMAILS);
      }
      
      return query;
    })
    .selectFrom("ActiveUser")
    .innerJoin("ResourcePermission", "ActiveUser.id", "ResourcePermission.userId")
    .where("ResourcePermission.siteId", "=", siteId)
    .where("ResourcePermission.deletedAt", "is", null)
    .orderBy([{ column: "ActiveUser.id", order: "asc" }]);
};

```

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: we want it to return **EITHER** agency or isomer admins.

you are right that it's confusing - have updated it https://github.com/opengovsg/isomer/pull/1118/commits/6087bd110371163c21a179c9bc90856b4ae80bc5 to use explicit value

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: same issue as non-uniqueness on null value https://github.com/opengovsg/isomer/pull/1118#discussion_r1978864412

harishv7: Ah right... Hmm, that's an issue as it complicates our queries a lot

Might be out of scope for this PR, but can we do a unique constraint with COALESCE?

harishv7: We should have used something like -1 for global resource perms

harishv7: I mean smth like 

```
CREATE UNIQUE INDEX unique_user_site_role
ON "ResourcePermission" (userId, siteId, COALESCE(resourceId, -1), role);

```

This will automatically make all resourceId which is NULL to -1

adriangohjw: that might be possible as a quick hack, will look into it.

though im quite liking @njenwei [idea](https://github.com/opengovsg/isomer/pulls?q=assignee%3Aadriangohjw+is%3Aopen+%22USER+MGMT%22+assignee%3Aadriangohjw+is%3Aopen+%22USER+MGMT%22++draft%3Afalse+-review%3Aapproved+) of having a separate table `SitePermission` - it feels more self-explanatory and straightforward. Might be something to consider

harishv7: Yea it might be cleaner, but might be too much to handle in this PR? Also we need to backfill for existing sites we have during deployment. Check out the effort needed - are there places in our codebase to update checks etc

adriangohjw: yup it's too much to handle in this PR, esp. since this is kinda being used everywhere and we don't have much test coverage on other parts of the app haha.

will add this to a backlog

harishv7: Sounds good, if so maybe the unique constraint temporarily can simplify your code till we decouple site vs resource perms

apps/studio/src/server/modules/permissions/permissions.service.ts
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/1118/commits/f2bec85e5bb24efca8ee4d4badcbc9e9a1db7f83

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/1118/commits/c5dc7baa0cae9ddddb21d944ee10c2ce2d33b524

apps/studio/src/server/modules/user/user.service.ts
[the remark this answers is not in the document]
adriangohjw: no longer required, removed in https://github.com/opengovsg/isomer/pull/1118/commits/c5dc7baa0cae9ddddb21d944ee10c2ce2d33b524

apps/studio/src/server/modules/user/user.router.ts
adriangohjw: implemented in https://github.com/opengovsg/isomer/pull/1118/commits/d9ffcd67849db5dfb2c923fd12be8f52c0b99871

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +21,144 @@ export const isUserDeleted = async (email: string) => {
...
+  }
+}
+
+interface CreateUserProps {
+  email: User["email"]
+  name?: User["name"]
+  phone?: User["phone"]
+  role: ResourcePermission["role"]
+  siteId: ResourcePermission["siteId"]
+  trx?: SafeKysely // allows for transaction to be passed in from parent transaction
```
seaerchin, line 47: [question] jst from the naming - should this be a `tx` rather than accepting `tx | db`?

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +21,144 @@ export const isUserDeleted = async (email: string) => {
...
+
+  const isWhitelisted = await isEmailWhitelisted(email)
+  if (!isWhitelisted) {
+    throw new TRPCError({
+      code: "FORBIDDEN",
+      message: "There are non-gov.sg domains that need to be whitelisted.",
+    })
+  }
+
+  const executeInTransaction = async (tx: SafeKysely) => {
```
seaerchin, line 75: [issue] this is pretty confusing - could we instead strengthen the type that hte main function accepts to `transaction` to avoid this?

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +21,144 @@ export const isUserDeleted = async (email: string) => {
...
+interface CreateUserProps {
+  email: User["email"]
+  name?: User["name"]
+  phone?: User["phone"]
+  role: ResourcePermission["role"]
+  siteId: ResourcePermission["siteId"]
+  trx?: SafeKysely // allows for transaction to be passed in from parent transaction
+}
+
+export const createUser = async ({
```
seaerchin, line 50: [nit] this name is abit misleading because the method also adds permissions for the user; what do you feel about `createUserWithPermissionsForSite` (can shorten but you get the idea)

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -11,3 +21,144 @@ export const isUserDeleted = async (email: string) => {
...
+      .where("resourceId", "is", null) // because we are granting site-wide permissions
+      .selectAll()
+      .executeTakeFirst()
+
+    if (existingResourcePermission) {
+      throw new TRPCError({
+        code: "CONFLICT",
+        message: "User and permissions already exists",
+      })
+    }
```
seaerchin, line 108: [question] why do we auto-resolve conflicting emails but not resource permissions? my take is that we should just automatically update it to the new requested permission because this is what the user wanted (ie, similar to updating a user's resource permission)

apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -0,0 +1,296 @@
...
+      if (userId === ctx.user.id) {
+        throw new TRPCError({
+          code: "FORBIDDEN",
+          message: "You cannot update your own role",
+        })
+      }
+
+      const user = await db
+        .selectFrom("User")
+        .where("id", "=", userId)
```
seaerchin, line 254: [issue] should we also be filtering so that hte users are not soft deleted?

seaerchin:
before i forget, we should chedck to make sure `%40` isn't interpreted as `@` by our backend (though this is a very minor case and i'm ok w/ omitting)

apps/studio/prisma/schema.prisma
```diff
@@ -122,10 +122,13 @@ model User {
   createdAt DateTime  @default(now())
   updatedAt DateTime  @default(now()) @updatedAt
   deletedAt DateTime?
+  lastLoginAt DateTime?
 
   ResourcePermission ResourcePermission[]
   versions           Version[]
 
+  @@index([lastLoginAt])
```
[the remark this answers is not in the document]
adriangohjw: yeah i dont think there's a need for now, not even until at least 10k users / high DB load

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: good point, updated in https://github.com/opengovsg/isomer/pull/1118/commits/82e1e85e6a2a354171efea8cd57adb92f33c036f

apps/studio/src/server/modules/user/user.router.ts
adriangohjw: good point, updated in https://github.com/opengovsg/isomer/pull/1118/commits/28d4027a161319f9341700a0741245b495b6bc23

apps/studio/src/server/modules/whitelist/whitelist.router.ts
```diff
@@ -0,0 +1,25 @@
...
+    .output(isEmailWhitelistedOutputSchema)
+    .query(async ({ ctx, input: { siteId, email } }) => {
+      // Validate permissions as this endpoint is used for user management
+      // and allows checking if emails are whitelisted. This ensures proper
+      // access control even though the operation itself is read-only.
+      await validatePermissionsForManagingUsers({
+        siteId,
+        userId: ctx.user.id,
+        action: "create",
+      })
```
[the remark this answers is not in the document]
adriangohjw: oh `isEmailWhitelisted` is only used to check for emails when user is creating a new user! And exposing whether an email domain is whitelisted or not might potentially be a security gap, thus we only want to limit it to users who can create users (which are Admins and likely more trusted users)

apps/studio/src/schemas/user.ts
```diff
@@ -0,0 +1,95 @@
...
+  adminType: ADMIN_TYPE.optional().default("agency"),
+})
+
+export const countUsersOutputSchema = z.number()
+
+export const hasInactiveUsersInputSchema = z.object({
+  siteId: z.number().min(1),
+})
+
+export const hasInactiveUsersOutputSchema = z.boolean()
```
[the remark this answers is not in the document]
adriangohjw: good point! have consolidated 2 queries together so it can be reused and its more simplified https://github.com/opengovsg/isomer/pull/1118/commits/18ac5d0043c8c611235a63a2c4d5efac5a531237

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: removed this code in https://github.com/opengovsg/isomer/pull/1118/commits/ec9bdc45dcb5b851daff556ab200d8529abf5002

context: this check here was added prior to https://github.com/opengovsg/isomer/pull/1156 where the unique DB constraint do not work. Since it should be handled by the DB now, we can go ahead to remove it

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: good point, updated in https://github.com/opengovsg/isomer/pull/1118/commits/65a31f7059d452dd6b86e1c9b76be96eb575c8fa!

apps/studio/src/server/modules/user/user.service.ts
adriangohjw: good point, updated in https://github.com/opengovsg/isomer/pull/1118/commits/65a31f7059d452dd6b86e1c9b76be96eb575c8fa!

seaerchin (approved):

PR #1122, opened 2025-02-19
isom-1741 USER MGMT (backend profile management service)

adriangohjw:
## Problem

Add user onboarding validation with Singapore phone number requirements.

Closes https://linear.app/ogp/issue/ISOM-1741/backend-profile-management-service

## Solution

Added user onboarding validation that checks for valid Singapore phone numbers and user details, along with necessary UI updates and utility functions.

**Breaking Changes**

**Features**:

- Added `isOnboarded` flag to useMe hook
- Added user onboarding validation with Singapore phone number check
- Added user details update endpoint with Singapore phone validation
- Added Singapore phone number validation utility

**Improvements**:

- Included phone number in default user selection

## Before & After Screenshots

N/A

## Tests

Added comprehensive test coverage:

- User onboarding validation tests
- Singapore phone number validation tests
- User details update endpoint tests


apps/studio/src/features/me/api/isUserOnboarded.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  // Check if the phone number contains any whitespace
+  if (/\s/.test(phone)) {
+    return false
+  }
+
+  // Check if it's exactly 8 digits with valid starting number
+  return /^[689]\d{7}$/.test(phone)
+}
```
seaerchin, line 22: [nit] i think the first check isn't required but it's fine

apps/studio/src/features/me/api/isUserOnboarded.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!name || !phone) {
+    return false
+  }
+
+  if (!isSingaporePhoneNumber(phone)) {
+    return false
+  }
+
+  return true
+}
```
seaerchin, line 34: nit: we can probably compress this into `return !!name && !!phone && isSingaporePhoneNumberPhone`

apps/studio/src/features/me/api/isUserOnboarded.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (/\s/.test(phone)) {
+    return false
+  }
+
+  // Check if it's exactly 8 digits with valid starting number
+  return /^[689]\d{7}$/.test(phone)
+}
+
+export const isUserOnboarded = ({ name, phone }: isUserOnboardedProps) => {
+  if (!name || !phone) {
```
seaerchin, line 25: [BACK-COMPAT] We might want to run a query against our db to see how many users have phones but not name (and vice versa) then fill in that data

apps/studio/src/schemas/user.ts
```diff
@@ -50,3 +50,26 @@ export const updateInputSchema = z.object({
...
+    .refine((phone) => {
+      const cleanPhone = phone.replace(/\s+/g, "")
+      return (
+        cleanPhone.startsWith("6") ||
+        cleanPhone.startsWith("8") ||
+        cleanPhone.startsWith("9")
+      )
+    }, "Phone number must start with 6, 8, or 9")
+    .transform((phone) => phone.replace(/\s+/g, "")), // Clean up any remaining whitespace
+})
```
seaerchin, line 73: [FYI] i think zod has a `transform` method that might be exactly what you're looking for! [This](https://zod.dev/?id=relationship-to-refinements) allows us to do the removal of whitespace one and makes the pipeline clearer

apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -173,4 +175,22 @@ export const userRouter = router({
...
+      // They should be able to update their own details even without any resource permissions
+
+      await db
+        .updateTable("User")
+        .where("id", "=", ctx.user.id)
+        .set({ name, phone })
+        .returningAll()
+        .executeTakeFirst()
+
+      return true
```
seaerchin, line 194: [nit] should we minimally return the new properties to frontend?

seaerchin (approved):
haven't looked at tests yet but so far lgtm - code doesn't seem to be used though? is that coming in a downstream PR?

apps/studio/src/features/me/api/isUserOnboarded.ts
adriangohjw: yup already ran the check and there's a few that we have to manually patch (for phone numbers) - part of the acceptance criteria before this epic goes live

apps/studio/src/schemas/user.ts
adriangohjw: ah thanks for sharing, didn't know about this! fixed in https://github.com/opengovsg/isomer/pull/1122/commits/552f4792e4bf7d36f493e6533899f32b5687751a

apps/studio/src/server/modules/user/user.router.ts
adriangohjw: done!

apps/studio/src/features/me/api/isUserOnboarded.ts
adriangohjw: thanks! in https://github.com/opengovsg/isomer/pull/1122/commits/afe1b04842dc7de3c6e530e50d9659b5f6dfbf1f

apps/studio/src/features/me/api/isUserOnboarded.ts
adriangohjw: thanks, makes sense! https://github.com/opengovsg/isomer/pull/1122/commits/7d11671349ea5e91c548b478e5786b88382c85ad

adriangohjw:
note - will not merge yet until security review

PR #1123, opened 2025-02-19
isom-1739 USER MGMT (backend email notification service)

adriangohjw:
## Problem

Implement email service with provider abstraction and invitation email templates.

Closes https://linear.app/ogp/issue/ISOM-1739/backend-email-notification-service

## Solution

**Breaking Changes**:

**Features**:

- Email service
- Invitation email templates with role-specific content
- Comprehensive error logging
- Test coverage for templates

## Before & After Screenshots

Screenshot 2025-02-19 at 9 41 07 PM

## Tests

- Invitation email template tests for different roles


apps/studio/src/features/mail/service.ts
```diff
@@ -0,0 +1,40 @@
+import type { InvitationEmailTemplateData } from "./templates"
+import { createBaseLogger } from "~/lib/logger"
+import { isValidEmail } from "~/utils/email"
+import { sendMail } from "../../lib/mail"
+import { templates } from "./templates/templates"
+
+const logger = createBaseLogger({ path: "lib/mail/service" })
```
seaerchin, line 7: i think this path might be `/features/mail/service`!

apps/studio/src/features/mail/service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function sendInvitation(
+  email: string,
+  data: InvitationEmailTemplateData,
+): Promise<void> {
+  if (!isValidEmail(email)) {
```
seaerchin, line 13: [FYI] no action required on your end, but a short coming of the `isValidEmail` function is that the email is still treated as just another string. we could make use of a `Tagged` type from `type-fest` to say then `email is Tagged<string, "email">` so we get type safety and don't have to repeat this check again, but not very important

apps/studio/src/features/mail/service.ts
```diff
@@ -0,0 +1,40 @@
...
+    logger.error({
+      error: "Invalid email format",
+      email,
+    })
+    throw new Error("Invalid email format")
+  }
+
+  const template = templates.invitation({
```
seaerchin, line 21: I really like how you've done this template grouping! it makes alot of sense - just wanted to ask, is there some base requirement that templates adhere to? don't have to do anything yet but we might benefit from making `templates` a mapping of that type so we know all templates share some base guarantees

apps/studio/src/features/mail/service.ts
```diff
@@ -0,0 +1,40 @@
+import type { InvitationEmailTemplateData } from "./templates"
+import { createBaseLogger } from "~/lib/logger"
+import { isValidEmail } from "~/utils/email"
+import { sendMail } from "../../lib/mail"
+import { templates } from "./templates/templates"
+
+const logger = createBaseLogger({ path: "lib/mail/service" })
+
+export async function sendInvitation(
```
seaerchin, line 9: [FYI] This is created on teh frontend - you can check the dockerfile to confirm this, not a

seaerchin (approved):
No big comments from me, just fixing the base path is ok alr

apps/studio/src/features/mail/service.ts
adriangohjw: nice catch, fixed in https://github.com/opengovsg/isomer/pull/1123/commits/8ca4497b3bf2d921d54866793dec3546966326a7 !

apps/studio/src/features/mail/service.ts
adriangohjw: makes sense, added some form of type requirements for `templates` to ensure it strictly follows a base template in https://github.com/opengovsg/isomer/pull/1123/commits/b1296ebe250ee9b544dc4854735b71c14dc23fe1

PR #1125, opened 2025-02-20
test(resource.service): add some preliminary tests for resource service

seaerchin:
## Problem
1. we have no tests for resource service so hard to have confidence

## Solution
1. add specs for **a small number of methods first**
    - will add to linear and see whether this get prioritised; just didn't want the old PR to be up for super long and accumulate merge conflicts


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -46,23 +46,6 @@ const defaultFooterSelect = [
   "Footer.content",
 ] satisfies SelectExpression<DB, "Footer">[]
 
-export const getPages = () => {
```
harishv7, line 49: are we not using these methods?

harishv7 (approved):
lgtm, just verify if the remove methods are unused

apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts
```diff
@@ -0,0 +1,663 @@
...
+    .executeTakeFirstOrThrow()
+
+  return db
+    .updateTable("Version")
+    .where("Version.id", "=", publishedVersionId)
+    .set({
+      blobId,
+    })
+    .executeTakeFirstOrThrow()
+  n
```
adriangohjw, line 662: to remove this causing lint issue

dcshzj:
Merging with bypass, commits are lacking verified signatures

PR #1126, opened 2025-02-20
[component] entire CollectionCard / BlogCard should be clickable

sehyunidaaa:
## Problem

Only the title field of the CollectionCard / BlogCard was clickable, which was unexpected behaviour to users.

Linear ticket: https://linear.app/ogp/issue/ISO-193/collection-card-and-block-card-touch-targets-are-limited

## Solution

- Wrap components in a linkcomponent, similar to InfoCards > Card implementation
- Apply group-hover to cards, so that when you hover on any parts of the card, the text always underlines

## Before & After

Only showing BlogCard. Change has also been made to CollectionCard.

Before:

After:

### DIFF
--- a/apps/studio/public/assets/css/preview-tw.css
+++ b/apps/studio/public/assets/css/preview-tw.css
@@ -4122,6 +4122,10 @@ video {
   color: rgb(21 71 190 / var(--tw-text-opacity));
 }
 
+.group:where([data-rac])[data-hovered] .group-hover\:underline {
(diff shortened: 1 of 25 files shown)


adriangohjw (approved):
update preview-tw.css + add hover state storybook for the cards

sehyunidaaa:
Yikes I keep forgetting the preview-tw.css. I will add it to my component checklist! Thank you @adriangohjw !

PR #1131, opened 2025-02-20
isom-1739 USER MGMT

adriangohjw:
closes https://linear.app/ogp/issue/ISOM-1731/user-management

refer to subbranches for ease of review
- https://github.com/opengovsg/isomer/pull/1123
  - https://github.com/opengovsg/isomer/pull/1161
- https://github.com/opengovsg/isomer/pull/1118
- https://github.com/opengovsg/isomer/pull/1122
- https://github.com/opengovsg/isomer/pull/1136
- https://github.com/opengovsg/isomer/pull/1137
- https://github.com/opengovsg/isomer/pull/1138
- https://github.com/opengovsg/isomer/pull/1142

# Things to check against DB
1. constraints from https://github.com/opengovsg/isomer/pull/1156
2. no records in `User` have invalid emails or phone number (phone number should be 8 digits starting with 6,8,9 and no empty spaces)


adriangohjw:
@harishv7 re: https://github.com/opengovsg/isomer/pull/1123, should we also have a check to make sure that the emails we are sending out, are to whitelisted emails and domains? To prevent human error of sending to incorrect people

harishv7:
> @harishv7 re: #1123, should we also have a check to make sure that the emails we are sending out, are to whitelisted emails and domains? To prevent human error of sending to incorrect people

Good point, For user management we already have defensive checks before someone can add the email. However, since this is more of a generic email service, we need to still have another defensive check in the service itself to only send to whitelisted emails

seaerchin (approved):

PR #1137, opened 2025-02-23
isom-1746 USER MGMT (frontend "delete user" modal)

adriangohjw:
## Problem

Implement user removal functionality in Studio, allowing admins to remove users from sites.

Closes https://linear.app/ogp/issue/ISOM-1746/frontend-delete-user-modal

## Solution

Added a new RemoveUserModal component and associated functionality to handle user removal from sites in Studio.

**Breaking Changes**:

**Features**:

- Added RemoveUserModal component for user management in Studio
- Implemented user removal functionality with confirmation dialog

**Improvements**:

- Simplified siteId handling in user info content query
- Moved date formatting utilities to shared users utils
- Enhanced error handling for user removal

## Tests

- Added Storybook stories for RemoveUserModal component


apps/studio/src/schemas/user.ts
```diff
@@ -30,7 +30,10 @@ export const deleteInputSchema = z.object({
   userId: z.string(),
 })
 
-export const deleteOutputSchema = z.boolean()
+export const deleteOutputSchema = z.object({
+  id: z.string(),
+  email: z.string().email(),
+})
```
seaerchin, line 36: curious - why define an output schema? wouldn't this run the risk of runtime errors next time we expand and return more stuff (and don't adhere to the zod contract here) + additionally, we validate on input because that's comign from an untrusted source

however, our backend is a known trusted source to our frontend, so validation isn't qutie as useful here

seaerchin (approved):
lgtm

apps/studio/src/schemas/user.ts
adriangohjw: > next time we expand and return more stuff (and don't adhere to the zod contract here)

IMO being more explicitly typed here let us make better use of typescript and its type-checking + it improves readability on the maintainer on what they can expect from this route

PR #1139, opened 2025-02-24
isom-1727 bug external link all should have new tab icon open in new

adriangohjw:
## Problem

For buttons and collection/blog cards, external links 1) do not open in new tab and 2) show "new tab" icon at the end

image

Closes https://linear.app/ogp/issue/ISOM-1727/bug-external-link-all-should-have-new-tab-icon-open-in-new-tab

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- add a check in `LinkButton`
- add a check in the `CollectionCard` and `BlogCard` + add stories

## Tests

1. go to a component with a button e.g. Hero. Add the link to an external link e.g. `https://google.com`.
    - [ ] It should show a "open in new link" icon
    - [ ] after building, when clicked on, it should open new link. (alternately: can inspect element on studio to check for `target: _blank`
2. go to a collection and create a collection link. Add the link to an external link e.g. `https://google.com`.
    - [ ] It should show a "open in new link" icon
    - [ ] after building, when clicked on, it should open new link. (alternately: can inspect element on studio to check for `target: _blank`

### DIFF
--- a/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
+++ b/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
@@ -76,8 +76,16 @@ export const Hover: Story = {
 }
 
 export const ExternalLink: Story = {
-  args: generateArgs({ isExternalLink: true }),
+  args: generateArgs({
+    isExternalLink: true,
+    title: "This is a not-so-long title that will be truncated",
+  }),
 }
+
+// TODO: fix the external link icon showing up when the text is long
+// export const ExternalLinkLongText: Story = {
+//   args: generateArgs({ isExternalLink: true }),
+// }
 
 export const UndefinedDate: Story = {
   args: generateArgs({ isLastUpdatedUndefined: true }),
--- a/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
+++ b/packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
@@ -22,6 +22,16 @@ const meta: Meta<typeof BlogCard> = {
 export default meta
 type Story = StoryObj<typeof BlogCard>
 
+interface GenerateArgsProps {
+  shouldShowDate?: boolean
+  isLastUpdatedUndefined?: boolean
+  withoutImage?: boolean
+  title?: string
+  description?: string
+  tags?: Tag[]
+  isExternalLink?: boolean
+}
+
 const generateArgs = ({
   shouldShowDate = true,
   isLastUpdatedUndefined = false,
@@ -29,14 +39,10 @@ const generateArgs = ({
   title = "A journal on microscopic plastic and their correlation to the number of staycations enjoyed per millennials between the ages of 30-42, substantiated by research from IDK university",
   description = "We've looked at how people's spending correlates with how much microscopic plastic they consumed over the year. We've looked at how people's spending correlates with how much microscopic plastic they consumed over the year.",
   tags = [],
-}: {
+  isExternalLink = false,
+}: GenerateArgsProps): Partial<CollectionCardProps> & {
   shouldShowDate?: boolean
-  isLastUpdatedUndefined?: boolean
-  withoutImage?: boolean
-  title?: string
-  description?: string
-  tags?: Tag[]
-}): Partial<CollectionCardProps> & { shouldShowDate?: boolean } => {
+} => {
   return {
     lastUpdated: isLastUpdatedUndefined ? undefined : "December 2, 2023",
     category: "Research",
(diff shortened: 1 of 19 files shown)


packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.stories.tsx
```diff
@@ -68,6 +73,18 @@ export const Hover: Story = {
...
 }
 
+export const ExternalLink: Story = {
+  args: generateArgs({
+    isExternalLink: true,
+    title: "This is a not-so-long title that will be truncated",
+  }),
+}
+
+// TODO: fix the external link icon showing up when the text is long
```
harishv7, line 83: didn't understand this comment, what's the current behaviour when the text is long?

packages/components/src/templates/next/components/internal/BlogCard/BlogCard.stories.tsx
```diff
@@ -22,21 +21,27 @@ const meta: Meta<typeof BlogCard> = {
...
+  isExternalLink = false,
+}: GenerateArgsProps): Partial<CollectionCardProps> & {
   shouldShowDate?: boolean
-  isLastUpdatedUndefined?: boolean
-  withoutImage?: boolean
-  title?: string
-  description?: string
-  tags?: Tag[]
-}): Partial<CollectionCardProps> & { shouldShowDate?: boolean } => {
+} => {
```
adriangohjw, line 44: slight refactoring here since props is getting larger

packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.stories.tsx
```diff
@@ -21,21 +21,26 @@ const meta: Meta<typeof CollectionCard> = {
...
+  isExternalLink = false,
+}: GenerateArgsProps): Partial<CollectionCardProps> & {
   shouldShowDate?: boolean
-  isLastUpdatedUndefined?: boolean
-  withoutImage?: boolean
-  title?: string
-  description?: string
-  tags?: Tag[]
-}): Partial<CollectionCardProps> & { shouldShowDate?: boolean } => {
+} => {
```
adriangohjw, line 43: slight refactoring here since props is getting larger

packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.stories.tsx
adriangohjw: my bad, should have been clearer

| current | expected |
| - | - |
| image | image |

couldn't find a quick easy way to do this so leaving it as a TODO - esp. since not that common for titles to be so long

adriangohjw: i was hesistating on this too because it's not the most ideal, tho you might be right that it's better-than-nothing. will make the UI change, but still keep the commment to explain the desired outcome!

adriangohjw:
note - changes from https://github.com/opengovsg/isomer/pull/1139/commits/d02f2f41310ebe715287ed2f14982da4a9166a94 and https://github.com/opengovsg/isomer/pull/1139/commits/156b7f7d2fc95e45035b7434e0c6bd8d91a0ee3d are such that the default UI in most storybook will not have the external link icon

packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.stories.tsx
harishv7: Ok, cc @sehyunidaaa fyi

adriangohjw: realized it's not as simple to apply `line-clamp-3` to the text span as that would mean the icon is always on the next line, which isn't ideal when the title is too short and not truncated.

instead, updated it (https://github.com/opengovsg/isomer/pull/1139/commits/e1fdb2d52ec8a9ae5a86a99f400fd712e6f34c40, https://github.com/opengovsg/isomer/pull/1139/commits/4cf732c7f1f2e1519ba9e9892239fcd20f456b18, https://github.com/opengovsg/isomer/pull/1139/commits/f0e2e9d83a938560d556c81fd39b50f77fc268bf) such that
- there's a `useEffect` to check if the text is truncated before deciding the location of the icon
- and because it's a useEffect, we will need `use client` so i further split it into 2 `InternalLinkTitle` (rendered during SSG) and `externalLinkTitle` (rendered during client side)

harishv7: How do I check this on storybook? Tried on [this](https://665439911775ce21bf8b8fb4-uqoozszvjs.chromatic.com/?path=/story/next-internal-components-collectioncard--external-link&globals=viewport:sm), but still was unable to see the icon

adriangohjw: https://665439911775ce21bf8b8fb4-uqoozszvjs.chromatic.com/?path=/story/next-internal-components-blog-card--external-link-long-text&globals=viewport:sm

try this?

harishv7 (approved):

PR #1140, opened 2025-02-25
Enhance/collection alphabetical sorting

adriangohjw:
## Problem

Collections are sorted in reverse alphabetical order

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- refactor by extracting into standalone file
- use `localeCompare()` to better compare title

**Bug Fixes**:

- use `getTime()` when comparing time to ensure comparing value and not reference

## Before & After Screenshots

| before | after |
| - | - |
| image | image |

1. create a collection with 2 items. Both items should have same date and different title e.g. `Zebra` and `Yoyo`
2. publish the resources and site
3. collection should show `yoyo` before `zebra`


packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
```diff
@@ -0,0 +1,104 @@
...
+    // Rank items with no dates last
+    if (a.rawDate === undefined) {
+      return 1
+    }
+
+    if (b.rawDate === undefined) {
+      return -1
+    }
+
+    return a.rawDate.getTime() < b.rawDate.getTime() ? 1 : -1
```
adriangohjw, line 102: this part contains the main changes (the others are refactoring and extracting the into a standalone component)

packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    // Rank items with no dates last
+    if (a.rawDate === undefined) {
+      return 1
+    }
+
+    if (b.rawDate === undefined) {
+      return -1
+    }
```
seaerchin, line 100: what happens if the `rawDates` are both undefined? i think we should tie break by title here too?

seaerchin (approved):
lgtm but 1 minor extension - right now it looks like we tie break only when both dates are present but it feels like we should also tie break when both dates are absent

packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
adriangohjw: good point, thanks for pointing it out!

added it https://github.com/opengovsg/isomer/pull/1140/commits/711f51e9029e18e7c235b74c3312467359128bed

PR #1148, opened 2025-03-01
fix aspect ratio for collectionCards / contentPic

sehyunidaaa:
## Problem

Aspect ratio for components were variable for every screen size and content

## Solution

Fix aspect ratio for collection thumbnail and contentpic:

image

Infocards will be dealt with on a different PR.

### DIFF
--- a/packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
+++ b/packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
@@ -29,7 +29,7 @@ export const CollectionCard = ({
     <Link
       LinkComponent={LinkComponent}
       href={referenceLinkHref}
-      className="group flex border-collapse flex-col gap-3 border-b border-divider-medium py-5 first:border-t lg:flex-row lg:gap-6"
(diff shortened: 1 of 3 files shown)


packages/components/src/templates/next/components/internal/CollectionCard/CollectionCard.tsx
```diff
@@ -59,12 +59,12 @@ export const CollectionCard = ({
...
         )}
         {/* TODO: Feature enhancement? Filter by category when clicked */}
-        <p className="prose-label-md mt-3 text-base-content-subtle lg:mt-2">
+        <p className="prose-label-md mt-3 text-base-content-subtle md:mt-2">
           {category}
         </p>
       </div>
       {image && (
-        <div className="relative mt-3 min-h-40 w-full shrink-0 lg:ml-4 lg:mt-0 lg:h-auto lg:w-1/4">
+        <div className="relative mt-3 h-[160px] w-[200px] shrink-0 md:ml-4 md:mt-0">
```
seaerchin, line 67: double checking because this (`h-[160px] w-[200px]`) changes the styling directly so wanted to check if it's intended. ok with the change in breakpoint from `lg -> md`

seaerchin (approved):

PR #1156, opened 2025-03-05
fix unique constraint on user and resourcepermission

adriangohjw:
# Problem

The current unique index for ResourcePermission in Prisma, defined as `@@unique([userId, siteId, resourceId, role])`, has three key issues:

## 1. Standard Unique Constraints Treat NULL as Distinct, Allowing Duplicate Permissions

Unique constraints in SQL consider each `NULL` value as distinct, which allows multiple records with `NULL` values in the same indexed columns.

For example, both of these records can exist together:
```sql
{ "siteId": 1, "userId": 1, "role": "Admin", "resourceId": NULL }
{ "siteId": 1, "userId": 1, "role": "Editor", "resourceId": NULL }
```

This creates ambiguity in determining the correct permission (Admin or Editor). While we might assume that the highest permission should apply (Admin > Editor), this assumption may not always hold true.

It also complicates database queries since we would always need to check for and resolve this edge case.

### Solution: Use `NULLS NOT DISTINCT` to treat `NULL` values as equal.

Note: This feature is only available in PostgreSQL 15 and later.

This is also unsupported in Prisma, thus we have to insert a custom migration. This is referencing how https://github.com/opengovsg/isomer/pull/642 did it

## 2. Including "role" in the Unique Index Allows Multiple Role Entries

Since `role` is part of the unique index, multiple records with different roles can exist for the same `(siteId, userId, resourceId)`.

Example:
```sql
{ "siteId": 1, "userId": 1, "resourceId": 1, "role": "Admin" }
{ "siteId": 1, "userId": 1, "resourceId": 1, "role": "Editor" }
```

This results in the same issue as above: ambiguity in determining the correct role for a user.

### Solution: Remove `role` from the unique index.

This ensures that for a given `(siteId, userId, resourceId)`, only one entry exists.

## 3. Lack of a Unique Constraint on `deletedAt` Prevents Record Recreation

The system uses soft deletion by setting a timestamp in the deletedAt field. However, the current unique constraint does not include deletedAt, preventing previously deleted records from being reinserted.

Example:
```sql
-- When this exists
{ "siteId": 1, "userId": 1, "resourceId": 1, "role": "Editor", "deletedAt": "some_timestamp" }

-- This new record cannot be created:
{ "siteId": 1, "userId": 1, "resourceId": 1, "role": "Editor", "deletedAt": NULL }
```

### Solution: Add `deletedAt` to the unique index

This allows re-creation of soft-deleted records.

# Breaking Changes

There are no breaking changes, but a database audit is required to ensure compatibility.

This can be verified by running the following queries and confirming that both return 0 results.

## Check for Duplicate ResourcePermissions
```sql
SELECT "siteId", "userId", "resourceId", "deletedAt", COUNT(*)
FROM "ResourcePermission"
GROUP BY "siteId", "userId", "resourceId", "deletedAt"
HAVING COUNT(*) > 1;
```

## Check for Duplicate Users
```sql
SELECT "email", "deletedAt", COUNT(*)
FROM "User"
GROUP BY "email", "deletedAt"
HAVING COUNT(*) > 1;
```

### DIFF
--- a/apps/studio/prisma/seed.ts
+++ b/apps/studio/prisma/seed.ts
@@ -39,7 +39,7 @@ async function main() {
         })
         .onConflict((oc) =>
           oc
-            .column("email")
+            .columns(["email", "deletedAt"])
             .doUpdateSet((eb) => ({ email: eb.ref("excluded.email") })),
         )
         .returning(["id", "name"])
--- a/apps/studio/prisma/migrations/20250305210834_fix_unique_indexes_on_user_and_resourcepermission/migration.sql
+++ b/apps/studio/prisma/migrations/20250305210834_fix_unique_indexes_on_user_and_resourcepermission/migration.sql
@@ -1,0 +1,18 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[userId,siteId,resourceId,deletedAt]` on the table `ResourcePermission` will be added. If there are existing duplicate values, this will fail.
+  - A unique constraint covering the columns `[email,deletedAt]` on the table `User` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- DropIndex
+DROP INDEX "ResourcePermission_userId_siteId_resourceId_role_key";
+
+-- DropIndex
+DROP INDEX "User_email_key";
+
+-- CreateIndex
+CREATE UNIQUE INDEX "ResourcePermission_userId_siteId_resourceId_deletedAt_key" ON "ResourcePermission"("userId", "siteId", "resourceId", "deletedAt");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "User_email_deletedAt_key" ON "User"("email", "deletedAt");
--- a/apps/studio/prisma/migrations/20250305210937_custom_migration/migration.sql
+++ b/apps/studio/prisma/migrations/20250305210937_custom_migration/migration.sql
@@ -1,9 +1,3 @@
--- Override Prisma's unique index with unique NOT DISTINCT index
-DROP INDEX IF EXISTS "Resource_siteId_parentId_permalink_key";
-
-CREATE UNIQUE INDEX IF NOT EXISTS "Resource_siteId_parentId_permalink_key" ON "Resource"("siteId", "parentId", "permalink") NULLS NOT DISTINCT;
-
-
 ---------------------------------
 -- This migration fixes uniqueness constraints for User and ResourcePermission.
 -- Reference: https://github.com/opengovsg/isomer/pull/1155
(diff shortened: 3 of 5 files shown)


apps/studio/src/server/modules/site/__tests__/site.router.test.ts
```diff
@@ -102,30 +102,6 @@ describe("site.router", async () => {
       expect(result).toEqual([])
     })
 
-    it("should only show a site once if there are multiple permissions for the same site and user", async () => {
```
adriangohjw, line 105: no longer needed because it's no longer possible for this to happen

apps/studio/prisma/custom/migration.sql
```diff
@@ -1,4 +1,19 @@
 -- Override Prisma's unique index with unique NOT DISTINCT index
 DROP INDEX IF EXISTS "Resource_siteId_parentId_permalink_key";
```
harishv7, line 2: @adriangohjw Overall great points. Some things to take note:

1. Is our application logic also enforcing one role per resource per user before it reaches DB layer?

2. We need to run an audit of the current DB state to ensure the new unique constraints can be created without conflict

3. Double check postgres version running on infra to be >= 15

4. For the new NULLS NOT DISTINCT, this will mean that NULL != NULL

For example, if previously, we have queries like `SELECT * FROM "ResourcePermission" WHERE "resourceId" IS NULL;` how will this change affect these queries? We do need to look through the codebase for such usages

For example, I could see such usages in `apps/studio/src/server/modules/permissions/permissions.service.ts`

I believe our tests should provide some level of confidence for this, but worth to look through as well

5. Do have a migration rollback plan, if we notice something odd after deploy, how can we enable a backwards compatible rollback? We will need to generate a down migration - can jot these down as part of deploy process

adriangohjw: > We need to run an audit of the current DB state to ensure the new unique constraints can be created without conflict

yup, added that to the PR description. If all is good, I will clean them up first before merging this in

> Double check postgres version running on infra to be >= 15

yup, we are using 15.6 on PROD + there's aldy existing `nulls not distinct` used in one other part of the app

> For example, if previously, we have queries like SELECT * FROM "ResourcePermission" WHERE "resourceId" IS NULL; how will this change affect these queries? We do need to look through the codebase for such usages

the `NULLS NOT DISTINCT` change only affects uniqueness constraints, not query behavior. So queries like `SELECT * FROM "ResourcePermission" WHERE "resourceId" IS NULL;` won't be affected

adriangohjw: added the rollback plan in the description

apps/studio/prisma/seed.ts
adriangohjw, line 1: updated so that we re-use the existing `addUsersToSite` script that's being used for single source of truth

harishv7 (approved):
lgtm, just make sure staging, prod env do not have conflicts before running migrations

adriangohjw:
update for STG: backed up RDS snapshot + cleaned up the records

adriangohjw:
update for PROD: backed up RDS snapshot

PR #1168, opened 2025-03-11
chore: add csv parsing

seaerchin:
## Problem
prod ops give user info in a gsheets, which is convenient to export as csv - this, however, si tedious to type out so having a csv parser is useful

## Solution
add csv parsing

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -114,6 +114,7 @@
     "@trpc/server": "10.45.2",
     "@uidotdev/usehooks": "^2.4.1",
     "ajv": "^8.17.1",
+    "csv-parser": "^3.2.0",
     "date-fns": "^4.1.0",
     "date-fns-tz": "^3.1.3",
     "dd-trace": "^5.24.0",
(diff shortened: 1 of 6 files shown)


apps/studio/prisma/scripts/data.csv
adriangohjw, line 1: maybe `addUsersToSite.csv` instead so it's clearer it's for that script?

apps/studio/package.json
```diff
@@ -114,6 +114,7 @@
     "@trpc/server": "10.45.2",
     "@uidotdev/usehooks": "^2.4.1",
     "ajv": "^8.17.1",
+    "csv-parser": "^3.2.0",
```
adriangohjw, line 117: unrelated but im starting to see a benefit of splitting out `scripts` into a standalone project i this monorepo instead of having Studio package.json bloats up ahaha

adriangohjw (commented):
actually im not sure if we should continue to make improvement to this script post-user management since these should be done by users themselves?

seaerchin:
> actually, im not sure if we should continue to make improvement to this script post-user management since these should be done by users themselves? esp. many sites have 10 or less users
> 
> Adding this kinda set a precedent that user managament should continue to be supported by Isomer

i don't mind closing this AHHA i just raised a PR cos i added this functionality to create a few sites. i'll close first adn we can revisit if needed!

PR #1169, opened 2025-03-12
feat(audit-logs): add initial service + methods to log various events

seaerchin:
## Problem
see related issue

## Solution
1. add various service methods to log different events
    - i didn't add `metadata` because i'm unsure what qualifies but feel free to extend the methods to add them yourself

### DIFF
--- a/apps/studio/prisma/types.ts
+++ b/apps/studio/prisma/types.ts
@@ -34,9 +34,18 @@ declare global {
       _IsomerSiteWideComponentsProps["footerItems"],
       "JSONB"
     >
-    interface AuditLogDeltaJsonContent {
-      before: Record<string, unknown> | null
-      after: Record<string, unknown> | null
+    interface NewLogEvent {
+      before: null
+      after: Record<string, unknown>
     }
+    interface DeleteLogEvent {
+      before: Record<string, unknown>
+      after: null
+    }
+    interface FullLogEvent {
+      before: Record<string, unknown>
+      after: Record<string, unknown>
+    }
+    type AuditLogDeltaJsonContent = FullLogEvent | NewLogEvent | DeleteLogEvent
   }
 }
--- a/apps/studio/prisma/types.ts
+++ b/apps/studio/prisma/types.ts
@@ -34,9 +34,9 @@ declare global {
       _IsomerSiteWideComponentsProps["footerItems"],
       "JSONB"
     >
-    type AuditLogDeltaJsonContent = Tagged<
-      { before: Record<string, unknown>; after: Record<string, unknown> },
-      "JSONB"
-    >
+    interface AuditLogDeltaJsonContent {
+      before: Record<string, unknown> | null
+      after: Record<string, unknown> | null
+    }
   }
 }
--- a/apps/studio/prisma/types.ts
+++ b/apps/studio/prisma/types.ts
@@ -34,7 +34,7 @@ declare global {
       _IsomerSiteWideComponentsProps["footerItems"],
       "JSONB"
     >
-    interface NewLogEvent {
+    interface CreateLogEvent {
       before: null
       after: Record<string, unknown>
     }
@@ -46,6 +46,9 @@ declare global {
       before: Record<string, unknown>
       after: Record<string, unknown>
     }
-    type AuditLogDeltaJsonContent = FullLogEvent | NewLogEvent | DeleteLogEvent
+    type AuditLogDeltaJsonContent =
+      | FullLogEvent
+      | CreateLogEvent
+      | DeleteLogEvent
   }
 }
--- a/apps/studio/src/server/modules/audit/audit.service.ts
+++ b/apps/studio/src/server/modules/audit/audit.service.ts
@@ -7,18 +7,30 @@ import type {
   Resource,
   Transaction,
   User,
+  Version,
 } from "../database"
+
+type FullResource = Resource & (Blob | undefined)
+
+interface CreateDelta {
+  before: null
+  after: FullResource
+}
+interface DeleteDelta {
+  before: FullResource
+  after: null
+}
+interface UpdateDelta {
+  before: FullResource
+  after: FullResource
+}
 
 interface ResourceEventLogProps {
   eventType: Extract<
     AuditLogEvent,
     "ResourceCreate" | "ResourceUpdate" | "ResourceDelete" | "ResourceMove"
   >
-  delta: {
-    // NOTE: Can be undefined because folders/collections don't have blobs
-    before: Resource & (Blob | undefined)
-    after: Resource & (Blob | undefined)
-  }
+  delta: CreateDelta | DeleteDelta | UpdateDelta
   by: User
   ip?: string
 }
@@ -61,5 +73,30 @@ export const logAuthEvent: AuditLogger<AuthEventLogProps> = async (
   await tx
     .insertInto("AuditLog")
     .values({ eventType, delta, userId: by.id, ipAddress: ip })
+    .execute()
+}
+
+interface PublishEventLogProps {
+  by: User
+  delta: {
+    // NOTE: `null` if this is the first publish
+    before: (FullResource & { versionNumber: number }) | null
+    after: FullResource & { versionNumber: number }
+  }
+  eventType: Extract<AuditLogEvent, "Publish">
+  ip?: string
+}
+export const logPublishEvent: AuditLogger<PublishEventLogProps> = async (
+  tx,
+  { by, delta, eventType, ip },
+) => {
+  await tx
+    .insertInto("AuditLog")
+    .values({
+      eventType,
+      delta,
+      userId: by.id,
+      ipAddress: ip,
+    })
     .execute()
 }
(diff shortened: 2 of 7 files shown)


seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1169?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1170** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1169** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1167** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1167?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/studio/prisma/types.ts
```diff
@@ -34,9 +34,18 @@ declare global {
...
+    }
+    interface DeleteLogEvent {
+      before: Record<string, unknown>
+      after: null
+    }
+    interface FullLogEvent {
+      before: Record<string, unknown>
+      after: Record<string, unknown>
+    }
+    type AuditLogJsonContent = FullLogEvent | NewLogEvent | DeleteLogEvent
```
dcshzj, line 49: nit: maybe can call it CreateLogEvent, DeleteLogEvent and UpdateLogEvent? To align with the naming of CRUD + what you named it in the audit logs service.

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -0,0 +1,181 @@
...
+  { eventType, delta, by, ip },
+) => {
+  await tx
+    .insertInto("AuditLog")
+    .values({ eventType, delta, userId: by.id, ipAddress: ip })
+    .execute()
+}
+
+interface LoginDelta {
+  before: VerificationToken
```
dcshzj, line 51: issue: should this be `null`? Seems like we are _creating_ a user session rather than _updating_.

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -0,0 +1,181 @@
...
+  after: VerificationToken
```
dcshzj, line 52: suggestion: was thinking perhaps we should also use `User` in the login to make it similar to the logout? This might make it more logical when comparing between the session creation and session destruction.

Separately, I think we might need to expand what the User object has to capture the session information, because it might be possible for a user to create multiple sessions across different devices. Ideally during login, we can give the session an ID, then use it to identify which session is being destroyed later.

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// NOTE: logout just calls `session.destroy` and we only have
+// `userId` or sgid info in session.
```
dcshzj, line 56: question(non-blocking): wanted to get confirmation, we won't be logging session expiries yeah?

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -0,0 +1,181 @@
...
+  Blob,
+  DB,
+  Resource,
+  ResourcePermission,
+  Transaction,
+  User,
+  VerificationToken,
+} from "../database"
+
+type FullResource = Resource & (Blob | undefined)
```
dcshzj, line 12: question: seems like we might be omitting changes to site config, navbar and footer? Or should they be captured as a separate AuditLogEvent?

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -0,0 +1,181 @@
...
+    // We don't want to store the `version` because it is a pointer
+    // to the blob/resource
+    // we will instead store the full data here so it is an accurate snapshot
+    before: (FullResource & { versionNumber: number }) | null
+    after: FullResource & { versionNumber: number }
+  }
+  eventType: Extract<AuditLogEvent, "Publish">
+  ip?: string
+}
+export const logPublishEvent: AuditLogger<PublishEventLogProps> = async (
```
dcshzj, line 91: question: related to the site config, navbar and footer question above, how will we be recording publishes due to changes in those records?

dcshzj, line 91: question(non-blocking): will resource moves, which we currently auto-publish, be recorded as two separate audit log events? I am guessing yes, but wanted to confirm with you.

apps/studio/prisma/types.ts
seaerchin: sounds good

apps/studio/src/server/modules/audit/audit.service.ts
seaerchin: eh not quite! this is because our login is a 2 step process
1. request for otp (audit log not called, verification token created)
2. verify otp (audit log called, verification token updated) 

in other words, we're discarding step 1 here - this is because we rate limit + no user has actually logged in yet, so i felt this wasn't useful to keep. 

lmk if you have the same thoughts

apps/studio/src/server/modules/audit/audit.service.ts
seaerchin: the `VerificationToken` here consists of `user.email` with the `ip` (see ss). because of this, it's possible for the user to have multiple session over devices **but** this will be captured under teh `ip` section (distinct ip) though there is a short-coming in that the user isn't logged directly; i'll add it in!

image

apps/studio/src/server/modules/audit/audit.service.ts
seaerchin: session expiry calls `logout` isn't it?

apps/studio/src/server/modules/audit/audit.service.ts
seaerchin: oh good catch - for now, we don't have user capability to add it in so it'll have to be through the admin endpoint. i think it'll be a different `AuditLogEvent` because we want to capture the full db item

apps/studio/src/server/modules/audit/audit.service.ts
seaerchin: yes, i think each "event" should be recorded as a unit of change to some table. since this involves 2 tables, we shud log as 2 events

seaerchin: good question - i actually missed this out. lemme think abit more HAHA

dcshzj: Yup thanks! Means my understanding is the same.

apps/studio/src/server/modules/audit/audit.service.ts
dcshzj: Yup I think when we should create an audit log makes sense and aligned with what I understand as well.

But I'm a bit not quite understanding about the verification token part. When the user successfully verifies the OTP, we are recording the `before` with the verification token + the total number of attempts made, then the `after` would be the same verification token + the total number of attempts made, plus the User object. However, if we are reading only the AuditLogs table, this means that there is supposed to be an earlier entry that is missing.

So this means that the state of items are meant to be something like NULL -> Object -> NULL, and the state changes are the audit log events:
1. NULL -> Object indicates creation (in this case a session is created, i.e. login)
2. Object -> NULL indicates deletion (in this case the session is destroyed, i.e. logout)

If we are representing Login as a session creation (which happens only after successful verification of OTP), then it should be the case of `before` being NULL and `after` being the object we want to store.

apps/studio/src/server/modules/audit/audit.service.ts
dcshzj: Hmm I think not really. The session is set using a cookie with an expiry date, so when the cookie expires it is automatically removed by the browser, and Studio treats the user as if they have never logged in before, so `logout` is not called.

seaerchin: eh yea you're right, i m mistaken!

apps/studio/src/server/modules/audit/audit.service.ts
dcshzj: Clarified offline, intent is more to capture DB-level changes and regardless of the `before` and `after`, we still fulfil the agency's intent of capturing log ins via the Login AuditLogEvent.

Caveat is that the number of attempts in the verification token should be X in `before` and X+1 in `after`. Meaning if the user enters the OTP correctly on the first try, `before` should be recorded as 0 and `after` should be recorded as 1.

dcshzj (approved):

seaerchin:
### Merge activity

* **Mar 19, 6:32 AM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1169).

PR #1172, opened 2025-03-14
Fix/collection card empty description

adriangohjw:
## Problem

Summary field is compulsory but agencies can enter pure whitespaces. In those case, it will look ugly and we say no to ugly sites!!!!

Closes https://opengovproducts.slack.com/archives/C083KGWTX5X/p1741932388403459

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- check if its "empty" using trim + add storybook

## Tests

1. create a collection -> create a article page -> set summary to pure white spaces e.g. `   ` 
2. publish collection and check that there's no awkward gap in the UI (refer to storybook)


adriangohjw:
> Do we intend to apply the same logic to Content page?

@sehyunidaaa hmmm makes sense, should we?

seaerchin (approved):

packages/components/src/templates/next/layouts/Article/Article.stories.tsx
adriangohjw, line 1: change is to extract args generation logic into a reusable function + add new storybook

PR #1173, opened 2025-03-17
feat: add script to migrate Isomer Classic to Next

dcshzj:
## Problem

We have many sites that we will need to migrate in the future, and we will need to be able to scale up this migration process.

## Solution

**Breaking Changes**

**Features**:

- Added a new script that is able to automatically convert an Isomer Classic site into Isomer Next.
    - A migration report will also be created for the agency to self-rectify.

**Not done yet**:

- Automatically importing the site contents into Isomer Studio.


seaerchin:
need to update for tag categories also

PR #1175, opened 2025-03-17
feat(audit-logging): add audit logging to collection router + tests 

seaerchin:
## Problem
See the attached issue for the problem 

## Solution
- add audit logging to folder router 

## Bug-fixes
- constrain `collectionRouter.updateCollectionLink` to only update collection links
- fix incorrect handling for `updateBlobById` when the resource is published
    - we create a `draftBlob` correctly but we don't use the `draftBlobId` of the newly created blob but isntead, the stale value of `page.draftBlobId` (which is `null`)
    - previously, we returned a `UpdateResult`, which led to a silent failure. this has been updated to `returningAll` because we should always expect >= 1 result
- removed duplicate (and incorrect) updates of `updateCollectionLink`

PR #1176, opened 2025-03-18
feat: add audit logging to email router with tests

dcshzj:
## Problem

Closes ISOM-1773.
Closes ISOM-1790.

## Solution

**Breaking Changes**

**Features**:

- Added audit logging capabilities to the email router.

**Improvements**:

- Changed the DB queries on me.router.ts to use via Kysely instead of Prisma directly.

## Tests

- [ ] View the AuditLogs table using TablePlus on staging.
- [ ] Perform a successful login to Isomer Studio using an email address that has not logged in before.
- [ ] Verify that the AuditLogs has 2 new entries - UserCreate and Login events

### DIFF
--- a/apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
+++ b/apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
@@ -144,14 +144,13 @@ describe("auth.email", () => {
       })
 
       // Act
-      try {
-        await caller.verifyOtp({
+      await expect(
+        caller.verifyOtp({
           email: TEST_VALID_EMAIL,
           token: INVALID_OTP,
-        })
-      } catch (e) {
-        // Discard error since it is expected
-      }
+        }),
+      ).rejects.toThrowError()
+
       const result = caller.verifyOtp({
         email: TEST_VALID_EMAIL,
         token: VALID_OTP,
--- a/apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
+++ b/apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
@@ -129,8 +129,7 @@ describe("auth.email", () => {
       const auditLogs = await db.selectFrom("AuditLog").selectAll().execute()
       expect(auditLogs).toHaveLength(1)
       expect(auditLogs[0]?.eventType).toBe(AuditLogEvent.Login)
-      expect(auditLogs[0]?.delta.before!.attempts).toBe(0)
-      expect(auditLogs[0]?.delta.after!.attempts).toBe(1)
+      expect(auditLogs[0]?.delta.before!.attempts).toBe(1)
     })
 
     it("should successfully set session on a subsequent valid OTP", async () => {
@@ -145,10 +144,14 @@ describe("auth.email", () => {
       })
 
       // Act
-      await caller.verifyOtp({
-        email: TEST_VALID_EMAIL,
-        token: VALID_OTP,
-      })
+      try {
+        await caller.verifyOtp({
+          email: TEST_VALID_EMAIL,
+          token: INVALID_OTP,
+        })
+      } catch (e) {
+        // Discard error since it is expected
+      }
       const result = caller.verifyOtp({
         email: TEST_VALID_EMAIL,
         token: VALID_OTP,
@@ -167,8 +170,7 @@ describe("auth.email", () => {
       const auditLogs = await db.selectFrom("AuditLog").selectAll().execute()
       expect(auditLogs).toHaveLength(1)
       expect(auditLogs[0]?.eventType).toBe(AuditLogEvent.Login)
-      expect(auditLogs[0]?.delta.before!.attempts).toBe(1)
-      expect(auditLogs[0]?.delta.after!.attempts).toBe(2)
+      expect(auditLogs[0]?.delta.before!.attempts).toBe(2)
     })
 
     it("should throw if OTP is not found", async () => {
@@ -180,7 +182,9 @@ describe("auth.email", () => {
       })
 
       // Assert
-      await expect(result).rejects.toThrowError("Invalid login email")
+      await expect(result).rejects.toThrowError(
+        "Please request for another OTP",
+      )
       await expect(
         db.selectFrom("AuditLog").selectAll().execute(),
       ).resolves.toHaveLength(0)
--- a/apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
+++ b/apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
@@ -8,6 +8,7 @@ import { describe, expect, it } from "vitest"
 
 import { env } from "~/env.mjs"
 import * as mailLib from "~/lib/mail"
+import { AuditLogEvent, db } from "~/server/modules/database"
 import { prisma } from "~/server/prisma"
 import { createTokenHash } from "../../auth.util"
 import { emailSessionRouter } from "../email.router"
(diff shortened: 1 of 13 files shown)


apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
```diff
@@ -100,6 +101,7 @@ describe("auth.email", () => {
 
     it("should successfully set session on valid OTP", async () => {
```
seaerchin, line 102: [optional] maybe we want to have a test asserting the wrong OTP -> correct OTP flow, to ensure that our audit logs do capture # of attempts

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -79,6 +86,21 @@ export const emailSessionRouter = router({
...
+          },
+        },
+      )
+
+      if (!oldVerificationToken) {
+        throw new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Invalid login email",
+        })
+      }
```
seaerchin, line 102: [nit] this might be abit confusing to our users because the fingerprint is `email | ip`! hence, this case only happens when their ip chnages. wdyt about "Please request for another OTP"

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -95,34 +117,32 @@ export const emailSessionRouter = router({
...
+            before: {
+              token: oldVerificationToken,
             },
-            select: defaultMeSelect,
-          })
+            after: {
+              token: {
+                ...oldVerificationToken,
+                attempts: oldVerificationToken.attempts + 1,
+              },
```
seaerchin, line 136: not quite indicative of what's in our db - we delete the token on successful login. you can check this in `verifyToken`; because our audit logs are meant to be as reflective of our "state of the world" (ie, our db), wdyt about setting `after: null`

apps/studio/src/server/modules/auth/email/__tests__/email.service.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      const user = await db.transaction().execute(async (tx) => {
+        return upsertUser({ tx, email: TEST_EMAIL })
+      })
+
+      // Assert
+      const auditLogs = await db
+        .selectFrom("AuditLog")
+        .selectAll()
+        .executeTakeFirst()
+      expect(user).toBeDefined()
```
seaerchin, line 46: nit: maybe should assert hte `emailName` portion but very minor

seaerchin (commented):
i think we should fix the `verificationToken` on login, otherwise it's ok

apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
```diff
@@ -123,6 +125,13 @@ describe("auth.email", () => {
       await expect(result).resolves.toMatchObject(expectedUser)
       // Session should have been set with logged in user.
       expect(session.userId).toEqual(expectedUser.id)
+      // Audit log should have been created.
+      const auditLogs = await db
+        .selectFrom("AuditLog")
+        .selectAll()
+        .executeTakeFirst()
+      expect(auditLogs).toBeDefined()
+      expect(auditLogs?.eventType).toBe(AuditLogEvent.Login)
```
adriangohjw, line 134: thoughts on
- checking that only 1 record is created? ref: https://github.com/opengovsg/isomer/pull/1175#discussion_r2004845813
- also that the JSON logged is correct as much as possible. ref: https://github.com/opengovsg/isomer/pull/1175#discussion_r2004859976

?

apps/studio/src/server/modules/auth/email/__tests__/email.service.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      // Act
+      const user = await db.transaction().execute(async (tx) => {
+        return upsertUser({ tx, email: TEST_EMAIL })
+      })
+
+      // Assert
+      const auditLogs = await db.selectFrom("AuditLog").selectAll().execute()
+      expect(user).toBeDefined()
+      expect(user.email).toBe(TEST_EMAIL)
+      expect(auditLogs).toHaveLength(0)
```
adriangohjw, line 29: nit - right now user-related test is sandwiched between audit logs setup and test. Perhaps re-order the code?

apps/studio/tests/integration/helpers/seed/index.ts
```diff
@@ -486,7 +486,7 @@ export const setupUser = async ({
   userId?: string
   email: string
   phone?: string
-  isDeleted: boolean
+  isDeleted?: boolean
```
adriangohjw, line 489: thanks for changing this haha been wanting to change this since it was quite annoying but kept forgetting

apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
dcshzj: Let me add for 1.

For 2, I think the JSON logged should be covered under the audit logs service unit test. This part is just the caller for that function hence I only opted to test the event type, but the shape is not this service's responsibility.

apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
dcshzj: Good idea, added in 6df2f491137cb7321046a905680739bb66f3f381.

apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
dcshzj: Done for 1 in 6df2f491137cb7321046a905680739bb66f3f381.

apps/studio/src/server/modules/auth/email/__tests__/email.service.test.ts
dcshzj: Updated in 6df2f491137cb7321046a905680739bb66f3f381.

apps/studio/src/server/modules/auth/email/__tests__/email.service.test.ts
dcshzj: Hmm I do that in the next line tho with `user.email`.

apps/studio/src/server/modules/auth/email/email.router.ts
dcshzj: Ah I forgot about the IP part, updated in 6df2f491137cb7321046a905680739bb66f3f381.

apps/studio/src/server/modules/auth/email/email.router.ts
dcshzj: Yeah I think setting `after: null` makes sense. Updated in 6df2f491137cb7321046a905680739bb66f3f381.

apps/studio/src/server/modules/auth/email/__tests__/email.service.test.ts
seaerchin: email name is the auto generated `name` but not very important

apps/studio/src/server/modules/auth/email/__tests__/email.router.test.ts
```diff
@@ -126,12 +126,51 @@ describe("auth.email", () => {
...
+      await prisma.verificationToken.create({
+        data: {
+          expires: new Date(Date.now() + env.OTP_EXPIRY * 1000),
+          identifier: TEST_OTP_FINGERPRINT,
+          token: VALID_TOKEN_HASH,
+        },
+      })
+
+      // Act
+      try {
```
seaerchin, line 147: nit: can we do `expect(caller.verifyOtp).rejects.toThrowError()`

seaerchin (approved):

PR #1177, opened 2025-03-18
feat: add audit logging to folder router with tests

dcshzj:
## Problem

Closes ISOM-1777.
Closes ISOM-1786.

## Solution

**Breaking Changes**

**Features**:

- Added audit logging capabilities to the folder router.

## Tests

- [ ] Log in to Isomer Studio.
- [ ] Create a new folder in any site.
- [ ] Verify that an additional audit log entry has been created in the AuditLog table which is of ResourceCreate type.
- [ ] Edit the folder's name.
- [ ] Verify that an additional audit log entry has been created in the AuditLog table which is of ResourceUpdate type.

### DIFF
--- a/apps/studio/src/server/modules/folder/__tests__/folder.router.test.ts
+++ b/apps/studio/src/server/modules/folder/__tests__/folder.router.test.ts
@@ -16,7 +16,7 @@ import {
 } from "tests/integration/helpers/seed"
 
 import { createCallerFactory } from "~/server/trpc"
-import { db, ResourceType } from "../../database"
+import { AuditLogEvent, db, ResourceType } from "../../database"
 import { folderRouter } from "../folder.router"
 
 const createCaller = createCallerFactory(folderRouter)
@@ -28,6 +28,7 @@ describe("folder.router", async () => {
 
   beforeEach(async () => {
     await resetTables(
+      "AuditLog",
       "Blob",
       "Resource",
       "Site",
@@ -59,6 +60,9 @@ describe("folder.router", async () => {
       await expect(result).rejects.toThrowError(
         new TRPCError({ code: "UNAUTHORIZED" }),
       )
+      await expect(
+        db.selectFrom("AuditLog").selectAll().execute(),
+      ).resolves.toEqual([])
     })
 
     it("should throw 409 if permalink already exists", async () => {
@@ -84,6 +88,9 @@ describe("folder.router", async () => {
           message: "A resource with the same permalink already exists",
         }),
       )
+      await expect(
+        db.selectFrom("AuditLog").selectAll().execute(),
+      ).resolves.toEqual([])
     })
 
     it("should throw 404 if `siteId` does not exist", async () => {
@@ -136,6 +143,9 @@ describe("folder.router", async () => {
           message: "Parent folder does not exist",
         }),
       )
+      await expect(
+        db.selectFrom("AuditLog").selectAll().execute(),
+      ).resolves.toEqual([])
     })
 
     it("should throw 400 if `parentFolderId` is not a folder", async () => {
@@ -163,6 +173,9 @@ describe("folder.router", async () => {
           message: "Resource ID does not point to a folder",
         }),
       )
+      await expect(
+        db.selectFrom("AuditLog").selectAll().execute(),
+      ).resolves.toEqual([])
     })
 
     it("should create a folder even with duplicate permalink if `siteId` is different", async () => {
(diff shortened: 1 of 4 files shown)


apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -130,33 +151,71 @@ export const folderRouter = router({
...
+            .where("Resource.siteId", "=", Number(siteId))
+            .where("Resource.type", "in", [
+              ResourceType.Folder,
+              ResourceType.Collection,
+            ])
+            .executeTakeFirst()
+
+          if (!oldResource) {
+            return undefined
+          }
```
seaerchin, line 174: [issue] won't this just bubble up as `undefined`? i think better to just throw `NOT_FOUND` ba

apps/studio/src/server/modules/folder/__tests__/folder.router.test.ts
```diff
@@ -550,6 +613,9 @@ describe("folder.router", async () => {
             "You do not have sufficient permissions to perform this action",
         }),
       )
+      await expect(
+        db.selectFrom("AuditLog").selectAll().execute(),
+      ).resolves.toEqual([])
     })
 
     it("should return undefined if the resourceId is not a folder", async () => {
```
seaerchin, line 621: might need to update this test if we decide to throw

seaerchin (approved):
not sure if we should throw `NOT_FOUND` when the user isn't found - perhaps `BAD_REQUEST` or `UNAUTHORISED` is a better fit but minor issue

another minor issue re: whether return `undefined` or throw - i think throwing here mkaes more sense cos our frontend can handle this via toast and it is an unexpected case too

apps/studio/src/server/modules/folder/folder.router.ts
dcshzj: Updated in 2ba86e239e4eabc16804821831254b47a74b7e85.

apps/studio/src/server/modules/folder/__tests__/folder.router.test.ts
dcshzj: Updated in 2ba86e239e4eabc16804821831254b47a74b7e85.

apps/studio/src/server/modules/folder/__tests__/folder.router.test.ts
```diff
@@ -59,6 +60,9 @@ describe("folder.router", async () => {
       await expect(result).rejects.toThrowError(
         new TRPCError({ code: "UNAUTHORIZED" }),
       )
+      await expect(
+        db.selectFrom("AuditLog").selectAll().execute(),
+      ).resolves.toEqual([])
```
adriangohjw, line 65: super nit - checking count is 0 might be more readable than checking for empty array

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -65,28 +67,47 @@ export const folderRouter = router({
...
-                message: "A resource with the same permalink already exists",
-              })
-            }
-            throw err
+        const folder = await db.transaction().execute(async (tx) => {
+          const user = await tx
+            .selectFrom("User")
+            .where("id", "=", ctx.user.id)
+            .selectAll()
+            .executeTakeFirstOrThrow(() => new TRPCError({ code: "NOT_FOUND" }))
```
adriangohjw, line 75: thoughts: perhaps move this out of the transaction since it's read-only and that the transaction won't be too blocking?

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -130,33 +151,70 @@ export const folderRouter = router({
...
+            .where("Resource.id", "=", resourceId)
+            .where("Resource.siteId", "=", Number(siteId))
+            .where("Resource.type", "in", [
+              ResourceType.Folder,
+              ResourceType.Collection,
+            ])
+            .set({
+              permalink,
+              title,
+            })
```
adriangohjw, line 190: since we already fetch the `oldResource`, should we find the newResource by using `oldResource.id`? otherwise the filtering logic is duplicated - in the future someone might update one without updating another one

apps/studio/src/server/modules/folder/__tests__/folder.router.test.ts
dcshzj: Updated in 9fff24c507dd19fca1f26ccd2fba956c074169d8.

apps/studio/src/server/modules/folder/folder.router.ts
dcshzj: No strong opinions on this one, initially thought would be good to keep all in one transaction to maintain data integrity due to interleaving but I think extremely unlikely. Updated in 9fff24c507dd19fca1f26ccd2fba956c074169d8.

apps/studio/src/server/modules/folder/folder.router.ts
dcshzj: No opinions on this one, updated in 9fff24c507dd19fca1f26ccd2fba956c074169d8.

PR #1186, opened 2025-03-19
feat: audit log - add user router and tests

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1781/[be]-add-audit-logging-to-userrouter
Closes https://linear.app/ogp/issue/ISOM-1782/[misc]-add-integration-tests-for-userrouter

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- added audit logging for User route
- Note - we are only soft-deleting ResourcePermission, so currently PermissionUpdate won;t be used


adriangohjw (commented):
i do feel like some of the tests are on the repetitive side, but IMO audit log is a pretty important feature, so i err on the side of caution

but do LMK if yall think otherwise!

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -154,3 +192,108 @@ export const getUsersQuery = ({ siteId, adminType }: GetUsersQueryProps) => {
...
+      .where("id", "=", userId)
+      .selectAll()
+      .executeTakeFirstOrThrow()
+
+    const updatedUser = await db
+      .updateTable("User")
+      .where("id", "=", userId)
+      .set({ name, phone })
+      .returningAll()
+      .executeTakeFirstOrThrow()
```
seaerchin, line 287: question - i noticed this repeated pattern of separating the checks from the update but i was just curious to find out why, because the same can be done w/ a `returningAll` and i'm not sure the distinction is useful to make (for the end user at least) whereas the code does read simpler without this distinction

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -154,3 +192,108 @@ export const getUsersQuery = ({ siteId, adminType }: GetUsersQueryProps) => {
...
+  phone,
+}: UpdateUserDetailsProps) => {
+  if (byUserId !== userId) {
+    throw new TRPCError({
+      code: "FORBIDDEN",
+      message: "You are not allowed to update this user's details",
+    })
+  }
+
+  const updatedUser = await db.transaction().execute(async (tx) => {
```
seaerchin, line 275: [nit] can we just do the return here because the 2 return blocks below do the same thing hahahahah

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -154,3 +192,108 @@ export const getUsersQuery = ({ siteId, adminType }: GetUsersQueryProps) => {
...
+    const user = await tx
+      .selectFrom("User")
+      .where("id", "=", userId)
+      .selectAll()
+      .executeTakeFirstOrThrow()
+
+    const updatedUser = await db
```
seaerchin, line 282: [issue] shuold we be using `tx` here?

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
```diff
@@ -51,20 +53,29 @@ describe("user.service", () => {
...
+      await resetTables("User", "ResourcePermission", "Site", "AuditLog")
       const { site } = await setupSite()
       siteId = site.id
+
+      const creator = await setupUser({
+        name: "creator",
+        email: "creator@open.gov.sg",
+        isDeleted: false,
+      })
+      creatorUserId = creator.id
```
seaerchin, line 72: nit - i think more useful next time to just have `creator` then assign in the `beforeEach` rather than having a `creatorUserId`

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
```diff
@@ -160,6 +187,41 @@ describe("user.service", () => {
...
+      expect(userAuditLogs[0]).toMatchObject({
+        eventType: "UserCreate",
+        delta: expect.objectContaining({
+          before: null,
+          after: expect.objectContaining({
+            id: createdUser.id,
+            email: TEST_EMAIL,
+          }),
+        }),
+      })
```
seaerchin, line 207: question - because we log the db object directly, do you think it's useful to query the object then match it here?

apps/studio/src/server/modules/user/__tests__/user.router.test.ts
seaerchin, line 1: skimmed this! mostly looks like a bunch of additional validation for audit logs so no big issues from me; i was wondering if a helper method might help but i'm not too tied to it, because i value clarity alot in test files

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -109,10 +138,19 @@ export const createUserWithPermission = async ({
...
+    return resourcePermission
+  }
+
+  const executeInTransaction = async (tx: Transaction<DB>) => {
+    const user = await createUserInTransaction(tx)
+    const permission = await createPermissionInTransaction(tx, user.id)
+    return { user, resourcePermission: permission }
   }
 
   if (tx) {
```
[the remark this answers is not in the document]
seaerchin: FYI, i just did a small refactor wher ei force the tx because we only use it in one place that passes a tx too

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -154,3 +192,108 @@ export const getUsersQuery = ({ siteId, adminType }: GetUsersQueryProps) => {
...
+  userId,
+  name,
+  phone,
+}: UpdateUserDetailsProps) => {
+  if (byUserId !== userId) {
+    throw new TRPCError({
+      code: "FORBIDDEN",
+      message: "You are not allowed to update this user's details",
+    })
+  }
```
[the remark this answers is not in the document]
seaerchin: updated to do ^, but as an aside i couldn't find out where `ctx.user` was being created, which seemed odd. in `createContext` and `createUserContext`, they both don't query the db for the user.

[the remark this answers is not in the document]
seaerchin: nvm, it's being done in `createAuthedSession`; iron session just seals it into the cookie object and everytime we create session, `getSession` is just `unsealing` with the pw etc [provided](https://github.com/vvo/iron-session/blob/bcf3aec414856c0b3952be1150df9a0b1649d880/src/core.ts#L308)

apps/studio/src/server/modules/permissions/permissions.service.ts
```diff
@@ -145,3 +146,79 @@ export const validatePermissionsForManagingUsers = async ({
...
+
+  return await db.transaction().execute(async (tx) => {
+    const sitePermissionToRemove = await tx
+      .selectFrom("ResourcePermission")
+      .where("userId", "=", userId)
+      .where("siteId", "=", siteId)
+      .where("resourceId", "is", null) // because we are updating site-wide permissions
+      .where("deletedAt", "is", null) // ensure deleted persmission deletedAt is not overwritten
+      .selectAll()
+      .executeTakeFirst()
```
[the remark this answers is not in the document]
adriangohjw: oh im currently only removing the "site" resource permission and not individual resource permission (since we don't support that at the moment too)*

yup, the existence of the unique constraint ensured that there can only be one record that fits this criteria - it technically cannot have more than 1 non-deleted permission per user per site so `executeTakeFirst` is safe

_* yeah ideally we should refactor to add SitePermission so as to improve code clarity instead of relying on `resourceId` null for that  _

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
adriangohjw: oh my thought process is that since `creator` is not being used beyond its `creatorUserId`, and that `creatorUserId` sole purpose is to pass it into the auditlog, there's no need to expose the `creator` to the tests

apps/studio/src/server/modules/user/user.service.ts
```diff
@@ -81,25 +90,45 @@ export const createUserWithPermission = async ({
         name: name || email.split("@")[0] || "",
         phone,
       })
-      .onConflict((oc) =>
-        oc
-          .columns(["email", "deletedAt"])
-          .doUpdateSet((eb) => ({ email: eb.ref("excluded.email") })),
-      )
-      .returning(["id", "email", "name", "phone", "deletedAt"])
+      .onConflict((oc) => oc.columns(["email", "deletedAt"]).doNothing())
```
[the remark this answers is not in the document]
adriangohjw: ah, previously this code was copied over from `createUser` script when doing User Mgmt, which have I had some question mark as to why we updated the email on conflict - i ended up not updating it since it worked and won't break anything (just strange tho!)

but yes it isn't possible with the unique constraint we have on email anyway

with audit log, it's important to update this to `doNothing()` since updating the email (even tho no theoretical change in value) will cause the `updatedAt` to be updated, which might be confusing

adriangohjw:
@seaerchin im not proud of this (https://github.com/opengovsg/isomer/pull/1186/commits/a5d3545142b6279cb3f68ffe99fd611e179c599f) since it can be a huge performance bottleneck once we introduce resource-level permissions, but i think it's ok for now - WDYT?

apps/studio/src/server/modules/permissions/permissions.service.ts
```diff
@@ -145,3 +146,79 @@ export const validatePermissionsForManagingUsers = async ({
...
+
+    // NOTE: this is technically impossible because we're executing
+    // inside a tx and this is the same resource which was fetched earlier
+    if (!deletedSitePermission) {
+      throw new TRPCError({
+        code: "INTERNAL_SERVER_ERROR",
+        message:
+          "Something went wrong while attempting to move your resource, please try again later",
+      })
+    }
```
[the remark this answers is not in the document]
seaerchin: i don't think that will happen since the `tx` will rollbcak as a whole! also, the audit log should only store **updates to the db** - in this case, nothing has happened, so we shouldn't log it 

this is because the audit log is meant to be used by external agencies for their own internal reconciliation, not by us (we are secondary consumers)

but yea no impact here so both ways look ok to me

apps/studio/src/server/modules/user/user.service.ts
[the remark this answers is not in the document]
seaerchin: ok, sounds good!

apps/studio/src/server/modules/user/__tests__/user.service.test.ts
seaerchin: not strongly opinionated here, willing to accept both. leaning towards db object because it's easier to read and this one requires understanding that `before/after` has this specific shape

seaerchin:
> @seaerchin im not proud of this ([a5d3545](https://github.com/opengovsg/isomer/commit/a5d3545142b6279cb3f68ffe99fd611e179c599f)) since it can be a huge performance bottleneck once we introduce resource-level permissions, but i think it's ok for now - WDYT?

wa it feels very iffy and i'm scared that we forget about it, onliy for it to bite us - i'll file an issue and lets work on it if/when it becomes a problem

seaerchin (approved):
overall ok, the db query is looking very sus but i filed an issue [ISOM-1819: \[MISC\] Relook at db queries for permissions](https://linear.app/ogp/issue/ISOM-1819/[misc]-relook-at-db-queries-for-permissions) so we can track.

the case for deletion for logging won't happen so imo we should combine but it's still clear as-is, just abit verbose so fine w/ leavin git too

PR #1195, opened 2025-03-26
Feat/add wizgov widget

adriangohjw:
## Problem

IPOS and a number of upcoming sites are using AskGov for their FAQ. We will need to add WizGov in order for them to embed their widget

## Solution

**Features**:

- similar implementation to vica
- to add to CSP too https://github.com/opengovsg/isomer-next-infra/pull/102


packages/components/src/templates/next/components/internal/Wizgov/WizgovWidget.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { WizgovWidgetClient } from "./WizgovWidgetClient"
+
+// Reference: https://github.com/opengovsg/wizgov
+export const WizgovWidget = (props: WizgovWidgetProps) => {
+  const { site, ...rest } = props
+
+  const scriptUrl =
+    site.environment === "production"
+      ? "https://script.wiz.gov.sg/widget.js"
+      : "https://script-staging.wiz.gov.sg/widget.js"
```
seaerchin, line 11: FYI, we need to add this for our CSP!

packages/components/src/templates/next/layouts/Skeleton/Skeleton.tsx
```diff
@@ -111,7 +112,9 @@ export const Skeleton = ({
       )}
 
       {/* Ensures that the webchat widget only loads after the page has loaded */}
+      {/* Note: did not account for both being added to the config as it's a very unlikely scenario and there's "correct" way to handle this */}
       {site.vica && <VicaWidget site={site} {...site.vica} />}
+      {site.wizgov && <WizgovWidget site={site} {...site.wizgov} />}
```
seaerchin, line 117: just double checking - woudl having both `vica` and `wizgov` cause the widgets to overlay each other?

packages/components/src/interfaces/internal/Wizgov.ts
```diff
@@ -0,0 +1,13 @@
+import type { IsomerSiteProps } from "~/types"
+
+export interface WizgovProps {
+  "data-agency": string
+}
```
seaerchin, line 5: where did we get this prop from?

seaerchin (approved):
lgtm code-wise, a few questions re: style

1. should we consolidate all our widgets ids and script ids into 1 place? no strong opinion here because it doesn't look like it'll. leak beyond the component, so ok to leave as-is
2. what's the diff between `WizgovWidget` and `WizgovWidgetClient`? as far as i can tell, it's just forwarding props and getting the correct `scriptUrl`? both look fine to me but just wondering why separate here
3. code-style - perhaps we should make `type WizgovWidgetProps = Pick<IsomerSiteProps, "environment">`. this makes it easier to test if we ever wnat to integrate on staging but if not, feel free to ignore this too!

packages/components/src/templates/next/layouts/Skeleton/Skeleton.tsx
adriangohjw: probably! but given that this is admin-only and most site owners will only want 1, the risk is very minimal - also dom't think there's a "right" way to handle it. Disabling both or causing build to fail sounds off too

packages/components/src/interfaces/internal/Wizgov.ts
adriangohjw: can refer to here: https://github.com/opengovsg/wizgov

FYI checked with Ziyang from AskGov that no longer require CSS even though the docs say so

packages/components/src/templates/next/components/internal/Wizgov/WizgovWidget.tsx
adriangohjw: yup! there's a corresponding PR for that https://github.com/opengovsg/isomer-next-infra/pull/102

adriangohjw:
@seaerchin 

> what's the diff between WizgovWidget and WizgovWidgetClient? as far as i can tell, it's just forwarding props and getting the correct scriptUrl? both look fine to me but just wondering why separate here

we will be generating the wizgov staging/prod link - reason why we split into client and non-client is that `site` can be quite large and passing it entirely into a `use client` component will cause page size to be huge. Splitting like this let us pre-rendered during SSG as much as possible

packages/components/src/templates/next/components/internal/Wizgov/WizgovWidgetClient.tsx
```diff
@@ -0,0 +1,45 @@
...
+  scriptTag.id = scriptId
+  scriptTag.async = true
+  scriptTag.type = "text/javascript"
+  scriptTag.src = scriptUrl
+  scriptTag.referrerPolicy = "origin"
+  document.body.appendChild(scriptTag)
+}
+
+export const WizgovWidgetClient = (props: WizgovWidgetClientProps) => {
+  const { "data-agency": dataAgency, scriptUrl } = props
```
dcshzj, line 27: minor nit: can do the object destructuring within the method signature directly so the code is more straightforward

packages/components/src/templates/next/components/internal/Wizgov/WizgovWidgetClient.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    // Only inject the script after everything else has finished loading
+    // This is to replicate Next.js lazyOnload behaviour (as recommended for widgets)
+    if (document.readyState === "complete") {
+      reloadWizgovScript(scriptUrl)
+    } else {
+      window.addEventListener("load", () => reloadWizgovScript(scriptUrl))
+      return () =>
+        window.removeEventListener("load", () => reloadWizgovScript(scriptUrl))
+    }
```
dcshzj, line 41: question(non-blocking): out of curiosity, is the Wizgov script able to load multiple times but only still show one single icon, [similar to WOGAA](https://github.com/opengovsg/isomer/blob/main/packages/components/src/templates/next/components/internal/Wogaa/Wogaa.tsx)?

Asking cos using the ScriptComponent (next/script) would be more ideal (not sure if more optimised, never checked) so if we can use that instead of our own useEffect would be good.

dcshzj (approved):

packages/components/src/templates/next/layouts/Skeleton/Skeleton.tsx
seaerchin: i was thinking we choose `wizgov` if it exists, otherwise vica but i think it's rare enough that we can leave it as is

seaerchin:
> @seaerchin
> 
> > what's the diff between WizgovWidget and WizgovWidgetClient? as far as i can tell, it's just forwarding props and getting the correct scriptUrl? both look fine to me but just wondering why separate here
> 
> we will be generating the wizgov staging/prod link - reason why we split into client and non-client is that `site` can be quite large and passing it entirely into a `use client` component will cause page size to be huge. Splitting like this let us pre-rendered during SSG as much as possible

sounds good! a point of confusion for me was that we're still passing `...rest` into the client, which seems like it'll still get the  whole `site` object?

adriangohjw:
@seaerchin anyway ii thought about it and decided to remove the Client and pass the `site.environment` instead of `site` into the component - this can reduce complexity of the change

> sounds good! a point of confusion for me was that we're still passing ...rest into the client, which seems like it'll still get the whole site object?

@seaerchin oh it won't since the `...rest` was the pure wizgov props. Nevertheless refactor to make it clearer too

@dcshzj tested the WOGAA method on staging but it doesn't work (same UX issue as VICA last time) 😢 

anyway added it to staging site for testing (https://test-isomer-vica-ncss-staging.isomer.gov.sg/)

PR #1200, opened 2025-03-28
chore: deploy to `uat` on every release

seaerchin:
## Problem
we shuold deploy to `uat` on every release so that the environment stays up to date with `production` and `uat` users can enjoy latest featureset

## Solution
copy deployment workflow for `uat` over to production deploy workflow.

concretely, this means we will deploy to `uat` on every release and testing of new features (via pushing to `uat`) will have to be off `main` or it will reset the state of the environment

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -44,3 +44,37 @@ jobs:
 
     secrets:
       DD_API_KEY: ${{ secrets.DD_API_KEY_GITHUB_ACTIONS }}
+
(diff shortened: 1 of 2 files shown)


adriangohjw (approved):

adriangohjw:
Curious - is there still a point for `uat` branch? Does it make sense that we always deploy `uat` off `main` branch?

seaerchin:
> Curious - is there still a point for `uat` branch? Does it make sense that we always deploy `uat` off `main` branch?

a big one is to actually test features before we merge into `main`!

PR #1203, opened 2025-03-29
fix(pdf-uploads): add UI to show when pdf uploads have completed

seaerchin:
## Problem
1. currently, pdf items have no UI to show when they have been uploaded due to a regression 

## Solution
- add in UI for pdf uploads. this also shifts to using an external library for exponential backoff.
- this fixes an issue with validation for file uploads, where we never trigger `shouldValidate` when using `setValue`. this results in validation only being ran at `save` button click, which resulted in issues in the following sequence of events:
    - upload an item
    - click save
    - remove previously uploaded item
    - click save (save is now toggleed to disabled)
- at this point, we cannot save the newly uploaded item anymore because revalidation is not being triggered when we `setValue` and hence, the save button (which triggers validation and was previously disabled) is always disabled. 
- add in a toggle to file attachment to let the consumer specify if they should fetch teh resource. we do this because the image requires fetching teh resource to display whereas for files, we can just show the href

## Video

https://github.com/user-attachments/assets/1ef0bf61-0a8a-4446-a2f7-db444bc3e038


apps/studio/src/components/PageEditor/FileAttachment.tsx
```diff
@@ -23,13 +24,14 @@ export const FileAttachment = ({
   siteId,
   maxSizeInBytes,
   acceptedFileTypes,
+  shouldFetchResource = true,
```
adriangohjw, line 27: curious - seem like default is `true` but the only place that uses this component hardcode it to be `false` - not too sure what's the intention of this props and when/why to use it

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -98,6 +101,8 @@ const LinkEditorModalContent = ({
 
   const isEditingLink = !!linkText && !!linkHref
 
+  console.log("errors", errors)
```
adriangohjw, line 104: remove

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -211,34 +218,29 @@ export const LinkEditorModal = ({
...
+          <AttachmentData data={curHref} onClick={() => setHref("")} />
+        ) : (
+          <FileAttachment
+            maxSizeInBytes={MAX_FILE_SIZE_BYTES}
+            acceptedFileTypes={FILE_UPLOAD_ACCEPTED_MIME_TYPE_MAPPING}
+            siteId={siteId}
+            setHref={(href) => setHref(href ?? "")}
+            shouldFetchResource={false}
+          />
+        )
```
adriangohjw, line 243: no need to change but thought to point out that naming does feel a bit counterintuitive, that we use `FileAttachment` when link types isn't File

apps/studio/src/features/editing-experience/components/form-builder/hooks/useAssetUpload.ts
```diff
@@ -0,0 +1,42 @@
...
+          const response = await fetch(`${assetsBaseUrl}${src}`)
+          if (!response.ok) {
+            throw new Error(`Unable to read from ${src}`)
+          }
+          return src
+        },
+        {
+          startingDelay: baseTimeoutMs,
+          numOfAttempts: retries,
+          delayFirstAttempt: true,
```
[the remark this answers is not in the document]
seaerchin: yeap!

apps/studio/src/features/editing-experience/components/form-builder/hooks/useAssetUpload.ts
```diff
@@ -0,0 +1,42 @@
...
+        },
+      )
+      setIsLoading(false)
+      return src
+    } catch (e) {
+      console.error(e)
+      setIsLoading(false)
```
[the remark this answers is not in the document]
seaerchin: ah ok, will do for code clarity!

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
seaerchin: oops, thanksf or catch!

apps/studio/src/components/PageEditor/FileAttachment.tsx
seaerchin: there are 2 places that use this component - `LinkEditor`, where it is set to `false` and `JsonFormsImageControl`, where it is `true`!

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
seaerchin: yea it's not ideal - do you have any suggestions? cos this covers basically everything we wanna upload so `File` felt like the most general term. wasn't specifically referring to `LINK_TYPES` here

apps/studio/src/features/editing-experience/components/form-builder/hooks/useAssetUpload.ts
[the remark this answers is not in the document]
seaerchin: it means 3 total attempts! i stuck with this cos got `startingDelay` but lmk if you want me to swap to 4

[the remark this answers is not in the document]
seaerchin: eh to double check, in the old setup, it also goes up to 3 attempts (0 fail -> retry -> 1 fail -> retry -> 2 fail -> retry)s

PR #1208, opened 2025-04-01
wizgov widget add data topic props

adriangohjw:
## Problem

Closes https://opengovproducts.slack.com/archives/C07CWUNUL68/p1743496259124609

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- add support for `data-topic` for wizgov

## Tests

1. add a `data-topic` into the wizgov config and it should filter the topics

### DIFF
--- a/packages/components/src/templates/next/components/internal/Wizgov/WizgovWidget.tsx
+++ b/packages/components/src/templates/next/components/internal/Wizgov/WizgovWidget.tsx
@@ -46,13 +46,5 @@ export const WizgovWidget = ({
     }
   }, [])
 
-  return (
-    <div
-      id="wizgov-widget"
-      data-agency={wizgovProps["data-agency"]}
-      {...(!!wizgovProps["data-topic"] && {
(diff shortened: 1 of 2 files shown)


packages/components/src/templates/next/components/internal/Wizgov/WizgovWidget.tsx
```diff
@@ -46,5 +46,13 @@ export const WizgovWidget = ({
...
   }, [])
 
-  return <div id="wizgov-widget" data-agency={wizgovProps["data-agency"]} />
+  return (
+    <div
+      id="wizgov-widget"
+      data-agency={wizgovProps["data-agency"]}
+      {...(!!wizgovProps["data-topic"] && {
+        "data-topic": wizgovProps["data-topic"],
+      })}
```
dcshzj, line 55: suggestion: we can spread the `wizgovProps` directly since they are the same. Any `undefined` will also not appear in the final output

dcshzj (approved):

PR #1212, opened 2025-04-01
Feat - collection sorting

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1836/add-collection-sorting

## Solution

**Breaking Changes**

  **- Need to check against DB to patch old records to make sure their sorting behaviour is followed**

**Features**:

- add option to pass in `sortBy` and `sortDirection`
- pass in values from collection page config

**Improvements**:

- shifted sorting logic into standalone function for ease of unit test
- add basic unit test for sorting for code confidence
- update studio "custom index page" dev button to respect CollectionPage schema + update to desc first

### Setup

Create a new collection with 5 items
1. Alice + today date
2. Bob + today date
3. Charlie + yesterday date
4. Charlie + tmr date
5. David + today date

### test case 1 (`defaultSortBy: "date", defaultSortDirection: "asc"`) 

retrigger build
order should be:
1. Charlie + yesterday date
2. Alice + today date
3. Bob + today date
4. David + today date
5. Charlie + tmr date

### test case 2 (`defaultSortBy: "date", defaultSortDirection: "desc"`) 

retrigger build
order should be:
1. Charlie + tmr date
2. Alice + today date
3. Bob + today date
4. David + today date
5.  Charlie + yesterday date

### test case 3 (`defaultSortBy: "title", defaultSortDirection: "asc"`) 

retrigger build
order should be:
1. Alice + today date
2. Bob + today date
3. Charlie + yesterday date
4. Charlie + tmr date
5. David + today date

### test case 4 (`defaultSortBy: "title", defaultSortDirection: "desc"`) 

retrigger build
order should be:
1. David + today date
2. Charlie + yesterday date
3. Charlie + tmr date
4. Bob + today date
5. Alice + today date

### test case 5 (`defaultSortBy: "category", defaultSortDirection: "asc"`) 

retrigger build
order should be (category + title)
1. 2001 + Alice
2. 2002 + Alice
3. 2002 + Bob
4. 2003 + Alice

### test case 6 (`defaultSortBy: "category", defaultSortDirection: "desc"`) 

retrigger build
order should be:
1. 2003 + Alice
2. 2002 + Alice
3. 2002 + Bob
4. 2001 + Alice


packages/components/src/templates/next/layouts/Collection/utils/__tests__/sortCollectionItems.test.ts
```diff
@@ -0,0 +1,246 @@
...
+      createItem({ title: "Newest", date: new Date("2023-12-31") }),
+      createItem({ title: "Also No Date" }),
+      createItem({ title: "Oldest", date: new Date("2023-01-01") }),
+    ]
+
+    // Act
+    const sorted = sortCollectionItems({ items })
+
+    // Assert
+    expect(sorted[0]?.title).toBe("Newest")
```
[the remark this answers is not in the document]
adriangohjw: yup! tho i was thinking some other assertions check for other fields, so keeping it consistent in this case might be easier to read/maintain 

https://github.com/opengovsg/isomer/blob/781a0367def763684427b2cbddfa30d3de8f7b0d/packages/components/src/templates/next/layouts/Collection/utils/__tests__/sortCollectionItems.test.ts#L154-L170

no strong opinion on this, happy to change!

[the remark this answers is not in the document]
harishv7: I felt the other example is more nested comparison hence being explicit was easier to read

for this case seems to be just a sequence of titles hence I felt the iterative way was actually more readable for me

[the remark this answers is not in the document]
adriangohjw: fixed in https://github.com/opengovsg/isomer/pull/1212/commits/ad546925b5bb17bfcd20ccbee52a3fe5ac76c49a

packages/components/src/templates/next/layouts/Collection/utils/__tests__/sortCollectionItems.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    // Act
+    const sorted = sortCollectionItems({ items })
+
+    // Assert
+    const expectedTitles = ["Alice", "Bob", "Charlie"]
+    expect(sorted.map((item) => item.title)).toEqual(expectedTitles)
+  })
+
+  describe("should sort by title with date as tiebreaker (newest first) when sortBy is title and", () => {
```
harishv7, line 118: I think the title is incomplete "title and ..."

packages/components/src/templates/next/layouts/Collection/utils/__tests__/sortCollectionItems.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    // Act
+    const sorted = sortCollectionItems({ items })
+
+    // Assert
+    const expectedTitles = ["Alice", "Bob", "Charlie"]
+    expect(sorted.map((item) => item.title)).toEqual(expectedTitles)
+  })
+
+  it("should place items without dates at the end", () => {
```
harishv7, line 85: how did we arrive at this btw? Why not at the front or add a default new Date()?

harishv7: title should be "should place items without dates at the end **sorted alphabetically**"

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -0,0 +1,96 @@
...
+
+// Sort by last updated date, tiebreaker by title
+const sortCollectionItemsByDate = ({
+  items,
+  sortDirection = "desc",
+}: Omit<SortCollectionItemsProps, "sortBy">) => {
+  return items.sort((a, b) => {
+    // If both items have no dates, sort by title ascending as a tiebreaker
+    if (a.rawDate === undefined && b.rawDate === undefined) {
+      return a.title.localeCompare(b.title)
```
harishv7, line 21: for localeCompare, there might be options to configure iirc

like sensitivity, and numeric

e.g. what if the items start with numbers "1 Item One", "2 Item Two" etc.

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -0,0 +1,96 @@
...
+        const exhaustiveCheck: never = sortDirection
+        return exhaustiveCheck
+    }
+  }) as AllCardProps[]
+}
+
+export const sortCollectionItems = ({
+  items,
+  sortBy = "date",
+  sortDirection = "asc",
```
harishv7, line 85: hmm if the default sortby is date, shouldn't this be desc?

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -0,0 +1,96 @@
...
+      return -1
+    }
+
+    // If both items have same date, sort by title ascending as a tiebreaker
+    if (a.rawDate.getTime() === b.rawDate.getTime()) {
+      return a.title.localeCompare(b.title)
+    }
+
+    switch (sortDirection) {
+      case "asc":
```
harishv7, line 39: when we say ascending it means older dates first right?

if so, shouldn't a < b return -1?

This will mean a comes before b

E.g. a : 1st April, b: 2nd April

Output should be 1 Apr, 2 Apr (in ascending order)

packages/components/src/templates/next/layouts/Collection/utils/__tests__/sortCollectionItems.test.ts
adriangohjw: > how did we arrive at this btw? Why not at the front or add a default new Date()?

this is an existing behaviour actually

if i recall correctly, the reason why we did this for the edge case of migrated collections that contains both date-variant and non-date variant, and non-date variant are placed at the end partially for aesthetic reasons _(otherwise it's looks jank to have them interweaved)_. This edge case will not be possible for collections created on Studio

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -0,0 +1,96 @@
...
+  }) as AllCardProps[]
+}
+
+// Sort by title, tiebreaker by last updated date
+const sortCollectionItemsByTitle = ({
+  items,
+  sortDirection = "asc",
+}: Omit<SortCollectionItemsProps, "sortBy">) => {
+  return items.sort((a, b) => {
+    // Rank items with no dates last
```
harishv7, line 56: why are we ranking by date first for a sort by title logic?

the date should only be a tiebreaker when the titles are equal right?

harishv7: i.e. 2 conditional flows:

if titles are different -> localeCompare

if titles are same -> use date as tiebreaker

if both have no dates -> consider as equal, else rank item with no date as last

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1212/commits/7d23cf61368a2abe30354d2baf70b31c79dac01d

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
adriangohjw: you are right, updated in https://github.com/opengovsg/isomer/pull/1212/commits/6cfdf9e375c49b71045ce599b4cf5c3817999ffd

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
adriangohjw: hmm i don't think it matter since they are mutually exclusive case (just happy-path first v.s. sad-path guard-clausish)?

nevertheless have updated the code to do happy path first since it reads better logically + added variable for conditions to improve readability too! https://github.com/opengovsg/isomer/pull/1212/commits/c819ab6cce71e4664237fb97fafa21afc09269f8

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    if (bothHaveDates && !bothSameDatetime) {
+      // Type assertion because TS control-flow narrowing only works when
+      // check is done inline and not when we define the variable
+      const aDate = a.rawDate as unknown as Date
+      const bDate = b.rawDate as unknown as Date
+
+      switch (sortDirection) {
+        case "asc":
+          return aDate.getTime() >= bDate.getTime() ? 1 : -1
```
harishv7, line 30: nit: since we have `!bothSameDatetime` technically don't need to check for equality, but ok with this

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+
+    if (bothHaveDates && a.title === b.title) {
+      // Type assertion because TS control-flow narrowing only works when
+      // check is done inline and not when we define the variable
+      const aDate = a.rawDate as unknown as Date
+      const bDate = b.rawDate as unknown as Date
+
+      return aDate.getTime() < bDate.getTime() ? 1 : -1
```
harishv7, line 75: if same title, same date, shouldnt we return 0?

harishv7 (approved):
mostly lgtm, some minor comments

packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
adriangohjw: didn't do that because was thinking it shouldn't affect the sorting order since it will still be determistic, but addressed it in to improve clarity https://github.com/opengovsg/isomer/pull/1212/commits/a7bada148ac0ce87a01edc7d51d1e7ec1a4f4f49

adriangohjw:
@harishv7 isorangers just shared that another use case that will be useful is sorting by category as sometimes agencies uses that as a "year" grouping. Adding it in the same PR here https://github.com/opengovsg/isomer/pull/1212/commits/fb869e4088c73bca348d3293de9df4bf48c1c1c0 - for your review again. thank you!

dcshzj (approved):

PR #1218, opened 2025-04-02
feat: add query params to collection filters

dcshzj:
## Problem

Agencies have been asking for enhancements about the collection filters:
- Having a permanent URL to share a particular set of active filters with other people
- Clicking on an article, then pressing back should retain the active filters
- Ordering of filters is not intuitive when it comes to numbers

## Solution

**Breaking Changes**

**Features**:

- Revamped the way the collection filters are calculated and stored by using the query params as the source of truth, then deriving the other states from it.
- Enhanced the ordering of filters to take into account numbers (1 -> 2 -> 3 -> 22 rather than 1 -> 2 -> 22 -> 3).

**Bug fixes**:

- Fixed issue with phantom collection cards appearing when applying collection filters.

## Before & After Screenshots

- Live demo is available on the NNA staging site: https://staging.d2112ghehokgef.amplifyapp.com/available-courses/

## Tests

- [ ] Go to a collection
- [ ] Activate any combination of filters, search and pagination
- [ ] Copy and paste the page's URL into another tab/window
- [ ] Verify that the same filters are being applied
- [ ] Verify that you can continue to interact with the collection filters
- [ ] Click on any one of the articles
- [ ] Press the browser back button
- [ ] Verify that the previous filters are still being applied

### DIFF
--- a/packages/components/src/hooks/useQueryParams.ts
+++ b/packages/components/src/hooks/useQueryParams.ts
@@ -1,5 +1,10 @@
 import { useEffect, useState } from "react"
 
+const getQueryParams = (params: URLSearchParams) => {
+  const entries = Array.from(params.entries())
+  return Object.fromEntries(entries)
+}
+
 interface UpdateQueryParams {
   oldParams: Record<string, string>
   newParams: Record<string, string | undefined>
@@ -16,8 +21,7 @@ export const useQueryParams = (): [
   // Parse initial query params from the URL when component mounts
   useEffect(() => {
     const params = new URLSearchParams(window.location.search)
-    const entries = Array.from(params.entries())
-    setQueryParams(Object.fromEntries(entries))
+    setQueryParams(getQueryParams(params))
   }, [])
 
   const updateQueryParams = ({ oldParams, newParams }: UpdateQueryParams) => {
@@ -30,11 +34,13 @@ export const useQueryParams = (): [
       }
     }
 
-    const newUrl = window.location.pathname + "?" + params.toString()
+    const url = new URL(window.location.pathname, window.location.origin)
+    url.search = params.toString()
+    const newUrl = url.toString()
+
     // NOTE: We do not want to add a new entry to the history stack
     window.history.replaceState({}, "", newUrl)
-
-    setQueryParams(Object.fromEntries(Array.from(params.entries())))
+    setQueryParams(getQueryParams(params))
   }
 
   return [queryParams, updateQueryParams]
--- a/packages/components/src/hooks/useQueryParams.ts
+++ b/packages/components/src/hooks/useQueryParams.ts
@@ -1,0 +1,41 @@
+import { useEffect, useState } from "react"
+
+interface UpdateQueryParams {
+  oldParams: Record<string, string>
+  newParams: Record<string, string | undefined>
+}
+
+// A hook to manage a page's query params without adding new entries to the
+// history stack
+export const useQueryParams = (): [
+  Record<string, string>,
+  (params: UpdateQueryParams) => void,
+] => {
+  const [queryParams, setQueryParams] = useState({})
+
+  // Parse initial query params from the URL when component mounts
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search)
+    const entries = Array.from(params.entries())
+    setQueryParams(Object.fromEntries(entries))
+  }, [])
+
+  const updateQueryParams = ({ oldParams, newParams }: UpdateQueryParams) => {
+    const params = new URLSearchParams(oldParams)
+    for (const key in newParams) {
+      if (newParams[key] == undefined || newParams[key] === "") {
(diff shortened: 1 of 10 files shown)


packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from "react"
+
+interface UpdateQueryParams {
+  oldParams: Record<string, string>
+  newParams: Record<string, string | undefined>
+}
+
+// A hook to manage a page's query params without adding new entries to the
```
adriangohjw, line 8: Curious - any reason why don't want to add to history stack?

packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from "react"
+
+interface UpdateQueryParams {
+  oldParams: Record<string, string>
+  newParams: Record<string, string | undefined>
```
seaerchin, line 5: why not just disallow `undefined` ah?

packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,41 @@
...
+    const params = new URLSearchParams(oldParams)
+    for (const key in newParams) {
+      if (newParams[key] == undefined || newParams[key] === "") {
+        params.delete(key)
+      } else {
+        params.set(key, newParams[key])
+      }
+    }
+
+    const newUrl = window.location.pathname + "?" + params.toString()
```
seaerchin, line 33: i think we should use the `URL` constructor

seaerchin (approved):
i'm assuming we don't use `pushState` cos we want to dynamically update the url and then let the collection read from it

1 thing to change prior to merge - should we use the `URL` constructor for constructing a new URL?

packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,41 @@
...
+    // NOTE: We do not want to add a new entry to the history stack
+    window.history.replaceState({}, "", newUrl)
+
+    setQueryParams(Object.fromEntries(Array.from(params.entries())))
```
adriangohjw, line 37: nit - seems like this pattern appears in the `useEffect` above - how about putting it in a function with readable name?

packages/components/src/hooks/useQueryParams.ts
dcshzj: Updated in d19d90171df010ff3d6fe6aa380106700d509998.

packages/components/src/hooks/useQueryParams.ts
dcshzj: Updated in d19d90171df010ff3d6fe6aa380106700d509998.

packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from "react"
+
+interface UpdateQueryParams {
+  oldParams: Record<string, string>
+  newParams: Record<string, string | undefined>
+}
+
+// A hook to manage a page's query params without adding new entries to the
+// history stack
+export const useQueryParams = (): [
```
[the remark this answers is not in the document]
seaerchin: not really - i'm just afraid that the package has handled cases which we haven't thought about. off the top of my head, we've only handled url encoding + back button behaviour, which _should_ be ok but yea i can't say for sure

packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,47 @@
...
+] => {
+  const [queryParams, setQueryParams] = useState({})
+
+  // Parse initial query params from the URL when component mounts
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search)
+    setQueryParams(getQueryParams(params))
+  }, [])
+
+  const updateQueryParams = ({ oldParams, newParams }: UpdateQueryParams) => {
```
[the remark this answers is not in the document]
dcshzj: No strong preference, updated in eeefa7e901431bb3417390e4fc49ff32ea5ece3e.

The state should be the main SOT here, hence opted to update based on the state rather than back from the query params again.

packages/components/src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,47 @@
...
+  newParams: Record<string, string | undefined>
+}
+
+// A hook to manage a page's query params without adding new entries to the
+// history stack
+export const useQueryParams = (): [
+  Record<string, string>,
+  (params: UpdateQueryParams) => void,
+] => {
+  const [queryParams, setQueryParams] = useState({})
```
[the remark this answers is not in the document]
dcshzj: Updated in c873e3ae4e78aa62bd6e6d558a0f8d7e5410a7ef.

PR #1219, opened 2025-04-03
chore: add site id to resource related audit logs

seaerchin:
## Problem
Audit logs right now don't have site id as a foreign key - this makes it difficult for us to get all changes related to a site without writing a json join query 

## Solution
1. add `siteId` + `site` to prisma schema
2. where applicable, add `siteId` to the service method


apps/studio/prisma/migrations/20250329080947_add_site_id_fk_to_audit_log/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "AuditLog" ADD COLUMN     "siteId" INTEGER;
+
+-- AddForeignKey
+ALTER TABLE "AuditLog" ADD CONSTRAINT "AuditLog_siteId_fkey" FOREIGN KEY ("siteId") REFERENCES "Site"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
adriangohjw, line 5: nit - instead of `ON DELETE SET NULL`, i think we should use `ON DELETE RESTRICT` to prevent Site record from being deleted. This will prevent accidental deletion of site by isoengineers, since we currently don't have any safeguards against accidental deletion of Site + we don't log site deletion

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -45,6 +45,7 @@ export interface ResourceEventLogProps {
   delta: ResourceCreateDelta | ResourceDeleteDelta | ResourceUpdateDelta
   by: User
   ip?: string
+  siteId: number
```
adriangohjw, line 48: should we use `Site["id"]` instead of `number`?
same for the other instances

adriangohjw (approved):
LGTM - i think we should also back-patch past records manually!

apps/studio/prisma/migrations/20250329080947_add_site_id_fk_to_audit_log/migration.sql
seaerchin: sounds good

PR #1231, opened 2025-04-09
feat: add error handling for Singpass authentication

dcshzj:
## Problem

There is no proper error handling for the Singpass authentication mechanism.

## Solution

**Breaking Changes**

**Improvements**:

- Add error handling screens.

## Before & After Screenshots

image

### DIFF
--- a/apps/studio/src/pages/sign-in/singpass/index.tsx
+++ b/apps/studio/src/pages/sign-in/singpass/index.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react"
+import { useEffect } from "react"
 import NextLink from "next/link"
 import { useRouter } from "next/router"
 import {
@@ -28,13 +28,13 @@ import { SIGN_IN } from "~/lib/routes"
 import { trpc } from "~/utils/trpc"
 
 const SingpassSignInPage: NextPageWithLayout = () => {
-  const [isSingpassError, setIsSingpassError] = useState<boolean>()
   const {
     isOpen: isUnableToUseSingpassModalOpen,
     onOpen: openUnableToUseSingpassModal,
     onClose: closeUnableToUseSingpassModal,
   } = useDisclosure()
   const router = useRouter()
+  const isSingpassError = Boolean(router.query.error)
   const { data, isLoading, isError } =
     trpc.auth.singpass.getUserProps.useQuery()
 
@@ -44,15 +44,11 @@ const SingpassSignInPage: NextPageWithLayout = () => {
 
   useEffect(() => {
     if (isError) {
-      void router.push(`${SIGN_IN}?error=Unable to retrieve user data`)
+      void router.push(
+        `${SIGN_IN}?error=${encodeURIComponent("Unable to retrieve user data")}`,
+      )
     }
   }, [isError, router])
-
-  useEffect(() => {
-    if (router.query.error) {
-      setIsSingpassError(true)
-    }
-  }, [router.query.error])
 
   if (isLoading || !data) {
     return <FullscreenSpinner />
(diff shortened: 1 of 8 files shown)


apps/studio/src/pages/sign-in/singpass/index.tsx
```diff
@@ -1,30 +1,58 @@
...
-  const { data, isLoading } = trpc.auth.singpass.getName.useQuery()
+  const { data, isLoading, isError } = trpc.auth.singpass.getName.useQuery()
 
   const handleBackToLogin = async () => {
     await router.push(SIGN_IN)
   }
 
+  useEffect(() => {
+    if (isError) {
+      void router.push(`${SIGN_IN}?error=Unable to retrieve user data`)
```
adriangohjw, line 46: nit - there's no issue with this since we are the one passing in the error message, but might wanna wrap the error message in `encodeURIComponent` as a best practise

apps/studio/src/pages/sign-in/singpass/index.tsx
```diff
@@ -1,30 +1,58 @@
...
 import { FullscreenSpinner } from "~/components/FullscreenSpinner"
 import { IsomerLogo } from "~/components/Svg"
+import { ISOMER_SUPPORT_LINK } from "~/constants/misc"
 import { SingpassLoginButton } from "~/features/sign-in/components"
 import { UnableToUseSignpassModal } from "~/features/sign-in/components/UnableToUseSignpassModal"
 import { SIGN_IN } from "~/lib/routes"
 import { trpc } from "~/utils/trpc"
 
 const SingpassSignInPage: NextPageWithLayout = () => {
+  const [isSingpassError, setIsSingpassError] = useState<boolean>()
```
adriangohjw, line 31: question - do we need a `useState` and `useEffect` or can we replace it with somehting like this?
```ts
const isSingpassError = Boolean(router.query.error)
```

apps/studio/src/pages/sign-in/singpass/index.tsx
```diff
@@ -1,30 +1,58 @@
...
 
   const handleBackToLogin = async () => {
     await router.push(SIGN_IN)
   }
 
+  useEffect(() => {
+    if (isError) {
+      void router.push(`${SIGN_IN}?error=Unable to retrieve user data`)
+    }
+  }, [isError, router])
```
adriangohjw, line 48: question - since the only conditional statement is `onError`, do we need a `useEffect` or can we leverage on useQuery() native `onError` callback hook. Something like

```ts
const { data, isLoading } = trpc.auth.singpass.getName.useQuery(undefined, {
    onError: () => {
      void router.push(`${SIGN_IN}?error=Unable to retrieve user data`)
    },
  })
  ```

apps/studio/src/pages/sign-in/singpass/index.tsx
dcshzj: Updated in a116f4da5ed9d10d8d78ae1f8bab668ac4f67b04.

apps/studio/src/pages/sign-in/singpass/index.tsx
dcshzj: Updated in a116f4da5ed9d10d8d78ae1f8bab668ac4f67b04.

apps/studio/src/pages/sign-in/singpass/index.tsx
dcshzj: Hmm `onError` is [marked as deprecated](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery), hence didn't use this method.

adriangohjw: ah didn't know about this, thanks for pointing out!

adriangohjw (approved):

PR #1237, opened 2025-04-10
isom-1882 improve release process

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1882/set-up-branch-protection-for-release-branch

### 1. Non-straightforward approach to allowing GH Action to push to protected branch
- This is unless we have a Personal Access Token (PAT) or a Github App.
- PAT is tied to individual account, so it isn't ideal as someone leaving the team/org will cause the action to fail
- Github App is possible, though it requires some setup + org-level developer setting. Personally, don't think it's worth the trade-off

### 2. Hardcoding `ISOMER_BUILD_REPO_BRANCH` cause node_modules caching issue
- Currently we cache based on `ISOMER_BUILD_REPO_BRANCH`
- By defaulting to hardcoded `release`, we face an issue as the old cache is always being used

### 3. (Good to fix) Releasing to STAGING is manual
- When testing feature branch or pre-release on `staging`, it requires one to remember to delete the respective old cache from S3 manually
- Issue: it's very manual + there has been instances that we forgot
- Examples:
   - https://opengovproducts.slack.com/archives/C06R4DX966P/p1743136609253279
   - https://opengovproducts.slack.com/archives/C06R4DX966P/p1744273896688979?thread_ts=1744271650.769819&cid=C06R4DX966P

## Solution

**Breaking Changes**

**Improvements**:

- Undoing https://github.com/opengovsg/isomer/pull/1210
   - removing GH action workflow
   - remove hardcoding of `release` branch
   - thus, we also remove the need for `release` branch
- get the latest version tag using a concatment of git command
  -  note: assumption here is that we follow the semantic versioning scheme, which is pretty safe asssumption
  - this also let us not have to rely on the github API
- for builds with a overwriting branch name, we add the short hash of that commit to the branch name, allowing one to do the following without having to mentally remember to remove existing cache from S3:
   - continue building and testing on a feature branch
   - run pre-release checks on staging easily

### DIFF
--- a/tooling/build/scripts/publisher.sh
+++ b/tooling/build/scripts/publisher.sh
@@ -21,6 +21,7 @@ if [ -z "$ISOMER_BUILD_REPO_BRANCH" ]; then
   # awk '{print $2}': Prints the second column of the last line, which is the tag name.
   # sed 's/refs\/tags\///': Removes the 'refs/tags/' prefix from the tag name.
   ISOMER_BUILD_REPO_BRANCH=$(git ls-remote --tags --sort='v:refname' https://github.com/opengovsg/isomer.git | tail -n1 | awk '{print $2}' | sed 's/refs\/tags\///')
+  IS_USING_RELEASE_TAG=true
 fi
 
 # Cloning the repository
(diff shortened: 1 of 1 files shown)


dcshzj:
Hmm there are still two pending issues that we are not solving, which I feel are more important:
1. Dependency on GitHub to clone the repository (GitHub goes down, none of the sites can be built).
2. Race conditions when multiple builds are running concurrently to build the first node_modules cache on the S3 bucket (some builds might fail due to conflict in writing to the S3 bucket).

I thought we aligned previously that this was only a temporary measure, and that the full solution should be that:
1. We host a copy of the repository somewhere in an S3 bucket (and by extension, the node_modules zip).
2. Upon release, the GitHub Actions job will update this copy in the S3 bucket and also update the node_modules zip.
3. Sites will always build by pulling contents from this S3 bucket and not GitHub.

For branches in development, we can either skip having this cache, or like what you proposed is to also put the commit hash as part of the cache key.

adriangohjw:
@dcshzj yup you are right
- github dependency: IMO not as big of a priority now and might be so a non-goal for this hotfix
- race conditions: yup, though it's not as common at the moment - only when multiple builds are triggered ~concurrently after a new release

intention is to have this PR to hotfix current release issues in point 1 and 2 (_3 is good to have_) since they will affect the next release, while i spend sometime thinking about the full solution we discussed as i have some concerns e.g. potential performance impact to build process, and if the repo artifact should be stored in `production` account (which is what we do for cache now) or it should be stored in a separate AWS account to be accessed by all environemnt

harishv7:
@dcshzj @adriangohjw can we have a deeper discussion on the "copy of repository" method? Not too sure what's the intent - both S3 and GH have same uptime SLA of 99.9% as well

For now, ok with this PR's intent to optimise release based on tags

harishv7 (approved):

PR #1238, opened 2025-04-14
chore: update vulnerability report URL in Footer component

adriangohjw:
## Problem

Currently using non-gogov link

## Solution

- update to go.gov.sg link


dcshzj:
FYI will be updating this in #1199

PR #1239, opened 2025-04-14
feat(singpass): set up e2e tests for singpass login

seaerchin:
## Problem
1. This sets up e2e scaffolding for our application, specifically for the singpass login flow. for test specs, refer to teh ticket.

## Solution
1. setup test containers for e2e tests - we don't do a full build but expose a command to start the services which are required for e2e testing
2. add `mockpass` to our dependencies for studio (via `compose`) - this is only done for singpass mocking purposes; in production, this is a third party service wihch we use via API
3.  scaffold `LoginPage` as a test fixture; we should reuse this next time to login if required.

## Follow-ups
- [x] build our docker container adn run it in `CI` 
- [x] run `test:e2e` as part of `CI` for release


apps/studio/.env.test
```diff
@@ -10,26 +10,16 @@ NEXT_PUBLIC_APP_ENV='test'
 
 # Used for E2E testing
 DATABASE_URL=postgres://root:root@localhost:5432/test
-
+CI=false
```
dcshzj, line 13: question: eh, why do we need to disable and are there any downstream impact?

apps/studio/docker-compose.yml
```diff
@@ -12,5 +12,19 @@ services:
     volumes:
       - postgres-volume:/var/lib/postgresql/data
 
+  mockpass:
+    build: https://github.com/isomerpages/mockpass.git
```
dcshzj, line 16: issue: I know we forked because mockpass got some issues, but are we able to either:
1. Have it hosted on `opengovsg` org (we might deprecate `isomerpages` alr in the future, just making sure we don't forget)
2. Use a specific branch inside the original `mockpass` repo, then reference it from here?

apps/studio/tests/common.ts
```diff
@@ -0,0 +1,190 @@
+import { dirname, join } from "path"
+import { fileURLToPath } from "url"
+import type { StartedNetwork, StartedTestContainer } from "testcontainers"
+import { GenericContainer, Wait } from "testcontainers"
+import { z } from "zod"
+
+type ContainerType = "database" | "studio"
+export const CONTAINER_CONFIGURATIONS: {
```
dcshzj, line 8: question(non-blocking): sorry a bit dumb, how come we don't need to do the same for the mockpass container here?

apps/studio/tests/e2e/singpass.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // NOTE: The function for verification of otp does a comparison between the hash of the submitted token
+  // and the VerificationToken.token in db.
+  await loginPage.fillToken(editorEmail)
+  await signinButton.click()
+
+  await loginPage.defaultMockpassLogin()
+  const continueButton = page.getByRole("link", {
+    name: "Continue to Isomer Studio",
+  })
```
dcshzj, line 55: issue: I think this should be the one that we are asserting, to ensure that users see the "Continue to Isomer Studio" page if it is the first time they log in.

apps/studio/.env.test
seaerchin: this was a mistake, lemme set to `true`

apps/studio/docker-compose.yml
seaerchin: sure, i'll raise a PR back to mockpass for this

apps/studio/tests/common.ts
seaerchin: we don't need this, lemme remove it. we do `setup/teardown` via `docker compose` for e2e, and we only setup postgres for integration

apps/studio/tests/e2e/singpass.test.ts
seaerchin: heya, could you explain what you mean here

dcshzj: Yup, I think the main distinction between the first-time login and subsequent login user is that they are presented with a screen with the big blue button that says "Continue to Isomer Studio", which I think we should be doing the assertion here.

However, we are asserting that the user sees the edit profile details modal, which is not really true to the intent of this test because the user can be an existing user before we toggled on the Singpass login, or the user might be a returning user that has closed the edit profile modal.

seaerchin:
fyi @dcshzj - some actionables from me before this PR can go in: 

1. need to update `mockpass` behaviour to match singpass spec 
2. need to update `mockpass` build to docker so that we can use it for local...

apps/studio/tests/e2e/singpass.test.ts
seaerchin: ok, thanks for this explanation - so to doublecheck my understanding, this button only shows up on the first login? on subsequent logins, they just get shown all the sitse they have access to right?

dcshzj: Yup that's correct, and the edit profile details modal if they never added their details before

dcshzj (approved):

PR #1240, opened 2025-04-14
chore: add error logging for Singpass auth

dcshzj:
## Problem

We need to add error logging and monitoring for the new 2FA flows, so we know and can respond to issues that users might face.

## Solution

**Breaking Changes**

**Improvements**:

- Add logging with context to the authentication flows, so that they can be monitored via alarms on Datadog.
- Also add unit tests for the added/updated flows.

### DIFF
--- a/apps/studio/src/server/modules/auth/singpass/__tests__/singpass.router.test.ts
+++ b/apps/studio/src/server/modules/auth/singpass/__tests__/singpass.router.test.ts
@@ -1,0 +1,370 @@
+import { TRPCError } from "@trpc/server"
+import { resetTables } from "tests/integration/helpers/db"
+import {
+  applySession,
+  createMockRequest,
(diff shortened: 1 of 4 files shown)


apps/studio/src/server/modules/auth/singpass/__tests__/singpass.router.test.ts
seaerchin, line 1: nits for code standardisation that @adriangohjw raised before that i thought made sense - since this is a `router.test` file, should we instead write the error code rather than simply saying "throw error"?

apps/studio/src/server/modules/auth/singpass/singpass.router.ts
seaerchin, line 1: [enhancement] up to you to decide, but we can tag this trace in dd?

PR #1241, opened 2025-04-14
Feat/god mode - create new sites

adriangohjw:
### This is a fun-ctional PR!

> _[God Mode: a general purpose term for a cheat code in video games that makes a player invincible](https://en.wikipedia.org/wiki/Glossary_of_video_game_terms#God_mode)_

- There are some on-call support load needed from now and then e.g. creating new sites
- While these actions are simple, they require context switching + setting up OGP VPN, environment etc.
- This new initiative of God Mode serves to be the bedrock for building internal tools for admins. Instead of building scripts to be run on an adhoc basis by oncalls, this allows all isomer admins to self service
- in the case of this PR, it allows isomer admins to create sites on their own instantly without relying on engineers
- Ideally eventually we will sunset some of these features as they move to proper studio features

## Solution

**Features**:

- Bedrock for God Mode template (that can be replicatred and extended for other features)
- Moving "create site" function from script to a button in Isomer Studio
- Create a permission service that checks if user is isomer admin

https://github.com/user-attachments/assets/fd8e8e2e-a7f0-4a95-95b2-edb360269e60

### DIFF
--- a/apps/studio/src/server/modules/site/site.service.ts
+++ b/apps/studio/src/server/modules/site/site.service.ts
@@ -309,7 +309,6 @@ export const createSite = async ({ siteName }: CreateSiteProps) => {
       )
       .executeTakeFirstOrThrow()
 
-    console.log(`Added site ${siteName} with id ${siteId} to database`)
     return siteId
   })
 


apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -213,3 +217,109 @@ export const clearSiteNotification = async ({
...
+      })
+
+      .onConflict((oc) =>
+        oc.column("draftBlobId").doUpdateSet((eb) => ({
+          draftBlobId: eb.ref("excluded.draftBlobId"),
+        })),
+      )
+      .executeTakeFirstOrThrow()
+
+    console.log(`Added site ${siteName} with id ${siteId} to database`)
```
seaerchin, line 312: nit: now don't need this already since it got shifted into bacskend

seaerchin (approved):
lgtm, didn't look too closely because i assumed we shifted from scripts -> backend. 

main things i checked was:
1. this route is blocked on frontend for non-admins 
2. backend route blocks api calls if non-admin

apps/studio/src/server/modules/site/site.service.ts
adriangohjw: nice catch! copied the script over directly and didn't remove - fixed in https://github.com/opengovsg/isomer/pull/1241/commits/9d644e0e5666ab89c4eb441d59adafebe844e762

PR #1242, opened 2025-04-14
Feat/god mode mass codebuild

adriangohjw:
## Problem

- An extension of https://github.com/opengovsg/isomer/pull/1241 to demo what's possible
- To solve the minior annoyance of having to manually trigger all the builds on codebuild

## Solution

**Breaking Changes**

**Features**:

- add a page to republish site / all sites

Note: demo video shows error toast because it's not connected to AWS

https://github.com/user-attachments/assets/721183d0-9090-476e-8c79-186e3101a5cb


apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -328,4 +339,66 @@ export const siteRouter = router({
 
       return createSite({ siteName })
     }),
+  publishOne: protectedProcedure
```
seaerchin, line 342: nit: can we stick to using `publish`

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -328,4 +339,66 @@ export const siteRouter = router({
...
+          }),
+      )
+
+    const sites = await db
+      .selectFrom("Site")
+      .selectAll()
+      .where("codeBuildId", "is not", null)
+      .execute()
+
+    return db.transaction().execute(async (tx) => {
```
seaerchin, line 391: [nit] i'm unsure of how important this is, given that the # of sites we have is low and unlikely to exceed 1000 

but it's important to note that this might hold the transaction for an extended period of time and be potentially harmful to transactions that need to update audit log/version. i think a quick and easy fix would just be to use a `map` since order doesn't really matter here and we can do a `Promise.all` to avoid holding an extended lock.

wdyt?

adriangohjw: yeah i think it makes sense to not put in transaction since in this use case, it doesn't matter even if they it throw an error after triggering build for only a portion of the sites - will fix it!

adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1242/commits/5508ae50874b0bded944d1879a48565a9c50ce9a

adriangohjw:
FYI: build failure unrelated to this change

seaerchin (approved):

PR #1245, opened 2025-04-15
Singpass fallback - Send notification

adriangohjw:
## Problem

Closes https://www.notion.so/opengov/Isomer-Next-Decision-Matrix-for-Disabling-Singpass-as-2FA-1d177dbba78880289845fc20f53f8503#1d277dbba78880f1b071ef622311fac6

## Solution

**Improvements**:

- create new template for email notification for publishing
- send email notification on publish (for Page and Resource, not Config since it's only isomer-admin only)
  - decide to always send it, instead of only when it's singpass is disabled, for a more consistent UX 
- refactor ways to getSiteName to use `config.siteName` instead of `name`


dcshzj:
I think needs broader team discussion, not too sure what others think about the proposed mitigation factors

adriangohjw:
will open new PR

PR #1247, opened 2025-04-16
fix: storybook rendering changes (out of nowhere 😔)

adriangohjw:
## Problem

Storybook starts failing after https://github.com/opengovsg/isomer/pull/1199 e.g. https://www.chromatic.com/build?appId=66543928751c844d159a31ef&number=3059&captureStack=capture-v7

suspect its due to the infrastracture upgrade from chromatic since issue started after that

image

## Solution

**Bug Fixes**:

- switch the order of env and manual overiding to not have it be overwritten e.g. https://www.chromatic.com/test?appId=66543928751c844d159a31ef&id=67ffce99c6d2e77ecc58d6d7
- create `generateAssetBaseUrl` so that we can set storybook's `NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME` as empty string for storybook so that it can fallback to use its static asset handling

**Improvements**:

- remove the need to reference `assetsBaseUrl` since we already set up static assets in storybook in https://github.com/opengovsg/isomer/pull/1222/files
- have `generateAssetBaseUrl()` as a single source of truth for asset base url

### DIFF
--- a/apps/studio/src/utils/generateAssetUrl.ts
+++ b/apps/studio/src/utils/generateAssetUrl.ts
@@ -1,7 +1,11 @@
 import { env } from "~/env.mjs"
 
+export const generateAssetBaseUrl = (): string => {
+  return env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME
+    ? `https://${env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME}`
+    : ""
+}
+
 export const generateAssetUrl = (url: string): string => {
-  return url.startsWith("/")
-    ? `https://${env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME}${url}`
-    : url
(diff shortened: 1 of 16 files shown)


apps/studio/src/utils/generateAssetUrl.ts
```diff
@@ -1,7 +1,11 @@
 import { env } from "~/env.mjs"
 
+export const generateAssetBaseUrl = (): string => {
```
dcshzj, line 3: question(non-blocking): curious why use a function instead of a constant?

dcshzj (approved):

apps/studio/src/utils/generateAssetUrl.ts
adriangohjw: oops, was trying out different way to fix this weird storybook issue and forgot to update. thanks for spotting, updated!

PR #1249, opened 2025-04-17
Fix collection sorting schema display

adriangohjw:
## Problem

Closes sorting option being shown with wrong value

image

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- set "hidden" rendering priority to 99999999999
- Add descriptive name to option (even if they won't be seen now, since its actually hidden)

## Before & After Screenshots

**AFTER**:

no longer being rendered at all

image

## Tests

1. Create a collection
2. Add a custom index page
3. Click on editing header
4. Check -> sorting options should not be displayed under the tags option

### DIFF
--- a/apps/studio/src/constants/formBuilder.ts
+++ b/apps/studio/src/constants/formBuilder.ts
@@ -3,7 +3,7 @@ export const JSON_FORMS_RANKING = {
   ArrayControl: 4,
   BooleanControl: 2,
   ConstControl: 2,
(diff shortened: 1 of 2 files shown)


adriangohjw:
chromatic failure unrelated to this PR
potential fix here: https://github.com/opengovsg/isomer/pull/1247

seaerchin (approved):

PR #1250, opened 2025-04-17
isom-1852 collection widget on homepage latest articles

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1852/collection-widget-on-homepage-latest-articles

## Solution

**Breaking Changes**

This has breaking change as studio will try to render the component in preview and crash as there's no sitemap. There's an attempt to fix it in https://github.com/opengovsg/isomer/pull/1277

**Features**:

- added collection widget

**Improvements**:

- largely copied the component design and setup from infocards since it's supposed to be 90% similar
- moved some utils and regex from Studio to Template for reusability and single source of truth

## Tests

- deployed to staging for preview: https://github.com/opengovsg/isomer/pull/1250#issuecomment-2818023719


adriangohjw:
@sehyunidaaa added storybook for homepage + published on staging for your review

https://test-isomer-vica-ncss-staging.isomer.gov.sg/
image

apps/studio/src/features/editing-experience/constants.ts
```diff
@@ -41,12 +44,11 @@ export const TYPE_TO_ICON: Record<
...
+  [COLLECTION_WIDGET_TYPE]: TbDatabase,
   // TODO: Add in these new block types
   // table: BiTable,
   // divider: DividerIcon,
   // iframe-gmap,
   // iframe-formsg
   // iframe-youtube
 }
-
-export const REFERENCE_LINK_REGEX = /\[resource:(\d+):(\d+)\]/
```
adriangohjw, line 52: moved it to component lib since we also move `getResourceIdFromReferenceLink` there

packages/components/src/templates/next/components/complex/CollectionWidget/CollectionWidget.tsx
```diff
@@ -0,0 +1,199 @@
...
+      },
+      3: {
+        grid: "md:grid-cols-2 lg:grid-cols-3",
+      },
+    },
+  },
+  defaultVariants: {
+    numberOfCards: NUMBER_OF_PAGES_TO_DISPLAY,
+  },
+})
```
adriangohjw, line 63: FYI: largely copied over from infocards since the design is supposed to be the same. Decided to duplicate since there are some minor differences tho

packages/components/src/templates/next/components/complex/CollectionWidget/CollectionWidget.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const isExternalLink = !!referenceLinkHref && isExternalUrl(referenceLinkHref)
+
+  return (
+    <Link
+      href={referenceLinkHref}
+      className={compoundStyles.cardContainer()}
+      LinkComponent={LinkComponent}
+      isExternal={isExternalLink}
+    >
+      {/* TODO: Add fallback image if displayThumbnail but image is not present e.g. link variant */}
```
adriangohjw, line 88: clarifying with Rachel & Sehyun - but for now good enough for Factually use case since they all have thumbnails

packages/components/src/templates/next/components/complex/CollectionWidget/CollectionWidget.tsx
```diff
@@ -0,0 +1,199 @@
...
+  LinkComponent,
+  collectionReferenceLink,
+  customTitle,
+  customDescription,
+  displayThumbnail,
+  displayCategory,
+  buttonLabel,
+  shouldLazyLoad,
+}: CollectionWidgetProps): JSX.Element => {
+  const collectionId = getResourceIdFromReferenceLink(collectionReferenceLink)
```
adriangohjw, line 142: Thought about adding a check that the siteId and collectionId indeed belongs to this site, but decided to skip it as am thinking this error will only happen by manual error by isomer team edits

packages/components/src/templates/next/components/complex/CollectionWidget/CollectionWidget.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const collectionPages = getCollectionPages({
+    site,
+    collectionParent,
+  })
+
+  if (collectionPages.length === 0) {
+    return <></>
+  }
+
+  return (
```
adriangohjw, line 155: largely dup. from infocards too, with some tweaks

packages/components/src/templates/next/components/complex/CollectionWidget/utils/getCollectionPages.ts
```diff
@@ -0,0 +1,42 @@
...
+    sortDirection:
+      collectionParent.collectionPagePageProps?.defaultSortDirection ??
+      COLLECTION_PAGE_DEFAULT_SORT_DIRECTION,
+  })
+
+  if (items.length === 0) {
+    throw new Error(
+      `CollectionWidget: No collection items found for reference link ${collectionParent.permalink}`,
+    )
+  }
```
adriangohjw, line 39: just a failsafe to force codebuild failureas there's no graceful fallback and this should not happen - would rather it trigger codebuild failure alarms so that we can investigate - open to better ideas

packages/components/src/templates/next/components/complex/CollectionWidget/utils/getCollectionParent.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    if (currentNode.children && currentNode.children.length > 0) {
+      queue.push(...currentNode.children)
+    }
+  }
+
+  throw new Error(
+    `CollectionWidget: No collection parent found for collection ID ${collectionId}`,
+  )
```
adriangohjw, line 39: just a failsafe to force codebuild failureas there's no graceful fallback and this should not happen - would rather it trigger codebuild failure alarms so that we can investigate - open to better ideas

packages/components/src/templates/next/components/complex/index.ts
```diff
@@ -12,3 +12,5 @@ export { default as KeyStatistics } from "./KeyStatistics"
 export { Map } from "./Map"
 export { Video } from "./Video"
 export { DynamicDataBanner } from "./DynamicDataBanner"
+export { CollectionWidget } from "./CollectionWidget"
+export { LogoCloud } from "./LogoCloud"
```
adriangohjw, line 16: was left out previously

packages/components/src/templates/next/layouts/Collection/Collection.tsx
```diff
@@ -1,59 +1,14 @@
...
 import { Skeleton } from "../Skeleton"
 import CollectionClient from "./CollectionClient"
 import {
   getAvailableFilters,
   getCollectionItems,
+  processCollectionItems,
   shouldShowDate,
 } from "./utils"
 
-const processedCollectionItems = (
```
adriangohjw, line 18: moved to standalone util file for reusability

packages/components/src/interfaces/complex/DynamicDataBanner.ts
```diff
@@ -10,7 +10,7 @@ export const DYNAMIC_DATA_BANNER_TYPE = "dynamicdatabanner"
 // Hardcoded for now because
 // 1. MUIS is the only use case and there's always 6 prayer timeslots
 // 2. No other known use cases have been identified
-export const NUMBER_OF_DATA = 6
+export const DYNAMIC_DATA_BANNER_NUMBER_OF_DATA = 6
```
adriangohjw, line 13: purpose: to improve clarity as `NUMBER_OF_DATA` can be misleading, esp. when i started working on this PR and we were considering giving users the option to determine 3/6 pages to display

packages/components/src/templates/next/components/complex/CollectionWidget/CollectionWidget.tsx
sehyunidaaa: @adriangohjw 

- is the numberOfCards something that will be exposed to users?
- can we get the cardImageContainer to have 3:2 aspect ratio across all viewports? it's something I'm adding to infocards too (PR in review). will leave comment on chromatic as well

adriangohjw: donezo 👍

apps/studio/public/assets/css/preview-tw.css
adriangohjw, line 1: mostly unrelated, and are forgotten from the infocards image resizing PR

packages/components/src/types/sitemap.ts
```diff
@@ -19,8 +20,18 @@ interface IsomerBaseSitemap {
 }
 
 interface IsomerPageSitemap extends IsomerBaseSitemap {
-  layout: Exclude<IsomerPageLayoutType, "file" | "link">
+  layout: Exclude<IsomerPageLayoutType, "collection" | "file" | "link">
 }
+
+export interface IsomerCollectionPageSitemap extends IsomerBaseSitemap {
+  layout: Extract<IsomerPageLayoutType, "collection">
+  // TODO: Reconsider how this is done as currently every item in the sitemap has the same props
```
dcshzj, line 28: thought: I thought that this setup makes sense as we can make layout-specific sitemap entries that's still type-safe, so we can extract these extra props if we detect the sitemap entry layout is different.

No action required for now, just curious about this TODO what are we able to potentially do

packages/components/src/interfaces/complex/CollectionBlock.ts
```diff
@@ -0,0 +1,74 @@
...
+  {
+    type: Type.Literal(COLLECTION_BLOCK_TYPE, {
+      default: COLLECTION_BLOCK_TYPE,
+    }),
+    collectionReferenceLink: Type.String({
+      title: "Collection",
+      description: "The collection to display pages from",
+      format: "link",
+      pattern: REF_INTERNAL_HREF_PATTERN,
+    }),
```
dcshzj, line 20: issue: hmm using the reference link means this is only available for Studio sites and not GitHub ones that we are still supporting. Should we support using permalinks as well?

packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
```diff
@@ -0,0 +1,221 @@
...
+        <h2 className={compoundStyles.headingTitle()}>
+          {customTitle ?? collectionParent.title}
+        </h2>
+        <p>{customDescription ?? collectionParent.summary}</p>
+      </div>
+
+      <div className={compoundStyles.grid({ numberOfCards })}>
+        {collectionPages.map((card, idx) => (
+          <SingleCard
+            key={idx}
```
dcshzj, line 192: issue: I think we should use `card.id` instead of the index, as index is not recommended to be a key in React

packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionParent.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface GetCollectionParentProps {
+  site: IsomerSiteProps
+  collectionId: string
+}
+
+export const getCollectionParent = ({
+  site,
+  collectionId,
+}: GetCollectionParentProps): IsomerCollectionPageSitemap => {
```
dcshzj, line 19: question: why not `getSitemapAsArray` then find the node from the array?

tooling/build/scripts/generate-sitemap.js
```diff
@@ -90,6 +90,16 @@ const getSiteMapEntry = async (fullPath, relativePath, name) => {
     tags: schemaData.page.tags,
   }
 
+  if (schemaData.layout === "collection") {
```
dcshzj, line 93: issue: I think we also need to do this for `index.ts` inside `publishing` for it to take effect for Studio sites

adriangohjw: thanks! updated in https://github.com/opengovsg/isomer/pull/1250/commits/2c6bd3cefd3a7633d3bb77e367f015c60bb34ebe

packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1250/commits/406048d8b434aaeeae82b6c55360835e6f9fc009

packages/components/src/interfaces/complex/CollectionBlock.ts
adriangohjw: updated to allow legacy too https://github.com/opengovsg/isomer/pull/1250/commits/6fc1dde274642120e563e3d5c17b8d477cc92715

packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionParent.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1250/commits/b130d7de96f59669511e9622d489bb75c45ffac3 - didn't occur to me that we do that. makes a lot of sense esp. since we using it for generating collection aldy. thanks!

packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionParent.ts
```diff
@@ -0,0 +1,29 @@
...
+  const items = site.siteMap.children?.flatMap((child) =>
+    getSitemapAsArray(child),
+  )
```
dcshzj, line 16: minor nit: I think we should be able to do `getSitemapAsArray(site.siteMap)` directly

packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
```diff
@@ -0,0 +1,221 @@
...
+  LinkComponent,
+  shouldLazyLoad,
+  numberOfCards,
+}: CollectionBlockSingleCardProps): JSX.Element => {
+  const isExternalLink = !!referenceLinkHref && isExternalUrl(referenceLinkHref)
+
+  const renderImage = () => {
+    const hasImage = image?.src !== undefined
+    const imageSrc = hasImage ? image.src : site.logoUrl
+    const imageAlt = hasImage ? image.alt : `Site logo for ${site.siteName}`
```
dcshzj, line 98: nit: can add a comment here that we fallback to the site logo if the image is missing

dcshzj (approved):

PR #1251, opened 2025-04-21
Add growthbook for isomer core

adriangohjw:
## Problem

Some features should only be limited to core team e.g. god mode

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- duplicate growthbook flag from `isomer_admins` to `isomer_admins_and_migrators`
- TODO: remove migrators' email from `isomer_admins` after this is launched


apps/studio/src/hooks/useIsUserIsomerUsers.ts
```diff
@@ -0,0 +1,29 @@
...
+  includeMigrators,
+}: UseIsUserIsomerUsersProps) => {
+  const {
+    me: { email },
+  } = useMe()
+
+  const { users } = useFeatureValue<GrowthbookIsomerUsersFeature>(
+    includeMigrators
+      ? ISOMER_ADMINS_AND_MIGRATORS_FEATURE_KEY
+      : ISOMER_ADMINS_FEATURE_KEY,
```
dcshzj, line 24: question: hmm any reason why we decided to split into two feature flags? I thought the current design is quite extensible whereby the `users` key is the admins, then we can create a new key `migrators` (or can be called `guests`, since we can consider interns in the future) within the same feature flag.

Right now means we would have to duplicate the list of Isomer admins.

adriangohjw: ah no particular reason, i didn't realize growthbook config is a JSON - i tot it returns an array that we pass into `users`. Closed this PR and created a new one https://github.com/opengovsg/isomer/pull/1252

PR #1253, opened 2025-04-22
chore: reduce dependabot PR limit

dcshzj:
## Problem

Dependabot is opening too many PRs and we don't have the bandwidth to look into them.

## Solution

**Breaking Changes**

**Improvements**:

- Reduce the limit on the number of PRs open by dependabot from 10 to 5.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: daily
     versioning-strategy: increase
-    open-pull-requests-limit: 10
+    open-pull-requests-limit: 5
     groups:
       aws-sdk:
         applies-to: version-updates


.github/dependabot.yml
```diff
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: daily
     versioning-strategy: increase
-    open-pull-requests-limit: 10
+    open-pull-requests-limit: 5
```
seaerchin, line 8: i think the strategy should be to auto merge minor version increases and wait for major version increases tbh

dcshzj: The auto merging is done via mergify

dcshzj (commented):
> IMO shouldn't matter that it opens up many PRs since they are labeled dependencies and if need be, we can just ignore it (https://github.com/opengovsg/isomer/pulls?q=is%3Aopen+is%3Apr+-label%3Adependencies)

We shouldn't be keeping too many open PRs, in fact it is ideal if we keep it less than 20 so it's easier to identify and focus on getting PRs merged. Right now at 10 (including all the draft PRs we have) it crowds out all the other legitimate PRs which makes it hard to know which one actually needs review.

> Plus, the upside of having a higher number means that in a situation where there's are a few major version updates (e.g. now with react, trpc, prisma...), it wouldn't block PRs for non-breaking smaller updates from being opened and automatically closed

We should close those PRs if we don't intend to update at that point in time. Problem now is that we have too many open PRs which is making it really difficult to push for code reviews.

.github/dependabot.yml
seaerchin: ye, i think that should be the way. we already auto merge non-major version upgrades. if the problem is that we don't have capacity to address major version ugprades, then we should just disable it.

seaerchin (approved):

PR #1254, opened 2025-04-22
feat(index-pages): create index pages automatically

seaerchin:
## Problem
1. index pages should be automatically created if they don't exist at present
2. we should show the UI for index pages 
3. we should remove the admin create index page button for folders as the index page will always be generated now

Closes ISOM-1890
Closes ISOM-1891
Closes ISOM-1792

## Solution
1. reuse `HomepageRow` styling for `IndexpageRow`
2. on load of the folder, we call `getIndexPage`; if the index page doesn't exist, we create it then
4. write through the updates to `title` for folder, so that index page and folder will remain in sync

## Videos

https://github.com/user-attachments/assets/04ed5e04-8788-44b0-b685-8ef9a5ca22b2


apps/studio/src/features/dashboard/components/IndexpageRow/IndexpageRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+      gap="0.75rem"
+      px="1.25rem"
+      py="0.5rem"
+      w="full"
+      bg="base.canvas.default"
+      border="1px solid"
+      borderColor="base.divider.medium"
+      borderRadius="8px"
+      layerStyle="focusRing"
+      _active={{}}
```
dcshzj, line 34: nit: can remove

apps/studio/src/features/dashboard/components/IndexpageRow/IndexpageRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+        Edit page
+      </Text>
+      <IconButton
+        as="div"
+        aria-hidden
+        variant="clear"
+        pointerEvents="none"
+        colorScheme="neutral"
+        icon={<BiChevronRight />}
+        aria-label="edit homepage"
```
dcshzj, line 75: issue: should be `edit page` or `edit index page`

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +265,58 @@ export const folderRouter = router({
         return pick(result, defaultFolderSelect)
       },
     ),
+
+  getIndexpage: protectedProcedure
+    .input(getIndexpageSchema)
+    .query(async ({ input: { resourceId, siteId } }) => {
+      // TODO: add permissions checking
```
dcshzj, line 272: issue: I think should add this perms check before merge? A bit scary to skip cos Maps team had an incident due to missing perms check. I think should be quite similar to the perms for editing normal pages

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +265,58 @@ export const folderRouter = router({
...
+              parentId: resourceId,
+              draftBlobId: indexPageBlob.id,
+              title,
+              type: ResourceType.IndexPage,
+              permalink: INDEX_PAGE_PERMALINK,
+              siteId,
+            })
+            .returning(["id", "draftBlobId"])
+            .executeTakeFirstOrThrow()
+
```
dcshzj, line 311: issue: I think we are missing an audit log entry here?

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +265,58 @@ export const folderRouter = router({
...
+        let indexPage = await tx
+          .selectFrom("Resource")
+          .where("Resource.parentId", "=", resourceId)
+          .where("Resource.type", "=", "IndexPage")
+          .select(["id", "draftBlobId"])
+          .executeTakeFirst()
+
+        // NOTE: Might want to create on click in
+        // rather than via the view
```
dcshzj, line 289: issue: actually I do agree with this, feels a bit weird to me that the index page is created when you are browsing the folder with no intention of touching the index page.

suggestion: I think the `getIndexpage` can return `undefined` if the folder index page does not exist, then the frontend can call a separate mutation to create the index page on click before redirecting the user to that newly created index page. If already exists, we can maintain the current behaviour as a normal link.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -195,7 +195,6 @@ export const resourceRouter = router({
         .selectFrom("Resource")
         .select(["title", "permalink", "type", "id", "parentId"])
         .where("Resource.type", "!=", ResourceType.RootPage)
-        .where("Resource.type", "!=", ResourceType.IndexPage)
```
dcshzj, line 198: question: hmm why did we remove this? `getChildrenOf` looks like it is used in `useResourceQuery` which is used in `ResourceSelector` as part of move items flow + adding links. I think for adding links, they should be linking to the folder rather than to the index page directly.

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -97,6 +106,27 @@ export const folderRouter = router({
...
+              parentId: folder.id,
+              draftBlobId: indexPageBlob.id,
+              title: folderTitle,
+              type: ResourceType.IndexPage,
+              permalink: INDEX_PAGE_PERMALINK,
+              siteId,
+            })
+            .returning(["id", "draftBlobId"])
+            .executeTakeFirstOrThrow()
+
```
dcshzj, line 129: issue: might want to also do the audit log for the creation of the index page

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -202,6 +232,21 @@ export const folderRouter = router({
               throw err
             })
 
+          // NOTE: update the index page's title so that they stay in sync
+          if (title) {
```
dcshzj, line 236: question(non-blocking): hmm seems like the guard clause is not needed here? `title` is a required field so it should always be defined?

seaerchin:
* **#1280** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1279** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1278** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1276** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1276?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1254** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1254?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1254?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/studio/src/server/modules/resource/resource.router.ts
seaerchin: this got removed because it's a bug fix. this method controls what's viewable from the sidebar, and we should do this in the subsequent method for `useResourceQuery`

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: ah good catch, i put it defensively without checking the type

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: no worries, i'll add it in hehe

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: yeap, done

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: added!

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: updated so that it throws an error, which we catch and then create the folder's index page. the initial decision was chosen because:

1. less waterfall timing (insteads of 2 requests, it's just one)
2. no need to handle loading states 

but given that they are separate responsibilities, it also makes sense to shift it out. i've went with this!

dcshzj: Clarified offline, I think my main concern is just cos audit logs now log everything, so someone (like our team) can be casually browsing the site and inevitably created a new index page, which might raise questions from the agency if they scrutinise the audit logs in very great detail (but assumption here is that they actually do).

I think the other way of creating only when clicking into it won't really solve this issue anyway, so I'm ok to keep the current behaviour. We can revisit if really got people make a lot of noise about it (highly doubt they will).

dcshzj (approved):

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +274,111 @@ export const folderRouter = router({
...
+                code: "BAD_REQUEST",
+                message: "User could not be found",
+              }),
+          )
+
+        await tx
+          .updateTable("Resource")
+          .where("id", "=", String(indexPage.id))
+          .set({ draftBlobId: indexPageBlob.id })
+          .execute()
```
adriangohjw, line 338: not sure if im missing out anything but why do we need this if we already set the `draftBlobId` in 309-320?

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +274,111 @@ export const folderRouter = router({
...
+    .input(getIndexpageSchema)
+    .query(async ({ ctx, input: { resourceId, siteId } }) => {
+      await validateUserPermissionsForResource({
+        siteId,
+        action: "read",
+        userId: ctx.user.id,
+        resourceId,
+      })
+
+      return db.transaction().execute(async (tx) => {
```
adriangohjw, line 361: not necessary to place them in a TX since this theres only DB reads

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +274,111 @@ export const folderRouter = router({
...
+        const { title } = await tx
+          .selectFrom("Resource")
+          .where("id", "=", resourceId)
+          .select("title")
+          .executeTakeFirstOrThrow()
+
+        const indexPage = await tx
+          .selectFrom("Resource")
+          .where("Resource.parentId", "=", resourceId)
+          .where("Resource.type", "=", "IndexPage")
```
adriangohjw, line 371: use `ResourceType.IndexPage`

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +274,111 @@ export const folderRouter = router({
...
+          delta: { before: null, after: indexPage },
+          by,
+          siteId,
+        })
+
+        return indexPage
+      })
+    }),
+
+  getIndexpage: protectedProcedure
```
adriangohjw, line 351: very nit - should this be `getIndexPage` with a uppercase `P` for standardization?

apps/studio/src/features/dashboard/components/IndexpageRow/IndexpageRow.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { BiChevronRight, BiFile, BiSolidCircle } from "react-icons/bi"
+
+import { trpc } from "~/utils/trpc"
+
+interface IndexpageRowProps {
+  siteId: number
+  resourceId: string
+}
+
+export const IndexpageRow = ({ siteId, resourceId }: IndexpageRowProps) => {
```
adriangohjw, line 14: question: thoughts on using `z.infer<typeof getIndexpageSchema>` as the type instead of redefining it as `IndexpageRowProps`? since the values passed in will be used to pass into `folder.getIndexpage`

apps/studio/src/features/dashboard/components/IndexpageRow/IndexpageRow.tsx
```diff
@@ -0,0 +1,94 @@
...
+              size="xs"
+              variant="clear"
+              colorScheme={data?.draftBlobId ? "warning" : "success"}
+            >
+              <BadgeLeftIcon fontSize="0.5rem" as={BiSolidCircle} />
+              <Text textStyle="legal">
+                {data?.draftBlobId
+                  ? ResourceState.Draft
+                  : ResourceState.Published}
+              </Text>
```
adriangohjw, line 64: question: from what im understanding, `data` will only be undefined on the first load as it will be created if not found and be then found. If that's the case, should we then check for `isLoaded={!isLoading || !data}` in line 34?

adriangohjw (commented):
will be good to write some basic tests for the API endpoints, or create tickets in the linear backlog

apps/studio/src/schemas/folder.ts
```diff
@@ -51,3 +51,12 @@ export const editFolderSchema = baseEditFolderSchema.superRefine(
...
 )
+
+export const baseIndexPageSchema = z.object({
+  resourceId: z.string(),
+  siteId: z.number().min(1),
+})
+
+export const createIndexPageSchema = baseIndexPageSchema
+
+export const getIndexpageSchema = baseIndexPageSchema
```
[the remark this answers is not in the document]
seaerchin: hmm i think good to standardise ba. i'll remove this method, this slipped my mind - thanks for reminding!

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -39,3 +39,20 @@ export const createDefaultPage = ({
     }
   }
 }
+
+export const createIndexPage = (title: string) => {
```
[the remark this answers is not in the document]
seaerchin: good call!

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -220,4 +274,111 @@ export const folderRouter = router({
...
+          .selectFrom("User")
+          .where("id", "=", ctx.user.id)
+          .selectAll()
+          .executeTakeFirstOrThrow(
+            () =>
+              new TRPCError({
+                code: "BAD_REQUEST",
+                message: "User could not be found",
+              }),
+          )
```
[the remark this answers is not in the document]
seaerchin: updated to use your `createIndexPage` method

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: sounds good

apps/studio/src/features/dashboard/components/IndexpageRow/IndexpageRow.tsx
seaerchin: good call, updated

apps/studio/src/features/dashboard/components/IndexpageRow/IndexpageRow.tsx
seaerchin: ah sounds good, i missed this

adriangohjw (approved):

seaerchin:
### Merge activity

* **May 8, 12:40 AM EDT**: @seaerchin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1254).

PR #1267, opened 2025-04-23
Fix/truncated 100th item

adriangohjw:
## Problem

ordered list beyond 100th item in an article page are truncated off

image

re: https://opengovproducts.slack.com/archives/C07CWUNUL68/p1745399835855599

## Solution

**Bug Fixes**:

- increase the padding slightly 

## Tests

- refer to storybook


seaerchin (commented):
lgtm but please wiat on sehyun as this changes the padding-start; not sure if we want to care when the # of items exceeds our new padding but i think this is good enough

PR #1277, opened 2025-04-24
Isom-1852 collection widget (studio preview RFC)

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1852/collection-widget-on-homepage-latest-articles

Currently collection block component needs to display content that are exclusively on Studio due to a sitemap generation and preview constraint

This is a working POC as a RFC to show how we might be able to do that while keeping Template and Studio as separated as possible

> [!NOTE]
> There's some overlaps with https://linear.app/ogp/issue/ISOM-1893/collection-widget-on-homepage-studio but it is not. Even if we ship 1893, we wouldn't be able to generate the collection items (since theres no sitemap still) so we still need a studio-specific view 

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- create an optional `studioProps` that can be passed in from Studio into `PreviewWithoutSitemap`
- when `studioProps` is present, render studio-specific content


dcshzj:
Haven't reviewed deeply yet but wanted to mention that it was a long-held belief that components should not be aware of the existence of Studio, so hmm I think we should have a deeper discussion about whether we should open up that possibility.

adriangohjw:
> Haven't reviewed deeply yet but wanted to mention that it was a long-held belief that components should not be aware of the existence of Studio, so hmm I think we should have a deeper discussion about whether we should open up that possibility.

yup agree! thats why i labeled it as RFC - not too sure where we want to draw the line as this rendering behaviour is strictly studio only. I tot about creating a separate studio-only collection block component but that also requires it to have some knowledge that its studio

PR #1282, opened 2025-04-29
isom-1885 add link to index pages on navbar

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1885/add-link-to-index-pages-on-navbar

Also did some refactoring but placed them in a separate branch and PR to reduce scope of PR https://github.com/opengovsg/isomer/pull/1266

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- add support for link for navbar parent item
- add storybook

## Tests

Should be checked via storybook, tho this is how you can verify too:
1. go to studio, open the homepage (or any page)
2. preview the navbar and interact with it - both desktop and mobile


packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
```diff
@@ -27,12 +36,81 @@ const mobileItemStyles = tv({
...
+      })}
+    >
+      <Link
+        LinkComponent={LinkComponent}
+        isExternal={isExternalUrl(referenceLinkHref)}
+        showExternalIcon={isExternalUrl(referenceLinkHref)}
+        isWithFocusVisibleHighlight
+        href={referenceLinkHref}
+        className={nestedItem({
+          className: `group/parent-item ${focusVisibleHighlight()}`,
```
seaerchin, line 93: [question] does this actually work? i thought we had to use the [syntax](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state) of `group-[state]`?

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
```diff
@@ -27,12 +36,81 @@ const mobileItemStyles = tv({
...
+const ParentItemLink = ({
+  name,
+  referenceLinkHref,
+  isExternal,
+  LinkComponent,
+}: ParentItemLinkProps) => {
+  // This is a hack to ensure that the rightArrow is always at the end of the last word even on smaller screens
+  const words = name.trim().split(" ")
+  const allButLastWord = words.slice(0, -1).join(" ")
+  const lastWord = words[words.length - 1]
```
seaerchin, line 77: [clarification] what happens iwthout this hack? does the right arrow then show between the lines on overflow?

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
```diff
@@ -27,12 +36,81 @@ const mobileItemStyles = tv({
...
+      >
+        <span className="row-gap-0 flex flex-row flex-wrap items-baseline gap-1">
+          Pages in
+          <span className="prose-headline-base-medium">{allButLastWord}</span>
+          <span className="prose-headline-base-medium flex flex-row items-center gap-1">
+            {lastWord}
+            {!isExternal && (
+              <BiRightArrowAlt
+                aria-hidden
+                className="text-[1rem] transition ease-in group-hover/parent-item:translate-x-1"
```
seaerchin, line 105: [question] i thuoght the group hover syntax had to be `group-hover` (from the [docs](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state))

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
```diff
@@ -27,12 +36,81 @@ const mobileItemStyles = tv({
...
+  isExternal,
+  LinkComponent,
+}: ParentItemLinkProps) => {
+  // This is a hack to ensure that the rightArrow is always at the end of the last word even on smaller screens
+  const words = name.trim().split(" ")
+  const allButLastWord = words.slice(0, -1).join(" ")
+  const lastWord = words[words.length - 1]
+
+  return (
+    <div
```
seaerchin, line 80: [nit] won't this mean that clicking on the div **won't** have an effect? if that's the case, wdyt about setting the `link` on the outer item?

seaerchin (commented):
minor comments re: styling, no issues wrt code

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
adriangohjw: ah yes but since there's 2 group being used here, we use `group-hover/{name}` to specify which group it belongs to (https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-nested-groups)

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
seaerchin: haha both lgtm... ._.

packages/components/src/templates/next/components/internal/Navbar/NavItem.tsx
```diff
@@ -99,18 +100,42 @@ export const NavItem = forwardRef<HTMLButtonElement, NavbarItemProps>(
...
+        LinkComponent={LinkComponent}
+        isExternal={isExternal}
+        showExternalIcon={isExternal}
+        isWithFocusVisibleHighlight
+        href={referenceLinkHref}
+        className="group inline-flex w-fit items-center gap-1.5 hover:text-brand-interaction-hover hover:no-underline"
+      >
+        {name}
+        {!isExternal && (
+          <BiRightArrowAlt className="text-[1.5rem] transition ease-in group-hover:translate-x-2.5" />
```
seaerchin, line 133: [no action required] i think we shuold've extracted this out as a component hahahahhaha

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
adriangohjw: oh i believe the confusion will be because the right arrow is aligned with the dropdown arrow. @sehyunidaaa can chip in on this design requirement

seaerchin (approved):

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavItemAccordion.tsx
sehyunidaaa: yeahh @adriangohjw  i think nicer if it's right after the last word (so not aligned to the chevrons)

but unrelated why is this chevron so tiny!! 
image

PR #1283, opened 2025-04-29
isom-1901 add an optional button to navigation bar

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1901/add-an-optional-button-to-navigation-bar

> [!NOTE]
> This PR only address the UI changes but will not be able to be used. Refer to https://github.com/opengovsg/isomer/pull/1286 on how we will migrate existing data structure over. 

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add support for optional CTA

## Tests

you can verify this via storybook, but to verify in usage:
1. go to studio and choose one site
2. open up the admin config setting and update navbar json to add the following as one of the key-value pair
   ```
   "callToAction": { "url": "https://google.com", "label": "cool CTA" }
   ```
3. click save
4. go to the studio preview for one of the page (or published site) and verify it is working


seaerchin (commented):
not sure if i'm understanding this pr correctly - the initial entrypoint is via `navbar` on the `callToAction` prop, but it's not being passed in anywhere? (ie, this won't be available for users to use?)

adriangohjw:
> not sure if i'm understanding this pr correctly - the initial entrypoint is via `navbar` on the `callToAction` prop, but it's not being passed in anywhere? (ie, this won't be available for users to use?)

ah good catch! 🤦 

that will be done in https://github.com/opengovsg/isomer/pull/1286 - im proposing a change in interface for NavBar.content stored

packages/components/src/templates/next/components/internal/Navbar/Navbar.tsx
```diff
@@ -43,6 +44,19 @@ export const Navbar = ({
         assetsBaseUrl: site.assetsBaseUrl,
         lazyLoading: false, // will always be above the fold
       }}
+      callToAction={
```
seaerchin, line 47: nit: prefer `&&` over ternary here cos both end up with `undefined`, which makes the `&&` easier to read

seaerchin (commented):
(no issues, just making sure i understand how this prop is passed via the migration pr before i approve)

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavbarMenu.tsx
```diff
@@ -42,6 +44,19 @@ export const MobileNavMenu = forwardRef<HTMLDivElement, MobileNavMenuProps>(
...
+              <div className="border-y border-b-base-divider-subtle bg-base-canvas-alt px-6 py-3">
+                <LinkButton
+                  href={callToAction.referenceLinkHref}
+                  isExternal={callToAction.isExternal}
+                  className="h-fit w-full justify-center"
+                  isWithFocusVisibleHighlight
+                  LinkComponent={LinkComponent}
+                >
+                  {callToAction.label}
+                </LinkButton>
```
seaerchin, line 57: [issue] because of our types, we can have a link button shown without a `referenceLinkHref` - should we tighten this up?

packages/components/src/interfaces/internal/Navbar.ts
```diff
@@ -27,9 +32,16 @@ export interface BaseNavbarProps {
...
+  callToAction?: NavbarCallToAction & {
+    url: string
+  }
   site: IsomerSiteProps
 }
 
 export interface NavbarClientProps extends BaseNavbarProps {
+  callToAction?: NavbarCallToAction & {
+    referenceLinkHref?: string
+    isExternal: boolean
```
seaerchin, line 44: [nit] `isExternal` is a computed property of `referenceLinkHref`, so more in favour of just computing it at compile time rather than storing it as data

seaerchin (commented):
if i'm understanding this PR correctly, we now have this shape:

```ts
navbar: {
  // what we used to have
  items: [oldNavBar], 
  // proposed change
  callToAction:  { referenceLinkHref, isExternal, label }
}
```

we've updated the shape of `Navbar` itself, but at **runtime**, we're still passing in `navBarItems`. to verify this, you can go to `template/page.tsx`, which renders the actual page we see

packages/components/src/interfaces/internal/Navbar.ts
adriangohjw: oh `NavbarClientProps` are all computed at build time

adriangohjw:
> if i'm understanding this PR correctly, we now have this shape:
> 
> ```ts
> navbar: {
>   // what we used to have
>   items: [oldNavBar], 
>   // proposed change
>   callToAction:  { referenceLinkHref, isExternal, label }
> }
> ```
> 
> we've updated the shape of `Navbar` itself, but at **runtime**, we're still passing in `navBarItems`. to verify this, you can go to `template/page.tsx`, which renders the actual page we see

@seaerchin 

```ts
navbar: {
  // what we used to have
  items: [oldNavBar], 
  // proposed change
  callToAction:  { url, label }
  // { referenceLinkHref, isExternal } are generated on the fly when converting it to a "use client" component for Navbar
}
```

seaerchin (approved):
lgtm, for reference, the template changes to ensure that `navbar` is passed in rather than `navbaritems` at runtime is done via the downstream pr. 

1. first, we migrate all usages of `navbarItems` -> `navbar`
2. we then update the code to use `navbar` instead of `navbarItems`

adriangohjw:
note to self: fix chromatic before merging next week

PR #1284, opened 2025-04-30
feat: add toggle to switch to new index page

dcshzj:
## Problem

Closes ISOM-1897.

## Solution

**Breaking Changes**

**Features**:

- Add a new toggle for converting custom index pages into the new index page.

## Screenshots

image
image

## Tests

- [ ] Create a new index page
- [ ] Use the raw JSON editor mode to convert the `layout` of the page from `index` to `content` and remove the `childpages` key. Save the changes.
- [ ] Verify that the infobox appears at the top.
- [ ] Click on preview changes, verify that you should be able to see a preview of the index page and you are not able to click into anything inside the editor.
- [ ] Verify that the "Keep old version" button brings you correctly back to the original page state.
- [ ] Verify that the "Accept this change" opens up a confirm changes modal
- [ ] Verify that the "Accept changes" button causes the index page changes to be persisted.
- [ ] Use the raw JSON editor mode to verify that the `layout` of the page is now changed back to `index` and that there is a `childpages` key.

### DIFF
--- a/apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
@@ -159,7 +159,7 @@ export default function RootStateDrawer() {
     // NOTE: This is defined but we just do the assertion here
     // because the type for `DEFAULT_BLOCKS` is possibly undefined
     // so `ts` cannot infer
-    if (!DEFAULT_BLOCKS.childpage) return
+    if (!DEFAULT_BLOCKS.childrenpages) return
 
     const newPageState = {
       ...savedPageState,
@@ -167,7 +167,10 @@ export default function RootStateDrawer() {
     // NOTE: This layout needs to be outside, otherwise it is treated as a
     // string type rather than the array of layout consts
     newPageState.layout = "index"
-    newPageState.content = [...savedPageState.content, DEFAULT_BLOCKS.childpage]
+    newPageState.content = [
+      ...savedPageState.content,
+      DEFAULT_BLOCKS.childrenpages,
+    ]
     setPreviewPageState(newPageState)
     setIsPreviewingIndexPage(true)
   }, [savedPageState, setPreviewPageState])
--- a/apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
@@ -167,7 +167,7 @@ export default function RootStateDrawer() {
     // NOTE: This layout needs to be outside, otherwise it is treated as a
     // string type rather than the array of layout consts
     newPageState.layout = "index"
-    newPageState.content = [DEFAULT_BLOCKS.childpage]
+    newPageState.content = [...savedPageState.content, DEFAULT_BLOCKS.childpage]
     setPreviewPageState(newPageState)
     setIsPreviewingIndexPage(true)
   }, [savedPageState, setPreviewPageState])
--- a/apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
@@ -1,11 +1,22 @@
 import type { DropResult } from "@hello-pangea/dnd"
-import { useCallback } from "react"
-import { Box, Button, Flex, Icon, Text, VStack } from "@chakra-ui/react"
+import { useCallback, useState } from "react"
+import {
+  Box,
+  Button,
+  Flex,
+  Icon,
+  Text,
+  useDisclosure,
+  VStack,
+} from "@chakra-ui/react"
 import { DragDropContext, Droppable } from "@hello-pangea/dnd"
-import { useToast } from "@opengovsg/design-system-react"
+import { Infobox, useToast } from "@opengovsg/design-system-react"
 import { ISOMER_USABLE_PAGE_LAYOUTS } from "@opengovsg/isomer-components"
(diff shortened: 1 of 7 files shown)


apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -122,7 +154,56 @@ export default function RootStateDrawer() {
     ],
   )
 
-  const pageLayout: string = savedPageState.layout
+  const handleConversionToIndexPage = useCallback(() => {
```
seaerchin, line 157: sorry, this got swapped downstream because now we are changing to using `blob.content` for the childpages...
 
because of this, you'd need to find the `childpage` block (`type === 'childpage'`) or just use the `currActiveIdx`

seaerchin: can rebase onto `ref/index-pages-in-content` for this change :sadge:

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -122,7 +154,56 @@ export default function RootStateDrawer() {
...
+    const newPageState = {
+      ...savedPageState,
+      content: [],
+      childpages: {
+        layout: "boxes",
+        summary: false,
+        thumbnail: false,
```
seaerchin, line 164: FYI, we shuold use `DEFAULT_CHILDPAGE_BLOCK`

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -122,7 +154,56 @@ export default function RootStateDrawer() {
     ],
   )
 
-  const pageLayout: string = savedPageState.layout
+  const handleConversionToIndexPage = useCallback(() => {
+    const newPageState = {
+      ...savedPageState,
+      content: [],
```
seaerchin, line 160: [issue] won't this essentially discard all their content? i think we shuold preserve this as we'll be causing irreversible data loss otherwise

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -122,7 +154,56 @@ export default function RootStateDrawer() {
...
+    onConfirmConvertIndexPageModalClose,
+    pageId,
+    previewPageState,
+    savePage,
+    setDrawerState,
+    setSavedPageState,
+    siteId,
+  ])
+
+  const pageLayout: string = previewPageState.layout
```
seaerchin, line 206: [nit] can we omit the `string` typing so it infers the correct type

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -134,168 +215,272 @@ export default function RootStateDrawer() {
     savedPageState.content[0]?.type === "hero"
 
   const hasChildpagesBlock = pageLayout === ISOMER_USABLE_PAGE_LAYOUTS.Index
+  const isCustomIndexPage =
+    type === ResourceType.IndexPage && pageLayout !== "index"
```
seaerchin, line 219: nit: this should be `isCustomContentIndexPage` or sometihng similar because we can have index pages which are custom (ie, have content) and have `layout: index`

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -134,168 +215,272 @@ export default function RootStateDrawer() {
...
+            w="full"
+          >
+            <Text textStyle="body-2">
+              You’re previewing what you’ll see once you accept the change.
+            </Text>
+          </Infobox>
+        )}
+
+        {/* Fixed Blocks Section */}
+        <Disable when={isPreviewingIndexPage}>
```
seaerchin, line 281: skimmed, assumed that they are the same

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
dcshzj: Rebased!

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
dcshzj: Updated to use `DEFAULT_BLOCKS.childpage` since `DEFAULT_CHILDPAGE_BLOCK` is for server-side use

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
dcshzj: Hmm yeah this was intended, can see the copy used:

image

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
dcshzj: Updated!

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
dcshzj: Updated!

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
dcshzj: Updated in 64ef2a0fa0fe05d4cdd39f8b2507df40466176ae!

seaerchin (approved):

PR #1285, opened 2025-05-01
isom-1906 oobee must fixes

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1906/oobee-must-fixes

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- make all alt text compulsory
- centralize `ajv` to be imported from `utils/ajv`
- add `ajv-errors` to supplement and allow for custom ajv message
- intercept hard break operations to split block instead
- if URL label is url itself, overwrite it so that aria-label will not throw error

### Compulsory alt text and checks for common poor alt text

https://github.com/user-attachments/assets/bd921c22-8aba-4b87-8f6d-43fb8124daa8

### aria-label improvement for URL 

| state | oobee screenshots |
| - | - |
| before | Screenshot 2025-05-02 at 11 38 38 PM |
| after | Screenshot 2025-05-02 at 7 47 55 PM |

### aria-label

1. Copy the content over from this prose block (https://staging-studio.isomer.gov.sg/sites/9/pages/18143)
2. inspect HTML and check that the link has an aria-label to it
2. publish and run oobee. it should not have `Clickable elements (i.e. elements with mouse-click interaction) must have accessible labels`

### Alt text

1. add a component block with image
2. try various lousy alt text and they should show error message e.g. leave it blank, enter blank space, enter common terms like "image" or "photo"

### Preventing hardbreak

1. go to prose editor
2. try to add hard break via Cmd+Enter or Shift+Enter. You should create a new paragraph instead of hardbreaking to the next line


apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
```diff
@@ -76,7 +76,17 @@ export const BASE_EXTENSIONS: Extensions = [
...
-  HardBreak,
+  HardBreak.extend({
+    // Prevent the default behavior of the hard break and split it into a new paragraph instead
+    // Why: Hard break introduces touch targets of under 24px between lines
+    // This is a WCAG violation if the link is a clickable link
+    addKeyboardShortcuts() {
+      return {
+        "Mod-Enter": () => this.editor.commands.splitBlock(),
+        "Shift-Enter": () => this.editor.commands.splitBlock(),
+      }
```
seaerchin, line 87: [question] i think our parsing rules for `ajv` for `prose` might destructively interfere with this rule (i can't quite remember what but i recall we had a rule for table that it can only have 1 header row or something similar)

could i trouble you to check the below sequence of steps
1. enable header col/row for table
2. use mod or shift enter to split the table into 2 
3. for the new table below, **also enable header col/row**
4. now, delete the hard break -> does this cause validation issues?

packages/components/src/interfaces/complex/Image.ts
```diff
@@ -20,11 +20,21 @@ export const generateImageSrcSchema = ({
 
 export const ImageSrcSchema = generateImageSrcSchema({})
 
+// Note: ajv pattern does not support the use of patternFlag like "i" for case-insensitive
+// Thus, we manually add the case-insensitive flag to the regex pattern
+export const ALT_TEXT_REGEX_PATTERN =
```
seaerchin, line 25: [nit] can i trouble you to add 1 line above this to explain what the regex does?

packages/components/src/templates/next/components/internal/Link/utils/generateAriaLabel.ts
```diff
@@ -0,0 +1,41 @@
...
+  } catch {
+    return false
+  }
+}
+
+export const generateAriaLabel = ({
+  label,
+  textContent,
+  isExternal = false,
+}: GenerateAriaLabelProps): string | undefined => {
```
seaerchin, line 29: [issue] should we pass in `href` rather than `textContent`?

packages/components/src/templates/next/components/internal/Link/utils/generateAriaLabel.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (label) return label
+
+  if (!textContent) return undefined
```
seaerchin, line 32: [nit] no actionable, since this is parity but i was wondering if just defaulting to `This is a link to an unknown url` would make sense or would it cause more confusion

my guess is that to screen readers (and my assumption is that the MOP here is visually handicapped), at least they know it's a link (but this would alr be done by the html tag) but hte value add seems marginal so ok eitherways

packages/components/src/templates/next/components/internal/Link/utils/generateAriaLabel.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  if (isUrl(textContent) && isExternal) {
+    return `Link to ${getDomainFromUrl(textContent)} (opens in new tab)`
+  }
+
+  if (isUrl(textContent)) {
+    return `Link to ${getDomainFromUrl(textContent)}`
+  }
```
seaerchin, line 40: [nit] similar to the above, if we have `""`, this will read `Link to (opens in new tab)`; we shuold either have a check above or just default

packages/components/src/templates/next/components/internal/Link/utils/getTextContent.ts
```diff
@@ -0,0 +1,21 @@
...
+export const getReactNodeText = (node: React.ReactNode): string => {
+  let text = ""
+
+  React.Children.forEach(node, (child) => {
+    if (child === null) return
+    if (typeof child === "boolean") return
+    if (typeof child === "undefined") return
+
+    if (typeof child === "string" || typeof child === "number") {
+      text += child.toString()
```
seaerchin, line 12: [issue] i think we should make `text` a string and join later. this is because if we hvae `string` or `number` child here, this would jsut add the raw string to the `text` **wihtout spaces**.

as an example, if we have 2 children: `[hello, world]` - this returns `"helloworld"`

apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
adriangohjw: re: convo

no longer relevant as users can no longer add new header col/row on prose blocks

packages/components/src/templates/next/components/internal/Link/utils/generateAriaLabel.ts
adriangohjw: ah the textContent is the text that's being hyperlinked - it can be a non-url like [this is an example](https://www.google.com) where the textContent is `this is an example`

packages/components/src/templates/next/components/internal/Link/utils/generateAriaLabel.ts
adriangohjw: text content being falsey is a very edge case e.g. hyperlinking empty spaces (TBH i don't even know if that's possible?). So we can just return undefined and have it not set the aria label

To share more:
- [this hyperlink's text is not a URL but link to google](https://www.google.com) -> does not trigger Oobee must-fix error
- [www.google.com](https://www.google.com) -> this trigger Oobee must-fix error

packages/components/src/templates/next/components/internal/Link/utils/getTextContent.ts
adriangohjw: fxed!

packages/components/src/templates/next/components/internal/Link/utils/generateAriaLabel.ts
seaerchin: sgtm

seaerchin (approved):

PR #1279, opened 2025-04-25
feat(index-pages): render thumbnail for content pages when available

seaerchin:
## Problem
see attached issue

## Solution
- update css for content page
- update stories to have image

PR #1278, opened 2025-04-25
feat(index-page): row layout

seaerchin:
## Problem
See attached ticket

## Solution
- Add a `RowsLayout` component that renders when teh `layout` selected is row 
- Add the `imageObjectSchema` to our `articlePageSchema` -> not tackling updated design of article page yet, will be done in subsequent ticket
- fixed infocard `layout` to `index`; this was guaranteed before (have to look at `IndexPage.layout`), so this is not changing anything

## Videos 
see storybook, attached videos here are also from story book

https://github.com/user-attachments/assets/d79a543e-bf17-42f6-8692-85fc8ae2585f

PR #1287, opened 2025-05-05
refactor: make index page a content block rather than a fixed block

seaerchin:
## Problem
We want to make child pages a content block rather than a fixed block so that, in the future, we can drag the child pages around

## Solution
1. Shift child pages from being an extra property on page to being an `IsomerComponent` and part of `content`. 
2. As part of this change, we also added a conditional on both `Add Block` + the delete block button 

## Videos

https://github.com/user-attachments/assets/6b877df4-4b1d-439f-8856-818cfb8293b0


seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1287?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1287** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1287?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1287?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1280** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1279** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1278** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1276** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1276?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1254** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1254?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1281](https://github.com/opengovsg/isomer/pull/1281) <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1281?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/components/src/templates/next/layouts/IndexPage/IndexPage.tsx
```diff
@@ -54,14 +56,20 @@ const IndexPageLayout = ({
...
+            <ChildrenPages
+              permalink={page.permalink}
+              site={site}
+              LinkComponent={LinkComponent}
+              variant={CHILDPAGE_LAYOUT_OPTIONS.Rows}
+              showSummary={true}
+              showThumbnail={false}
+              type="childpage"
+            />
+          )}
```
dcshzj, line 72: nit: leave it up to you on this one but thought would be cleaner if we updated `content` directly with the `childpages` block if one doesn't exist yet, so we skip this direct call to the `ChildrenPages` block from this layout.

packages/components/src/templates/next/render/render.tsx
```diff
@@ -23,6 +23,7 @@ import {
   Video,
 } from "../components"
 import { LogoCloud } from "../components/complex/LogoCloud"
+import { ChildrenPages } from "../components/internal"
```
dcshzj, line 26: issue(non-blocking): I think should move it out to become a `complex` block. The `internal` components are meant for those that do not have a JSON schema attached to them.

packages/components/src/interfaces/internal/ChildrenPages.ts
```diff
@@ -1,9 +1,9 @@
 import type { Static } from "@sinclair/typebox"
 
-import type { ChildpageMetaSchema } from "~/schemas/childpage"
+import type { ChildpageSchema } from "../complex"
 import type { IsomerSiteProps, LinkComponentType } from "~/types"
 
-export interface ChildrenPagesProps extends Static<typeof ChildpageMetaSchema> {
+export interface ChildrenPagesProps extends Static<typeof ChildpageSchema> {
```
dcshzj, line 6: question: sorry a bit confused, what's the difference between this `ChildrenPagesProps` and the `ChildpageProps` type? Seems like they can be merged.

apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -61,6 +72,51 @@ export function EditorDrawerProvider({
...
+      onError: (error) => {
+        toast({
+          title: "Failed to update blocks",
+          description: error.message,
+          ...BRIEF_TOAST_SETTINGS,
+        })
+      },
+    })
+
+  useEffect(() => {
```
dcshzj, line 92: question: might be lacking a bit of context here so for my understanding, this `useEffect` is needed because we want to ensure that if the user is editing an index page, they would be able to have access to the child pages block?

I assume this is so that we can keep things backward compatible without having to run a migration of existing index pages to add this child pages block?

^ if this is so, then actually why not just do the migration directly so we don't need to keep this part of the code?

apps/studio/src/server/modules/folder/folder.router.ts
```diff
@@ -315,4 +320,75 @@ export const folderRouter = router({
 
       return { title, ...indexPage }
     }),
+
+  insertChildpageBlock: protectedProcedure
```
dcshzj, line 324: question: hmm any reason why this new mutation was created? Seems like it is quite a duplicate with the `updatePageBlob` procedure in page.router, and frontend should be able to be adjusted slightly to use that procedure instead.

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -177,31 +175,16 @@ export default function RootStateDrawer() {
...
+      <VStack w="100%" h="100%" gap="1rem">
+        {/* Custom Blocks Section */}
+        <Flex flexDirection="row" w="100%">
+          <VStack gap="0.25rem" align="start" flex={1}>
+            <Text textStyle="subhead-1">Custom blocks</Text>
+            <Text textStyle="caption-2" color="base.content.medium">
+              Use blocks to display your content in various ways
+            </Text>
+          </VStack>
+          {pageLayout !== ISOMER_USABLE_PAGE_LAYOUTS.Index && (
```
dcshzj, line 187: question: do we need to also hide the "Add a new block" placeholder if the page has no existing blocks?

apps/studio/src/server/modules/folder/folder.router.ts
seaerchin: also can, added because i wanted this to be backend focused (ie, we always add the same block) but i'll swap to `updatePageBlob` instead!

apps/studio/src/contexts/EditorDrawerContext.tsx
seaerchin: 1. we can't do this migration before-hand because it'll result in studio showing an unknown block
2. if we do the migration after, we'd need to comms out to users so that they'd avoid data corruption (ie, i do the migration whilst they're updating, resulting in studio rejecting the update cos the stored state isn't equivalent to the user's state).
3. if we run the migration at night **to avoid impacting users**, there will be a time lag where the children pages aren't generated on end sites - because of this, we'd need the intermediate state between feature release and migration

packages/components/src/templates/next/render/render.tsx
seaerchin: ok sounds good!

packages/components/src/templates/next/layouts/IndexPage/IndexPage.tsx
seaerchin: ok, that's doable tooj1

apps/studio/src/contexts/EditorDrawerContext.tsx
dcshzj: Hmm sorry might not be understanding, I thought that adding the childpage block is backward compatible such that if the block is not there, then we still generate it on the components' side. If the block is there, then no additional generation necessary. In both cases, it should be transparent to the user?

We can introduce the component first so Studio knows about it, then later on perform the migration to add the childpages block to existing index pages.

packages/components/src/interfaces/internal/ChildrenPages.ts
seaerchin: merged - i also standardized to `childrenpages` ba

apps/studio/src/contexts/EditorDrawerContext.tsx
seaerchin: [resolved offline] - we're keeping this `useEffect` temporarily whilst we add the migration, then we'll remove it afterwards

apps/studio/src/features/editing-experience/components/ComplexEditorStateDrawer.tsx
```diff
@@ -300,13 +303,15 @@ export default function ComplexEditorStateDrawer(): JSX.Element {
...
           >
             <HStack spacing="0.75rem">
-              <IconButton
-                icon={<BiTrash fontSize="1.25rem" />}
-                variant="outline"
-                colorScheme="critical"
-                aria-label="Delete block"
-                onClick={onDeleteBlockModalOpen}
-              />
+              {component.type !== DEFAULT_CHILDREN_PAGES_BLOCK.type && (
```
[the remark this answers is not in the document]
seaerchin: can i ask why? i don't have too strong an opinion here, but this `type` is what's used for `renderComponent` and by extension, every derived function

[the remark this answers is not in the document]
dcshzj: Hmm the `renderComponent` is using the `childrenpages` literal directly tho, so I thought is ok if we use `childrenpages` directly here too.

The name of the constant `DEFAULT_CHILDREN_PAGES_BLOCK` feels more like a sample placeholder block rather than one used as part of an enum (not sure the exact phrasing here). If the name is like `CHILDREN_PAGES_BLOCK_TYPE` then would make a bit more sense.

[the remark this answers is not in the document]
seaerchin: hmm can, i'll maek this change

seaerchin:
### Merge activity

* **May 8, 2:53 AM EDT**: @seaerchin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1287).

PR #1288, opened 2025-05-05
Support/add searchpage to create new site

adriangohjw:
## Problem

Currently when we allow user to self migrate on studio, we don't create the Search page record. On-call engineer has to manually go to tableplus to create this

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- create Search page resource and blob as part of the process
- refactor `createSite` by modularizing functions to improve readability

### DIFF
--- a/apps/studio/src/server/modules/site/site.service.ts
+++ b/apps/studio/src/server/modules/site/site.service.ts
@@ -2,12 +2,16 @@ import { TRPCError } from "@trpc/server"
 import { ISOMER_ADMINS, ISOMER_MIGRATORS } from "~prisma/constants"
 import { addUsersToSite } from "~prisma/scripts/addUsersToSite"
 
-import type { DB, Transaction } from "../database"
+import type { DB, Resource, Transaction, Version } from "../database"
 import type {
   CrudResourceActions,
   PermissionsProps,
 } from "../permissions/permissions.type"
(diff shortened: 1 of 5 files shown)


apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -303,14 +311,44 @@ export const createSite = async ({ siteName }: CreateSiteProps) => {
...
+
+    await tx
+      .insertInto("Resource")
+      .values({
+        draftBlobId: String(blobId),
+        permalink: "search",
+        siteId,
+        type: ResourceType.Page,
+        title: "Search",
+      })
```
seaerchin, line 337: do we need to set this to `published` so that it'll automatically be picked up on the first build?

adriangohjw: thought about it and i don't have a preference - i can add it in 👍

dcshzj: If we are publishing the search page, can we also publish the homepage as well? So we can have a guarantee that builds will always pass right from the time it is created

dcshzj:
Can re-request for reviews once the changes have been made, thanks!

seaerchin (approved):

PR #1289, opened 2025-05-06
feat: allow enabling of Singpass for selected users

dcshzj:
## Problem

Closes ISOM-1883.

## Solution

**Breaking Changes**

**Features**:

- Added email attribute to Growthbook to allow for selective enabling of Singpass using the email address.

## Tests

- [ ] Add your email address to Growthbook and force the value to be `false` for the `is-singpass-enabled` feature flag.
- [ ] Attempt to log in to Isomer Studio
- [ ] Verify that you did not need to authenticate with Singpass
- [ ] Use another email that is not in the list inside Growthbook to log in to Isomer Studio
- [ ] Verify that you had to authenticate with Singpass

### DIFF
--- a/apps/studio/src/types/growthbook.ts
+++ b/apps/studio/src/types/growthbook.ts
@@ -1,0 +1,3 @@
+export interface GrowthbookAttributes {
+  email: string
+}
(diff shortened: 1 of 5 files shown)


apps/studio/src/types/growthbook.ts
```diff
@@ -0,0 +1,3 @@
+export interface GrowthbookAttributes {
+  email: string
```
seaerchin, line 2: [question] why use `email` as idnetifier over something likle `uuid`

dcshzj: Hmm we don't know the user's `uuid` until they authenticated with Singpass, and we maintain that we are identifying users on Studio purely using their emails

seaerchin (approved):

PR #1290, opened 2025-05-06
feat: introduce new Blockquote component

dcshzj:
## Problem

Closes ISOM-1851.

## Solution

**Breaking Changes**

**Features**:

- A new Blockquote complex component has been introduced.
    - All types of pages on Studio can use this new component, but the homepage layout has a special design that is different from the rest.

## Before & After Screenshots

**Article**:

image

**Content**:
image

**Homepage**:
image

## Tests

- [ ] Go to Studio
- [ ] Verify that you can create and edit the Quote component inside an article, content and home pages.


packages/components/src/templates/next/components/complex/Blockquote/Blockquote.tsx
```diff
@@ -0,0 +1,97 @@
...
+        image: "h-24 min-h-24 w-24 min-w-24 rounded-full",
+      },
+    },
+  },
+  defaultVariants: {
+    layout: "default",
+  },
+})
+
+const compoundStyles = createBlockquoteStyles()
```
seaerchin, line 49: [nit] why not create teh style inside the compnent then pass the `variant` to the initial creation?

seaerchin (approved):

packages/components/src/templates/next/components/complex/Blockquote/Blockquote.tsx
dcshzj: Omg lol TIL we can do that, updated in bda8f7d190b8e0928d5723dffdec44148100a97b.

lohjunxiang: Senior aura farming.

PR #1291, opened 2025-05-06
isom-1893 collectionblock studio

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1893/collection-widget-on-homepage-studio

## Solution

a follow-up from the discussion from https://github.com/opengovsg/isomer/pull/1277 to update localisedsitemap
- replicating how we generate summary during build time for getLocalisedSitemap
- aritifically overwrite collection's children if its homepage

Added the brief discussion notes into ADR [here](https://www.notion.so/opengov/6aa98291511345c68e4b04b3e47dacd0?v=aaf927bda82344a9a938a0f151040f0b&p=1eb77dbba78880f48e70dc24a974d374&pm=s)

**Breaking Changes**

  - Details ...


seaerchin:
can rebase or merge? then i'll review after

adriangohjw:
merging test cases from https://github.com/opengovsg/isomer/pull/1363 to ensure no breaking changes

apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts
```diff
@@ -856,6 +890,403 @@ describe("resource.service", () => {
...
+      // Assert: Find Collection in the sitemap
+      const collectionNode = result.children
+        ?.at(0)
+        ?.children?.find((child) => child.id === collection.id)
+      expect(collectionNode?.title).toBe(collection.title)
+      expect(collectionNode?.summary).toBe(`Pages in ${collection.title}`)
+      expect(collectionNode?.image?.src).toBeUndefined()
+    })
+
+    it("should include any nested collections if resourceId is a RootPage", async () => {
```
adriangohjw, line 1206: all tests other than this was from https://github.com/opengovsg/isomer/pull/1363

apps/studio/tests/integration/helpers/seed/index.ts
adriangohjw, line 1: changes from https://github.com/opengovsg/isomer/pull/1363

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -362,25 +365,40 @@ export const getLocalisedSitemap = async (
...
-        .where("state", "=", "Published")
-        .leftJoin("Version", "Version.id", "publishedVersionId")
+        .where("Resource.siteId", "=", siteId)
+        .where("Resource.type", "in", [
+          ResourceType.Folder,
+          ResourceType.Collection,
+          ResourceType.IndexPage,
+        ])
+        .where("Resource.state", "=", ResourceState.Published)
+        .leftJoin("Version", "Version.id", "Resource.publishedVersionId")
```
seaerchin, line 379: [perf] can we try this on staging (or on local?) - i had a similar query and the performance wasn't great so i'm worried that this iwll negatively impact load times

apps/studio/src/utils/sitemap.ts
```diff
@@ -110,3 +115,46 @@ export const getSitemapTree = (
...
+      ...sitemap,
+      children: Array.from({ length: NUMBER_OF_CARDS_IN_COLLECTION_BLOCK }).map(
+        (_, idx) => ({
+          id: `collection-card-${idx}`,
+          title: "Article title",
+          summary: "Article summary",
+          permalink: "/",
+          layout: ISOMER_USABLE_PAGE_LAYOUTS.Article,
+          lastModified: new Date().toISOString(),
+          category: "Category of article",
```
seaerchin, line 135: [question] don't have to do now, but wondering if we should use the children's title/summary if any

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: [stupid question] won't this result in more data than required for nested resources? eg, i'm previewing a page C at `A -> B -> C`, this will still get the list of all resources that are collections/folders/index pages?

adriangohjw: yes it will, and thats intentional (unfortunately)

this is because collection can be pretty nested down (e.g. root -> 1 or more folders -> collection). Without fetching all the way down, it will crash since studio cannot find the collection in the sitemap

adriangohjw: @seaerchin tried on staging (locally)
- for the root page of the largest site, its between 180ms to 480ms (median 300ms)
- for smaller sites (<100ms)

this is including network call (whats observed from the dev. console)

seaerchin: i'm ok with approving this - 480ms seems reasonable as an upper bound, but we need to be clear on:
1. what is the expected upper bound of resources we expect to get (ie, for direct children, what's the highest amount)
2. what's our deepest level for resources (ie, how nested do we allow our resources to be?)

i'm ok with this because it shouldn't impact load times negatively **at this point**, but i'm not sure if we need to do more work wrt this query (out of scope for this)

cc @harishv7 or @dcshzj if yall have objections

seaerchin (approved):
lgtm - but please hold off on merging until someone sanity checks me hahahahaha

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: @harishv7 

using site 1 for staging, total resources: 5476 with max depth of 7
```sql
WITH RECURSIVE tree AS (
  SELECT
    id,
    "parentId",
    1 AS depth
  FROM "Resource"
  WHERE "parentId" IS NULL and "siteId" = 1 and "type" in ('Folder', 'Collection', 'IndexPage')

  UNION ALL

  SELECT
    i.id,
    i."parentId",
    t.depth + 1
  FROM "Resource" i
  JOIN tree t ON i."parentId" = t.id
  where "siteId" = 1
)
SELECT MAX(depth) AS max_depth FROM tree;
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -362,25 +365,40 @@ export const getLocalisedSitemap = async (
         .leftJoin("Blob as published", "Version.blobId", "published.id")
         .select(() => [headerSql, thumbnailSql, ...defaultResourceSelect]),
     )
-    .with("childCousinIndexPages", (eb) => {
-      return eb
+    // Step 3: Get all nested folders and collections
+    .withRecursive("nestedResources", (eb) =>
```
harishv7, line 369: Actually can explain a bit more on why do we need this? My understanding was that we were injecting dummy data into the localisedSitemap we have fetched to render the collection block with 3 placeholder items

adriangohjw: @harishv7 here https://github.com/opengovsg/isomer/pull/1291#discussion_r2113582834 !

You are right that the children is using the dummy blocks, but the title and description aren't - they are using the values from the Collection (or Index Page, if any)

adriangohjw: that's a possible different approach, though personally I thought that was a bit less straightforward and different from how things are currently done (via the same single method), as opposed to this approach of just only tweaking the query (which is easier to test for regression via unit test too). Esp. since in most cases, the number of nestings wouldn't be as deep as this test case.

harishv7: Discussed offline, understand we want to show the users some preview of "actual" content when they use a collection block as have fully dummy data will not be ideal. Given the below benchmarks I think it's reasonable for the use case as well. 

Will go ahead to approve the PR

harishv7 (approved):

PR #1293, opened 2025-05-06
isom-1904 support infobar styling dark (selective display on Studio based on layout)

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1904/support-infobar-styling-dark

We currently have no way to selectively show a component on one layout but not another layout, which was required in order to only show "variant" on homepage but not content pages

## Solution

- add a metadata to the field `layouts` in Infobar
- pass in `layout` into FormBuilder
- Creating a higher order component (HOC) `withLayoutFilter` that intercepts and check for `layout` in `layouts`

**Breaking Changes**

  - Details ...

## Before & After Screenshots

the "variant" field is shown in infobar only for Homepage and not content pages

https://github.com/user-attachments/assets/d1ed7415-1d73-4158-866a-a1061dc4e2fe


seaerchin (commented):
this change feels abit off to me because studio is now aware of a component level rendering logic. 

i'm not entirely sure if that's the way forward because it breaks some assumptions which we've put in place - wrt that, did we ask design why content pages can't have dark infobar...

adriangohjw:
@seaerchin yeah i thought about it too and my thought process:
- studio should be able to know what is component doing
- relatively harmless since the fallback is to show, if no checks are done
- we kinda aldy doing it with format to display "radio" or "hidden"

we can also extend this similar approach to support component fields that are meant to be experimental and selectively rolled out

dcshzj:
Hmm I looked through and thought it's a bit weird also that we are selectively showing fields depending on the layout. If we go back to the principles of the schema of components, it is meant to be common across all the layouts and not have fields that are only specific to certain layouts (which is why all layouts go through the same entry point of `renderPageContent`).

In this case if we want to have specific schema-level differences in components, it should be a new component (but under the hood the rendering can use the same base component).

adriangohjw:
@dcshzj my thoguth process is that in n this case there's still a default value for those fields, just not exposed to users similar to how we implement `hidden`. Felt a bit overengineered to split them into 2 components + it makes migrating them from one to another more troublesome (e.g. opening this fields up to other layouts)

dcshzj:
Hmm for the format `hidden`, ideally we should stop using that alr cos it only exacerbates the studio-template disparity issue, and keep it reserved for the `type` key. It was not intended to hide fields that will affect the component props from the user.

I think if we proceed with this change, it breaks our assumption that the schema for components is consistent for all layouts. Right now, components and layouts are on the same level, so the schema used for a component does not change when the layout changes, but this change means that when we change the layout, we will also need to change the shape of the component schema. What I mean about the changing of the shape is that in one layout, a prop that is expected to change some presentation of the component does not do anything.

Are we able to push back on design for this? Either we do it properly whereby this styling is also applicable to all layouts, or have a new homepage-only component to fit this use case.

adriangohjw:
@dcshzj  In Figma, we actually do have the design for it when it's used in both layouts (last row). If im not wrong, we only releasing it on Homepage for now as a staged rollout at this moment since this is rather experimental

image

seaerchin:
my 2c here - the components should be completely agnostic of surrounding context - all it takes is what's required to render. why we want to hold this line is that components are _just components_ - this allows us to make our components/studio decoupled from each other. because of this, our current structure is:

components (render only) -> schema (**actual invariants and options**) -> studio (render schema, even higher up optiosn (on a site level, for example)

what we might want to do here is to declare two separate schemas **for the same component**. one schema has this option declared as `format: hidden`, the other has it as a `Radio`. this shifts the burden of deciding this logic to schemas (and by extension, studio as that decides what schema to render).

let me know if this makes sense @adriangohjw and cc @dcshzj for opinions

adriangohjw:
@seaerchin  agree with the component <> template <> studio thoughts.

> what we might want to do here is to declare two separate schemas for the same component. one schema has this option declared as format: hidden, the other has it as a Radio. this shifts the burden of deciding this logic to schemas (and by extension, studio as that decides what schema to render).

though curious how is this ideologically different from defining layouts to render the field in, like what was done too? And also, using `format` for `radio` and `hidden` is also the template telling studio what to do. Not sure if its a naming thing e.g. calling this "onlyShowOn"?  

Plus, in this approach, we are also using the same component

dcshzj:
Hmm it's different right? The `format` of `radio` is giving a hint of how the component form field should be rendered, but this case is different because we are effectively changing the shape of a schema depending on the layout (we are just hiding on the frontend but this has the same idea). Components are just components and while there are rendering differences depending on the layout, that does not extend to the actual underlying schema (aka the props of the component).

When it is a different component on the schema level, components no longer need to make that distinction on the props that it chooses to take, rather that is left to the consumer of the schema to determine which component to use inside the layout. In this case, the consumer is Studio, and Studio currently has the ability to pick and choose the components that are available to each layout. Components give the full set of layouts and components for consumers to use, Studio is the one that picks and choose which are available to the user.

adriangohjw:
refer to https://github.com/opengovsg/isomer/pull/1304 instead

PR #1295, opened 2025-05-07
Chore - infocards validation (min. cards 1 + non-empty title)

adriangohjw:
## Problem

RE: https://opengovproducts.slack.com/archives/C07CWUNUL68/p1746610816727699?thread_ts=1746610673.649569&cid=C07CWUNUL68

## Solution

**Breaking Changes**

  - agencies who have empty-title or 0-card infocards will not be able to edit

**Improvements**:

- add regex pattern for title - NOTE: the error message will be "ugly" until https://github.com/opengovsg/isomer/pull/1285
- add min. cards


seaerchin (approved):
no issues with this, but have we done a quick scan over prod to check if there are many instances of such issues?

adriangohjw:
> no issues with this, but have we done a quick scan over prod to check if there are many instances of such issues?

@seaerchin my thoughts is that we can just them be invalid and be a screen test -> agency will either fix it, or complain to us and we handle on case-by-case basis

WDYT?

seaerchin:
> > no issues with this, but have we done a quick scan over prod to check if there are many instances of such issues?
> 
> @seaerchin my thoughts is that we can just them be invalid and be a screen test -> agency will either fix it, or complain to us and we handle on case-by-case basis
> 
> WDYT?

can also, but you need to be aware of long tail of components that won't adhere. additionally, the error message might not be super informative until we click in, which for agencies will be confusing

adriangohjw:
@seaerchin yeah i think its worth for us to have a ticket to make `NON_EMPTY_STRING_REGEX` compulsory for all compulsory string (or a wrapper for it) 😢

PR #1297, opened 2025-05-07
display category before each tag on article header

sehyunidaaa:
## Problem

The article page shows tags for an article, but not the article category. So we have cases where unclear tags like 'No' are on their own.

Closes [ISOM-1922]

## Solution

Display each category before a tag.

## Before

image

## After

image

### DIFF
--- a/apps/studio/docker-compose.yml
+++ b/apps/studio/docker-compose.yml
@@ -13,8 +13,7 @@ services:
       - postgres-volume:/var/lib/postgresql/data
 
   mockpass:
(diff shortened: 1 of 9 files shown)


adriangohjw (approved):

adriangohjw:
@sehyunidaaa not quite sure if the gap-4 was intentional but it doesn't do anything so i updated it - LMK if i understrood it wrongly!

sehyunidaaa:
eh you're right, it wasn't doing much. thank you @adriangohjw!!

PR #1298, opened 2025-05-07
isom-1898 image gallery slideshow

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1898/image-gallery-slideshow

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- new image galley component
- ~skipped the image preload optimization part - i feel like its not an issue at this point so decided to not overengineer~ **update**: decided to still add it otherwise navigation can feel sluggish, and not somethign im proud of. And since we rarely update components after shipping it, I rather do it well now (https://github.com/opengovsg/isomer/pull/1298#issuecomment-2870035938)

FYI @sehyunidaaa - lemme know otherwise
- set number of images between 2 and 30.
- there's only design for 3 or 5 preview images in Figma. I have standardized to use the height for 3 images if there's 1-3 preview images, and the height for 5 images if there's 4-5 preview images.
   - Note: 1 preview image only happens in case someone saved an invalid schema directly via DB or from github->studio migration. it should NOT be possible on studio, but adding it to storybook nevertheless
- i have skip the image preview in studio for now image
- side-note: im feeling half-half about transition animation after implementing it - it feels a bit more sluggish, but maybe we just have to lower the transition? just pointing out - leave it up to you to make the final call 

### Without transition

https://github.com/user-attachments/assets/b06cacf0-c449-464a-acf2-01fa9cf80d2f

### With transition

https://github.com/user-attachments/assets/1e079229-a4e9-437d-b4e0-406f3c8478be

## Tests

Go to Studio to create a new Image Gallery block in a Content page, and add more than 5 images
- [ ] navigate with the left/right button -> it should navigate correctly
- [ ] go to the first image and click left button -> you should be at the last image
- [ ] go to the last image and click right button -> you should be at the first image
- [ ] click any image directly via preview below -> it should navigate correctly
- [ ] at any point of time, the image shown should be in the center of the preview, unless it's the first 2 images or last 2 images


packages/components/src/interfaces/complex/ImageGallery.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { IsomerSiteProps } from "~/types"
+import { AltTextSchema, ImageSrcSchema } from "./Image"
+
+export const IMAGE_GALLERY_TYPE = "imagegallery"
+
+const SingleImageSchema = Type.Object({
+  src: ImageSrcSchema,
+  alt: AltTextSchema,
+  caption: Type.Optional(Type.String({ maxLength: 100 })),
+})
```
seaerchin, line 13: [comment, no actionable] i wonder if we should consolidate this together iwth our existing image schema to avoid an explosion of similar yet subtly different schemas

adriangohjw: thought about it - felt a bit hard since it's aldy not standardized at this moment, from schema shape to name of fields e.g. `src`, `imageSrc`

sehyunidaaa:
@adriangohjw thanks for the quick turnaround

> there's only design for 3 or 5 preview images in Figma. I have standardized to use the height for 3 images if there's 1-3 preview images, and the height for 5 images if there's 4-5 preview images.

Eh sorry this was unclear on Figma. The ones with 5 and 3 previews are for desktop and tablet breakpoints respectively. If there are less than 5 or 3 images, the place where preview boxes would have been can be left blank. Can see some examples [on this Frame](https://www.figma.com/design/rsKQdmtyoOWawyAv1LkJJK/Isomer-Next-Component-Library?node-id=7347-12621&t=hdgv22OqNQ4bPwlc-1).

> i have skip the image preview in studio for now

Okie! Can you please add to the Linear ticket as a comment? 

> im feeling half-half about transition animation after implementing it - it feels a bit more sluggish, but maybe we just have to lower the transition? 

Eh yeah does feel a bit slow yeah. But I do like having some transition cos otherwise it feels a bit abrupt. Can we try a lower value like 150ms?

adriangohjw:
## Updated to have better transition by using preloading

refer to the network tab to see how images are preloaded

| scenario | demo |
| - | - |
| entering and clicking next button | https://github.com/user-attachments/assets/a272b554-e858-4204-8cc5-81ab3875cc37 |
| entering "previous" button | https://github.com/user-attachments/assets/c185131c-c59d-4f20-8a9e-e82b127ff1bb |
| entering image in preview sequence | https://github.com/user-attachments/assets/f4dbb6dd-3c33-40d5-a570-491b460037d4 |

packages/components/src/templates/next/components/complex/ImageGallery/ImageGalleryClient.tsx
```diff
@@ -0,0 +1,295 @@
...
+      // Image has already been loaded
+      if (image.complete) return
+
+      // Image not loaded yet, force load it!
+      if (image.getAttribute("loading") === "lazy") {
+        image.removeAttribute("loading")
+        const src = image.src
+        image.src = ""
+        image.src = src
+      }
```
seaerchin, line 89: [comment] not sure if it's possible to do, but the lazy load appears to be a function of the current index and the `maxPreviewImages` - should be just pass in the logic here to the `lazyLoad` prop on `ImageClient`? feels abit clearer and saves us the trouble of dom manipulation

packages/components/src/templates/next/components/complex/ImageGallery/utils/getPreviewIndices.ts
```diff
@@ -0,0 +1,30 @@
...
+  numberOfImages,
+  currentIndex,
+  maxPreviewImages,
+}: GetPreviewIndicesProps) => {
+  // If there are less than or equal to max images, show all of them
+  if (numberOfImages <= maxPreviewImages) {
+    return Array.from({ length: numberOfImages }, (_, i) => i)
+  }
+
+  const numberOfImagesBeforeCurrent = Math.floor(maxPreviewImages / 2)
```
seaerchin, line 17: not sure i understand this line here. if i have an image array [0, 1, 2, 3, 4], and i'm at position 4, the number of images before current should be 4 isn't it? similarly if i'm at position 1, it's should be 0.

if the direction is descending, then we can just flip the logic so i'm not entirely sure i understand why it's `maxPreviewImages / 2 `

seaerchin, line 17: not sure i understand this line here. if i have an image array [0, 1, 2, 3, 4], and i'm at position 4, the number of images before current should be 4 isn't it? similarly if i'm at position 1, it's should be 0.

if the direction is descending, then we can just flip the logic so i'm not entirely sure i understand why it's `maxPreviewImages / 2`

seaerchin, line 17: not sure i understand this line here. if i have an image array [0, 1, 2, 3, 4], and i'm at position 4, the number of images before current should be 4 isn't it? similarly if i'm at position 1, it's should be 0.

not sure i understand why it's `maxPreviewImages / 2` rather than just the `pos`

packages/components/src/templates/next/components/complex/ImageGallery/ImageGalleryClient.tsx
```diff
@@ -0,0 +1,295 @@
...
+      if (isTransitioning) return
+      setIsTransitioning(true)
+      setCurrentIndex((current) =>
+        direction === "next"
+          ? (current + 1) % images.length
+          : (current - 1 + images.length) % images.length,
+      )
+      setTimeout(() => setIsTransitioning(false), TRANSITION_DURATION)
+
+      if (direction === "next") {
```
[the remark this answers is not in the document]
adriangohjw: no need because as we go to the next image, the prev image (which is the current image) has already been loaded

packages/components/src/interfaces/complex/ImageGallery.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  {
+    title: "Image Gallery component",
+  },
+)
+
+export type ImageGalleryProps = Static<typeof ImageGallerySchema> & {
+  site: IsomerSiteProps
+  shouldLazyLoad?: boolean
+}
```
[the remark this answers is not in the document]
adriangohjw: in the implementation, we will eager load all images (main image + preview) or none. This is on the basis that if the component should be eager loaded (which means they are the first component on the page), then both the main image and preview images will always be visible and have to be eager loaded

in the subsequent implementation, we do have a check to always lazy load images that aren't visible (those in preview but not visible), even if we pass in `shouldLazyLoad: false`

packages/components/src/templates/next/components/complex/ImageGallery/ImageGalleryClient.tsx
adriangohjw: not sure if i understand correctly. `maxPreviewImages` is determined by the breakpoints and not a config, and we will only know when the component is rendering

packages/components/src/templates/next/components/complex/ImageGallery/ImageGalleryClient.tsx
```diff
@@ -0,0 +1,295 @@
...
+    },
+    [images.length, isTransitioning, preloadNextImage],
+  )
+
+  const navigateToImageByIndex = (index: number) => {
+    if (isTransitioning) return
+    setIsTransitioning(true)
+    setCurrentIndex(index)
+    setTimeout(() => setIsTransitioning(false), TRANSITION_DURATION)
+  }
```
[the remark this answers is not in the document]
adriangohjw: good point, updated!

packages/components/src/templates/next/components/complex/ImageGallery/constants.tsx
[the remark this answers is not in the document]
adriangohjw: updated!

packages/components/src/templates/next/components/complex/ImageGallery/ImageGalleryClient.tsx
```diff
@@ -0,0 +1,295 @@
...
+      const lastIndexInPreview = previewIndices[previewIndices.length - 1]
+      if (!lastIndexInPreview) return // should never happen since there's always at least one image in the preview sequence
+      preloadImage(lastIndexInPreview + numberOfImagesAheadOfCurrentPreview)
+    },
+    [preloadImage, previewIndices],
+  )
+
+  const navigateToImageByDirection = useCallback(
+    (direction: "prev" | "next") => {
+      if (isTransitioning) return
```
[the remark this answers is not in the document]
adriangohjw: practically, it shouldn't affect since the transition is pretty fast as there's no data fetching

but technically, will remove this since we also set the buttons to be disabled when it is transitioning -> this should be cleaner

packages/components/src/templates/next/components/complex/ImageGallery/utils/getPreviewIndices.ts
adriangohjw: ah this do some computes 

### What it does:
Creates a "sliding window" of image previews (e.g., 3 or 5 images).

### How it works:
- If we have fewer images than the window size, it shows all of them.
- Otherwise (we have more images than the window size), it tries to center the currentIndex within this window
- It smartly adjusts the window to ensure it doesn't go off the edges (e.g., showing negative indices or indices beyond the last image). So, if currentIndex is near the start or end, the window shifts to stay within bounds.

### To your question:
You've got a good point - the name `numberOfImagesBeforeCurrent` is a bit confusing.
- This line `const numberOfImagesBeforeCurrent = Math.floor(maxPreviewImages / 2)` isn't counting the actual images before currentIndex in the whole array. Instead, it's figuring out how many images should appear before the currentIndex within the small, visible preview window to keep the currentIndex roughly centered.
- If maxPreviewImages is 5, it tries to put 2 images before the current one in the preview.
- If maxPreviewImages is 3, it tries to put 1 image before the current one in the preview.

The rest of the function then uses this to position that "preview window" correctly within the full list of images, making sure it doesn't go out of bounds.

packages/components/src/templates/next/components/complex/ImageGallery/utils/getPreviewIndices.ts
seaerchin, line 1: [comment] i can't help but feel like this is abit confusing and probably clearer expressed with modulo arithmetic, but given its narrow scope, i'm fine w/ just keeping this as is

seaerchin (commented):
To make sure i'm understanding the behaviour of the pr correctly:

we have 2 forms of images
1. the big one shown when the user clicks in 
2. the preview we see down below

the big one is eagerly preloaded when user is +/- 1 or 2 positions away (depending on the window size) 

all the images in the preview we see down below is loaded when the user hovers over the preview (and we load in window size)?

adriangohjw:
@seaerchin 

 > we have 2 forms of images
> 
> 1. the big one shown when the user clicks in
> 2. the preview we see down below

correct

> the big one is eagerly preloaded when user is +/- 1 or 2 positions away (depending on the window size)
> all the images in the preview we see down below is loaded when the user hovers over the preview (and we load in window size)?

correct + also as they onMouseEnter the "next"/"previous" button

seaerchin (approved):

PR #1300, opened 2025-05-08
chore: add check for immediate children

seaerchin:
## Problem
[see slack thread](https://opengovproducts.slack.com/archives/C06R4DX966P/p1746684298741139)
  

## Solution
1. first, we pre-generate the list of full URLs 
2. we then generate the list of resources **that are actually going to be in our sitemap** - the source of truth for this shuold be based off publishing. 
3. based off `publishing/index`, we will only create index pages for **folders that have some children** (regardless of nesting depth).

## Caveats
1. because we only take `immediateChildren` + `childCousinIndexPages`, this isn't quite enough to get the full list of pages under a folder. this means that for folders with `draft` index pagse (or no index pages) but deeply nested children, there will be a difference in preview

### DIFF
--- a/apps/studio/src/utils/sitemap.ts
+++ b/apps/studio/src/utils/sitemap.ts
@@ -27,7 +27,8 @@ const getResourcesWithFullPermalink = (
   const children = resources
     .filter((resources) => {
       return parent.type === ResourceType.RootPage
-        ? resources.parentId === null
+        ? resources.parentId === null &&
+            resources.type !== ResourceType.RootPage
         : resources.parentId === parent.id
     })
     .map((child) => {
@@ -68,7 +69,10 @@ const getSitemapTreeFromArray = (
       )
     })
     .filter((resource) => {
-      return resource.parentId === parentId
+      return (
+        resource.parentId === parentId &&
+        resource.type !== ResourceType.IndexPage
+      )
     })
 
   // TODO: Sort the children by the page ordering if the FolderMeta resource exists
--- a/apps/studio/src/utils/sitemap.ts
+++ b/apps/studio/src/utils/sitemap.ts
@@ -2,8 +2,14 @@ import type { IsomerSitemap } from "@opengovsg/isomer-components"
 import { ResourceType } from "~prisma/generated/generatedEnums"
 
 import type { Resource } from "~prisma/generated/selectableTypes"
-import { INDEX_PAGE_PERMALINK } from "~/constants/sitemap"
 
+export const PAGE_RESOURCE_TYPES = [
+  ResourceType.Page,
+  ResourceType.CollectionPage,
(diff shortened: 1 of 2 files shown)


apps/studio/src/utils/sitemap.ts
```diff
@@ -14,35 +20,69 @@ type ResourceDto = Omit<
...
   if (resources.length === 0) {
     return undefined
   }
 
   // Get the immediate children of the resource with the given parent ID
-  const children = resources.filter((resource) => {
-    if (parentId === null) {
+  const children = resources
+    .filter((resource) => {
+      const hasPageChildren = resources.some((possibleChild) => {
```
dcshzj, line 59: minor nit: should be `hasPageDescendants`, children here seem to imply 1 level down only

apps/studio/src/utils/sitemap.ts
```diff
@@ -14,35 +20,69 @@ type ResourceDto = Omit<
...
-    return (
-      resource.parentId === parentId &&
-      resource.permalink !== INDEX_PAGE_PERMALINK
-    )
-  })
+    })
+    .filter((resource) => {
+      if (parentId === null) {
+        return (
+          resource.parentId === null &&
```
dcshzj, line 72: nit: outside of where I can comment on this diff but I think we can remove the exclusion of `FolderMeta` and `CollectionMeta` here since it would have been excluded earlier under `PAGE_RESOURCE_TYPES`.

apps/studio/src/utils/sitemap.ts
```diff
@@ -65,30 +105,36 @@ const getSitemapTreeFromArray = (
...
 
     return {
       id: String(resource.id),
       layout: "content", // Note: We are not using the layout field in our previews
-      title: titleOfPage || resource.title,
-      summary: summaryOfPage ?? `Pages in ${resource.title}`,
+      title: resource.title,
+      summary: `Pages in ${resource.title}`,
       lastModified: new Date() // TODO: Update this to the updated_at field in DB
         .toISOString(),
```
dcshzj, line 134: nit: can update this also since the information should be in `resource` now anyway

dcshzj (approved):

apps/studio/src/utils/sitemap.ts
seaerchin: sounds good, lemme pudate

apps/studio/src/utils/sitemap.ts
seaerchin: uypdated the checks here to `resource.parentId === null`

apps/studio/src/utils/sitemap.ts
seaerchin: updated

PR #1305, opened 2025-05-12
isom-1849 support more hero banner styling options

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1849/support-more-hero-banner-styling-options

## Solution

**Features**:

- added new style option for hero
- added responsive "scroll for more" UI

**Improvements**:

- minor refactoring

**Note**
- did not give user the option to select the theme for now on studio
- temporarily hiding the 2 new options from studio display via `format: hidden` config in schema

## Screenshots

Sticky pulsing "scroll for more" button

https://github.com/user-attachments/assets/17acbbfa-ddb2-4e33-bfd6-e859fbabc209

(pls refer to storybook for the rest)

## Testing (Chromatic)

As the component relies on `window.innerHeight` to determine layout/position, we are hitting a common gotcha in Chromatic or headless browser environments.

Here's what’s likely happening:
1. Chromatic sets a fixed viewport size when running stories (default is 900x600 unless you configure it).
2. When your component accesses window.innerHeight, it gets that value — e.g., 600px.
3. If your story's content overflows vertically, and you didn't constrain it with styles, Chromatic may render a taller page and scroll to capture it, resulting in:
   - The component reacting to a smaller window.innerHeight,
   - But Chromatic capturing a larger screenshot (possibly stitched) of the scrollable area.
   - So while it tries to take a single screenshot, it can behave like stitching if the rendered height exceeds the browser window and scrolling occurs.


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsAnyOfControl.tsx
```diff
@@ -42,14 +42,22 @@ export function JsonFormsAnyOfControl({
...
+      const option = String(
+        anyOfRenderInfo.label || anyOfRenderInfo.schema.const,
+      )
+
+      return {
+        label: option.charAt(0).toUpperCase() + option.slice(1),
+        value: option,
+      }
+    })
+    .filter((option) => option !== null)
```
seaerchin, line 60: [question] abit obtuse of me, but what necessitates this change? wouldn't the correct fix be on the priority so that `hidden` gets rendered above `anyOf` (or is the issue that a child schema has `format: hidden`)

packages/components/src/interfaces/complex/Hero.ts
```diff
@@ -69,12 +77,22 @@ export const HeroSchema = Type.Composite(
...
+          Type.Literal(HERO_STYLE.gradient, { title: "Gradient (Default)" }),
+          Type.Literal(HERO_STYLE.block, { title: "Block" }),
+          Type.Literal(HERO_STYLE.largeImage, {
+            title: "Large image",
+            format: "hidden",
+          }),
+          Type.Literal(HERO_STYLE.floating, {
+            title: "Floating",
+            format: "hidden",
+          }),
```
seaerchin, line 89: [clarification] i'm assuming the `largeImage` and `floating` is what necessitates the additional code in `JsonFormsAnyOf`? 

if so, wondering why we don't just remove the extra variants from the schema _but_ allow it in our types. next time if we want then can just add back

packages/components/src/presets/next/index.ts
```diff
@@ -20,6 +20,20 @@ const config: Config = {
       screens: {
         xl: "1240px",
       },
+      keyframes: {
+        buttonPulse: {
+          "0%, 100%": { backgroundColor: "rgba(0, 0, 0, 0.65)" },
+          "50%": { backgroundColor: "rgba(0, 0, 0, 0.325)" },
```
seaerchin, line 26: pog

seaerchin: [low impact, no actionable]  wondering if we care about their [motion preference](https://tailwindcss.com/docs/transition-property#supporting-reduced-motion) but given that our codebase as a whole doesn't really have this, i'm ok with skipping also (majority of the time this has no impact)

packages/components/src/templates/next/components/complex/Hero/Hero.stories.tsx
seaerchin, line 1: skipped

packages/components/src/templates/next/components/complex/Hero/HeroBlock.tsx
seaerchin, line 1: skimmed, assuming we're just copy pasting from teh old `HeroBlock` in `Hero.tsx`

packages/components/src/templates/next/components/complex/Hero/HeroGradient.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <div className="w-full content-center bg-gradient-to-r from-[rgba(0,0,0,85%)] to-[rgba(0,0,0,10%)] xl:from-[rgba(0,0,0,100%)]">
+        <div
+          className={`${ComponentContent} flex flex-row justify-start py-16 text-start text-base-content-inverse`}
+        >
+          <div className="xl:max-w-50% flex w-full flex-col gap-9 sm:w-3/5">
+            <div className="flex flex-col gap-6">
+              <h1 className="prose-display-xl break-words">{title}</h1>
+              {subtitle && <p className="prose-title-lg-regular">{subtitle}</p>}
+            </div>
+            {buttonLabel && buttonUrl && (
```
seaerchin, line 38: [clarification] not sure if i'm seeing things wrong but this conditional looks like it applies across both the main and secondary button? ie, we'd only show the secondary button fi the main button si shown. this makes sense to me but we need to make sure that wording is clear on studio so that users aren't confused by this behaviour

packages/components/src/templates/next/components/complex/Hero/HeroLargeImage/HeroLargeImage.tsx
```diff
@@ -0,0 +1,72 @@
...
+                LinkComponent={LinkComponent}
+                isWithFocusVisibleHighlight
+              >
+                {secondaryButtonLabel}
+              </LinkButton>
+            )}
+          </div>
+        )}
+      </div>
+      <ImageContainer imageSrc={backgroundSrc} imageAlt={title} />
```
seaerchin, line 69: [sensecheck] this isnt being eagerly loaded because it's below the fold right?

packages/components/src/templates/next/components/complex/Hero/HeroLargeImage/ImageContainer.tsx
```diff
@@ -0,0 +1,66 @@
...
+    if (!imageRef.current) return
+
+    const imageRect = imageRef.current.getBoundingClientRect()
+
+    if (imageRect.top + SCROLL_THRESHOLD >= window.innerHeight) {
+      setShouldShowButton(false)
+      return
+    } else {
+      setShouldShowButton(true)
+    }
```
seaerchin, line 32: [minor nit, feel free to ignore] can just directly `setShouldShowButton(condition)`

packages/components/src/templates/next/components/complex/Hero/HeroLargeImage/ScrollForMoreButton.tsx
```diff
@@ -0,0 +1,31 @@
+interface ScrollForMoreButtonProps {
+  isFixed: boolean
+}
+
+export const ScrollForMoreButton = ({ isFixed }: ScrollForMoreButtonProps) => {
+  return (
+    <div
+      className={`${isFixed ? "fixed" : "absolute"} bottom-16 left-1/2 -translate-x-1/2 animate-slide-up-fade-in rounded-full`}
```
seaerchin, line 8: [sense-check] when i had a conditional like this in my code, it didn't _quite_ work wrt styling. can we double check that it's workign here (also, if you know why it didn't work last time, it'd be much appreciated! i swapped over to using `tv` to get it to work)

seaerchin (commented):
largely ok, wondering if we should extract out shared styling into a `tv` style then we can set `variants` there, but i'm ok with the current code now

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsAnyOfControl.tsx
adriangohjw: childschem has hidden! in this case the option itself `largeImage` and `floating` should not be shown - this is a workabout without having to use growthbook to hide it from other agencies

packages/components/src/templates/next/components/complex/Hero/HeroGradient.tsx
adriangohjw: ah yes, i don't quite know the context/rationale behind this logic, but its been there for for Block and Gradient variance already, so I copied it

packages/components/src/templates/next/components/complex/Hero/HeroLargeImage/HeroLargeImage.tsx
adriangohjw: it is! it renders <ImageClient> under the hood with a hardcoded `lazyLoading={false}` - safe assumption here is that hero will always be the first block

packages/components/src/interfaces/complex/Hero.ts
adriangohjw: correct me if im wrong: if we do that, then we cannot enable it for agencies as it will be considered invalid schema, and they cannot do any changes for homepage as a whole?

packages/components/src/templates/next/components/complex/Hero/HeroLargeImage/ScrollForMoreButton.tsx
adriangohjw: was that a build-specific issue? because checked that this is working on storybook. nevertheless updated it to use tv - i think its cleaner too

packages/components/src/presets/next/index.ts
adriangohjw: nice catch, updated, thanks!

packages/components/src/interfaces/complex/Hero.ts
seaerchin: only if they have this option selected - if they don't currently have it, then it's forward compat

packages/components/src/templates/next/components/complex/Hero/HeroGradient.tsx
seaerchin: i think it makes sense and it reads as "don't show the secondary button if i don't hvae a primary button" which seems to be reasonable

packages/components/src/templates/next/components/complex/Hero/HeroLargeImage/HeroLargeImage.tsx
seaerchin: ah oops, this is `ImageContainer` and not `ImageClient`

packages/components/src/interfaces/complex/Hero.ts
adriangohjw: re: Slack convo

- Only 2 agencies are being onboarded for now, and both will definitely use this — so we're treating it as a template-only setup for the time being.
- Per discussion with Rachel, adding GrowthBook logic to hide fields per agency feels like over-engineering at this stage. We'll just hide the fields for now and expose them later when we open it up to all agencies.
- If either agency changes and save this by mistake and wants to revert it, we'll manually help them revert since the new options aren't visible by them yet.

seaerchin (approved):
lgtm!

to make sure we are aligned: 
1. changes in `variant` from hero is controlled by `Studio`, via a radio. for all our users, they will only see the default gradient/block styling, which is compat
2. when this pr rolls out, **post release**, we'll update the two agencies to have the new variant (floating/large image)
3. the component for hero is already updated on teh component side so the site will build without issue.

PR #1331, opened 2025-05-20
Potential fix for code scanning alert no. 9: Prototype-polluting assignment

adriangohjw:
Potential fix for [https://github.com/opengovsg/isomer/security/code-scanning/9](https://github.com/opengovsg/isomer/security/code-scanning/9)

To fix the issue, we need to prevent prototype pollution by ensuring that untrusted input cannot be used to modify `Object.prototype`. The best approach here is to replace the plain object `tagCategories` with a `Map`, which does not inherit from `Object.prototype` and is immune to prototype pollution. Alternatively, we could validate the `category` key to reject unsafe values like `__proto__`, but using a `Map` is a more robust and future-proof solution.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


adriangohjw:
fyi: this is autogenerated by github

seaerchin (approved):

PR #1342, opened 2025-05-21
chore(deps): add ignore rules for major version updates in dependabot

adriangohjw:
due to it keep creating new PRs everyday 😢


seaerchin (approved):

seaerchin:
should probably also ignore `eslint` given our experience on monday...

PR #1343, opened 2025-05-22
feat(bold-components): full image variant for infocards

seaerchin:
## Problem
see attached ticket

## Solution
**studio**
1. on studio, add `hidden` for the new schema (some changes wrt design, but will clear with isodesigners)

**components**
1. split different components and styles out into their own files 
2. for styling specifically, spilt single infocard and overall infocardS (note the s) styling into 2 parts
3. use 2 grids to achieve specified styling where if there are specifically 5 items, we do a 3/2 split on `lg`; this does have a downside - on tablet specifically where we tile into 2 col grids, there will be an empty space between the first grid (3 items) and the second grid (2 items); i'll check with iso-designers on this 
4. pass `variant` down into sub-components so they can also change style based on whether it's a `bold` variant


packages/components/src/templates/next/components/complex/InfoCards/common.ts
```diff
@@ -0,0 +1,172 @@
+import { INFOCARD_VARIANT } from "~/interfaces/complex/InfoCards"
```
harishv7, line 1: nit: should we have the file named as styles.ts?

packages/components/src/templates/next/components/complex/InfoCards/common.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export const createInfoCardsStyles = tv({
+  extend: singleInfoCardStyle,
+  slots: {
+    container: `${ComponentContent} flex flex-col`,
+    headingContainer: "flex flex-col",
+    headingTitle: "prose-display-md break-words text-base-content-strong",
+    headingSubtitle: "text-base-content",
+    grid: "grid grid-cols-1",
+    urlButtonContainer: "mx-auto block pt-8 sm:pt-12", // temp: following headingContainer's mb
```
harishv7, line 86: what's the temp comment abt?

harishv7 (approved):

PR #1349, opened 2025-05-22
Singpass fallback - Disabling singpass send notification

adriangohjw:
## Problem

Closes https://www.notion.so/opengov/Isomer-Next-Decision-Matrix-for-Disabling-Singpass-as-2FA-1d177dbba78880289845fc20f53f8503#1d277dbba78880f1b071ef622311fac6

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- on `publishPageResource` when singpass is down, send an email to publisher and all site admins

**Improvements**:

- refactor `ISOMER_SUPPORT_EMAIL` and `ISOMER_SUPPORT_LINK`

### DIFF
--- a/apps/studio/src/features/mail/templates/templates.ts
+++ b/apps/studio/src/features/mail/templates/templates.ts
@@ -4,8 +4,12 @@ import type {
   BaseEmailTemplateData,
   EmailTemplate,
   InvitationEmailTemplateData,
+  PublishAlertContentPublisherEmailTemplateData,
+  PublishAlertSiteAdminEmailTemplateData,
 } from "./types"
+import { ISOMER_SUPPORT_EMAIL, ISOMER_SUPPORT_LINK } from "~/constants/misc"
 import { env } from "~/env.mjs"
+import { getResourceStudioUrl } from "~/utils/resources"
 
 export const invitationTemplate = (
   data: InvitationEmailTemplateData,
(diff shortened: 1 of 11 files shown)


apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -542,6 +542,7 @@ export const pageRouter = router({
       })
       return publishPageResource(
         ctx.logger,
+        ctx.gb,
```
seaerchin, line 545: [question] if all we're passing in growthbook for is to determine whether singpass is enabled, shouldn't we do it at a higher level (eg: router) so that the service method can be agnostic?

apps/studio/src/features/mail/templates/templates.ts
```diff
@@ -50,11 +54,47 @@ export const invitationTemplate = (
   }
 }
 
+export const publishAlertContentPublisherTemplate = (
+  data: PublishAlertContentPublisherEmailTemplateData,
+): EmailTemplate => {
+  const { recipientEmail, siteName, resource } = data
+  const resourceStudioUrl = getResourceStudioUrl(resource)
```
seaerchin, line 61: [nit] `studioResourceUrl`

apps/studio/src/features/mail/templates/templates.ts
```diff
@@ -50,11 +54,47 @@ export const invitationTemplate = (
...
+<p>Best,</p>
+<p>Isomer team</p>`,
+  }
+}
+
+export const publishAlertSiteAdminTemplate = (
+  data: PublishAlertSiteAdminEmailTemplateData,
+): EmailTemplate => {
+  const { recipientEmail, publisherEmail, siteName, resource } = data
+  const resourceStudioUrl = getResourceStudioUrl(resource)
```
seaerchin, line 77: [nit] `studioReosurceUrl`

seaerchin (approved):
lgtm, minor nits

seaerchin:
not sure if we're doing - i thought we were going to be sending notifs on login also when singpass is down?

adriangohjw:
bugbot run

apps/studio/src/features/mail/templates/templates.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1349/commits/9faba3d4858aff37c312dbc18501fc9bf8821e57 !

apps/studio/src/features/mail/templates/templates.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1349/commits/9faba3d4858aff37c312dbc18501fc9bf8821e57 !

adriangohjw:
failing test is because we are using an incorrect (non-existent) `NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY` for `.env.test` and the fallback is `false`

code is working as intended - i will update GB to include the key

adriangohjw:
> not sure if we're doing - i thought we were going to be sending notifs on login also when singpass is down?

@seaerchin ah yes, added it here https://github.com/opengovsg/isomer/pull/1349/commits/6af12fdf95597facae1881fc6da0b6ba127c1efe

adriangohjw:
update: will be solved after https://github.com/opengovsg/isomer/pull/1483 merged

PR #1355, opened 2025-05-24
add temporary workaround for VAPT testers in isGovEmail function

adriangohjw:
## Problem

We wanna do VAPT and grant testers ADMIN access but they cannot be added as their email doesn't end with `.gov.sg`

## Solution

Hardcoded to return true for `isGovEmail` for `vapt` environment and their emails

### DIFF
--- a/apps/studio/src/utils/email.ts
+++ b/apps/studio/src/utils/email.ts
@@ -1,9 +1,20 @@
 import isEmail from "validator/lib/isEmail"
 
+import { env } from "~/env.mjs"
+
 /**
  * Returns whether the passed value is a valid government email.
  */
 export const isGovEmail = (value: unknown) => {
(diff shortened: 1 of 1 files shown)


apps/studio/src/utils/email.ts
```diff
@@ -1,9 +1,21 @@
...
 
+import { env } from "~/env.mjs"
+
 /**
  * Returns whether the passed value is a valid government email.
  */
 export const isGovEmail = (value: unknown) => {
+  // Temp. workaround to allow VAPT testers to be added as admins
+  if (
+    env.NEXT_PUBLIC_APP_ENV === "vapt" &&
```
harishv7, line 11: set reminder to remove end of month or so

harishv7 (approved):

PR #1362, opened 2025-05-26
Fix getmetadata invalid site url

adriangohjw:
## Problem

We had an issue last week where `url` doesn't have `https://` and thus calling site build to fail with an unknown error

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- wrap in `try/catch` for better error handling
- add checks for `http://` or `https://` and auto append if missing


packages/components/src/engine/metadata.ts
```diff
@@ -60,9 +60,19 @@ const getMetaImage = (props: IsomerPageSchemaType) => {
...
+      const siteUrl =
+        props.site.url.startsWith("http://") ||
+        props.site.url.startsWith("https://")
+          ? props.site.url
+          : `https://${props.site.url}`
+      return new URL(props.page.permalink, siteUrl).toString()
+    } catch {
+      return props.page.permalink
+    }
+  })()
```
seaerchin, line 75: actually i disagree with this approach; i'd rather we just throw an error here so that we can guarantee that the `site.url` in our db is a well-formed url. 

if we go with this approach, we should also disallow `http://`

seaerchin (commented):
should we just throw a better error instead during URL construction?

packages/components/src/engine/metadata.ts
harishv7: I think @seaerchin  has a good point, we should rather fix and be opinionated about the expected input as the schema is the contract to be followed/adhered to. When it does not match the expected input, it should error out

Else, we will end up with schemas of site that vary between each other.

As per URL standards it should be https:// or http://. But for Isomer, we will strictly disallow http, so it should throw as well for http

adriangohjw: okay will update!

PR #1375, opened 2025-05-27
Fix/hero floating width

adriangohjw:
## Problem

Floating variance is not full width if the text is too short

Ref: https://opengovproducts.slack.com/archives/C06R4DX966P/p1748325229517369

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- adding full width but capped with calculated max-width

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- refer to storybook

### DIFF
--- a/apps/studio/prisma/generated/generatedTypes.ts
+++ b/apps/studio/prisma/generated/generatedTypes.ts
@@ -1,147 +1,154 @@
-import type { ColumnType, GeneratedAlways } from "kysely";
-export type Generated<T> = T extends ColumnType<infer S, infer I, infer U>
-  ? ColumnType<S, I | undefined, U>
-  : ColumnType<T, T | undefined, T>;
-export type Timestamp = ColumnType<Date, Date | string, Date | string>;
+import type { ColumnType, GeneratedAlways } from "kysely"
 
-import type { ResourceState, ResourceType, RoleType, AuditLogEvent } from "./generatedEnums";
+import type {
+  AuditLogEvent,
+  ResourceState,
+  ResourceType,
+  RoleType,
+} from "./generatedEnums"
 
-export type AuditLog = {
-    id: GeneratedAlways<string>;
-    userId: string;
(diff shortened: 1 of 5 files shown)


apps/studio/prisma/generated/generatedTypes.ts
adriangohjw, line 1: unrelated lint and fix

packages/components/src/templates/next/components/complex/Hero/HeroFloating.tsx
```diff
@@ -61,7 +61,7 @@ export const HeroFloating = ({
         />
       </div>
       {/* Floating container */}
-      <div className="mx-6 -mt-[1.5rem] md:mx-10 md:-mt-[10rem] lg:mx-0 lg:-mt-[17.85rem]">
+      <div className="mx-6 -mt-[1.5rem] w-full max-w-[calc(100%-3rem)] md:mx-10 md:-mt-[10rem] md:max-w-[calc(100%-5rem)] lg:mx-0 lg:-mt-[17.85rem] lg:max-w-none">
```
seaerchin, line 64: [question] why do we need to subtract the 3 rem and 5 rem for the `max-w`? went to look at the designs - seems like the empty white space is 6 units/10 units, which shuold be 1.5rem and 2.5 rem respectively

seaerchin (approved):
minor question; tested with single word title + description; the hero floating box still looked good

packages/components/src/templates/next/components/complex/Hero/HeroFloating.tsx
adriangohjw: oh cos we using `w-full` which excludes margin, effectively putting the margin outside the box and not being visible

adriangohjw:
@sehyunidaaa i have added another example to show how it looks like on homepage!

sehyunidaaa (approved):

PR #1376, opened 2025-05-27
Feat/add new AskGov embed

adriangohjw:
## Problem

Askgov is deprecating wizgov for the new askgov widget.
Note: During this transition period, wizgov will still be supported. 

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add askgov widget implementation (replicate of wizgov widget)

note: require updating CSP on infra side


packages/components/src/interfaces/internal/Askgov.ts
```diff
@@ -0,0 +1,10 @@
+import type { IsomerSiteProps } from "~/types"
+
+export interface AskgovProps {
+  "data-agency": string
+  "data-topic"?: string
```
seaerchin, line 5: [question]: what happens if `data-topic` is left empty or omitted? (ie, if i put `""` or if it's `undefined`)

packages/components/src/templates/next/components/internal/Askgov/AskgovWidget.tsx
```diff
@@ -0,0 +1,50 @@
...
+  environment,
+  ...askgovProps
+}: AskgovWidgetProps) => {
+  const scriptUrl =
+    environment === "production"
+      ? "https://script.ask.gov.sg/widget.js"
+      : "https://script-staging.ask.gov.sg/widget.js"
+
+  const reloadAskgovScript = () => {
+    const scriptId = "isomer-askgov-script"
```
seaerchin, line 18: [nit] no real issue, but prefer it to be outside of the component in `SCREAMING_SNAKE_CASE`

seaerchin (approved):

packages/components/src/interfaces/internal/Askgov.ts
adriangohjw: Oh most cases it will be empty and not used. This is only for niche cases where agencies wanna pre-filter and only show questions from certain topics

PR #1392, opened 2025-06-04
Fix safari grid problem for infocards

adriangohjw:
## Problem

https://opengovproducts.slack.com/archives/C07CWUNUL68/p1749016488478229

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- adding `grid-auto-rows: max-content` and `grid-template-rows: max-content;`. more details here: https://opengovproducts.slack.com/archives/C06R4DX966P/p1749021958760469

some related links:
- https://stackoverflow.com/questions/44770074/css-grid-row-height-safari-bug

## Tests

1. go to browserstack and select Safari 17 (on Sonoma OS)
2. go to a newly built page with infocards - it should not look jank on both desktop and mobile


apps/studio/public/assets/css/preview-tw.css
```diff
@@ -2082,6 +2082,10 @@ video {
   list-style-type: none;
 }
 
+.auto-rows-max {
```
harishv7, line 2085: the tests specifically mention safari 17. is that the minimum version we support?

adriangohjw: weirdly, it only affects safari 15, 16, 17 and not the earlier versions (<14). But yes, it works for all 3 of them

PR #1398, opened 2025-06-06
migration(index-page): add children page blocks to all existing folders

seaerchin:
## Problem
we want to add children pages block to any block that does not have it at present. 

Closes [insert issue #]

## Solution
- this PR adds a migration to do that and verifies the changes via tests
- this works via getting all index pages with `layout: index`, then filtering for those without an existing `childrenpages` block 
- when we get this list, we spread the existing content and add the default children page at the last index

### DIFF
--- a/apps/studio/prisma/scripts/migrations/addChildrenBlock/addChildrenBlock.ts
+++ b/apps/studio/prisma/scripts/migrations/addChildrenBlock/addChildrenBlock.ts
@@ -28,7 +28,7 @@ export const up = async () => {
           "type" = 'IndexPage'
       ) AS "indexContent"
     WHERE
-      ("content"->'content')::text NOT ILIKE '%"type": "childrenpages"%';
+      ("content"->'content')::text NOT ILIKE '%"type": "childrenpages"%' and "content" ->> 'layout' = 'index';
 `
 
   const result = await sqlQuery.execute(db)
--- a/apps/studio/prisma/scripts/migrations/addChildrenBlock/addChildrenBlock.ts
+++ b/apps/studio/prisma/scripts/migrations/addChildrenBlock/addChildrenBlock.ts
@@ -1,0 +1,58 @@
+import { DEFAULT_CHILDREN_PAGES_BLOCK } from "@opengovsg/isomer-components"
+
+import { db, sql } from "~/server/modules/database"
+import { updateBlobById } from "~/server/modules/resource/resource.service"
+
+export const up = async () => {
+  // Step 1: select all index pages and their blobs where
+  // the text `"type": "childrenpages"` isn't found
+  const sqlQuery = sql<{
+    id: string
+    siteId: string
+    content: PrismaJson.BlobJsonContent
+  }>`
+    SELECT
+      *
+    FROM
+      (
+        SELECT
+          COALESCE("db"."content", "pb"."content") AS "content",
+          "Resource"."id", 
+          "Resource"."siteId"
+        FROM
+          "Resource"
+          LEFT JOIN "Version" ON "Version"."id" = "Resource"."publishedVersionId"
+          LEFT JOIN "Blob" AS "pb" ON "pb"."id" = "Version"."blobId"
+          LEFT JOIN "Blob" AS "db" ON "db"."id" = "Resource"."draftBlobId"
+        WHERE
+          "type" = 'IndexPage'
+      ) AS "indexContent"
+    WHERE
+      ("content"->'content')::text NOT ILIKE '%"type": "childrenpages"%';
+`
+
+  const result = await sqlQuery.execute(db)
+
+  for (const r of result.rows) {
+    const existing = r.content.content
+    const updated = [...existing, DEFAULT_CHILDREN_PAGES_BLOCK]
+    const newContent = {
+      ...r.content,
+      content: updated,
+    }
+
+    await db.transaction().execute((tx) =>
+      updateBlobById(tx, {
+        pageId: Number(r.id),
+        content: newContent,
+        siteId: Number(r.siteId),
+      }),
+    )
+  }
+}
+
+// NOTE: Uncomment the below to run the migration
+// await up()
+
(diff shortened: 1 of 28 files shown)


adriangohjw:
bugbot run

seaerchin:
> The migration assumes `r.content.content` is always an array. If it's null, undefined, or not an array, the spread operator will cause a runtime error, leading to migration failure. Add validation to ensure `r.content.content` is an array before attempting to spread it.

this is a code level guarantee given by our ajv validator

apps/studio/prisma/scripts/migrations/addOrdering/addOrdering.ts
```diff
@@ -0,0 +1,89 @@
...
+          COALESCE("db"."content", "pb"."content") AS "content",
+          "Resource"."id", 
+          "Resource"."siteId"
+        FROM
+          "Resource"
+          LEFT JOIN "Version" ON "Version"."id" = "Resource"."publishedVersionId"
+          LEFT JOIN "Blob" AS "pb" ON "pb"."id" = "Version"."blobId"
+          LEFT JOIN "Blob" AS "db" ON "db"."id" = "Resource"."draftBlobId"
+        WHERE
+          "type" = 'IndexPage'
```
dcshzj, line 28: question: to confirm, all our index pages are indeed using `IndexPage` resource type? There is no resource that has permalink `_index`, resource type `Page` or something and uses `index` layout.

apps/studio/prisma/scripts/migrations/addOrdering/addOrdering.ts
```diff
@@ -0,0 +1,89 @@
...
+      ) AS "indexContent"
+    WHERE
+      ("content"->'content')::text ILIKE '%"type": "childrenpages"%' and "content" ->> 'layout' = 'index' and ("content" -> 'content')::text NOT ILIKE '%"childrenPagesOrdering"%';
```
dcshzj, line 31: question(non-blocking): to read back this condition, it is looking for:
- Layout is `index`, and
- Page contains `childrenpages` block, and
- Page does not contain `childrenPagesOrdering`

Is this interpretation correct? Cos seems a bit different from the PR description that says we are adding the `childrenpages` block to all index pages if it doesn't exist.

apps/studio/prisma/scripts/migrations/addOrdering/addOrdering.ts
```diff
@@ -0,0 +1,89 @@
...
+        ...r.content,
+        content: updated,
+      }
+
+      await db.transaction().execute((tx) =>
+        updateBlobById(tx, {
+          pageId: Number(r.id),
+          content: newContent as any,
+          siteId: Number(r.siteId),
+        }),
```
dcshzj, line 74: question: if a page only has a published blob but not a draft blob, we will write these changes as a draft for the page. Will we be publishing for these pages as well or don't need?

apps/studio/prisma/scripts/migrations/addChildrenBlock/addChildrenBlock.ts
```diff
@@ -0,0 +1,58 @@
...
+        FROM
+          "Resource"
+          LEFT JOIN "Version" ON "Version"."id" = "Resource"."publishedVersionId"
+          LEFT JOIN "Blob" AS "pb" ON "pb"."id" = "Version"."blobId"
+          LEFT JOIN "Blob" AS "db" ON "db"."id" = "Resource"."draftBlobId"
+        WHERE
+          "type" = 'IndexPage'
+      ) AS "indexContent"
+    WHERE
+      ("content"->'content')::text NOT ILIKE '%"type": "childrenpages"%' and "content" ->> 'layout' = 'index';
```
dcshzj, line 31: question(non-blocking): ah I read the `addOrdering` file first, so to confirm, we will:
1. Run this `addChildrenBlock` first to add the `childrenpages` block to all index pages, then
2. Run the `addOrdering` script to add the `childrenPagesOrdering`?

apps/studio/prisma/scripts/migrations/addOrdering/addOrdering.ts
```diff
@@ -0,0 +1,89 @@
...
+  for (const r of result.rows) {
+    const last = r.content.content.at(-1)
+
+    // Skip if the last block is not a childrenpages block
+    if (last?.type !== "childrenpages") {
+      console.log(
+        `Skipping Resource ${r.id}: last block is not childrenpages (type: ${last?.type})`,
+      )
+      continue
+    }
```
dcshzj, line 45: question: do we need to find for the `childrenpages` block to add the `childrenPagesOrdering` property? Seems like potentially there can be pages that are missing that.

apps/studio/prisma/scripts/migrations/addOrdering/addOrdering.ts
seaerchin: that wouldn't match our secondary condition (`content.layout === 'index'`); based off the rfc for index pages, all index pages have either `layout: index` or `layout: content`

apps/studio/prisma/scripts/migrations/addOrdering/addOrdering.ts
seaerchin: i didn't publish because i followed our existing principle when creating a folder - published folder but we set the blob as `draft` to ensure user can see and know the changes (as they are teh content owner)

apps/studio/prisma/scripts/migrations/addChildrenBlock/addChildrenBlock.ts
seaerchin: the order is:

1. run `addOrdering` so that every **existing** index page **with childrenpages block but without ordering** now has ordering in their child block
2. run `addChildrenBlock` so that any **existing index page without childrenpages block** now has the default children block (with `ordering`) 

(together, this means that all index pages have childrenpagse + ordering)

PR #1420, opened 2025-06-12
feat: allow users to edit link title

seaerchin:
## Problem
we want to allow users to edit their link title and self service

## Solution
1. reuse the existing page settings modal and remove the restriction that it needs to be a `Page` 
2. update wording of the modal and associated toasts to remove `page` (cleared by design alr)
3. hide the url when it's a link

## Videos

https://github.com/user-attachments/assets/346046d4-eca3-4164-9c13-382613319d08

## Tests
1. click into a collection
2. create a collection link
3. try to update the title
4. it should work
5. repeat steps 2-4 but for a collection page
6. click into a foldre
7. create a folder page
8. try to update the title
9. it should work


adriangohjw:
bugbot run

PR #1422, opened 2025-06-12
fix(deploy): add github ref to image tag 

seaerchin:
## Problem
right now, our docker images are uniquely identified by the github sha. most of the itme, this is ok. however, when we want to rollback a commit, what happens is that we will attach a tag to the old version. this results in deploys failing as our tags are immutable on production

Closes [insert issue #]

## Solution 
- add `github.ref` to be part fo the image tag. when this is from a push, this will be a branch name. however, on release, this will be the `release_tag`; this means we can have 2 release tags that are differently only ihn the release version but having the same commit hash


adriangohjw:
bugbot run

PR #1424, opened 2025-06-12
feat(collection-links): alllow users to add thumbnails to items

seaerchin:
## Problem
see ticket

## Solution
1. add the `item` to the `LinkProps` in components
2. on studio, add this to the atom and to the backend where we send it to save. we also update the data being sent to teh preview iwth the image.

## Video

https://github.com/user-attachments/assets/5163b902-8fc5-4056-a974-68aa1b822f63

## Notes
- this PR isn't quite ideal because this change requires alot of manual work, whereas our ideal state is that we're hooked into json forms and the update is automatic (+ now our validation is out of sync on our backend). i can't quite remember why i needed to use the `atom`, but i'll try using the same `ajv` validator and see how it goes


adriangohjw (approved):

adriangohjw:
bugbot run

PR #1429, opened 2025-06-13
chore: add watch command

seaerchin:
## Problem
don't want to have ot manually run `watch` every time a thing change on components

## Solution
add `npm run watch:packages` command that watches our `build` whenever there are code changes so that it'll reflect on studio automatically


adriangohjw:
thank you for the kind work 🫡

PR #1440, opened 2025-06-16
VICA - add some default colours 

adriangohjw:
some of these colours should be set to a default value based on agencie's colour

closes https://linear.app/ogp/issue/ISOM-1954/add-colours-for-vica


seaerchin (approved):

packages/components/src/templates/next/components/internal/Vica/VicaWidget.tsx
```diff
@@ -13,6 +15,21 @@ export const VicaWidget = (props: VicaWidgetProps) => {
...
+      app-button-border-color="var(--color-brand-canvas-inverse)"
+      app-canvas-background-color="var(--color-brand-canvas-default)"
+      app-quick-reply-button-background-color="#FFFFFF" // hardcoded to be white for all agencies
+      app-auto-complete-background-color="var(--color-brand-canvas-default)"
+      app-auto-complete-hover-color="var(--color-brand-canvas-alt)"
+      app-auto-complete-foreground-color={twColors.gray["900"]}
+      app-auto-complete-divider-color={twColors.gray["500"]}
+      app-recommendations-background-color="var(--color-brand-canvas-default)"
+      app-recommendations-hover-color="var(--color-brand-canvas-alt)"
+      app-recommendations-foreground-color={twColors.gray["900"]}
```
seaerchin, line 32: [nit] should we use a color token rather than `twColors`? not a blocker, happy to merge this but might be confusing if we need to sync this to the agency's color down teh line

adriangohjw: makes sense, updated!

PR #1446, opened 2025-06-18
Chore(stories): add stories for collection link and thumbnail 

seaerchin:
## Problem
prior pr had no stories - this pr adds that in 

## Solution
- add stories


adriangohjw:
bugbot run

apps/studio/src/stories/Page/CollectionPage.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+    growthbook: [
+      createBannerGbParameters({
+        variant: "info",
+        message:
+          "This is a test banner that is very long. This is a test banner that is very long. This is a test banner that is very long. This is a test banner that is very long. This is a test banner that is very long.",
+      }),
+    ],
+  },
+}
+
```
adriangohjw, line 84: question: why do we need this for CollectionPage story?

apps/studio/src/stories/Page/CollectionPage.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+export const ExpandedProfileDropdown: Story = {
+  play: async ({ canvasElement }) => {
+    const screen = within(canvasElement)
+    const testUserSelector = await screen.findByText(/TU/i)
+    const testUserSelectorButton = testUserSelector.closest("button")
+    if (testUserSelectorButton) {
+      await userEvent.click(testUserSelectorButton)
+    }
+  },
+}
```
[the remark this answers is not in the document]
seaerchin: but more seriously speaking, the profile/banner are "global" components -> they get injected via the `layout`; we should check they remain the same across different paths (or just at the layout level is sufficient)

apps/studio/src/stories/Page/CollectionPage.stories.tsx
seaerchin: see above

PR #1447, opened 2025-06-18
feat: take sort order from index pages

seaerchin:
**NOTE: THIS PR IS PENDING A TEST ON STAGING**

## Problem
we want the siderail to be sorted via the index page's sorting order also 

## Solution
1. sort the index page at the publishing level 
2. we take the index page order preferentially then fallback to folder meta if found. 
3. not removing the existing sort order on `ChildrenPages` -> this sort in publishing is kidna magical and i'd prefer to keep the extra sort in components just for guarantees

### DIFF
--- a/tooling/build/scripts/publishing/index.ts
+++ b/tooling/build/scripts/publishing/index.ts
@@ -304,6 +304,8 @@ function generateSitemapTree(
   const children = [...existingChildren, ...danglingDirectories]
 
   // Get the page sorting order from the FolderMeta resource
+  // TODO: delete this once `FolderMeta` is removed
+  /** @deprecated use `pageOrderFromIndex` instead; we should remove this once `FolderMeta` is removed from db */
   const pageOrderFromMeta = resources.find(
     (resource) =>
       resource.type === "FolderMeta" &&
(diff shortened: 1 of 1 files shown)


tooling/build/scripts/publishing/index.ts
```diff
@@ -313,6 +313,27 @@ function generateSitemapTree(
...
+  const pageOrderFromIndex = resources
+    .find(
+      (resource) =>
+        resource.type === "IndexPage" &&
+        resource.fullPermalink ===
+          (pathPrefixWithoutLeadingSlash.length === 0
+            ? INDEX_PAGE_PERMALINK
+            : `${pathPrefixWithoutLeadingSlash}/${INDEX_PAGE_PERMALINK}`),
+    )
+    ?.content?.content?.at(-1)
```
dcshzj, line 325: suggestion: I think we should use `.find` instead to find for the `childrenPages` block?

tooling/build/scripts/publishing/index.ts
```diff
@@ -304,7 +304,7 @@ function generateSitemapTree(
   const children = [...existingChildren, ...danglingDirectories]
 
   // Get the page sorting order from the FolderMeta resource
-  const pageOrder = resources.find(
+  const pageOrderFromMeta = resources.find(
```
dcshzj, line 307: nit: can mark this as deprecated if we have confirmed that we will stick to the original ADRs and remove FolderMeta

tooling/build/scripts/publishing/index.ts
seaerchin: sounds good; i did this because it's always the last block but i think making the code clearer sounds godo to me too

dcshzj (approved):

PR #1450, opened 2025-06-19
chore: update to use the org wide secret

seaerchin:
## Problem
- still have 2 usages of our own `DD_API_KEY` for GHA to configure test visibility; this has been replaced with our org wide secret

## Solution
- use org wide secret
- reusable workflow for `aws_deploy` passes in `DD_API_KEY_GITHUB_ACTIONS` as the `secrets.DD_API_KEY`

## Notes 
will proceed to delete our own secret after this has been merged

### DIFF
--- a/.github/workflows/aws_deploy.yml
+++ b/.github/workflows/aws_deploy.yml
@@ -105,10 +105,6 @@ on:
         description: "Intercom app ID"
         required: false
         type: string
-    secrets:
-      DD_API_KEY:
-        description: "Datadog API key for uploading sourcemaps"
-        required: false
 
 permissions:
   id-token: write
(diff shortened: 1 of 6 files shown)


seaerchin:
> Hmm I thought the DD_API_KEY is no longer needed in aws_deploy so we should be able to remove all instances?

- this is kept for clarity (as in not using the same name as the org wide one), unless you mean to just remove this and reference the `secret` directly
- if the latter is what you meant, since `aws_deploy` is a reusable workflow, it **cannot** have [secrets](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) unless we use `inherit`
- i didn't go for that option because i wanted all the dependencies to be directly on the workflow itself

dcshzj:
Hmm I meant more like cos the workflow does not require the secret anymore (don't see it being used anywhere inside the file), so we might be able to just remove it entirely?

PR #1454, opened 2025-06-19
chore: fix image showing errors when empty

seaerchin:
## Problem
- right now, when a user adds image/alt text and then deletes both of them (ie, both image/alt are empty), we still show an error and prevent them from saving
- this is because the object containing both properties (`image` and `alt`) still exists as an empty object but our schema specifies that the object has to be either `undefined | { image, alt }` 

## Solution
- update props to our object control
- `handleChange` so that when `isEmpty(object)`, we set `object` itself to `undefined` 

 ## Videos

https://github.com/user-attachments/assets/d8214ecf-809a-47ae-823c-2365af1c6018

## Tests
- [ ] create an article
- [ ] click into the article's header
- [ ] upload an image
- [ ] the bottom save button should be grayed out
- [ ] delete teh image the button shuold be allowed
- [ ] add alt text (without an image)
- [ ] the button should still be grayed out
- [ ] when both are there, the button shuold be allowed

(basically is opposite of XOR for imgae/alt)

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
@@ -90,11 +90,15 @@ const withContextToControlWithDetailProps = (
     ctx,
     props,
   }: JsonFormsStateContext & ControlWithDetailProps) {
-    // NOTE: provides `handleChange`
+    // NOTE: provides `handleChange` for our method.
+    // Unfortunately, the `ctx` is typed as `any` here
+    // and requires suprresion.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
     const dispatchProps = ctxDispatchToControlProps(ctx.dispatch)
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsObjectControl.tsx
```diff
@@ -61,4 +71,36 @@ export function JsonFormsObjectControl({
...
+const withContextToControlWithDetailProps = (
+  Component: ComponentType<ControlWithDetailProps>,
+) =>
+  function WithContextToControlProps({
+    ctx,
+    props,
+  }: JsonFormsStateContext & ControlWithDetailProps) {
+    // NOTE: provides `handleChange` for our method.
+    // Unfortunately, the `ctx` is typed as `any` here
+    // and requires suprresion.
```
dcshzj, line 95:
```suggestion
    // and requires suppression.
```

dcshzj (approved):

seaerchin:
### Merge activity

* **Aug 7, 4:37 AM UTC**: @seaerchin merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1454).

PR #1460, opened 2025-06-19
Improve lighthouse TBT (reduce JS build size)

adriangohjw:
## Problem

**Lighthouse TBT score for mobile is high**

- Some observation: as compared to a Classic site, a Next site has a better performance on desktop but worst off performance on mobile (speficially total blocking time (TBT)). Upon [research](https://uploadcare.com/blog/what-is-total-blocking-time/), high TBT is often associated with JS and the need to process them. This is because the browser need to parse and evaluate the script, and the large the script, the more likely it takes more than 50ms, which is what lighthouse consideres as "long task". The goal is to reduce the number of chunks, or reduce the size of chunks.
image

- Unfortunately, NextJS static export is made to optimize for general speed and browser-cached for navigation speed [(e.g. `next/link` prefetch)](https://www.charlievuong.com/the-hidden-pitfalls-of-the-next-js-link-on-website-speed) and less of the initial first load. Thus, compared to other static framework like Jekyll (isomer classic), it loads entire `react-dom` and `next` (~60kb gzipped each) on first render. This means Lighthouse benchmarking, which is always using a cold first load, will always suffer in comparison

- nevertheless, there are areas where we can control, which is the JS chunks generated by ourselves through isomer-components **(this PR mainly targets this)** . However, because of the way in `template` we are importing `RenderEngine` which statically import all components e.g. `DynamicDataBanner`, this means these are included even though they aren't used. This is made worst by the fact all of them share the same route `/app/[[...permalink]]/page.tsx` which means the same bundle JS chunk is used for every page, even though they aren't necessarily needed for that route e.g. `CollectionClient` is needed for collection but not for homepage
image

- also, unrelated but all of a sudden ([compared to marc](https://mccy-muis-base.pages.dev/)h), 3rd party scripts start to be blocking 😢 
image

## Solution

**Improvements**:

use [nextjs bundle analyzer](https://nextjs.org/docs/app/guides/package-bundling) to understand output files (also useful [reference](https://www.charlievuong.com/fix-lighthouse-reduce-unused-javascript-using-next-js-bundle-analyzer))

- remove datadogrum since we are no longer using it (we are using clarity anyway)
- make interweave-ssr polyfill (cause for the `parse5/lib`) runs only on server and not client 
- use `sideEffects: false` for `packages/components` to treeshake imports that are unused and not necessary for runtime e.g. `sinclair` and interfaces (types)

together, this (building site locally on my laptop)
- reduces the main JS chunk by ~50% (250kb -> 120kb), while moving interweave-ssr polyfill to its standalone module
- reduces the TBT from 300-500ms to consistently 70ms

> [!NOTE]
> this PR addresses the lower hanging fruit and does not addressing TBT caused by 1) 3rd party scripts, or 2) splitting the main JS chunk into smaller chunks based on routes, or 3) the fundamental "issue" of using the `RenderEngine` approach

## Before & After Screenshots

**BEFORE**:

Screenshot 2025-06-20 at 5 40 52 AM

**AFTER**:

Screenshot 2025-06-20 at 3 28 28 AM

Screenshot 2025-06-20 at 2 36 57 PM


adriangohjw:
note: i will follow up with a separate PR to remove DD from our infra

packages/components/src/templates/next/components/internal/BaseParagraph/BaseParagraph.tsx
```diff
@@ -1,15 +1,17 @@
...
 import { Link } from "../Link"
 
-// NOTE: We need this polyfill as interweave uses a DOM to perform the
-// conversion of HTML to React components
-polyfill()
+// This will be tree-shaken out of client bundles
+if (typeof window === "undefined") {
+  void import("interweave-ssr").then(({ polyfill }) => {
+    polyfill()
+  })
```
dcshzj, line 13: nit: can retain the original comment on why we need this polyfill

adriangohjw: @dcshzj done! https://github.com/opengovsg/isomer/pull/1460/commits/737026483e20f1ed2bd8db9375e80372beccaf50

PR #1466, opened 2025-06-23
Fix collectionblock preview

adriangohjw:
## Problem

Closes https://opengovproducts.slack.com/archives/C0924LBA0PR

## Solution

**Bug Fixes**:

- revert changes on how `id` is generated for collection that has index pages

## Tests

1. Go to studio and create a collection with an index page
2. Go to root page and add a valid collectionblock to the homepage (linking to the collection ID as reference) - page should render

### DIFF
--- a/apps/studio/src/utils/sitemap.ts
+++ b/apps/studio/src/utils/sitemap.ts
@@ -75,16 +75,7 @@ const getSitemapTreeFromArray = (
     const summaryOfPage = indexPage?.summary
 
     return {
-      id:
-        resource.type === ResourceType.Collection
-          ? // Note: Current publishing script still use collection id
-            // as the id even if there is an index page
-            resource.id
-          : // NOTE: this is done for compat reasons as
-            // our publishing script uses the index page id
-            // if it can find one and if it cannot,
-            // then uses the base `resourceId`
-            (indexPage?.id ?? resource.id),
+      id: resource.id,
       layout:
         resource.type === ResourceType.Collection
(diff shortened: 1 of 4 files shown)


apps/studio/src/utils/sitemap.ts
```diff
@@ -75,11 +75,16 @@ const getSitemapTreeFromArray = (
...
+      id:
+        resource.type === ResourceType.Collection
+          ? // Note: Current publishing script still use collection id
+            // as the id even if there is an index page
+            resource.id
+          : // NOTE: this is done for compat reasons as
+            // our publishing script uses the index page id
+            // if it can find one and if it cannot,
+            // then uses the base `resourceId`
+            (indexPage?.id ?? resource.id),
```
dcshzj, line 87: question: sorry missing a bit of context here, shouldn't we always use the Folder/Collection resource ID instead of the index page? Why do we need to use the index page's ID?

From what I understand that's what publishing script does, because the Folder/Collection resource is technically guaranteed to exist but not the index page (only an app-level guarantee).

adriangohjw: oh actually im  also not 100% sure - i realized this was changed but looking back i also don't 100% rmb/understand why it was changed. I only know the changes seem to be more about Folder rather than Collection, thus i made this change - was thinking of waiting for @seaerchin to be back but at least this PR can unblock the preview issue for agencies

dcshzj (approved):

PR #1469, opened 2025-06-23
Loosen character limits on image (+ image gallery) and infobar

sehyunidaaa:
## Problem

We impose character limits on Isomer Next conservatively. To minimise ways to abuse a component, we take the minimum number of characters that make sense. This is usually 50 for very short text (button), 100 for short text (titles), 150-200 for longer text (description) and so on. 

These conservative limits make sure that the component looks balanced when the limits are maxed. However, as we migrate more sites we start seeing more sites with content that they cannot afford to shorten. These two components -- image and infobar -- are recent examples.

Closes [ISOM-1967]

## Solution

- Increase infobar description limit from 100 -> 200
- Increase image caption limit from 100 -> 250
- Increase image gallery's image caption limit from 100 -> 250
- Image's caption has a smaller 
- In Image Gallery, captions will be truncated to 3 lines on mobile to avoid obstructing the image too much. Front-load your captions with important details.

However, Isomer Next's principles on concise and easy-to-scan content still stand.

### DIFF
--- a/packages/components/src/interfaces/complex/Image.ts
+++ b/packages/components/src/interfaces/complex/Image.ts
@@ -46,7 +46,7 @@ export const ImageSchema = Type.Object(
     caption: Type.Optional(
       Type.String({
         title: "Caption",
-        maxLength: 100,
+        maxLength: 250,
         format: "textarea",
       }),
     ),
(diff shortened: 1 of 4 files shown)


adriangohjw (commented):
need to update preview-tw

PR #1474, opened 2025-06-24
Improve lighthouse TBT (update GTM to use "src" instead of inline)

adriangohjw:
## Problem

Why `dangerouslySetInnerHTML` performs worse
- even if it eventually injects GTM asynchronously (j.async = true), the execution of that whole IIFE **runs immediately**, during the main thread’s execution window.
- That adds script evaluation time to your page load, which: 1) Shows up in "Long Tasks" in Chrome performance profiler, and 2) Contributes to TBT and INP in Lighthouse

## Solution

- using `src` and `afterinteractive` instead of inline

 ### pros (as compared to existing inline method)

improved performance and UX
- delaying the loading of GTM to reduce render-blocking resources
- ensures that critical content is available to users without delay. 

### pros (as compared to existing inline method)
_note, this is our tradeoff_

- Delayed tag firing: Tags that rely on GTM may fire later than expected, potentially missing early user interactions e.g. **missing tracking of users who bounced too early**
   - tested on Preview mode on GTM dashboard and its still firing events 

### Impact

**1. Performance**
- as a standalone, there's minimal improvement
- however, when coupled with https://github.com/opengovsg/isomer/pull/1460, it consistently scores >90 for a live site

**2. Best practices**
- i think its because clarity's loading is also delayed (until after page hydrates), it no longer shows up in "best practices"

image

Note:
- this is based off one of the better sites
- even on sites with not-so-good-performance score 80+ for performance, as compared to their 70+ currently on WOGAA.

### DIFF
--- a/packages/components/src/templates/next/components/internal/GoogleTagManager/GoogleTagManagerHeader.tsx
+++ b/packages/components/src/templates/next/components/internal/GoogleTagManager/GoogleTagManagerHeader.tsx
@@ -10,13 +10,8 @@ const GoogleTagManagerHeaderScript = ({
   return (
     <ScriptComponent
       id={`_next-gtm-init-${gtmId}`}
-      dangerouslySetInnerHTML={{
-        __html: `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm, just wanted to check that if `gtmId` is empty, we don't error out on the user right?

PR #1476, opened 2025-06-25
feat(aws): scan images with inspector and fail deploy if threshold exceeded

seaerchin:
## Problem
see corresponding ticket 

## Solution
1. add aws inpsector on infra side (corresponding `pulumi up` has been performed on `staging`)
2. scan the buiilt docker image iwth `aws inspector` then check if it's over thereshold (we have 2 critical vulnerability - 1 in nextjs, 1 in golang but no known exploits...)
3. fail the deploy if over threshold + upload assets 

## TODOs
- [ ] `pulumi up` on `production/uat` once environment is up to date

### DIFF
--- a/.github/workflows/aws_deploy.yml
+++ b/.github/workflows/aws_deploy.yml
@@ -173,11 +173,27 @@ jobs:
         with:
           artifact_type: "container"
           artifact_path: "${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}:${{ env.IMAGE_TAG }}"
-          critical_threshold: 1
-          high_threshold: 1
-          medium_threshold: 1
-          low_threshold: 1
-          other_threshold: 1
+          critical_threshold: 5
+          high_threshold: 10
+
+      # Display Inspector results in the GitHub Actions terminal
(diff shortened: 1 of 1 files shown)


.github/workflows/aws_deploy.yml
```diff
@@ -164,6 +163,49 @@ jobs:
...
+      # from the GitHub actions job summary page.
+      - name: Upload Scan Results
+        id: upload-scan-results
+        uses: actions/upload-artifact@v4
+        with:
+        path: |
+          ${{ steps.inspector.outputs.inspector_scan_results }}
+          ${{ steps.inspector.outputs.inspector_scan_results_csv }}
+          ${{ steps.inspector.outputs.inspector_scan_results_markdown }}
+          ${{ steps.inspector.outputs.artifact_sbom }}
```
adriangohjw, line 196: question: where is this being uploaded and where can we see the content?

seaerchin: github actions [summary](https://github.com/opengovsg/isomer/actions/runs/15869259433?pr=1476); alternatively, you can click onto the step itself and the artifact link will be there!

fun fact: this is also what we use to uploda e2e test artifacts

PR #1484, opened 2025-07-01
Fix/revert gtm inline

adriangohjw:
## Problem

Fixed missing GTM data caused by https://github.com/opengovsg/isomer/pull/1474

## Solution

- reverting changes 
- instead, will set GTM and Clarity to load on "Window Loaded" in GTM setting instead, as per these few resources:
   - https://www.wpspeedfix.com/how-to-reduce-total-blocking-time/#Move-JavaScript-to-Google-Tag-Manager-and-Use-the-Window-Loaded-Trigger
   - https://web.dev/articles/tag-best-practices#choose_an_appropriate_trigger_event


dcshzj (approved):
> instead, will set GTM and Clarity to load on "Window Loaded" in GTM setting instead

This needs to be done by the agency yeah? Can add into the site launch runbook and the on-call runbook to tell the agency to enable this setting?

PR #1487, opened 2025-07-02
chore: add hadolint action for linting

seaerchin:
## Problem
we need to add dockerfile linting to comply with ssp. 

## Solution
- add hadolint as a ci action. the ci action for hadolint internally calls `hadolint` via CLI and stores the output as a variable which [it then exits with ](https://github.com/hadolint/hadolint-action/blob/master/hadolint.sh#L42)
- i cross checked this locally to make sure that `hadolint` exits with an error code (non zero basically) 
image
- set this to be on `true` but not blocking on pushes to `main` etc. we should be able to fix these later rather than have them block releases


adriangohjw (approved):
nice! wondering if we should also add a recommended .vscode extension?

PR #1489, opened 2025-07-03
isom-1867 prune inactive users (service + email)

adriangohjw:
## Problem

> [!NOTE]
> this PR does not include the scheduler aspect of it - will do that in another branch and PR 

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add service to remove inactive users from sites
- add email template for sending to them when deactivating

### DIFF
--- a/apps/studio/src/features/mail/templates/__tests__/templates.test.ts
+++ b/apps/studio/src/features/mail/templates/__tests__/templates.test.ts
@@ -1,7 +1,8 @@
 import { RoleType } from "~prisma/generated/generatedEnums"
 
+import { ISOMER_SUPPORT_EMAIL, ISOMER_SUPPORT_LINK } from "~/constants/misc"
 import { env } from "~/env.mjs"
-import { invitationTemplate } from "../templates"
+import { accountDeactivationTemplate, invitationTemplate } from "../templates"
 
 describe("invitationTemplate", () => {
   const mockData = {
@@ -54,5 +55,181 @@ describe("invitationTemplate", () => {
         role: "InvalidRole" as RoleType,
       }),
     ).toThrow("Unknown role. Please check the role type.")
+  })
+})
+
+describe("accountDeactivationTemplate", () => {
+  const mockData = {
+    recipientEmail: "test@example.com",
+    sitesAndAdmins: [],
+  }
+
+  it("should generate correct subject line", () => {
+    // Arrange
+    const template = accountDeactivationTemplate(mockData)
+
+    // Assert
+    expect(template.subject).toBe(
+      "[Isomer Studio] Your account has been deactivated due to inactivity",
+    )
+  })
+
+  it("should generate correct body content with recipient email", () => {
+    // Arrange
+    const template = accountDeactivationTemplate(mockData)
+
+    // Assert
+    expect(template.body).toContain("Hi test@example.com")
+  })
+
+  it("should include inactivity message with correct days", () => {
+    // Arrange
+    const template = accountDeactivationTemplate(mockData)
+
+    // Assert
+    expect(template.body).toContain(
+      "Your Isomer Studio account has been removed as you have not logged in for over 90 days.",
+    )
+  })
+
+  it("should include reason for deactivation", () => {
+    // Arrange
+    const template = accountDeactivationTemplate(mockData)
+
+    // Assert
+    expect(template.body).toContain(
+      "This is a standard security measure to protect your site data.",
+    )
+  })
+
+  it("should include content preservation and site accessibility reassurance message", () => {
+    // Arrange
+    const template = accountDeactivationTemplate(mockData)
+
+    // Assert
+    expect(template.body).toContain(
+      "Your content and previous contributions have been preserved",
+    )
+    expect(template.body).toContain(
+      "Your site(s) will continue to be accessible to visitors, and all your work remains intact.",
+    )
+  })
+
+  it("should include instructions for regaining access", () => {
+    // Arrange
+    const template = accountDeactivationTemplate(mockData)
+
+    // Assert
+    expect(template.body).toContain(
+      "To regain access to your site(s), please follow the instructions below",
+    )
+  })
+
+  it("should include site name and admin emails in instructions when admins exist", () => {
+    // Arrange
+    const template = accountDeactivationTemplate({
+      recipientEmail: "test@example.com",
+      sitesAndAdmins: [
+        {
+          siteName: "Test Site 1",
+          adminEmails: ["admin1@example.com", "admin2@example.com"],
+        },
+      ],
+    })
+
+    // Assert
+    const expectedSiteInstructions = `<p><b>Test Site 1</b></p>
+          <p>To regain access, please contact one of your site's administrators:</p>
+          <ul><li>admin1@example.com</li><li>admin2@example.com</li></ul>
+        `
+    expect(template.body).toContain(expectedSiteInstructions)
+  })
+
+  it("should show support email message when no admins exist", () => {
+    // Arrange
+    const template = accountDeactivationTemplate({
+      recipientEmail: "test@example.com",
+      sitesAndAdmins: [
+        {
+          siteName: "Test Site 1",
+          adminEmails: [],
+        },
+      ],
+    })
+
+    // Assert
+    const expectedSiteInstructions = `<p><b>Test Site 1</b></p>
+        <p>There are no administrators for this site. To be added back, please send an email to <a href="${ISOMER_SUPPORT_LINK}">${ISOMER_SUPPORT_EMAIL}</a> with your line manager in CC for approval.</p>
+      `
+    expect(template.body).toContain(expectedSiteInstructions)
+  })
+
+  it("should handle multiple sites with admins", () => {
+    // Arrange
+    const template = accountDeactivationTemplate({
+      recipientEmail: "test@example.com",
+      sitesAndAdmins: [
+        {
+          siteName: "Site A",
+          adminEmails: ["admin1@example.com"],
+        },
+        {
+          siteName: "Site B",
+          adminEmails: ["admin2@example.com", "admin3@example.com"],
+        },
+      ],
+    })
+
+    // Assert
+    const expectedSiteInstructionForSiteA = `<p><b>Site A</b></p>
+          <p>To regain access, please contact one of your site's administrators:</p>
+          <ul><li>admin1@example.com</li></ul>
+        `
+    expect(template.body).toContain(expectedSiteInstructionForSiteA)
+
+    const expectedSiteInstructionForSiteB = `<p><b>Site B</b></p>
+          <p>To regain access, please contact one of your site's administrators:</p>
+          <ul><li>admin2@example.com</li><li>admin3@example.com</li></ul>
+        `
+    expect(template.body).toContain(expectedSiteInstructionForSiteB)
+  })
+
+  it("should handle multiple sites with mixed admin scenarios", () => {
+    // Arrange
+    const template = accountDeactivationTemplate({
+      recipientEmail: "test@example.com",
+      sitesAndAdmins: [
+        {
+          siteName: "Site with one admin",
+          adminEmails: ["admin@example.com"],
+        },
+        {
+          siteName: "Site with two admins",
+          adminEmails: ["admin1@example.com", "admin2@example.com"],
+        },
+        {
+          siteName: "Site without Admins",
+          adminEmails: [],
+        },
+      ],
+    })
+
+    // Assert
+    const expectedSiteInstructionForSiteWithOneAdmin = `<p><b>Site with one admin</b></p>
+          <p>To regain access, please contact one of your site's administrators:</p>
+          <ul><li>admin@example.com</li></ul>
+        `
+    expect(template.body).toContain(expectedSiteInstructionForSiteWithOneAdmin)
+
+    const expectedSiteInstructionForSiteWithTwoAdmins = `<p><b>Site with two admins</b></p>
+          <p>To regain access, please contact one of your site's administrators:</p>
+          <ul><li>admin1@example.com</li><li>admin2@example.com</li></ul>
+        `
+    expect(template.body).toContain(expectedSiteInstructionForSiteWithTwoAdmins)
+
+    const expectedSiteInstructionForSiteWithoutAdmins = `<p><b>Site without Admins</b></p>
+        <p>There are no administrators for this site. To be added back, please send an email to <a href="${ISOMER_SUPPORT_LINK}">${ISOMER_SUPPORT_EMAIL}</a> with your line manager in CC for approval.</p>
+      `
+    expect(template.body).toContain(expectedSiteInstructionForSiteWithoutAdmins)
   })
 })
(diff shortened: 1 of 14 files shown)


apps/studio/src/server/modules/user/__tests__/utils.ts
adriangohjw, line 1: refactored as `hasLoggedIn` doesn't seem to be used for this helper method anywhere

adriangohjw:
bugbot run

adriangohjw:
bugbot run

adriangohjw:
> ### Bug
> The `deactivateUser` function's query for `sitesAndAdmins` has multiple issues:
> 
> 1. **Missing sites with no admins:** The `LEFT JOIN` on `ResourcePermission` combined with an `INNER JOIN` on `User` and subsequent filters effectively becomes an `INNER JOIN`. This incorrectly excludes sites where the deactivated user was the last active administrator (or where all remaining administrators are also being deactivated), preventing the deactivation email from notifying users about these sites and directing them to support, despite the email template's design to handle sites with no administrators.

fixed in https://github.com/opengovsg/isomer/pull/1489/commits/d4d359d4ae113b1a4836781945b6bb779bb280bd

> 2. **Duplicate admin emails:** The `array_agg(User.email)` does not use `DISTINCT`, leading to duplicate admin emails in the `adminEmails` array if an admin has multiple permissions on the same site.

will not happen -> each user will only have one permission due to DB relations

> 3. **`NOT IN` with empty array:** The `WHERE "ResourcePermission.userId" NOT IN userIdsToDeactivate` clause may behave unexpectedly if `userIdsToDeactivate` is an empty array.

fixed in https://github.com/opengovsg/isomer/pull/1489/commits/49a139af6e9452aee9b54d1c108b3cc3b721026d

apps/studio/src/features/mail/templates/__tests__/templates.test.ts
```diff
@@ -56,3 +57,179 @@ describe("invitationTemplate", () => {
...
+    expect(template.body).toContain(expectedSiteInstructionForSiteA)
+
+    const expectedSiteInstructionForSiteB = `<p><b>Site B</b></p>
+          <p>To regain access, please contact one of your site's administrators:</p>
+          <ul><li>admin2@example.com</li><li>admin3@example.com</li></ul>
+        `
+    expect(template.body).toContain(expectedSiteInstructionForSiteB)
+  })
+
+  it("should handle multiple sites with mixed admin scenarios", () => {
```
seaerchin, line 197: [test] what happens if you're the admin :sadge: (or if you're the only admin)

apps/studio/src/server/modules/user/__tests__/inactiveUsers.service.test.ts
seaerchin, line 1: i skipped reviewing this

apps/studio/src/server/modules/user/inactiveUsers.service.ts
seaerchin, line 1: [question] why split out rather than just putting it in `users.service`? not very important, but migbht cause confusion next time

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+import type { User } from "../database"
+import type { AccountDeactivationEmailTemplateData } from "~/features/mail/templates/types"
+import { sendAccountDeactivationEmail } from "~/features/mail/service"
+import { createBaseLogger } from "~/lib/logger"
+import { db, RoleType, sql } from "../database"
+import { PG_ERROR_CODES } from "../database/constants"
+import { MAX_DAYS_FROM_LAST_LOGIN } from "./constants"
+
+export const HOURS_IN_MS = 60 * 60 * 1000
```
seaerchin, line 11: [super nit] if exporting this, we should probably put in a shared `constants.ts` but this is v minor

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  const sgNow = new Date(now.getTime() + sgOffsetMs)
+
+  // Subtract daysAgo and zero out time (SG midnight)
+  sgNow.setDate(sgNow.getDate() - daysAgo)
+  sgNow.setHours(0, 0, 0, 0)
+
+  // Convert back to UTC
+  const utcDate = new Date(sgNow.getTime() - sgOffsetMs)
+  return utcDate
+}
```
seaerchin, line 31: [question] concern here is readability - this seems like it's trying to get the db time and then get the last logged in via that - should we instead use another small [library](https://day.js.org/docs/en/installation/installation)? this is backend anyways, so the user evner has to load it.

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        const siteIdsUserHasPermissionsFor = deletedPermissions.map(
+          (p) => p.siteId,
+        )
+
+        return await tx
+          .with("siteAdmins", (eb) =>
+            eb
+              .selectFrom("Site")
+              .leftJoin(
```
seaerchin, line 117: [question] why is this a `leftJoin`? the left table is `Site` and this just means that there will be sites without a joined `ResourcePermission`?

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+          .execute()
+
+        // For idempotency purposes since multiple nodes may run this function at the same time
+        if (deletedPermissions.length === 0) return []
+
+        const siteIdsUserHasPermissionsFor = deletedPermissions.map(
+          (p) => p.siteId,
+        )
+
+        return await tx
```
seaerchin, line 113: [nit] actually we don't need the `return await` - can just directly return a `Promise` + the style mixing here (cos we `.then` below) makes it difficulit to parse. 

can we just do diff `awaits` then return later?

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  // If there are no sites and admins, we don't need to send an email
+  // Note: this is to ensure there's idempotency and not send duplicate emails
+  if (sitesAndAdmins.length === 0) return
+
+  try {
+    await sendAccountDeactivationEmail({
+      recipientEmail: user.email,
+      sitesAndAdmins,
+    })
+  } catch {
```
seaerchin, line 181: [issue] not a fan of this, because it seems to mix in account sending iwth deactivation. should we let the controller handle this?

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+          .select([
+            "baseSites.siteName",
+            sql`COALESCE("siteAdmins"."adminEmails", ARRAY[]::text[])`.as(
+              "adminEmails",
+            ),
+          ])
+          .execute()
+          .then(
+            (result) =>
+              result as AccountDeactivationEmailTemplateData["sitesAndAdmins"],
```
seaerchin, line 155: [nit] avoid the `.then` and shift the cast to `sql`:

```ts
  await db.select([
            "baseSites.siteName",
            sql<
              string[]
            >`COALESCE("siteAdmins"."adminEmails", ARRAY[]::text[])`.as(
              "adminEmails",
            ),
          ])
```

main benefit is locality of information - easy to see that the cast (in code) is the same cast in sql

apps/studio/src/server/modules/user/inactiveUsers.service.ts
seaerchin, line 1: i'm assuming we're keeping this file, and that only the executor will change (from node cron to a job runner)

apps/studio/src/features/mail/templates/__tests__/templates.test.ts
seaerchin: from reading later on, we'll send out the mail still but say that they should contatc isomser admin; notsure if i saw a test case and i lazy reread lol

apps/studio/src/server/modules/user/inactiveUsers.service.ts
adriangohjw: good point! https://github.com/opengovsg/isomer/pull/1489/commits/559ed4d8991a71637286f68ad77b82c1faa1d657

apps/studio/src/server/modules/user/inactiveUsers.service.ts
adriangohjw: good point! https://github.com/opengovsg/isomer/pull/1489/commits/c770fe47fdacb7cb7fbe63bfdc42bb28c4856706

apps/studio/src/server/modules/user/inactiveUsers.service.ts
adriangohjw: good reccommendation! https://github.com/opengovsg/isomer/pull/1489/commits/a437d6d068b25f5ca3d1246ae006a3bdda26da10

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,198 @@
...
+    .execute()
+}
+
+interface DeactivateUserProps {
+  user: User
+  userIdsToDeactivate: User["id"][] // used for idempotency purposes
+}
+export const deactivateUser = async ({
+  user,
+  userIdsToDeactivate,
```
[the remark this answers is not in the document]
adriangohjw: `usersIdToDeactivate` is less so to avoid duplicates, but to handled the edge case that the list of users we will be asking the user to reach out to remains correct regardless of ordering

e.g. SiteA has AdminA and AdminB. Assuming both will be deactivated and we don't pass in `usersIdToDeactivate `,  AdminA will receive an email asking them to reach out to AdminB to restore their access. When in actual fact, AdminB will also be removed and unable to do so

but i agree with your proposed pipeline too! will update

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -72,127 +70,172 @@ export const getInactiveUsers = async ({
...
+    .execute()
+    .then((users) =>
+      users.map((user) => ({
+        user,
+        siteIds: deletedPermissions
+          .filter((permission) => permission.userId === user.id)
+          .map((permission) => permission.siteId),
+      })),
+    )
 }
```
seaerchin, line 110: [nit] just for consistency's sake, can we avoid mixing `await` and `then` syntax? 

```ts
const users = tx.updateTable(...).execute()
return users.map(...)
```

seaerchin (approved):
lgtm. minor comment to avoid mixing `async/await`

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,241 @@
...
+        })
+
+        for (const { user, siteIds } of deactivatedUsersAndSiteIds) {
+          if (siteIds.length === 0) continue
+
+          const sitesAndAdmins = await getSiteAndAdmins({
+            tx,
+            userId: user.id,
+            siteIds,
+          })
```
seaerchin, line 211: [question] we are still in the current `tx` so will the user who's being deactivated get an email also? not a big issue, i'm fine w leaving this in too lol

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,241 @@
...
+export const bulkDeactivateInactiveUsers = async (): Promise<void> => {
+  // not setting fromDaysAgo to MAX_DAYS_FROM_LAST_LOGIN because we want to
+  // deactivate users beyond the MAX_DAYS_FROM_LAST_LOGIN (e.g. 91 days). This could be due to:
+  // 1. legacy users who wasn't covered by the MAX_DAYS_FROM_LAST_LOGIN before the feature was introduced
+  // 2. users who weren't removed for whatever reason e.g. unintentional failed cron job that wasn't retried
+  const inactiveUsers = await getInactiveUsers({})
+
+  const deactivatedUsersAndSites: SendAccountDeactivationEmailsProps = []
+
+  try {
```
seaerchin, line 194: [readability] this `try/catch` chunk reads very oddly because what we're trying to do in the inner loop requires a context switch to know that we're also pushing the deactivated user into the above array. 

if we can allow for a more relaxed requirement on `getSitesAndAdmins`, wdyt about the following:

```ts
// shift the db query into this + handle the serialisation error here 
// we don't need to try/catch and rethrow then because this should fail upwards in our service
const deactivatedUsersAndSiteIds = await deactivateUsers(inactiveUsers)
// shift the iteration over `deactivatedUsersAndSiteIds` into this
const sitesAndAdmins = await getSiteAdmins(deactivatedUsersAndSiteIds)

// shift the length check into this
await sendAccountDeactivationEmails(deactivatedUslersAndSites)
```

main benefit is that the `service` reads more clearly (so we can follow the flow) then can read the individual nitty gritty later so that logic is clearer. my assumption here is that this has value because we'll shift into a job based system later rather than the existing cron, so the smaller utility functions can more easily be reused there.

if you don't think so, feel free to disregard

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -0,0 +1,241 @@
...
+            "ResourcePermission",
+            "ResourcePermission.siteId",
+            "Site.id",
+          )
+          .innerJoin("User", "User.id", "ResourcePermission.userId")
+          .where("Site.id", "in", siteIds)
+          .where("ResourcePermission.userId", "!=", userId) // don't want to ask users to ask themselves for permissions
+          .where("ResourcePermission.deletedAt", "is", null)
+          .where("ResourcePermission.role", "=", RoleType.Admin) // should only give the admin emails to request reactivation permissions from
+          .where("User.email", "not in", ISOMER_ADMINS_AND_MIGRATORS_EMAILS) // we don't want to send emails to admins and migrators
```
seaerchin, line 144: [side observation] this `ISOMER_ADMINS_AND_MIGRATORS_EMAILS` pattern comes up enough that we might benefit just from making a super admin table lol

seaerchin (commented):
good to haves that i feel might have value later on; none are blocking - i assume the only throwaway part is the cronjob init later on!

apps/studio/src/server/modules/user/inactiveUsers.service.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/1489/commits/6c8ef4c16b06d64aaece7914c7764ac4d8276f35

apps/studio/src/server/modules/user/inactiveUsers.service.ts
adriangohjw: makes sense - updated in https://github.com/opengovsg/isomer/pull/1489/commits/e12c45f3cb63da2289d834c3dd46e727b6ad106a ! thanks!

apps/studio/src/server/modules/user/inactiveUsers.service.ts
adriangohjw: good catch, updated in https://github.com/opengovsg/isomer/pull/1489/commits/e12c45f3cb63da2289d834c3dd46e727b6ad106a ! thanks!

PR #1491, opened 2025-07-07
isom-1867 prune inactive users cron (scheduler)

adriangohjw:
## Problem

## Solution

image

- add `node-cron`
- schedule the jobs on server start and stop at server end


apps/studio/src/server/cron/jobs/sendAccountDeactivationDayEmailJob.ts
seaerchin, line 1: to double confirm - this isn't used rgiht

seaerchin (approved):

apps/studio/src/server/cron/jobs/sendAccountDeactivationDayEmailJob.ts
adriangohjw: yup, not yet until scheduler publishing !

PR #1493, opened 2025-07-07
isom-1867 prune inactive users service (remove banner)

adriangohjw:
## Problem

no longer need this banner since users should be auto removed

## Solution

- removed code relating to inactive users warning and banner on user management page
- removed `user.count` query need to filter by `inactive` params


seaerchin (approved):
lgtm (but how will users know this new behaviour of ours (that we remove useres >90 days logged in)

adriangohjw:
> lgtm (but how will users know this new behaviour of ours (that we remove useres >90 days logged in)

will have to align with Product / Ops as to inform users

PR #1501, opened 2025-07-09
chore: add migration for collections to have index pages

seaerchin:
## Problem
Related to ISOM-1993

we want to have our tags on the collection index pages, but not every collection is guaranteed to have an index page

Closes [insert issue #]

## Solution
add migration so that every collection will have an index page

### DIFF
--- a/apps/studio/prisma/scripts/addCollectionIndexPage/index.ts
+++ b/apps/studio/prisma/scripts/addCollectionIndexPage/index.ts
@@ -1,0 +1,86 @@
+import type { CollectionPagePageProps } from "@opengovsg/isomer-components"
+import {
+  COLLECTION_PAGE_DEFAULT_SORT_BY,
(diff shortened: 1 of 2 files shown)


apps/studio/prisma/scripts/addCollectionIndexPage/index.ts
```diff
@@ -0,0 +1,86 @@
...
+import {
+  db,
+  jsonb,
+  ResourceState,
+  ResourceType,
+} from "~/server/modules/database"
+import { PG_ERROR_CODES } from "~/server/modules/database/constants"
+
+export const up = async () => {
+  const collectionsWithoutIndexPages = await db
```
dcshzj, line 20: question: can I check when we create a collection inside Studio right now, does the index page get automatically created as well?

dcshzj (approved):
The migration LGTM, just a clarification on the behaviour of the app when creating collections

seaerchin:
closing, using other PR

PR #1508, opened 2025-07-14
DGS Integration: Searchable Table POC (as a component) 🧪

adriangohjw:
## Problem

> [!NOTE]
> This is a POC on how we can replace searchable table content with DGS as the data source 

## Solution

- creating a new component called `DGSSearchableTable`
- this component will fetch data from DGS via the resource IDs and the stated header fields
- it will fetch the endpoint once with default limit to get the total number of records via `result.total`, and then fetch again using that as the new limit. I opted for this approach because it "one-off load everything" and allows us to do pure client side filtering and reuse existing `SearchableTableClient`.
   - The alternative will be to fetch data from DGS using their `q` params to query which is worst off from UX perspective because 1) user have to wait for debounce or press "enter" manually AND 2) less instant (wait on API call every time) - also possibly introduce unnecessary surge in load for DGS and causing downtime on both them and us

## Screenshots

(can also refer to Chromatic to interact with it https://www.chromatic.com/build?appId=665439911775ce21bf8b8fb4&number=2122)

Screenshot 2025-07-15 at 3 48 18 AM


packages/components/src/templates/next/components/complex/DGSSearchableTable/DGSSearchableTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+        setData(response)
+      } catch {
+        setIsError(true)
+      } finally {
+        setIsLoading(false)
+      }
+    }
+
+    void fetchAllData()
+  }, [])
```
adriangohjw, line 59: - refer to PR description on why we call the API twice
- putting in a useEffect this way also because otherwise it will fail during SSG buildtime

adriangohjw:
closing this off as we are keeping it as a layout https://github.com/opengovsg/isomer/pull/1515

PR #1512, opened 2025-07-16
fix: add additional ordering by createdAt in user query

adriangohjw:
## Problem

Closes https://opengovproducts.slack.com/archives/C07CWUNUL68/p1752655893507529

issue: order not preserved for pagination due to similarly sorted order (all no lastLoginAt) with no tie breaker

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- use user's createdAt as tiebreaker

### DIFF
--- a/apps/studio/src/server/modules/user/user.router.ts
+++ b/apps/studio/src/server/modules/user/user.router.ts
@@ -235,6 +235,7 @@ export const userRouter = router({
 
       return getUsersQuery({ siteId, adminType })
         .orderBy("ActiveUser.lastLoginAt", sql.raw(`DESC NULLS LAST`))
+        .orderBy("ActiveUser.createdAt", "desc")
         .select((eb) => [
           "ActiveUser.id",
           "ActiveUser.email",


apps/studio/src/server/modules/user/user.router.ts
```diff
@@ -235,6 +235,7 @@ export const userRouter = router({
 
       return getUsersQuery({ siteId, adminType })
         .orderBy("ActiveUser.lastLoginAt", sql.raw(`DESC NULLS LAST`))
+        .orderBy("ActiveUser.createdAt", "desc")
```
dcshzj, line 238: question: hmm why not show all the admins first, then order by email address?

(or can be just by email address, if there is an issue with the user "jumping" when the role is being changed)

dcshzj: Hmm yeah I was referring to the `lastLoginAt` also, cos I would assume that as an admin when doing the monthly access review, I will know the email address of the user to remove. So I will either Ctrl+F for it or look for the email address directly, and having the email addresses in alphabetical order would be easier to look through.

The last login date seems to only be relevant when last logged in >90 days ago, but that is also covered by the different text colour alr, so it felt like doing ordering based on minimally the email address would be useful.

adriangohjw: oh on that, i believe we had some discussion (i couldn't find the doc.) but we decided to not spend too much time on this because we show 25 rows per page and most agencies back then have less than 25 users so they should see everything on the first page

dcshzj: Yeah hmm I always thought it was weird that we are not sorting by the email address at least, so might be a slightly stronger opinion on this haha

Can we sort by email address? Then don't need the secondary sorting

adriangohjw: but now that some users are accustomed to lastLoginAt, im thinking will it be weird if we remove that as the primary sorting haha

dcshzj: I think it is ok? The lastLoginAt timestamp will keep changing so I think people don't rely on it to find the person they want to remove.

It would have been useful if people were to use the last logged in time to find and remove those not logged in for >90 days, but we will also be doing the inactive users script quite soon so that utility is no longer there. The script would have removed those users by then, so there would be less need to sort by the last logged in time.

adriangohjw: okay, no strong opinion on this. updated in https://github.com/opengovsg/isomer/pull/1512/commits/24a44324ae89ec81b045000b567d4a2d09f36583

dcshzj (approved):

PR #1490, opened 2025-07-07
isom-1867 prune inactive users (reminder email)

adriangohjw:
## Problem

Partially closes https://linear.app/ogp/issue/ISOM-1867/ac-3-create-script-to-automatically-prune-users-not-logged-in-90-days

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [x] No - this PR is backwards compatible

**Features**:

- add logic to bulk send warning emails
- add email template for account deactivation emails

PR #1513, opened 2025-07-16
isom-1867 prune inactive users

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1867/ac-3-create-script-to-automatically-prune-users-not-logged-in-90-days

## Solution

**Breaking Changes**

  - Details ...

Refer to:
- https://github.com/opengovsg/isomer/pull/1489
- https://github.com/opengovsg/isomer/pull/1490
- https://github.com/opengovsg/isomer/pull/1491
- https://github.com/opengovsg/isomer/pull/1493
- https://github.com/opengovsg/isomer/pull/1494

> [!CAUTION]
> DO NOT MERGE as the scheduler should only start running on/after 1st August as per discussion with Ops

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -124,7 +124,6 @@
     "ajv-errors": "^3.0.0",
     "date-fns": "^4.1.0",
     "date-fns-tz": "^3.1.3",
-    "dayjs": "^1.11.13",
     "dd-trace": "^5.24.0",
     "exponential-backoff": "^3.1.2",
     "filenamify": "^6.0.0",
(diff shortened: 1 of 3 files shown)


apps/studio/package.json
```diff
@@ -124,6 +124,7 @@
     "ajv-errors": "^3.0.0",
     "date-fns": "^4.1.0",
     "date-fns-tz": "^3.1.3",
+    "dayjs": "^1.11.13",
```
dcshzj, line 127: issue: I just saw that we are using `dayjs`. We should prefer to use `date-fns` instead cos `dayjs` configurations are set globally and not scoped to a specific execution, which can cause conflicts in other parts of the codebase.

dcshzj (approved):

PR #1514, opened 2025-07-16
isom-1965 enable collection widget for all on studio

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-1965/enable-collection-widget-for-all-on-studio

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- adding a new `collection-dropdown` format
- adding a JsonForms renderer for that format to fetch all collections and drop down. Note: this is mostly copied from the existing category dropdown and tweaked
- add new collection API endpoint + unit tests
- because this is a dynamic component, we need a way to "hijack" and display an "instruction" otherwise it won't render and will throw error

## Before & After Screenshots

https://github.com/user-attachments/assets/cb01f271-b2b7-4a19-bfa7-adb8196844db

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionDropdownControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionDropdownControl.tsx
@@ -13,7 +13,7 @@ import { getReferenceLink } from "~/utils/link"
 import { trpc } from "~/utils/trpc"
 
 export const jsonFormsCollectionDropdownControlTester: RankedTester = rankWith(
-  JSON_FORMS_RANKING.AnyOfControl,
+  JSON_FORMS_RANKING.CollectionDropdownControl,
   and(schemaMatches((schema) => schema.format === COLLECTION_DROPDOWN_FORMAT)),
 )
 
--- a/apps/studio/src/constants/formBuilder.ts
+++ b/apps/studio/src/constants/formBuilder.ts
@@ -15,6 +15,8 @@ export const JSON_FORMS_RANKING = {
   ChildrenPagesOrderingControl: 5,
   AllOfControl: 3,
   AnyOfControl: 3,
+  CategoryControl: 3,
+  CollectionDropdownControl: 3,
   ProseControl: 3,
   LinkControl: 3,
   RefControl: 3,
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryControl.tsx
@@ -16,7 +16,7 @@ import { trpc } from "~/utils/trpc"
 import { JsonFormsTextControl } from "./JsonFormsTextControl"
 
 export const jsonFormsCategoryControlTester: RankedTester = rankWith(
-  JSON_FORMS_RANKING.AnyOfControl,
+  JSON_FORMS_RANKING.CategoryControl,
   and(schemaMatches((schema) => schema.format === "category")),
 )
 


adriangohjw:
> ### Bug: Dropdown Control Incorrectly Sets Category Value
> The `JsonFormsCollectionDropdownControl`'s `handleChange` function incorrectly sets `linkAtom.category` with a collection reference value (e.g., `[resource:siteId:collectionId]`). This is a copy-paste error from a category control, as this component should handle collection-related data.

fixed in https://github.com/opengovsg/isomer/pull/1514/commits/34257a11e3cf6ce2671d1f385285b6f3577981d0

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionDropdownControl.tsx
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  return (
+    <SingleSelect
+      value={data}
+      name={label}
+      items={collections.map((collection) => {
+        return {
+          label: collection.title,
+          value: `[resource:${siteId}:${collection.id}]`,
```
dcshzj, line 42: suggestion: I think there is a util function somewhere that constructs this reference link, would be good to use that instead

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionDropdownControl.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { COLLECTION_DROPDOWN_FORMAT } from "@opengovsg/isomer-components"
+
+import Suspense from "~/components/Suspense"
+import { JSON_FORMS_RANKING } from "~/constants/formBuilder"
+import { siteSchema } from "~/features/editing-experience/schema"
+import { useQueryParse } from "~/hooks/useQueryParse"
+import { trpc } from "~/utils/trpc"
+
+export const jsonFormsCollectionDropdownControlTester: RankedTester = rankWith(
+  JSON_FORMS_RANKING.RefControl,
```
dcshzj, line 15: issue: should use a different ranking name

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -349,4 +350,21 @@ export const collectionRouter = router({
...
       },
     ),
+  getCollections: protectedProcedure
+    .input(getCollectionsSchema)
+    .query(async ({ ctx, input: { siteId } }) => {
+      await validateUserPermissionsForResource({
+        siteId,
+        action: "read",
+        userId: ctx.user.id,
+      })
```
dcshzj, line 360: issue: hmm I think it should be `validateUserPermissionsForSite`? Cos will be querying over the entire site's collections

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionDropdownControl.tsx
dcshzj: Eh haha bugbot is right, I meant that each tester ranking should be its own name, like here should be called `JSON_FORMS_RANKING.DropdownControl` in case we need to change it to a different value for some reason in the future

adriangohjw: thought about it but felt like they are different UI variations of multi-select (AnyOf)

the only difference is
1. which queries they call e.g. getCollections, getCategories, and 
2. how to display them (dropdown, radio etc.)

As such, don't think there's a need to differentiate them as it wouldn't and shouldn't affect the ranking order though

dcshzj: Hmm I think you might be reading too much into the comment. It's more of each form control should have its own ranking name so it's easier to see the 1-to-1 mapping. The actual numbers for the ranking sure, they can be the same number.

If we ever need to adjust the ranking integer for the dropdown control only, it's easy to switch, not that we would ever need to but just in case.

PR #1520, opened 2025-07-23
fix: avoid terminating the script early on error

dcshzj:
## Problem

GitHub Copilot sucks.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Don't terminate the script early on errors.

### DIFF
--- a/apps/studio/scripts/connectRds.sh
+++ b/apps/studio/scripts/connectRds.sh
@@ -2,8 +2,6 @@
 # Sets up a tunnel to the RDS instance using AWS Session Manager
 # Usage: bash ./scripts/connectRds.sh prod|staging|uat|vapt
 
-set -e
-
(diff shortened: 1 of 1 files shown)


apps/studio/scripts/connectRds.sh
```diff
@@ -2,8 +2,6 @@
 # Sets up a tunnel to the RDS instance using AWS Session Manager
 # Usage: bash ./scripts/connectRds.sh prod|staging|uat|vapt
 
-set -e
```
harishv7, line 5: This is for local usage right? Why is copilot causing early termination?

adriangohjw: eh why it works for me? 👁️👄👁️

adriangohjw (approved):

PR #1525, opened 2025-07-30
chore: use central generateSiteConfig helper for stories

dcshzj:
## Problem

When doing #1523, I realised how much duplication we are doing with the site config, which is not really ideal.

## Solution

**Breaking Changes**

**Improvements**:

- Use a common helper function to generate the site config, and allow for overrides when necessary.


seaerchin:
ssed the faliure here
image

PR #1528, opened 2025-07-31
fix: update PermissionsProvider to use site perms instead of resource perms

adriangohjw:
## Problem

Publish button not shown https://opengovproducts.slack.com/archives/C06R4DX966P/p1753944863938479?thread_ts=1753939513.304439&cid=C06R4DX966P

previously, did a fix on a buggy query that does not return resource perms even when resourceId is passed in. Turns out that bug was pulling site perms always which make it return a correct value

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- use site perms by force passing in `resourceId: null`

### DIFF
--- a/apps/studio/src/server/modules/permissions/__tests__/permissions.service.test.ts
+++ b/apps/studio/src/server/modules/permissions/__tests__/permissions.service.test.ts
@@ -1051,7 +1051,8 @@ describe("getResourcePermission", () => {
     expect(permissions[0]?.role).toBe(RoleType.Admin)
   })
 
-  it("should return resource-specific permissions when resourceId is provided", async () => {
+  // TODO: add this back in when we have resource-specific permissions
+  it.skip("should return resource-specific permissions when resourceId is provided", async () => {
     // Arrange
     const user = await setupUser({ email: "test@example.com" })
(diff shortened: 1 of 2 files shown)


apps/studio/src/features/permissions/PermissionsContext.tsx
```diff
@@ -30,11 +30,10 @@ export const PermissionsContext = createContext<ResourceAbility | PureAbility>(
 export const PermissionsProvider = ({
   children,
   siteId,
-  resourceId,
```
dcshzj, line 33: issue: hmm I feel a bit iffy about this, cos the intent is that we support granular permissions up till the final check where we get the available roles for the user, then we fallback to the site. This removal feels like it is done "somewhere in the middle of the stack".

I thought that this "discarding" of resourceId should still be done in `getResourcePermission`, then all the callers will retain the assumption that it is checking the correct permissions for the specific resource.

harishv7: +1, the code should still retain the architecture/design of granular perms. within the individual checks on backend we can discard the use of resourceId to return site-wide perms

PR #1530, opened 2025-07-31
Fix/collection dropdown only those with children

adriangohjw:
## Problem

collection dropdown selection include collection without children, which should not be selectable for collectionblock

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- extend API to allow one to filter only for collections with children

### DIFF
--- a/apps/studio/src/server/modules/collection/collection.router.ts
+++ b/apps/studio/src/server/modules/collection/collection.router.ts
@@ -353,7 +353,7 @@ export const collectionRouter = router({
     ),
   getCollections: protectedProcedure
     .input(getCollectionsSchema)
-    .query(async ({ ctx, input: { siteId } }) => {
+    .query(async ({ ctx, input: { siteId, hasChildren } }) => {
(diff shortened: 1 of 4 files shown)


adriangohjw:
bugbot run

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -353,20 +353,30 @@ export const collectionRouter = router({
...
-        .selectFrom("Resource")
+      let query = db.selectFrom("Resource")
+
+      if (hasChildren) {
+        query = query.innerJoin(
+          "Resource as children",
+          "Resource.id",
+          "children.parentId",
+        )
+      }
```
seaerchin, line 372: collections are allowed to have pages, links and collection meta under them. this assumes that all these resource types are valid children for collections. i'm assuming this is ok? (though we need to be aware that if we have only collection meta under a collection, the collection itself will be empty)

PR #1533, opened 2025-08-01
security: only allow filename of extension from the allowlist to be uploaded

adriangohjw:
## Problem

Currently someone can actually maliciously upload a unsafe file via the `getPresignedPutUrl` API directly.

We only restrict it via the frontend at the moment and not the API

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- add it to schema to validate the file extension to be one that's in our allowed list
- write test

## Tests

1. try to call the API directly (via Insonmia, maybe) with a filename that contains `.html` or `.js` -> it should be rejected

### DIFF
--- a/apps/studio/src/schemas/asset.ts
+++ b/apps/studio/src/schemas/asset.ts
@@ -1,5 +1,16 @@
 import { z } from "zod"
 
+import {
+  FILE_UPLOAD_ACCEPTED_MIME_TYPE_MAPPING,
+  IMAGE_UPLOAD_ACCEPTED_MIME_TYPE_MAPPING,
+} from "~/features/editing-experience/components/form-builder/renderers/controls/constants"
+
+// Combine allowed extensions from existing constants
+const ALLOWED_EXTENSIONS = [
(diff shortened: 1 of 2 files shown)


apps/studio/src/schemas/asset.ts
seaerchin, line 1: right now, the presigned put url flow is:
1. create the file key (directly from the filename) -> lets call this `test.jpg`
2. generate the presigned put url (this uploads directly to `s3://my-bucket/2/some_uuid/test.jpg`)
3. what if i upload a file called `malicious.js`? 

at present, we only validate the filename out from fronted, but we're still allowing (i think) a file (malicious.js) to be uploaded using that presigned url. the only issue is that the `malicious.js` will be renamed to `test.jpg` -> is this sufficient to prevent attacks?

seaerchin (approved):
lgmt

adriangohjw:
adding link to discussion with SecEngs for reference purposes: https://opengovproducts.slack.com/archives/C055FUACLJU/p1754294026894809

PR #1535, opened 2025-08-04
feat: allow OGP Maps to be embedded in Isomer Next

dcshzj:
## Problem

OGP Maps is not supported on Isomer Next.

## Solution

**Breaking Changes**

**Features**:

- Add support for OGP Maps.
    - CSP whitelist is on infra repo: https://github.com/opengovsg/isomer-next-infra/pull/191

## Tests

- [ ] Go to Studio and edit any content page.
- [ ] Add a new Map component to the page.
- [ ] Edit the embed and insert the following:
```
<iframe src="https://maps.gov.sg/scdf-aed" width="100%" height="600" frameborder="0"></iframe>
```
- [ ] Verify that the embed is able to be created, and the map loads as expected.

### DIFF
--- a/packages/components/src/utils/validation.ts
+++ b/packages/components/src/utils/validation.ts
@@ -38,6 +38,20 @@ const isValidOneMapEmbedUrl = (urlObject: URL) => {
   )
 }
 
+// Slug pattern obtained from:
+// https://github.com/opengovsg/maps/blob/8c3f2e6c9af1be78fac5fc69115858241ae1f720/apps/web/modules/map/schema.ts#L3-L16
+const OGP_MAPS_SLUG_REGEX_PATTERN = "[a-z0-9][a-z0-9_-]*[a-z0-9]"
+
+const isValidOGPMapsEmbedUrl = (urlObject: URL) => {
+  return (
+    urlObject.hostname === "maps.gov.sg" &&
+    urlObject.pathname.startsWith("/") &&
+    new RegExp(`^${OGP_MAPS_SLUG_REGEX_PATTERN}$`).test(
+      urlObject.pathname.slice(1),
+    )
+  )
+}
+
 export const isValidMapEmbedUrl = (url: string) => {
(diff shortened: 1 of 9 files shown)


packages/components/src/utils/validation.ts
```diff
@@ -38,6 +38,20 @@ const isValidOneMapEmbedUrl = (urlObject: URL) => {
   )
 }
 
+// Slug pattern obtained from:
+// https://github.com/opengovsg/maps/blob/8c3f2e6c9af1be78fac5fc69115858241ae1f720/apps/web/modules/map/schema.ts#L3-L16
```
adriangohjw, line 42: can update to `develop` instead of `8c3f2e6c9af1be78fac5fc69115858241ae1f720`?

packages/components/src/utils/validation.ts
```diff
@@ -38,6 +38,20 @@ const isValidOneMapEmbedUrl = (urlObject: URL) => {
...
+const OGP_MAPS_SLUG_REGEX_PATTERN = "[a-z0-9][a-z0-9_-]*[a-z0-9]"
```
adriangohjw, line 43: should we be doing the check here on what is a valid or invalid slug?

seems like the intention is to have valid URL for the embedding, but
1. we are duplicating the application logic from Maps, and
2. this is half the validation (doesn't validate the other regex and min/max strings) so feels neither here nor there

dcshzj: Updated in 7beee0e55a2c00faf56a996a8928089c40df1c7c.

packages/components/src/utils/validation.ts
dcshzj: Removed, but we should reference the actual commit hash because the line numbers can run if they make changes to the file in the future

adriangohjw (approved):

adriangohjw:
bugbot run

PR #1537, opened 2025-08-05
fix: remove inactive users only for production

dcshzj:
## Problem

We do not want to remove users for all envs, only production.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Schedule the remove inactive users cron job only for the production env.

### DIFF
--- a/apps/studio/src/server/cron/index.ts
+++ b/apps/studio/src/server/cron/index.ts
@@ -1,3 +1,4 @@
+import { env } from "~/env.mjs"
 import { createBaseLogger } from "../../lib/logger"
 import { deactivateInactiveUsersJob } from "./jobs"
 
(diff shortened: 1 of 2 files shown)


shazlithebestie:
I don't think this is a correct way to fix this - to not remove any UAT users. 

What I feel is a good way to fix this is to not notify UAT users that they're removed: 
- They're already told that the UAT environment can be removed at any time at log in
- UAT environment is not meant for permanent use

shazlithebestie (changes requested):
I don't think this is a correct way to fix this - to not remove any UAT users.

What I feel is a good way to fix this is to not notify UAT users that they're removed:

They're already told that the UAT environment can be removed at any time at log in
UAT environment is not meant for permanent use

adriangohjw (approved):

shazlithebestie (commented):
based from what I see, I kindda get the logic. lgtm if what @dcshzj comments is what it is. over to engineers to approve.

PR #1540, opened 2025-08-07
feat(collection-tags): implement functionality on studio

seaerchin:
## Problem
This PR adds editing functionality **for the new data structure which we've adopted** for collection tags. this will **not work** for existing tags as the data structure is different, which requires a migration from the old format to the new format.

## Solution
- add a new `TaggedSchema` and `TagCategoriesSchema`. We cannot reuse the `tag` namespace on `page` because it clashes iwth our existing one, and we need to preserve functionality for existing sites while rollling this out. 
- For the `TaggedSchema` specifically, i opted to use a 2 layer data structure (ie, i store the UUID of the tag category along iwth the UUIDs of the selected options for that tag category) rather than just a single layer data structure. This is for ease of implementation downstream (studio + components), where we need to figure out what are the selected options for each category. 
- for the json renderer, we do not use a `ArrayLayout` due to issues faced iwth `JsonFormsDispatch` resolving the `uischemas` so we handle state reconcilation on our end before calling `handleChange`

## Notes
- This handles moves of a page from 1 collection to another gracefully because we extract the selected options by the `uuid`. however, this **does** introduce bloat because we don't trim the (now extra) properties. this means that if a user shifts a page with tags from collection A to collection B then back to collection A, the original tags will still show. 
- i opted to not optimise for this case (ie, by trimming the now obsolete tags) because i don't think space creep will be a real thing (ie, that the obsolete tags eat up alot of storage space).


packages/components/src/types/page.ts
```diff
@@ -138,6 +180,20 @@ export const ContentPagePageSchema = Type.Composite([
...
+// should fully own the state of the collection/folder
+// but the content page header should not.
+// Doing a straight copy paste rather than `Type.Composite`
+// to avoid unexpected spillover of properties
+export const IndexPagePageSchema = Type.Composite([
+  Type.Object({
+    contentPageHeader: ContentPageHeaderSchema,
+  }),
+  imageSchemaObject,
+])
```
seaerchin, line 195: technically this is not needed since the index page of a folder won't have the changes here (`tagCategories`) but i felt that it still makes sense to separate

packages/components/tsconfig.json
```diff
@@ -39,7 +39,7 @@
     // "emitDecoratorMetadata": true /* Enables experimental support for emitting type metadata for decorators. */,
 
     "typeRoots": ["node_modules/@types", "src/types"],
-    "lib": ["es2021", "dom"],
+    "lib": ["es2022", "dom"],
```
seaerchin, line 42: required for `.at` to work

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -127,6 +136,44 @@ export const collectionRouter = router({
...
+            .executeTakeFirstOrThrow()
+            .catch((err) => {
+              if (get(err, "code") === PG_ERROR_CODES.uniqueViolation) {
+                throw new TRPCError({
+                  code: "CONFLICT",
+                  message: "A resource with the same permalink already exists",
+                })
+              }
+              throw err
+            })
```
seaerchin, line 168: we need to auto create collection index pagse going forward because this is relied on by our new collection tags feature (and by extension, collection categories)

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { useQueryParse } from "~/hooks/useQueryParse"
+import { trpc } from "~/utils/trpc"
+
+export const jsonFormsTaggedControlTester: RankedTester = rankWith(
+  JSON_FORMS_RANKING.ArrayControl,
+  and(schemaMatches((schema) => schema.format === "tagged")),
+)
+
+interface TaggedControlProps extends Omit<ControlProps, "data"> {
+  data: ArticlePagePageProps["tagged"]
```
dcshzj, line 24: thought: do you think it would be worth to not do this, but rather, stick to using `ControlProps` and in the `JsonFormsTaggedControl` renderer we use a type predicate function to do the narrowing to `ArticlePagePageProps["tagged"]`?

Reason is because while we control the schema and the tester for this form control, in theory it is possible for all types of pages to be using this form control (since they all come from the generic `<FormBuilder>` component. We are doing this for specific cases that have this field, so wanted to be extra safe in the handling of the data.

(don't **have** to do it btw, just raising a thought)

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
```diff
@@ -0,0 +1,71 @@
...
+export function JsonFormsTaggedControl({
+  data,
+  path,
+  description,
+  required,
+  handleChange,
+}: TaggedControlProps) {
+  const { siteId, linkId, pageId } = useQueryParse(collectionItemSchema)
+  // NOTE: Since this is only rendered inside a collection page or collection link,
+  // we should always have the `resourceId` specifier
```
dcshzj, line 36: issue(non-blocking): hmm hate to deliver bad news but we _might_ be reusing FormBuilder on site config, navbar and footer editing in the future, so this assumption might not entirely hold true. Are we able to default to having no tag values?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <VStack>
+      {tags?.map(({ id, label, options }) => {
+        const currentTagCategory = data?.find(({ id: tagId }) => tagId === id)
+
+        return (
+          <FormControl isRequired={required} gap="0.5rem">
+            <FormLabel description={description}>{label}</FormLabel>
+            <MultiSelect
```
dcshzj, line 51: question(non-blocking): just checking, creating new tag options or tag categories is out of scope yeah?

apps/studio/src/schemas/collection.ts
```diff
@@ -66,6 +66,11 @@ export const createCollectionSchema = z.object({
   parentFolderId: z.number().optional(),
 })
 
+export const getCollectionTagsSchema = z.object({
+  resourceId: z.number().min(1),
+  siteId: z.number().min(1),
+})
```
dcshzj, line 72: nit: think it's about time we centralise this, feels like this zod object combination is used very frequently

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -352,6 +399,38 @@ export const collectionRouter = router({
...
+        .where("siteId", "=", siteId)
+        .select("parentId")
+        .executeTakeFirstOrThrow()
+
+      const indexPage = await db
+        .selectFrom("Resource")
+        .where("type", "=", ResourceType.IndexPage)
+        .where("parentId", "=", parentId)
+        .select("id")
+        .executeTakeFirstOrThrow()
```
dcshzj, line 425: nit: I think we can combine these two queries, seems like the first one is a subquery of the second

packages/components/src/types/page.ts
```diff
@@ -10,6 +10,39 @@ import {
...
+const TagCategoriesSchema = Type.Object({
+  tagCategories: Type.Optional(Type.Array(TagCategorySchema)),
+})
+const TaggedSchema = Type.Optional(
+  // NOTE: This stores the `uuid` of the tag category
+  // to the array of the `uuid` of the values;
+  // we cannot just store a plain array of `uuid`
+  // because we need to render each category as a dropdown
+  Type.Array(
+    Type.Object({ id: TagValueSchema, values: Type.Array(TagValueSchema) }),
```
dcshzj, line 37: nit: can we rename these to be something like TagCategoryIdSchema and TagOptionIdSchema to distinguish their intent?

Or actually can make use of the `title` and `description` keys to elaborate more on each property of the schema, so someone reading the schema on [schema.isomer.gov.sg](https://schema.isomer.gov.sg) can make immediate sense of it

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { FormLabel, MultiSelect } from "@opengovsg/design-system-react"
+import { ArticlePagePageProps } from "@opengovsg/isomer-components"
+
+import { JSON_FORMS_RANKING } from "~/constants/formBuilder"
+import { collectionItemSchema } from "~/features/editing-experience/schema"
+import { useQueryParse } from "~/hooks/useQueryParse"
+import { trpc } from "~/utils/trpc"
+
+export const jsonFormsTaggedControlTester: RankedTester = rankWith(
+  JSON_FORMS_RANKING.ArrayControl,
```
dcshzj, line 19: nit: can give it a separate ranking constant, e.g. `JSON_FORMS_RANKING.TaggedControl`

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
```diff
@@ -0,0 +1,71 @@
...
+  and(schemaMatches((schema) => schema.format === "tagged")),
```
dcshzj, line 20: nit: don't need the `and` since there is only one here

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
dcshzj: As discussed offline, not worth doing now

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTaggedControl.tsx
seaerchin: double checked, the component doesn't render if the schema doesn't match the tester

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -352,6 +399,38 @@ export const collectionRouter = router({
...
+        )
+        .select("id")
+        .executeTakeFirstOrThrow()
+
+      const { content } = await getBlobOfResource({
+        db,
+        resourceId: indexPage.id,
+      })
+
+      return (content as unknown as CollectionPageSchemaType).page.tagCategories
```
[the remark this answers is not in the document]
seaerchin: actually in our exports to studio, should we just `SimplifyDeep` everything?

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -397,6 +399,88 @@ export const collectionRouter = router({
...
+            .select("parentId"),
+        )
+        .select("id")
+        .executeTakeFirst()
+
+      if (!indexPage) {
+        return []
+      }
+
+      const { draftBlobId, publishedVersionId } = await db
```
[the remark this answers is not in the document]
seaerchin: iirc i made this a function in the subsequent branch!

PR #1544, opened 2025-08-12
feat: add additional sort by last updated date

dcshzj:
## Problem

We sort articles by alphabetical order when they have the same published date, which causes confusion to site editors.

## Solution

**Breaking Changes**

**Features**:

- Use the last modified date as a tie breaker first before using the title.

## Before & After Screenshots
image

## Tests

- [ ] Create a new collection
- [ ] Create the first article with "Bbbbbb" and today's date as the article date
- [ ] Create the second article with "Cccccc" and today's date as the article date
- [ ] Verify on the built site that the ordering of the articles is "Cccccc" then "Bbbbbb"


adriangohjw:
bugbot run

adriangohjw (approved):

PR #1548, opened 2025-08-19
feat: introduce FormSG embed

dcshzj:
## Problem

Closes ISOM-1930.

## Solution

**Breaking Changes**

**Features**:

- Introduce new FormSG embed component.

## Tests

- [ ] Add a new FormSG embed component
- [ ] Paste the following sample embed code:

```
<div
  style="
    font-family: Sans-Serif;
    font-size: 15px;
    color: #000;
    opacity: 0.9;
    padding-top: 5px;
    padding-bottom: 8px;
  "
>
  If the form below is not loaded, you can also fill it in at
  <a href="https://form.gov.sg/685a2374f7482ad8c494f42d">here</a>.
</div>

<!-- Change the width and height values to suit you best -->
<iframe
  id="iframe"
  src="https://form.gov.sg/685a2374f7482ad8c494f42d"
  style="width: 100%; height: 500px"
></iframe>

<div
  style="
    font-family: Sans-Serif;
    font-size: 12px;
    color: #999;
    opacity: 0.5;
    padding-top: 5px;
  "
>
  Powered by <a href="https://form.gov.sg" style="color: #999">Form</a>
</div>
```

- [ ] Verify that you are able to see the form embedded correctly in the preview
- [ ] Verify that you cannot add any random things into the insert embed code


apps/studio/next.config.mjs
```diff
@@ -37,14 +37,17 @@ const ContentSecurityPolicy = `
     'self'
     https://intercom-sheets.com
     https://www.intercom-reporting.com
-    https://player.vimeo.com
     https://fast.wistia.net
     https://www.google.com
-    https://www.youtube.com
```
adriangohjw, line 43: to confirm: removing `https://www.youtube.com` is not breaking change?

packages/components/src/utils/validation.ts
```diff
@@ -21,6 +21,28 @@ export const REF_HREF_PATTERN =
...
 
+// Validation for form-related embed URLs
+export const isValidFormSGEmbedUrl = (url: string) => {
+  if (!url) {
+    return false
+  }
+
+  try {
+    const urlObject = new URL(url)
+    return urlObject.hostname === "form.gov.sg"
```
adriangohjw, line 32: nit: im not sure if there are other valid formsg url that aren't actually forms, so might wanna check for the slug

packages/components/src/utils/validation.ts
```diff
@@ -21,6 +21,28 @@ export const REF_HREF_PATTERN =
...
+  } catch (_) {
+    return false
+  }
+}
+
+export const FORMSG_EMBED_URL_REGEXES = {
+  formsg: "^https://form\\.gov\\.sg/[a-z0-9]*$",
+} as const
```
adriangohjw, line 40: should be `FORMS_` not `FORMSG_`

adriangohjw:
bugbot run

apps/studio/next.config.mjs
dcshzj: Hmm yeah it won't be. I made this change to sync both the CSP here and in our infra repo, plus our component explicitly replaces it with the nocookie version:

https://github.com/opengovsg/isomer/blob/c40ec4b519775f35ac3b690657dddb568e9840f0/packages/components/src/templates/next/components/complex/Video/Video.tsx#L7-L9

packages/components/src/utils/validation.ts
dcshzj: Hmm tbh this one gets a bit weird, cos the component is specifically for FormSG but our map/video components are more generic so the pattern we use here is more suitable for generic ones. For now I think we are not planning to embed other types of forms (e.g. Typeform) so hence decided to go with specifically FormSG here to standardise the naming

packages/components/src/utils/validation.ts
dcshzj: Yeah I'm not too sure also but was thinking of doing the same treatment here as with Maps.gov.sg to do no validation. Worst case is people embed weird stuff but it is still scoped to form.gov.sg so I think it's ok

packages/components/src/utils/validation.ts
```diff
@@ -21,6 +21,28 @@ export const REF_HREF_PATTERN =
...
+
+export const FORMSG_EMBED_URL_PATTERN = Object.values(FORMSG_EMBED_URL_REGEXES)
+  .map((re) => `(${re})`)
+  .join("|")
```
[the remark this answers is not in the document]
dcshzj: Why do I feel this was a summary of what @adriangohjw wrote LOL

packages/components/src/utils/validation.ts
adriangohjw: yeah though since we have `fomsg` in the key, i thought it will be more consistent to have it be `FORMS_`, to standardize with MAPS and VIDEOS, otherwise feels slightly weird to do `FORMSG_EMBED_URL_REGEXES`.formsg`

adriangohjw (approved):

packages/components/src/utils/validation.ts
adriangohjw: anyway i approve first, this is not really a blocker for me

PR #1549, opened 2025-08-20
fix: prevent CollectionLink from being linked to

dcshzj:
## Problem

Closes ISOM-2041.

## Solution

**Breaking Changes**

**Bug Fixes**:

- CollectionLinks are excluded from the query to get the children of a particular folder or collection.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [ ] Edit any page in Studio and use any component that involves a link (e.g. Text component)
- [ ] Insert a link to open the resource selector modal
- [ ] Navigate to a collection that contains CollectionLinks. Verify that the CollectionLinks do not appear as options to be selected in the resource selector modal


apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -230,13 +230,15 @@ export const resourceRouter = router({
...
+          .where("Resource.type", "!=", ResourceType.CollectionLink)
           .where("Resource.siteId", "=", Number(siteId))
           .$narrowType<{
             type: Exclude<
               ResourceType,
               | typeof ResourceType.RootPage
               | typeof ResourceType.FolderMeta
               | typeof ResourceType.CollectionMeta
+              | typeof ResourceType.CollectionLink
             >
```
adriangohjw, line 242: should we take this chance to instead whitelist `=` instead of blacklist `!=` what should be returned?

since adding things that they should not add is worse than not being able to find things they should add

adriangohjw (approved):

PR #1553, opened 2025-08-21
Feat/nextjs turbo build

adriangohjw:
## Problem

nextjs v15.5 released with beta release for turbopack production build https://nextjs.org/blog/next-15-5#turbopack-builds-beta

this reduces build time, thus saving us cost while improving experience of site owners

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- upgrade template's nextjs version to v15.5 -> testing on staging for a site with 1.5k pages on MEDIUM, reduces time from 100s to 80s. This is supposed to see larger impact for sites with larger number of cores

> [!NOTE]
> while nextjs says this reduces JS bundle size, this has yet to be verified on our end. Will verify this before release

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6098,9 +6098,9 @@
       }
     },
     "node_modules/@emnapi/runtime": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.3.1.tgz",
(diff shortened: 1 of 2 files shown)


adriangohjw:
sadness, it seems like JS chunks size (esp. first load) is larger

| before | after |
| - | - |
| image | image |

adriangohjw:
shutting down for now to not overwhelm open PRs

PR #1556, opened 2025-08-22
Feat/collectionblock highlighted categories

adriangohjw:
## Problem

Some agencies only want to highlight certain categories of a collection in collectionblock but unable to do so, thus they aren't using it and is manually updating it

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add an option to pass in `highlightedCategories` into the filter when fetching the collection items

> [!NOTE]
> keeping it hidden from studio from the time being as we should instead let them select from a list of collection tags, which is after we migrate category over 

---

> [!NOTE]
> Adds optional highlighted categories to CollectionBlock and filters collection items by those categories.
> 
> - **CollectionBlock**:
>   - Add optional `highlights` config with `categories` in `interfaces/complex/CollectionBlock` (hidden for now).
>   - Wire `highlights.categories` through `CollectionBlock.tsx` to `getCollectionPages`.
> - **Collection utils**:
>   - `getCollectionPages`: accept `categories` and pass to `getCollectionItems`.
>   - `getCollectionItems`: add `categories` param; filter items by category (defaults to include all when empty); preserve existing sort and mapping.
> - **Tests**:
>   - Extend `getCollectionPages` tests to cover filtering by highlighted categories.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c2c6b9254ea9586d57cd9bc7fbee1a75f41d2673. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
```diff
@@ -51,9 +53,11 @@ export const getCollectionItems = ({
...
     .filter(
       (item) =>
-        item.layout === "file" ||
-        item.layout === "link" ||
-        item.layout === "article",
+        (item.layout === "file" ||
+          item.layout === "link" ||
+          item.layout === "article") &&
+        (categories.length === 0 ||
+          categories.includes(item.category || CATEGORY_OTHERS)),
```
dcshzj, line 60: question: how urgent is this feature needed ah, cos I think there will be some changes to the logic surrounding the category and this change might make the switchover to the new style categories a little harder

If we can delay this to after/during the collection index page revamp, I think would be better?

seaerchin (commented):
lgtm codewise, but this might become an issue next time, so i think we should find a way to track it. 

not sure if growthbook will work here

adriangohjw:
close for now :(

PR #1558, opened 2025-08-22
Feat/footer add flickr and threads

adriangohjw:
## Problem

basic 5 mins feature request https://www.notion.so/opengov/Support-flickr-threads-whatsapp-telegram-icon-in-footer-social-media-20d77dbba788805eba5cdd3cd7dd421c?source=copy_link

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- add support for threads and flickr

### DIFF
--- a/packages/components/src/interfaces/internal/Footer.ts
+++ b/packages/components/src/interfaces/internal/Footer.ts
@@ -13,6 +13,8 @@ export const SocialMediaTypes = [
   "github",
   "tiktok",
   "whatsapp",
+  "flickr",
+  "threads",
 ] as const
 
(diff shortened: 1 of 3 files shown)


dcshzj (approved):
Ok with this PR, but I thought we had concerns about all these images being bundled for all sites regardless of whether it is being used, so this will have a negative performance impact

adriangohjw:
> Ok with this PR, but I thought we had concerns about all these images being bundled for all sites regardless of whether it is being used, so this will have a negative performance impact

yeah but it wasn't an issue i can replicate for all sites and builds - sometimes they got loaded in a separate bundle, and sometimes they aren't loaded at all

PR #1561, opened 2025-08-28
feat: introduce collapsible sidenav and refactor layouts

dcshzj:
## Problem

Prep work for doing navbar/footer/site settings pages.

## Solution

**Breaking Changes**

**Features**:

- Added a new collapsible sidenav layout.

**Improvements**:

- Renamed the layouts to be clearer on their purposes
- Refactored the two layouts that display the existing sidebar such that they come from the same place, as previously there was significant code duplication.

## Before & After Screenshots
image

## Tests

- [ ] Existing layouts should not change
- [ ] Going to `/sites/1/navbar` should show you how the collapsible sidenav will look like


seaerchin (commented):
some of the components look like they're prop-drilling and doing a repeated check on whether the `onClick` exists, which is fine for now; however, it feels abit odd that we define a context which does this (provide the on click) but yet each of the layout still needs to do this check.

apps/studio/src/contexts/CmsSidenavContext.tsx
seaerchin, line 1: nit: please remove as the corresponding hook is never called

PR #1563, opened 2025-08-28
feat(scheduled-publishing): Add BE endpoint for scheduled publishing

gautammmanek:
## Problem

Adds initial scheduled publishing endpoint, as part of the https://www.notion.so/opengov/Scheduled-Publishing-23f77dbba7888089a4eecd375faf350a?source=copy_link RFC

## Solution

1. Add initial schedulePageSchema, which is further edited in #1567, as input to the schedulePage endpoint
2. Add schedulePage endpoint, with (1) checks the user perms, (2) verifies the page is not already scheduled and (3) adds an audit log call with a special resource type ResourceSchedule to indicate a schedule change for a resource
3. Rewrite ResourceEventLogProps as a mapped type, which associates the event type with the resource expected (instead of just throwing everything into a single union)
4. Add integration tests for success/failure case

**Breaking Changes**

  - Details ...
**Migration needs to be run to add the scheduledAt nullable column**

## Tests

1. Integration test added to ensure the scheduledAt time is set correctly, and the audit log is created with the ResourceSchedule event type
2. Integration test added to ensure providing a scheduled timestamp in the past leads to an error being thrown


apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -332,7 +336,72 @@ export const pageRouter = router({
...
+            before: resource,
+            after: { ...resource, scheduledAt: input.scheduledAt },
+          },
+          eventType: AuditLogEvent.ResourceSchedule,
+        })
+      })
+      await sendScheduledPageEmail({
+        scheduledAt: input.scheduledAt,
+        recipientEmail: by.email,
+      })
```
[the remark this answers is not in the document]
gautammmanek: Nope, this shouldn't be inside the DB transaction, since it's a blocking side-effect. If we want to have at-least-once guarantees to account for edge cases, we can have a separate outbox table to track emails

adriangohjw:
bugbot run

apps/studio/src/features/mail/templates/templates.ts
```diff
@@ -74,6 +76,20 @@ export const loginAlertTemplate = (
...
 
+export const schedulePageTemplate = (
+  data: SchedulePageTemplateData,
+): EmailTemplate => {
+  const { recipientEmail, scheduledAt } = data
+  return {
+    subject: `[Isomer Studio] You scheduled a page to be published`,
+    body: `<p>Hi ${recipientEmail},</p>
+<p>You’ve scheduled a page to be published at a later time. Your page will publish at: <strong>${format(scheduledAt, "MMMM d, yyyy hh:mm a")}</strong>.</p>
+<p>Log in to Isomer Studio to change or cancel this.</p>
```
adriangohjw, line 87: nit: can we hyperlink `Isomer Studio` or add a hyperlink after that so they can click it? better UX

apps/studio/src/features/mail/templates/templates.ts
```diff
@@ -74,6 +76,20 @@ export const loginAlertTemplate = (
   }
 }
 
+export const schedulePageTemplate = (
+  data: SchedulePageTemplateData,
+): EmailTemplate => {
+  const { recipientEmail, scheduledAt } = data
+  return {
+    subject: `[Isomer Studio] You scheduled a page to be published`,
```
adriangohjw, line 84: nit: can we add the `resource.title` here in the subject? similar to `publishAlertContentPublisherTemplate` - better UX IMO

apps/studio/src/schemas/page.ts
```diff
@@ -56,6 +65,57 @@ export const reorderBlobSchema = z.object({
   ),
 })
 
+// Schema for scheduling a page (publishAt is derived from publishDate and publishTime)
+// On the client side, publishDate is selected via a date picker
+// and publishTime is selected via a time picker (in HH:mm format)
+export const MINIMUM_SCHEDULE_LEAD_TIME_MINUTES = 10
```
adriangohjw, line 71: nit: this isn't being used outside the file so should not have `export`

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -24,31 +24,40 @@ type FullResource =
...
-}
-interface ResourceDeleteDelta {
-  before: FullResource
-  after: null
-}
-interface ResourceUpdateDelta {
-  before: FullResource
-  after: FullResource
+// map each event type to its delta type
+interface EventDeltaMap {
```
adriangohjw, line 28: question: should this be something more specific to `Resource`? currently audit logs exist for other events beyond Resource, so this might not be as clear. Perhaps something like this might be clearer
```ts
interface ResourceEventDeltaMap {
  Create: {},
  Update: {},
  // ...
}
```

apps/studio/tests/integration/helpers/iron-session.ts
```diff
@@ -74,8 +74,8 @@ export const createMockRequest = (
 
   return {
     ...innerContext,
-    req: req as unknown as NextApiRequest,
-    res: res as unknown as NextApiResponse,
+    req: mocks.req as unknown as NextApiRequest,
+    res: mocks.res as unknown as NextApiResponse,
```
adriangohjw, line 78: curious: any reason for this change?

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2015,4 +2016,72 @@ describe("page.router", async () => {
       })
     })
   })
+  describe("scheduled publishing", () => {
```
adriangohjw, line 2019: can add 2 basic auth+authorization check?
1. should throw 401 if not logged in
2. should throw 403 if user does not have create access to site

can copy from other test cases - mainly to give more confidence in terms of security

---

can also add another test case for it throwing error if user tries to schedule within 10 minutes

adriangohjw: PS, can add line break between each section (Arrange, Act, Assert)? for readability purposes. thank you!

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2015,4 +2016,72 @@ describe("page.router", async () => {
...
+      // This should throw an error on the frontend or backend based on the value specified in MINIMUM_SCHEDULE_LEAD_TIME_MINUTES
+      await expect(
+        caller.schedulePage({
+          siteId: site.id,
+          pageId: Number(expectedPage.id),
+          publishDate: scheduledAtInPast,
+          publishTime: format(scheduledAtInPast, "HH:mm"),
+        }),
+      ).rejects.toThrowError()
+    })
```
adriangohjw, line 2085: can add Act and Assert?

adriangohjw: can add assertion for no audit logs created?

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
adriangohjw, line 2019: can rename it from `scheduled publishing` to the endpoint name itself `schedulePage` for consistency

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -332,7 +336,72 @@ export const pageRouter = router({
         return actualBlocks
       })
     }),
-
+  schedulePage: protectedProcedure
+    .input(schedulePageSchema)
+    .mutation(async ({ ctx, input }) => {
+      await bulkValidateUserPermissionsForResources({
+        siteId: input.siteId,
+        action: "update",
```
adriangohjw, line 344: issue: we should check if they have perms to `publish` not `update`

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -332,7 +336,72 @@ export const pageRouter = router({
         return actualBlocks
       })
     }),
-
+  schedulePage: protectedProcedure
+    .input(schedulePageSchema)
+    .mutation(async ({ ctx, input }) => {
```
adriangohjw, line 341: for consistency, can deconstruct the props here? `input: { ... }`

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -332,7 +336,72 @@ export const pageRouter = router({
...
+        // TODO: add logic to add the job to the job queue
+        await logResourceEvent(tx, {
+          siteId: input.siteId,
+          by,
+          delta: {
+            before: resource,
+            after: { ...resource, scheduledAt: input.scheduledAt },
+          },
+          eventType: AuditLogEvent.ResourceSchedule,
+        })
```
adriangohjw, line 398: instead of constructing the after ourselves `after: { ...resource, scheduledAt: input.scheduledAt }`, can we have the output returned from `updatePageById` be used instead? i think more accurate logging of before-after

adriangohjw (changes requested):

apps/studio/src/features/mail/templates/templates.ts
gautammmanek: Cool, added constructStudioRedirect() to handle the redirection to studio from the email

apps/studio/src/server/modules/page/page.router.ts
gautammmanek: Cool, I've modified updatePageById to return the result of the update & fixed the tests accordingly

apps/studio/src/server/modules/audit/audit.service.ts
gautammmanek: Good call, thanks!

apps/studio/tests/integration/helpers/iron-session.ts
gautammmanek: Not a 100% sure why, but ESlint kept throwing an unsafe destructuring error on this this

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
gautammmanek: All done, thanks!

gautammmanek:
Kept generated enums & types, and added CancelScheduleResource event type migration for audit logs to this PR so we only need to run one migration on prod

apps/studio/src/schemas/page.ts
```diff
@@ -56,6 +65,57 @@ export const reorderBlobSchema = z.object({
...
+    // combine publishDate and publishTime into a single Date object
+    const [hours, minutes] = publishTime.split(":").map(Number)
+    return {
+      ...rest,
+      scheduledAt: set(publishDate, {
+        hours,
+        minutes,
+        seconds: 0,
+        milliseconds: 0,
+      }),
```
dcshzj, line 92: question: wanted to check to be super clear, how do we know that the `publishDate` is in the correct time zone? Initial glance seems to be taking the system time zone which is UTC

apps/studio/src/schemas/page.ts
```diff
@@ -56,6 +65,57 @@ export const reorderBlobSchema = z.object({
...
+  })
+  .superRefine(({ scheduledAt }, ctx) => {
+    const earliestScheduleTime = add(new Date(), {
+      minutes: MINIMUM_SCHEDULE_LEAD_TIME_MINUTES,
+    })
+    if (isBefore(scheduledAt, earliestScheduleTime)) {
+      // if the scheduled date is before the earliest allowable date, show error on publishDate
+      if (startOfDay(scheduledAt) < startOfDay(earliestScheduleTime)) {
+        ctx.addIssue({
+          path: ["publishDate"],
```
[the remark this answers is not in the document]
gautammmanek: Fixed!

apps/studio/src/schemas/page.ts
gautammmanek: Ah, z.date() itself is timezone-aware, ie it selects the user's local timezone when the user is using the datepicker. If the client is in Singapore, won't be an issue, since the set() functionality sets the time according to the user's local timezone, not the UTC. Ergo, since everything is done in Sg time, should be fine for an sg client

HOWEVER, an issue here is if the client is outside singapore. Since the datepicker shows SGT +8, users will be expecting that whatever time they put in should be singapore time. However, it will incorrectly be their local timezone. Quick fix for this is to use fromZonedTime() to assert that whatever date provided IS IN SGT

apps/studio/src/schemas/page.ts
```diff
@@ -56,6 +66,52 @@ export const reorderBlobSchema = z.object({
...
+      const parsed = parse(time, "HH:mm", new Date())
+      return isValid(parsed) && format(parsed, "HH:mm") === time
+    }),
+  })
+  .transform(({ publishDate, publishTime, ...rest }) => {
+    // combine publishDate and publishTime into a single Date object
+    const [hours, minutes] = publishTime.split(":").map(Number)
+    return {
+      ...rest,
+      scheduledAt: fromZonedTime(
```
gautammmanek, line 88: Explicitly setting the time here to be zoned time in SGT

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2015,4 +2016,111 @@ describe("page.router", async () => {
...
+        }),
+      )
+    })
+    it("should set a scheduledAt time correctly for a page", async () => {
+      // Arrange
+      const now = new Date()
+      const { site, page: expectedPage } = await setupPageResource({
+        resourceType: "Page",
+      })
+      await setupAdminPermissions({
```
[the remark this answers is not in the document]
gautammmanek: Added in #1567 already!

PR #1565, opened 2025-08-28
feat: setup base navbar and footer editing pages

dcshzj:
## Problem

Second part of setting up the base for navbar and footer editing.

## Solution

**Breaking Changes**

**Features**:

- Added the navbar and footer routes as a base for editing those two parts of the site.
- Added a callback function inside the Frame component to allow running functions after the iframe preview is loaded.
- Added a `waitForElement` utility function to wait for the page in the iframe to fully render before doing any further actions.
- Added the ability to automatically open the navigation bar when going to the navbar editing page.
- Added the ability to automatically scroll to the footer when going to the footer editing page.

## Demo

https://github.com/user-attachments/assets/b7463c83-f470-4bdf-9746-79a78ee1560d

https://github.com/user-attachments/assets/ececc145-59fa-4b5e-a936-bc991ed29936

## Tests

- [ ] Navigate to Studio and go into any of the sites
- [ ] Change the URL to something like `/sites/1/navbar`, verify that the navigation bar editing page appears, and the navbar automatically appears in the preview
- [ ] Change the URL to somethng like `/sites/1/footer`, verify that the footer editing page appears, and the preview automatically scrolls to the footer

### DIFF
--- a/apps/studio/src/pages/sites/[siteId]/navbar.tsx
+++ b/apps/studio/src/pages/sites/[siteId]/navbar.tsx
@@ -17,7 +17,7 @@ const NavbarSettingsPage: NextPageWithLayout = () => {
         <p>Navigation menu</p>
       </GridItem>
       <GridItem colSpan={1}>
-        <EditNavbarPreview siteId={siteId} />
+        <EditNavbarPreview siteId={Number(siteId)} />
       </GridItem>
     </Grid>
   )


seaerchin (commented):
not an in depth code review, more a few questions wrt approach. the current one is fine, i think we can live iwth it as is and be happy with it too HAHA

1. actually the `Frame` component exposes a `ref` prop - what's happening now is the `Preview` component dictates the shape of the callback (albeit the `Document` passed to the callback has the same function as the `ref`)
2. by using the `ref` prop, we can extract our `handleIframeMount` into a generalised hook taking a `ref` (eg: `useGetElementBySelector(baseElement: HTMLElement, selector: string): HTMLElement`) -> the benefit here is that we can extend this for other items in the `Iframe`

**just to reiterate** - no issues w/ the approach taken here, only possible pitfall is that the `Observer` might have perf issues on large pages (since it watches on every dom mutation), but apart from that this lgtm

PR #1566, opened 2025-08-29
feat(scheduled-publishing): add scheduled publishing frontend

gautammmanek:
## Problem

Implement the schedule publishing modal flow for users to choose between (1) publishing the page now or (2) publishing at a specific date/time. Add the following components:
1. BaseSelect: A wrapper around chakra-react-select used for the <TimeSelect /> component
2. TimeSelect: A time-picker dropdown with 30min granularity, which allows for selecting a suitable schedule start time
3. ScheduledPublishingModal contains logic for validating form based on publishClientSchema & schedulePublishClientSchema
4. Update PublishButton to display the modal when clicked, based on the feature flag ENABLE_SCHEDULED_PUBLISHING_FEATURE_KEY 

**Breaking Changes**

  - Details ...

Feature flag ENABLE_SCHEDULED_PUBLISHING_FEATURE_KEY enabled for dev, but NOT enabled for other environments

## Video recordings

With feature flag enabled:

https://github.com/user-attachments/assets/fe646ee1-f4a5-4da3-8d87-fc424544720b

With feature flag disabled (growthbook > enable-scheduled-publishing flag toggled to off):

https://github.com/user-attachments/assets/6863d5c6-7c30-452e-9513-b8bbbdc085aa

## Tests

Chromatic tests to be added in a follow-up PR

**New dependencies**:

- chakra-react-select used for the dropdown


apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,429 @@
...
+  selectedDate: Date,
+  earliestSchedule: Date,
+) => {
+  if (isSameDay(selectedDate, earliestSchedule)) {
+    return {
+      hours: earliestSchedule.getHours(),
+      minutes: earliestSchedule.getMinutes(),
+    }
+  }
+  return undefined
```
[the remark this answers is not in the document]
gautammmanek: Changed to return null instead of undefined

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,429 @@
...
+}: {
+  earliestAllowableTime: { hours: number; minutes: number } | undefined
+}) => {
+  const optionsToShow = useMemo(() => {
+    // the array of pills to display in the section
+    const quickSelectTimes: { hours: number; minutes: number }[] = [
+      { hours: 0, minutes: 0 },
+      { hours: 9, minutes: 0 },
+      { hours: 13, minutes: 0 },
+      { hours: 17, minutes: 0 },
```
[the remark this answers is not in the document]
gautammmanek: Extracted to top of file

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,429 @@
...
+    defaultValues: {
+      publishMode: PublishMode.NOW,
+      pageId,
+      siteId,
+    },
+  })
+  const publishMode = methods.watch("publishMode")
+  // Validate form when publish mode changes or when the date/time selected changes
+  // so the banner can be shown to the users. Do NOT use trigger() since that
+  // triggers validation and causes error messages to show up
```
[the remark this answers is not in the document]
gautammmanek: Fixed

adriangohjw:
bugbot run

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/index.tsx
adriangohjw, line 1: should be `index.ts` instead of `index.tsx`

apps/studio/src/lib/growthbook.ts
```diff
@@ -1,5 +1,7 @@
 import type { GrowthBook } from "@growthbook/growthbook-react"
 
+export const ENABLE_SCHEDULED_PUBLISHING_FEATURE_KEY =
+  "enable-scheduled-publishing"
```
adriangohjw, line 4: question: seems like we are feature flagging this on a app-level instead of agency-level. if so, this seems to only be useful in quickly disabling this feature in the event it is buggy - is that the intention?

Otherwise, if the goal is to observe usage for specific agencies before mass rollout, perhaps we should switch it to a agency-level rollout (via the siteId in the URL)

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
adriangohjw, line 1: (personal) nit: this file feels a bit long - can move the other components e.g. `SchedulePublishDetails` into their own file in the same folder, and import into here? something like `UserPermissionModal`. IMO feels easier to trace

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,427 @@
...
+    </VStack>
+  )
+}
+
+const QuickSelectTimeSection = ({
+  earliestAllowableTime,
+}: {
+  earliestAllowableTime: { hours: number; minutes: number } | null
+}) => {
+  const optionsToShow = useMemo(() => {
```
adriangohjw, line 295: question: do we need useMemo here? since `earliestAllowableTime` is passed in and itself is memoized by the parent. Seems to be a rather trivial filtering action too

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,427 @@
...
+}
+
+/**
+ * if the date provided is equal to the earliestSchedule's date, the earliest allowable time should be set to the FIRST
+ * time slot after the current minimum allowable time
+ * @param selectedDate Date selected inside the datepicker
+ * @param earliestSchedule Earliest schedule time, based on the current date and MINIMUM_SCHEDULE_LEAD_TIME_MINUTES
+ * @returns
+ */
+const getEarliestAllowableTime = (
```
adriangohjw, line 201: can move into a `utils` folder in this file and write a simple unit test for this?

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,427 @@
...
+    <Modal onClose={onClose} {...rest}>
+      <form
+        onSubmit={methods.handleSubmit(
+          (res: z.output<typeof publishClientSchema>) => {
+            // publish immediately if publish mode is now, else schedule publish
+            if (res.publishMode === PublishMode.NOW) {
+              onPublishNow(res.pageId, res.siteId)
+            } else {
+              schedulePageMutation(res)
+            }
```
adriangohjw, line 127: nit - while we currenty only have `PublishMode.NOW` and `PublishMode.SCHEDULED`, perhaps we can use a switch statement here with exhaustic check, so that if/when we add another method, it will throw type error

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,427 @@
...
+    resetField,
+    watch,
+    control,
+    formState: { errors },
+  } = useFormContext<z.input<typeof schedulePublishClientSchema>>()
+
+  const publishDate = watch("publishDate")
+  const publishTime = watch("publishTime")
+  // this is the earliest date and time that the user can schedule a publish for
+  const { earliestAllowableTime, earliestSchedule } = useMemo(() => {
```
adriangohjw, line 225: question: what happens when user open this modal and then AFK and come back after 10 mins? are we hoping to not refresh the UI here and solely rely on backend to validate? _(to be clear, im okay with it since its trivial, but just want to confirm that its intended)_

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,427 @@
...
+          {format(
+            parse(getValues("publishTime"), "HH:mm", new Date()),
+            "hh:mm a",
+          )}
+        </Text>
+        , Singapore Standard Time, on{" "}
+        <Text display="inline" textStyle="subhead-2">
+          {format(getValues("publishDate"), "MMMM d, yyyy")}
+        </Text>
+        . Changes will be live on your site approximately 15-30 minutes after
```
adriangohjw, line 422: this seems inconsistent with the 5-10mins timing stated in "publish now" - any reason? since they should be similar

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
```diff
@@ -0,0 +1,427 @@
...
+              onClick={onClose}
+              variant="clear"
+              color="base.content.strong"
+            >
+              No, don't publish
+            </Button>
+            <Button type="submit" isLoading={isPublishingNow || isScheduling}>
+              {publishMode === PublishMode.NOW
+                ? "Publish now"
+                : "Schedule publish"}
```
adriangohjw, line 185: re: same comment about switch-case with exhaustive check

adriangohjw (commented):
## clarifying current behaviour

seems like current behaviour means:
1. there's no indication that it is currently scheduled
2. we can update/"Publish now" when it is already scheduled -> i rmb the RFC mentioned that should not be allowed? if we gonna allow it, we would also have to update "publish now" endpoint to clear `scheduledAt` field

## adding `index.ts`

also can add a `index.ts` for `components/Select`? consistency purpose 🙏 

## storybook

can add storybook for the modal? we use storybook as some sort of visual testing to prevent regression + make it easier for Design to review

apps/studio/src/lib/growthbook.ts
gautammmanek: Hmm, interesting, yeah let me add this in a follow up PR! TBH might stick to using a site-specific whitelist and an overall kill switch, and then we can deprecate both once confirmed everything is fine

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
gautammmanek: Sure, let me extract some stuff out of here to make it easier to trace

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/ScheduledPublishingModal.tsx
gautammmanek: Fair enough, simple filtering here doesn't need useMemo

apps/studio/src/lib/growthbook.ts
adriangohjw: think we can have single source of truth by using site-specific whitelist - we can just set it as empty array if we want to use it as a killswitch. will be simpler to implement and manage

gautammmanek: Alright!

apps/studio/src/features/editing-experience/components/ScheduledPublishingModal/QuickSelectTimeSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Badge } from "@opengovsg/design-system-react"
+import { format } from "date-fns"
+import { useFormContext } from "react-hook-form"
+
+import type { schedulePublishClientSchema } from "~/schemas/schedule"
+
+/**
+ * The times to display as 'quick-select-times' under the date/time picker
+ */
+const QUICK_SELECT_TIMES: { hours: number; minutes: number }[] = [
```
adriangohjw, line 12: super nit: `{ hours: number; minutes: number }` can be extracted into a type to be reused below in `QuickSelectTimeSection`

adriangohjw (approved):

PR #1569, opened 2025-09-01
refactor(components): derive `tags` from the `tagged` property

seaerchin:
## Problem
- We need to update the build script as well as the components to make use of the new `tagged` and `tagCategories` so that they can be used by our components.

## Solution
- make the `getCollectionItems` take in `tagCollections` then return either the `tags` derived from `tagCollections` or the one from `tags` on the page. we **shouldn't fold** `tagged` into `tags` because these mean 2 different things. the `tags` are teh actual data, whereas the `tagged` is just a uuid pointer
- for publishing, take a lazy approach and put `tagged` and `tagCategories` onto the `collectionpagePageProps`

### DIFF
--- a/packages/components/src/types/page.ts
+++ b/packages/components/src/types/page.ts
@@ -41,7 +41,9 @@ const TagCategorySchema = Type.Composite([
 // NOTE: can be optional because the categories might not exist
 const TagCategoriesSchema = Type.Object(
   {
-    tagCategories: Type.Optional(Type.Array(TagCategorySchema)),
+    tagCategories: Type.Optional(
+      Type.Array(TagCategorySchema, { format: "hidden" }),
+    ),
   },
   { format: "hidden" },
 )


packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
```diff
@@ -11,13 +12,15 @@ export interface GetCollectionItemsProps {
...
+  tagCategories?: CollectionPagePageProps["tagCategories"]
 }
 
 export const getCollectionItems = ({
   site,
   permalink,
   sortBy,
   sortDirection,
+  tagCategories,
 }: GetCollectionItemsProps): AllCardProps[] => {
```
dcshzj, line 24: nit: can add a TODO comment that we can do the optimisation on getting the filters directly from the UUID rather than doing another pass of the collection items using the `getAvailableFilters`

PR #1570, opened 2025-09-03
feat: introduce script to automatically get audit logs

dcshzj:
## Problem

The way we have been running audit logs is unsustainable and easily prone to be forgotten.

## Solution

**Breaking Changes**

**Features**:

- Add a script that can be used to automatically retrieve the audit logs required by agencies, which can be run at the start of the month for the sites that require them.

### DIFF
--- a/apps/studio/prisma/scripts/getAuditLogs.ts
+++ b/apps/studio/prisma/scripts/getAuditLogs.ts
@@ -17,7 +17,7 @@ const SITES_WITH_AUDIT_LOGS = [
   48, // muis.gov.sg
   50, // knowledgehub.clc.gov.sg
   53, // clc.gov.sg
-  63, // ipos.gov.sg
+  61, // ipos.gov.sg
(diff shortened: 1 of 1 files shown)


adriangohjw:
as discussed on slack:

> The state of the audit logs requests now is quite messy in the sense that there is no records of which sites requires what kind of audit logs. Different migrators are in charge of different sites, so they have visibility only over their own sites, and that means the requests become very ad-hoc. We may have promised agencies that we will send the audit logs, but there is no mechanism right now to ensure that happens. It was clear in August that nobody really knows which site needs the audit logs, so we now centralise this inside the script so there is accountability and proper tracking.

> Separately, we now have 7 sites that we need to do the audit logs for, and 2 more coming on board starting November, so the copy-pasting would need to be done 9x2=18 times and can be quite tedious. With the script we can do one shot for all the site, less error-prone.

seaerchin (approved):

PR #1577, opened 2025-09-08
Chore/restrict to migrators

seaerchin:
## Problem
1. we should auto generate the uuid key and hide it from migrators
2. the tag categories should be hidden from everyone who is not a migrator 

## Solution
1. for tag categories, we call the `useIsIsomerAdmin` hook then check that the user is part of the team 
2. for uuid, we hide the value (can still see via admin mode) but our rangers should interact via the dropdown

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsUuidControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsUuidControl.tsx
@@ -19,7 +19,10 @@ export const JsonFormsUuidControl = ({
   handleChange,
   path,
 }: ControlProps) => {
-  const uuid = useMemo(() => data || crypto.randomUUID(), [path])
+  const uuid = useMemo(
+    () => (typeof data === "string" ? data : crypto.randomUUID()),
+    [path],
+  )
   useEffect(() => {
     handleChange(path, uuid)
   }, [])


dcshzj (approved):

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsUuidControl.tsx
```diff
@@ -0,0 +1,34 @@
...
+  JSON_FORMS_RANKING.UuidControl,
+  schemaMatches((schema) => schema.format === "uuid"),
+)
+
+export const JsonFormsUuidControl = ({
+  data,
+  handleChange,
+  path,
+}: ControlProps) => {
+  const uuid = useMemo(() => data || crypto.randomUUID(), [path])
```
adriangohjw, line 22: question: why is dependency `[path]` when its not being used and the variable being used is `data`?

packages/components/src/types/page.ts
```diff
@@ -39,14 +39,11 @@ const TagCategorySchema = Type.Composite([
...
-  {
-    tagCategories: Type.Optional(
-      Type.Array(TagCategorySchema, { format: "hidden" }),
-    ),
-  },
-  { format: "hidden" },
-)
+const TagCategoriesSchema = Type.Object({
+  tagCategories: Type.Optional(
+    Type.Array(TagCategorySchema, { format: "tag-categories" }),
```
adriangohjw, line 44: can make `tag-categories` a const? feel like easier to debug between components and studio

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsUuidControl.tsx
seaerchin: i derped, removed; shouldn't need it since `uuid` is already based off `path`

packages/components/src/types/page.ts
seaerchin: @dcshzj had this idea that components **should only export schemas/types** - that's the contract between studio as a client and components as the server. to that end, we should not be exporting `const` and other values from components.

rather, these are _studio_ constructs that are derived from the schema. instead of doing the export from `components`, we should define a file in studio (or a folder or smth) that basically defines and exports all these `const` or default values since they are either artificial constructs or inferred from the schema. 

wdyt about this approach @adriangohjw ?

PR #1579, opened 2025-09-09
Feat/standalone dgs searchabletable selfserve

adriangohjw:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/apps/studio/src/components/PageEditor/constants.ts
+++ b/apps/studio/src/components/PageEditor/constants.ts
@@ -4,6 +4,7 @@ import {
   DEFAULT_CHILDREN_PAGES_BLOCK,
   DYNAMIC_DATA_BANNER_TYPE,
   IMAGE_GALLERY_TYPE,
+  SEARCHABLE_TABLE_TYPE,
 } from "@opengovsg/isomer-components"
 
 // TODO: add in default blocks for remaining
@@ -263,6 +264,15 @@ export const DEFAULT_BLOCKS: Record<
       },
     ],
   },
+  [SEARCHABLE_TABLE_TYPE]: {
+    type: `${SEARCHABLE_TABLE_TYPE}`,
+    title: "Sample DGS Table",
+    dataSource: {
+      type: "dgs",
+      resourceId: "d_1ebdc22551ae034d076722aa671e7f1b", // hardcoded
+    },
+    headers: [],
+  },
 }
 
 export const BLOCK_TO_META: Record<
@@ -398,6 +408,12 @@ export const BLOCK_TO_META: Record<
     usageText:
       "Get mailing list sign-ups or quick feedback by embedding a form directly on your page.",
     imageSrc: "/assets/block-images/FormSG.png",
+  },
+  [SEARCHABLE_TABLE_TYPE]: {
+    // TODO: Add image source
+    label: "Searchable Table",
+    description: "Display a searchable table with data from a DGS dataset",
+    usageText: "Display a searchable table.",
   },
 }
 
--- a/apps/studio/src/components/PageEditor/constants.ts
+++ b/apps/studio/src/components/PageEditor/constants.ts
@@ -453,7 +453,7 @@ export const CONTENT_ALLOWED_BLOCKS: AllowedBlockSections = [
   },
   {
     label: "Add a new section",
-    types: ["infocards", "infocols", "keystatistics"],
+    types: ["infocards", "infocols", "keystatistics", SEARCHABLE_TABLE_TYPE],
   },
   { label: "Embed external content", types: ["map", "video", "formsg"] },
 ]
--- a/packages/components/src/templates/next/components/complex/SearchableTable/DGS/DGSSearchableTable.tsx
+++ b/packages/components/src/templates/next/components/complex/SearchableTable/DGS/DGSSearchableTable.tsx
@@ -11,6 +11,7 @@ import { useDgsData } from "~/hooks/useDgsData"
 import { SearchableTableClient } from "../shared"
 
 export const DGSSearchableTable = ({
+  type,
   dataSource: { resourceId, filters, sort },
   title,
   headers,
@@ -59,6 +60,7 @@ export const DGSSearchableTable = ({
 
   return (
     <SearchableTableClient
+      type={type}
       title={title}
       headers={labels}
       items={items}
--- a/apps/studio/public/assets/css/preview-tw.css
+++ b/apps/studio/public/assets/css/preview-tw.css
@@ -1459,6 +1459,10 @@ video {
   -webkit-line-clamp: 3;
 }
 
+.\!block {
+  display: block !important;
+}
+
 .block {
   display: block;
 }
@@ -1479,12 +1483,20 @@ video {
   display: inline-flex;
 }
 
+.\!table {
+  display: table !important;
+}
+
 .table {
   display: table;
 }
 
 .table-row {
   display: table-row;
+}
+
+.\!grid {
+  display: grid !important;
 }
 
 .grid {
@@ -4005,6 +4017,10 @@ video {
 
 .\[--color\:theme\(colors\.red\.700\)\] {
   --color: #b91c1c;
+}
+
+.\[a-zA-Z0-9\:\\\\-\\\\\._\$\] {
+  a-z-a--z0-9: \\-\\.$;
 }
 
 .\[dgs\:123456\] {
--- a/apps/studio/src/features/editing-experience/constants.ts
+++ b/apps/studio/src/features/editing-experience/constants.ts
@@ -4,6 +4,7 @@ import {
   COLLECTION_BLOCK_TYPE,
   DYNAMIC_DATA_BANNER_TYPE,
   IMAGE_GALLERY_TYPE,
+  SEARCHABLE_TABLE_TYPE,
 } from "@opengovsg/isomer-components"
 import {
   BiChevronDown,
@@ -54,4 +55,5 @@ export const TYPE_TO_ICON: Record<
   [COLLECTION_BLOCK_TYPE]: TbDatabase,
   [IMAGE_GALLERY_TYPE]: BiImages,
   childrenpages: BiListUl,
+  [SEARCHABLE_TABLE_TYPE]: TbDatabase,
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -35024,6 +35024,21 @@
     "tooling/typescript": {
       "name": "@isomer/tsconfig",
       "version": "0.0.0"
+    },
+    "tooling/template/node_modules/@next/swc-win32-ia32-msvc": {
+      "version": "14.2.30",
+      "resolved": "https://registry.npmjs.org/@next/swc-win32-ia32-msvc/-/swc-win32-ia32-msvc-14.2.30.tgz",
+      "integrity": "sha512-pVZMnFok5qEX4RT59mK2hEVtJX+XFfak+/rjHpyFh7juiT52r177bfFKhnlafm0UOSldhXjj32b+LZIOdswGTg==",
+      "cpu": [
+        "ia32"
+      ],
+      "optional": true,
+      "os": [
+        "win32"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
     }
   }
 }
--- a/packages/components/src/interfaces/complex/SearchableTable.ts
+++ b/packages/components/src/interfaces/complex/SearchableTable.ts
@@ -4,7 +4,10 @@ import { Type } from "@sinclair/typebox"
 import type { IsomerSiteProps, LinkComponentType } from "~/types"
 import { DgsDataSourceSchema, NativeDataSourceSchema } from "../integration"
 
+export const SEARCHABLE_TABLE_TYPE = "searchabletable"
+
 const BaseSearchableTableSchema = Type.Object({
+  type: Type.Literal(SEARCHABLE_TABLE_TYPE, { default: SEARCHABLE_TABLE_TYPE }),
   title: Type.Optional(
     Type.String({
       title: "Title",
--- a/packages/components/src/interfaces/complex/index.ts
+++ b/packages/components/src/interfaces/complex/index.ts
@@ -54,6 +54,7 @@ export {
   type ChildrenPagesProps,
 } from "./ChildrenPages"
 export {
+  SEARCHABLE_TABLE_TYPE,
   SearchableTableSchema,
   type SearchableTableProps,
   type SearchableTableClientProps,
--- a/packages/components/src/schemas/components.ts
+++ b/packages/components/src/schemas/components.ts
@@ -31,6 +31,8 @@ import {
   OrderedListSchema,
   ParagraphSchema,
   ProseSchema,
+  SEARCHABLE_TABLE_TYPE,
+  SearchableTableSchema,
   TableSchema,
   UnorderedListSchema,
   VideoSchema,
@@ -57,6 +59,7 @@ export const IsomerComplexComponentsMap = {
   logocloud: LogoCloudSchema,
   [COLLECTION_BLOCK_TYPE]: CollectionBlockSchema,
   [IMAGE_GALLERY_TYPE]: ImageGallerySchema,
+  [SEARCHABLE_TABLE_TYPE]: SearchableTableSchema,
 }
 
 export const IsomerNativeComponentsMap = {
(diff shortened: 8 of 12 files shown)


adriangohjw:
bugbot run

apps/studio/src/components/PageEditor/constants.ts
```diff
@@ -263,6 +264,13 @@ export const DEFAULT_BLOCKS: Record<
       },
     ],
   },
+  [SEARCHABLE_TABLE_TYPE]: {
+    type: `${SEARCHABLE_TABLE_TYPE}`,
+    dataSource: {
+      type: "dgs",
+      resourceId: "d_3c55210de27fcccda2ed0c63fdd2b352", // hardcoded
+    },
+  },
```
adriangohjw, line 273: adding for type completeness: otherwise not useded as it cannot be added as a component

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsAnyOfControl.tsx
```diff
@@ -83,9 +83,21 @@ export function JsonFormsAnyOfControl({
...
+    if (indexOfFittingSchema >= 0 && options[indexOfFittingSchema]) {
+      setVariant(options[indexOfFittingSchema].label)
+      return
+    }
+
+    // Fallback to first option
+    if (options[0]) {
+      setVariant(options[0].label)
+    }
+  }, [options, indexOfFittingSchema])
```
adriangohjw, line 100: handling the schema of having intersect + union
```
Type.Intersect(
  [
    BaseSchema,
    Type.Union([NativeSchema, DGSSchema]),
  ],
)
```

apps/studio/src/features/editing-experience/data/databaseLayoutPreview.json
```diff
@@ -0,0 +1,65 @@
...
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"],
+        ["Cell copy", "Cell copy"]
+      ]
+    }
```
adriangohjw, line 33: even though this feature is to allow agency users to add DGS component, for preview we can use native searchable table to 1) reduce additional DGS API call + 2) reduce latency (no need to wait for API)

packages/components/src/templates/next/components/complex/SearchableTable/DGS/DGSSearchableTable.stories.tsx
adriangohjw, line 1: mostly refactoring into a `commonArgs` for reusability

packages/components/src/utils/dgs/fetchDgsMetadata.ts
```diff
@@ -4,19 +4,35 @@ interface FetchDgsMetadataProps {
...
 }
 
-type FetchDgsMetadataOutput = Pick<
+export type FetchDgsMetadataOutput = Pick<
   FetchDgsMetadataResponse["data"],
   "name" | "format"
 > & {
   size: string | undefined
+  columnMetadata:
+    | [string, string][] // 1st string is column field key, 2nd string is column title
```
adriangohjw, line 34: array instead of `Record<>` to ensure ordering is preserved

packages/components/src/templates/next/components/complex/SearchableTable/DGS/DGSSearchableTable.tsx
adriangohjw, line 1: main changes: if user does not specify title/headers (or both), we fetch them from DGS's metadata API to fallback to -> this is for a better UX and switch the default from _"you have to specify them or they won't show"_ (happy path) to _"it will use data from DGS unless you specify it to overwrite"_

packages/components/src/hooks/useDgsMetadata.ts
adriangohjw, line 1: simplified version referencing `packages/components/src/hooks/useDgsData/useDgsData.ts`

apps/studio/src/features/editing-experience/components/DatabaseEditorStateDrawer.tsx
adriangohjw, line 1: referencing `HeroEditorDrawer` with some tweaks

apps/studio/src/features/editing-experience/components/MetadataEditorStateDrawer.tsx
```diff
@@ -59,7 +60,22 @@ export default function MetadataEditorStateDrawer(): JSX.Element {
...
+    previewPageState.layout === ISOMER_USABLE_PAGE_LAYOUTS.Database
+      ? {
+          ...metadataSchema,
+          properties: Object.fromEntries(
+            Object.entries(
+              metadataSchema.properties as Record<string, unknown>,
+            ).filter(([key]) => key !== "database"),
+          ),
+        }
+      : metadataSchema
```
dcshzj, line 76: issue: this approach would violate the principle of Studio being headless, because we are directly modifying the contract that Studio is supposed to hold itself to

I was discussing with @seaerchin and he has a proposal: Can we shift to an approach whereby components are the ones that decide the schemas that are available to Studio, and Studio provides say a scope that it wants from components. For example:

**In Studio**
```ts
const scopedSchema = getSchema({ scope: 'page.database' })
```

**In components**
```ts
const getSchema = ({ scope }) => {
    // return only properties in the schema that matches the desired scope
    // using the `$scope` key
}
```

So we still maintain that Studio respects the contract that components (or basically the schema layer) provide in a way that no manipulation of the schema is done, while Studio can still maintain the ability to organise the UI based on its needs.

dcshzj (commented):
I reviewed only the two files that I have comments on, will review the rest in a bit

adriangohjw:
bugbot run

packages/components/src/hooks/useDgsData/useDgsData.ts
```diff
@@ -80,12 +78,14 @@ export const useDgsData = ({
...
   useEffect(() => {
     const fetchData = async () => {
+      setIsLoading(true)
       try {
         if (fetchAll) {
           await fetchAllRecords()
         } else {
           await fetchRecords()
         }
+        setIsError(false)
```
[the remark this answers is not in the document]
adriangohjw: nope, its handled by `setIsLoading(true)` outside the try-catch loop

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDgsDatasetIdControl.tsx
```diff
@@ -0,0 +1,311 @@
...
+                fontFamily="monospace"
+                placeholder="Paste dataset URL here"
+                value={inputValue}
+                onChange={(e) => {
+                  setInputValue(e.target.value)
+                  void register("datasetId").onChange(e)
+                }}
+                onBlur={register("datasetId").onBlur}
+                name={register("datasetId").name}
+                ref={register("datasetId").ref}
```
dcshzj, line 185: issue: we should spread the `register("datasetId")` here, then override just the `onChange`

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDgsDatasetIdControl.tsx
```diff
@@ -0,0 +1,311 @@
...
+            "This doesn't look like a valid link from data.gov.sg. Check that you have the correct link and try again.",
+        }),
+    }),
+    reValidateMode: "onChange",
+  })
+
+  // Update inputValue when initialValue changes (when modal opens)
+  useEffect(() => {
+    setInputValue(initialValueUrl)
+  }, [initialValueUrl])
```
dcshzj, line 102: issue: I think we don't need this `useEffect`? We can force `DgsDatasetIdModal` to re-render below inside `JsonFormsDgsDatasetIdControl` by doing something like this:

```ts
      {isDgsModalOpen && (<DgsDatasetIdModal
        isOpen={isDgsModalOpen}
        onClose={onDgsModalClose}
        onSave={(datasetId) => handleDatasetIdSave(datasetId)}
        initialValue={data || ""}
      />)}
```

This forces a re-render without the need for useEffect, same perf impact anyway since useEffect re-renders the entire component.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDgsDatasetIdControl.tsx
```diff
@@ -0,0 +1,311 @@
...
+    const extractedId = getDgsIdFromString({ string: debouncedInputValue })
+    if (extractedId) {
+      onClose()
+      onSave(extractedId) // Save only the ID, not the full URL
+    }
+  })
+
+  const handleClose = () => {
+    onClose()
+    setInputValue(initialValueUrl)
```
dcshzj, line 138: question: hmm is this intended?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDgsDatasetIdControl.tsx
```diff
@@ -0,0 +1,311 @@
...
+              </Box>
+            </VStack>
+
+            <Button variant="clear" onClick={onDgsModalOpen}>
+              Edit
+            </Button>
+          </HStack>
+
+          <FormErrorMessage>
+            {label} {errors}
```
dcshzj, line 303: nit: can standardise to use `getCustomErrorMessage(errors)`

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsDgsDatasetIdControl.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+export function JsonFormsDgsDatasetIdControl({
+  data,
+  label,
+  handleChange,
+  path,
+  description,
+  required,
+  errors,
+  schema: _schema,
```
dcshzj, line 245: nit: can remove this line if not used

apps/studio/src/features/editing-experience/utils/getDgsIdFromString.ts
```diff
@@ -0,0 +1,52 @@
...
+    // Handle direct ID format: d_abc123
+    if (DgsDatasetIdSchema.safeParse(string).success) {
+      return string
+    }
+
+    // Try to parse as URL
+    const parsedUrl = new URL(string)
+    if (parsedUrl.hostname !== "data.gov.sg") {
+      return null
+    }
```
dcshzj, line 26: question(non-blocking): for completeness, do we need to support `www.data.gov.sg` also?

apps/studio/src/features/editing-experience/utils/getDgsIdFromString.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    // Handle full URL format: https://data.gov.sg/datasets/d_abc123/view
+    if (parsedUrl.pathname.startsWith("/datasets/")) {
+      const pathParts = parsedUrl.pathname.split("/")
+      if (pathParts.length === 4 && pathParts[3] === "view" && pathParts[2]) {
+        const datasetId = pathParts[2]
+        if (DgsDatasetIdSchema.safeParse(datasetId).success) {
+          return datasetId
+        }
+      }
```
dcshzj, line 36: super nit: can reorganise this so it doesn't become an arrow anti-pattern, same for the next few lines

packages/components/src/hooks/useDgsData/useDgsData.ts
```diff
@@ -17,11 +17,10 @@ export const useDgsData = ({
   resourceId,
   limit,
   offset,
-  fields,
   filters,
   sort,
 }: UseDgsData) => {
-  const [isLoading, setIsLoading] = useState(true)
+  const [isLoading, setIsLoading] = useState(false)
```
dcshzj, line 23: question(non-blocking): hmm for my understanding, how come we need to set this to `false` as an initial value ah, I thought starting with `true` makes sense to always show the loading state first

dcshzj (commented):
Briefly looked through, ok to go after the remaining comments are resolved since the blast radius is extremely small

Also, seems like we are rolling this out to all sites? Should we gate this to some sites only? Otherwise we will keep getting support tickets on how to set up this SearchableTable using DGS.

apps/studio/src/features/editing-experience/components/RootStateDrawer.tsx
```diff
@@ -329,6 +333,25 @@ export default function RootStateDrawer() {
                     description="Title, subtitle, and Call-to-Action"
                     icon={TYPE_TO_ICON.hero}
                   />
+                ) : pageLayout === ISOMER_USABLE_PAGE_LAYOUTS.Database ? (
+                  <VStack gap="1rem" w="100%" align="start">
+                    <BaseBlock
+                      onClick={() => {
+                        setDrawerState({ state: "metadataEditor" })
```
seaerchin, line 340: comment: feels abit iffy to have this - should we have a function called `getMetadataType`

packages/components/src/schemas/scopedSchema.ts
seaerchin, line 1: clarification: i think this might be abit too complicated for our use case ._. hapoy to clarify afterwards :sadge:

seaerchin (commented):
i'm abit concerned about code duplication here - afaict alot of stuff is duplicated between studio and components (hooks, biz logic)

not sure if i'm missing something, but i thought studio should jsut import hte functions from componentes or have a shared folder for them. 

wdyt?

packages/components/src/interfaces/integration/dgs.ts
```diff
@@ -9,18 +11,34 @@ export const DgsDataSourceFieldsSchema = Type.Object({
...
+      {
+        // unlikely to be used for Studio users,
+        // so hiding it to reduce complexity of the UI for them
+        format: "hidden",
+      },
     ),
   ),
-  sort: Type.Optional(Type.String()),
+  sort: Type.Optional(
+    Type.String({
```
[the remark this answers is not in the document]
adriangohjw: no sir, its SQL query LOL https://guide.data.gov.sg/developer-guide/dataset-apis/search-and-filter-within-dataset#get-api-action-datastore_search

packages/components/src/hooks/useDgsData/types.ts
```diff
@@ -1,8 +1,11 @@
+// not using "fields" to prevent runtime errors,
+// as DGS returns errors for invalid fields,
+// which can change unpredictably if admins update the dataset schema.
 export interface DgsApiDatasetSearchParams {
   resourceId: string
   limit?: number
   offset?: number
-  fields?: string // comma separated list of fields to fetch
+  // fields?: string // comma separated list of fields to fetch
```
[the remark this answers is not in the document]
adriangohjw: oh i tot to explicitly add but commented out to give context on why we aren't using it despite it being available so no one in the future uses it

apps/studio/src/features/editing-experience/components/form-builder/hooks/useDgsMetadata.ts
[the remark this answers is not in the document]
adriangohjw: yeah i was a bit conflicted about this, but the intention seems to be different, so not sure want to tie them together. WDYT? should we just reuse and worry about this when it comes to that?

packages/components/src/hooks/useDgsData/useDgsData.ts
adriangohjw: true, makes sense. updated https://github.com/opengovsg/isomer/pull/1579/commits/a660d13b5d67a143cb1fb0508dec52b1e5610806

apps/studio/src/features/editing-experience/utils/getDgsIdFromString.ts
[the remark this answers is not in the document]
adriangohjw: @seaerchin updated https://github.com/opengovsg/isomer/pull/1579/commits/0bc4fd9eb4d5c28cca4ee45ed2e2639f21bcc648

PR #1580, opened 2025-09-09
fix: disable creating blocks on collection index pages

seaerchin:
## Problem
- our current check for disabling creating blocks on index pages uses the `page.layout` property
- unfortunately, this is **only** for `page.layout == index`, which doesn't hold true for collection index pages (`layout == collection`, which leads to users being able to add block but these blocks don't do anything (collection layout just renders the filters)

## Solution
- update the condition to prevent adding blocks whne layout is `collection`


adriangohjw (approved):
feels quite jank LOL but okay since hotfix

PR #1581, opened 2025-09-09
chore: remove add block chunk on collections

seaerchin:
## Problem
1. the add block chunk (w button) is based off same conditional and was still showing despite the top-most button being removed

## Solution
1. extract condition into single variable
2. add conditional here


adriangohjw (commented):
i still think its better (even tho more plain) if we remove the whole section tho haha

Image

sehyunidaaa:
this is just for migrators? @seaerchin 

if so ya can remove the whole section.

PR #1582, opened 2025-09-09
chore: shift over from migration script

seaerchin:
## Problem
- when we migrate github sites, we need to automatically shift their `tags` format into our new `tagged` format so that it'll maintain compat with our feature going forward.

## Solution
- reuse same logic from the migration script. this is laid out below:
  1. run a query on the site to get all collections containing pages with the `tags` property
  2. for each collection that has children with tags, we process the children items to extract out the tags 
  3. after we get all the tags, we build a mapping of `tag categories -> tag options` 
  4. next, we update all the blobs of each children item that has the `tags` to the mapping which we've defined earlier
  5. then we proceed to update the db. **an important change:** now, we **omit** the `tags` property. since this site is new, we can strip teh `tags` property and live with only `tagged`  

## Notes
- i didn't want to put this together with the migration script. this is because the migration script now **does not touch the content of the page**. it is primarily concerned with ensuring structure and translating from a github blob -> a db row, **not with ensuring the contents of the items seeded.**
- because of this, i think **post migration data changes** should be done outside of the script. this performance concern (i process more as i need to query for pages w/ tags whereas i already know beforehand via the script) to me is acceptable, considering that we are not optimising for perf here

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36765,6 +36765,10 @@
         "node": ">=8"
       }
     },
+    "node_modules/redis": {
+      "resolved": "packages/redis",
+      "link": true
+    },
     "node_modules/redux": {
       "version": "5.0.1",
       "resolved": "https://registry.npmjs.org/redux/-/redux-5.0.1.tgz",
@@ -42509,6 +42513,9 @@
         "url": "https://opencollective.com/vitest"
       }
     },
+    "packages/redis": {
+      "version": "0.0.0"
+    },
     "tooling/build": {
       "version": "0.0.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


tooling/migrate-tags/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+    },
+  )
+
+  return {
+    tagCategories,
+    labelToId,
+  }
+}
+
+export async function getCollectionsOfSiteWithTags(siteId: number) {
```
seaerchin, line 97: comment: technically this isn't quite `collections` but more any `Resource` that is referenced via `parentId` from a `Resource` with `tags` -> this assumption isn't captured anywhere, so if we have a folder + pages with tags (entirely possible, migrators missed out adding index page), what will happen is that the folder's index pages and child pages will be updated to contain `tagCategories`/`tagged` in a n**on-destructive fashion**

seaerchin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1582?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1582** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1582?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1499** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1584, opened 2025-09-09
fix(eslint-config): cleanup eslint config for isomer

gautammmanek:
## Problem

The current es-lint behavior is broken in a couple of ways:
1. Some packages have a malformed eslint lint command inside package.json, they should be invoked using "eslint ." rather than "eslint", otherwise it won't run the lint and just silently skip. This is currently happening for isomer/tooling/template & isomer/tooling/build
```
 Taken from current github CI:
 WARNING  no output files found for task build#lint. Please check your `outputs` key in `turbo.json`
 WARNING  no output files found for task isomer-base-template#lint. Please check your `outputs` key in `turbo.json`
```

3. The isomer/tooling/build folder lacks an eslint.config file, and there is no workspace-specific eslint config specified, or even a root eslint config to fall back on. Right now on CI this job is skipping silently, due to (1) above
4. VSCode should always run the eslint command as of the currently open directory, to prevent linting the entire codebase from the root. changeProcessCWD tells eslint to lint each each directory in isolation - this makes eslint much faster when linting subdirs

This PR does the following:
1. Fix for (1) by changing eslint -> eslint . to recursively lint all subdirectories
2. Add tooling/build/eslint.config.mjs, which specifies some temporary rules to deal with existing linting errors. These rules should be removed and the behaviour fixed in following PR's
4. Fix some obviously fixable lint errors (type import syntax, incorrect async specifiers) in isomer/tooling/builds, and adding eslint-disable-next-line specifiers to some tooling/build/template statements that are currently failing lint (chose to do this on the file level rather than the rules level since there are much fewer errors here)
5. Add a root eslint config that acts as a backup, just in case a package/workspace is missing an eslint. Also specify the paths for all the tsconfigs inside the individual projects, so regardless of the cwd, eslint knows how to resolve the path to the correct tsconfig

**Breaking Changes**

  - Details ...

**New dependencies**:

- globals package added, to inform eslint which global environment vars the js scripts under tooling/build have access to: https://eslint.org/docs/latest/use/configure/migration-guide#configuring-language-options


gautammmanek:
* **#1584** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1584?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1584?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tooling/build/eslint.config.mjs
seaerchin, line 1: how did we derive this config?

tooling/build/package.json
```diff
@@ -5,12 +5,15 @@
...
+    "lint:fix": "eslint . --fix"
   },
   "devDependencies": {
     "@isomer/eslint-config": "*",
     "@isomer/prettier-config": "*"
   },
-  "prettier": "@isomer/prettier-config"
+  "prettier": "@isomer/prettier-config",
+  "dependencies": {
+    "globals": "^11.12.0"
```
seaerchin, line 17: why do we need this globals dependency?

tooling/build/scripts/generate-sitemap.js
```diff
@@ -25,15 +25,15 @@ const getSchemaJson = async (filePath) => {
   try {
     const schemaContent = await fs.readFile(filePath, "utf8")
     return JSON.parse(schemaContent)
-  } catch (error) {
+  } catch {
```
seaerchin, line 28: nit: instead of just silently omitting, could we `console.log` since it might be useful for debugging from logs later?

tooling/template/app/[[...permalink]]/page.tsx
```diff
@@ -32,6 +32,7 @@ const getPatchedPermalink = async (
   props: DynamicPageProps,
 ): Promise<ParamsContent["permalink"]> => {
   const params = await props.params
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
   return params.permalink ?? [""]
```
Comment (seaerchin) [inline tooling/template/app/[[...permalink]]/page.tsx:36]: nit: i think can just remove since permalink is supposed to be guaranteed

tooling/build/eslint.config.mjs
gautammmanek: Base config exported from isomer/tooling/eslint, rules added to turn existing eslint errors into warnings. My goal was only to fix eslint, I need more context and exposure to the codebase before I can reliably fix the errors flagged up. For js & jsx files, only thing of note is to tell eslint that the scripts run in a server env, and have access to global vars of nodejs

tooling/template/app/[[...permalink]]/page.tsx
Comment (gautammmanek) [reply tooling/template/app/[[...permalink]]/page.tsx:36]: The comment above says to keep it: 
```
// Note: permalink should not be able to be undefined
// However, nextjs had some magic props passing going on that causes
// { permalink: [""] } to be converted to {}
// Thus the patch is necessary to convert it back if its undefined
```
Either way, not going to make any logic changes in this PR

seaerchin (approved):
lgtm, wasn't an issue w/ your PR (i tried this elsewhere, on another branch). tried it again and it works!

PR #1590, opened 2025-09-11
chore: add the sorting functionality back

seaerchin:
## Problem
1. we want to automatically sort tag options and categories 

## Solution
1. add the sorting functionality back

### DIFF
--- a/packages/components/src/templates/next/layouts/Collection/utils/getTagFilters.ts
+++ b/packages/components/src/templates/next/layouts/Collection/utils/getTagFilters.ts
@@ -27,15 +27,17 @@ export const getTagFilters = (
     }
   })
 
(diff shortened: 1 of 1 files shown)


seaerchin:
> Logic makes sense, can add some tests like how the sorting of collection items is tested, if got bandwidth

i didn't do this because this was the old version (ie, i removed this for collection tags). what i did was to check out that [commit hash](https://github.com/opengovsg/isomer/pull/1569/files#diff-4a500cb79c44a260cefc859f7791104a2aa3ed6861cfbfe06245a5b68b93b5b6) for that file and commit it (as this pr). 

lmk if you still want the tests because this will be removed by eow :kekw:

PR #1592, opened 2025-09-13
Refactor/fix: move components into layout

adriangohjw:
## Problem

Some components (e.g. Wogaa, GTM, Widgets like VICA/AskGov) should only be loaded once since they live in the shared layout across all pages.

Normally, this isn’t an issue because we use next/script, which pushes them into <head> so they don’t reload on navigation.

However, for certain widgets, we currently use a hacky approach where they’re unloaded and reloaded between page transitions, which causes flickering.

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- refactor and move common components into `layout.tsx`

> [!NOTE]
> we should also move things like navbar, notification, navbar etc. into layout, but we currently pass `layout` into SearchSG component, thus i skip that scope for this PR 

## Tests

rebuild sites
- [ ] it should load properly
- [ ] go to dev console and check and check on WOGAA/GTM -> they should be in the head
- [ ] for sites with WOGAA/AskGov, they should not flicker between navigation -> if you open them and navigate, it should remain open

### DIFF
--- a/packages/components/src/engine/index.ts
+++ b/packages/components/src/engine/index.ts
@@ -1,4 +1,5 @@
 export { RenderEngine, renderComponentPreviewText } from "./render"
+export { RenderLayoutComponents } from "./renderLayoutComponents"
 export {
   getMetadata,
   shouldBlockIndexing,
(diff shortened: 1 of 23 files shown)


seaerchin (approved):
lgtm, i referenced teh next docs for script loading [here](https://nextjs.org/docs/app/guides/scripts); it seems like we'll load in all the scripts on initial load (and we're using `next/script`) + they'll only be rendered once

PR #1599, opened 2025-09-16
fix: timezone conditions

gautammmanek:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


gautammmanek:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1599?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1600** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1599** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1599?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1572** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1572?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1594](https://github.com/opengovsg/isomer/pull/1594) <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1594?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1584** <a href="https://app.graphite.dev/github/pr/opengovsg/isomer/1584?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/studio/prisma/generated/generatedTypes.ts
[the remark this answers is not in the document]
gautammmanek: We actually have a fair bit of stopping/starting builds because of recency etc (based on the build algo), which makes it easier from a separation-of-concerns point of view when we handle this in a different table. This table can then be used for build-specific concerns (such as is an email sent etc, status of the build ...) for a source of truth

apps/studio/src/server/bullmq/queues/__test__/schedule-publish.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      await expect(
+        publishScheduledResource(
+          "test-job-id",
+          { resourceId: Number(page.id), siteId: site.id, userId: user.id },
+          0,
+        ),
+      ).rejects.toThrowError(
+        "You do not have sufficient permissions to perform this action",
+      )
+    })
```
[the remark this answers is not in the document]
gautammmanek: Ah yes the extreme edge case is a feature, not a bug!

HAHA, it's exactly to prevent the case when user A loses the right to publish because of changing roles etc, we should be more conservative and not allow the publish to succeed; ie we shouldn't reach a point where we need to worry about what's inside the queue when we downgrade user perms!

Let me know thoughts on this!

seaerchin (commented):
no issues w/ business logic, but some comments re: naming/api
1. the table is called `CodeBuildJobs`, but we store `ResourceId` and non-codebuild stuff inside. i was thinking that this table should store just a reference to the codebuild job, and the "non-codebuild" stuff could just be plain json to be consumed later. we'd enforce the structure via code rather than via db (weaker constraint but more flexibility). this allows for future extension using that same table (cos i think next time we'll use this table for future codebuild stuff too), and avoids complicating things like updating site config (no resourceId). 
2. also, i wanted to check in wrt the `lock.acquire` API - i'm not a fan of this pattern in general because it leaves resource acquisition/release to the caller and prone to error. if it's possible, could we do a minor refactor so that actions on redis have to go through something like `executeRedisAction(...)` and this automatically acquires/releases the lock?

apps/studio/src/server/modules/aws/codebuild.service.ts
```diff
@@ -143,12 +145,20 @@ export const shouldStartNewBuild = async (
...
     const command = new StartBuildCommand({ projectName: projectId })
-    const response = await client.send(command)
-    return response
+    const { build } = await client.send(command)
+    // in theory build should always be defined, but adding a check just in case
+    if (!build?.id || !build.startTime) {
+      logger.error({ build }, `Failed to obtain codebuild metadata`)
+      throw new TRPCError({
+        code: "INTERNAL_SERVER_ERROR",
+        message: "Failed to obtain codebuild metadata",
```
seaerchin, line 158: nit: as part of troubleshooting ease, should we add `projectName` here

apps/studio/src/server/modules/aws/codebuild.service.ts
```diff
@@ -143,12 +145,20 @@ export const shouldStartNewBuild = async (
...
+      })
+    }
+    return { id: build.id, startTime: build.startTime }
```
seaerchin, line 161: super ultra nit: why not just return `build`?

seaerchin (approved):

apps/studio/src/server/modules/aws/codebuild.service.ts
gautammmanek: Yessir, added projectId

apps/studio/src/server/modules/aws/codebuild.service.ts
gautammmanek: Ah, just a minor typescript thing here; id and startTime are typed as optional, but we need them required in the Build type. The alternative would be returning {...build, id: build.id, startTime: build.startTime}

gautammmanek:
For reference, have added the separation of the Redlock functionality into its own package in #1639 - which exposes an easy wrapper withRedlockClient

gautammmanek:
### Merge activity

* **Nov 4, 9:05 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1599).
* **Nov 4, 9:05 AM UTC**: @gautammmanek merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1599).

PR #1608, opened 2025-09-19
feat: add script to enable exporting and importing of sites

dcshzj:
## Problem

We have no way of doing an export and import of Isomer Studio sites across environments.

## Solution

**Breaking Changes**

**Features**:

- Add a quick and dirty tooling script to export and import between environments.


adriangohjw (approved):
approving cos non blocker, just questions to understand better

PR #1611, opened 2025-09-19
feat(site-settings): notification view

seaerchin:
## Problem

Add notification banner settings page to allow site administrators to configure site-wide notification banners.

## Solution

Created a new notification settings page that allows administrators to:
- Toggle notification banner on/off
- Set a notification banner title (required when enabled)
- Add an optional banner description

The notification settings are saved to the backend and will be visible to users until dismissed.

**Breaking Changes**

**Features**:

- New notification settings page with toggle to enable/disable banner
- Character-limited input fields for notification title and description
- Form validation to ensure notification text is provided when enabled
- Preview panel to visualize changes before saving
- Unsaved changes detection with confirmation modal

## Before & After Screenshots

**BEFORE**:

N/A - New feature

**AFTER**:

Settings page with notification banner configuration options

## Tests

Verify that:
- Notification settings can be toggled on/off
- Form validation works correctly for required fields
- Changes are saved to the backend and reflected on the site
- Unsaved changes prompt appears when navigating away

**New dependencies**:

None - uses existing design system components and hooks

### DIFF
--- a/apps/studio/src/server/modules/site/__tests__/site.router.test.ts
+++ b/apps/studio/src/server/modules/site/__tests__/site.router.test.ts
@@ -516,7 +516,7 @@ describe("site.router", async () => {
     })
   })
 
-  describe("getNotification", () => {
+  describe.skip("getNotification", () => {
     it("should throw 401 if user is not logged in", async () => {
       // Arrange
       const unauthedSession = applySession()
@@ -564,7 +564,7 @@ describe("site.router", async () => {
     })
   })
 
-  describe("setNotification", () => {
+  describe.skip("setNotification", () => {
     beforeEach(async () => {
       await resetTables("AuditLog")
     })


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsBooleanControl.tsx
```diff
@@ -25,14 +25,15 @@ export function JsonFormsBooleanControl({
...
+  required,
 }: ControlProps): JSX.Element {
   if (schema.const !== undefined) {
     return <></>
   }
 
   return (
     <Box>
-      <FormControl isRequired>
+      <FormControl isRequired={required}>
```
dcshzj, line 36: question(non-blocking): hmm I recall discussing with design on this before and it seems like having the words "(optional)" for a toggle from a design perspective seemed a little weird to users, hence it is set to required. I think the `defaultValue` should be able to ensure that a value is set, to fulfil the FormControl's requirement?

apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
```diff
@@ -87,11 +80,24 @@ export const BASE_EXTENSIONS: Extensions = [
...
     },
   }),
+  History,
+  Italic,
+  Paragraph,
+  Text,
+  Underline,
+]
+
+export const BASE_PROSE_EXTENSIONS: Extensions = [
```
dcshzj, line 90: nit: uhh now we have `BASE_EXTENSIONS` and `BASE_PROSE_EXTENSIONS` which I think the naming can be quite confusing. Can this be renamed to something else instead? Thinking perhaps just `PROSE_EXTENSIONS` should be sufficient (but can suggest a better name if you have)

apps/studio/src/pages/sites/[siteId]/settings/index.tsx
```diff
@@ -1,217 +1,20 @@
...
 const siteSettingsSchema = z.object({
   siteId: z.coerce.number(),
 })
 
 const SiteSettingsPage: NextPageWithLayout = () => {
-  const toast = useToast()
-  const trpcUtils = trpc.useUtils()
+  const router = useRouter()
   const { siteId } = useQueryParse(siteSettingsSchema)
+  void router.replace(`sites/${siteId}/settings/agency`)
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/settings/index.tsx:15]: question: I can't remember what we discussed about the rollout plan, are we doing full rollout here?

apps/studio/src/pages/sites/[siteId]/settings/notification.tsx
```diff
@@ -0,0 +1,137 @@
...
+    previousNotification ?? {
+      enabled: false,
+      title: "",
+      content: { type: "prose", content: [] },
+    },
+  )
+
+  useEffect(() => {
+    if (!isEnabled) {
+      router.push(`/sites/${siteId}/settings`)
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/settings/notification.tsx:75]: issue: seems like there is a redirect loop with the original settings page

packages/components/src/interfaces/internal/Notification.ts
```diff
@@ -2,26 +2,25 @@ import type { Static } from "@sinclair/typebox"
...
-        title: "Notification title",
-        description: "The title of the notification",
-        format: "hidden",
-      }),
-    ),
-    content: Type.Array(TextSchema, {
-      title: "Notification content",
-      description: "The content of the notification",
-      format: "hidden",
+    enabled: Type.Boolean({
```
dcshzj, line 9: question: hmm what do you think about doing an optional object instead, so something like this:

```ts
{
    notification: Type.Optional(Type.Object(
        {
            title: Type.String({...}),
            content: SimpleProseSchema
        }, {
            title: "Display a banner",
            description: "The site notification will always be visible on the site until it is dismissed by the user.",
        }
    ))
}
```

So the existence of the `notification` object determines if it's enabled or not.

packages/components/src/interfaces/internal/Notification.ts
```diff
@@ -2,26 +2,25 @@ import type { Static } from "@sinclair/typebox"
...
+      title: "Display a banner",
+      description:
+        "The site notification will always be visible on the site until it is dismissed by the user.",
+      default: false,
     }),
+    title: Type.String({
+      title: "Notification title",
+      description: "The title of the notification",
+      maxLength: 100,
```
adriangohjw, line 18: quick sense check re: discussion on character limit: we are setting a character limit on this?

apps/studio/src/server/modules/site/site.router.ts
adriangohjw, line 34: to remove since no longer used

packages/components/src/templates/next/components/native/Prose/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+        )
+      })
+    case "heading":
+      return !!content.content?.some((headingContent) => {
+        return headingContent.text.trim() !== ""
+      })
+    case "orderedList":
+    case "unorderedList":
+    case "table":
+      return true
```
adriangohjw, line 29: ask: add exhaustive check

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -86,27 +86,31 @@ export const getNotification = async (siteId: number) => {
...
 }
 
 interface SetSiteNotificationParams {
   tx: Transaction<DB>
   siteId: number
   userId: string
-  notification: string // TODO: Replace this with Tiptap schema once frontend refactored
+  title: string
+  content: Notification["content"]
+  enabled: boolean
```
adriangohjw, line 98: should we derive all these from `Notification` schema?

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -200,39 +200,32 @@ export const siteRouter = router({
...
-            userId: ctx.user.id,
-          })
-        }
+        return await setSiteNotification({
+          tx,
+          siteId,
+          userId: ctx.user.id,
+          title,
+          content,
+          enabled: !!enabled,
```
adriangohjw, line 222: question: is this for backward compat?

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -200,39 +200,32 @@ export const siteRouter = router({
...
+        })
```
adriangohjw, line 223: ask: actually since this is the only use case of `setSiteNotification`, can we do some refactoring to move `db.transaction()` into `setSiteNotification` instead?

apps/studio/src/pages/sites/[siteId]/settings/notification.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const isEnabled = useNewSettingsPage()
+  const router = useRouter()
+  const { siteId: rawSiteId } = useQueryParse(siteSchema)
+  const siteId = Number(rawSiteId)
+  const trpcUtils = trpc.useUtils()
+  const toast = useToast()
+  const [{ name }] = trpc.site.getSiteName.useSuspenseQuery({
+    siteId: Number(siteId),
+  })
+  const validateFn = ajv.compile<Notification>(NotificationSchema)
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/settings/notification.tsx:39]: suggestion: move this outside the component as its computationally expensive

apps/studio/src/pages/sites/[siteId]/settings/notification.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { ajv } from "~/utils/ajv"
+import { trpc } from "~/utils/trpc"
+
+const NotificationSettingsPage: NextPageWithLayout = () => {
+  const isEnabled = useNewSettingsPage()
+  const router = useRouter()
+  const { siteId: rawSiteId } = useQueryParse(siteSchema)
+  const siteId = Number(rawSiteId)
+  const trpcUtils = trpc.useUtils()
+  const toast = useToast()
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/settings/notification.tsx:35]: ask: initialize with `BRIEF_TOAST_SETTINGS`

apps/studio/src/pages/sites/[siteId]/settings/notification.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const [{ name }] = trpc.site.getSiteName.useSuspenseQuery({
+    siteId: Number(siteId),
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/settings/notification.tsx:37]: ultra nit: shouldn't need `Number` since we do that in line 33

apps/studio/src/pages/sites/[siteId]/settings/notification.tsx
Comment (seaerchin) [reply apps/studio/src/pages/sites/[siteId]/settings/notification.tsx:75]: leaving this as is first, confiring w rachel that we're ok with rolling out all at once. if that's the case, i'll remove all the flags and this should be fixed

packages/components/src/interfaces/internal/Notification.ts
seaerchin: no explicit guidance given on design side but i just set first. i'll check w/ @sehyunidaaa but i retained existing guidelines

apps/studio/src/server/modules/site/site.service.ts
seaerchin: done downstream!

apps/studio/src/server/modules/site/site.router.ts
seaerchin: this got updated downstream

apps/studio/src/server/modules/site/site.router.ts
seaerchin: sure

dcshzj (approved):
Approving first, but make sure to add back the feature flag on the `/sites/1/settings/index.tsx` so that if true, redirect to `/sites/1/settings/agency.tsx`, otherwise stay here

seaerchin:
### Merge activity

* **Oct 22, 8:35 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1611).
* **Oct 22, 8:38 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/opengovsg/isomer/1611) rebased this pull request as part of a merge.

PR #1614, opened 2025-09-23
chore: shift to json forms for rendering

seaerchin:
## Problem
In the previous iteration, we used handbuilt components for rendering. however, this is not ideal because we should take the definition from `JsonForms` then build it automatically.

## Solution
- defined new jsonforms component for rendering a disabled text componnet. (side note: this has the same priority as a text control but with format optimal, **which is intended**. this tiebreaks via the ordering)
- **created a custom schema** for UI rendering purposes via `TypeBox`. This does break encapsulation on component's - more on this later. 

## Notes
- as noted above, the jsonforms takes in a **custom** schema which is generated on studio's end via picking properties from the exported schema from components.
- i initially considered using a `$scope` property on the schema (on components) but this would require filtering on the actual schema itself **outside of typebox**. i felt that it was worth doing this idiomatically on typebox rather than busting it open and doing js logic to get the correct schema (eg; `schema.properties.filter((scope) => scope === "agency")`). 
- i'm more towards keeping it inside of typebox and i feel that this is still adhering to our schema -> if the `format` changes, we still do re-render accordingly. it's just that "what" to render is left to the consumer (here, Studio).

### DIFF
--- a/apps/studio/src/features/settings/SettingsHeader.tsx
+++ b/apps/studio/src/features/settings/SettingsHeader.tsx
@@ -16,8 +16,8 @@ export const SettingsHeader = ({
     <Flex justifyContent="space-between" w="100%">
       <Flex align="center">
         <Center
-          w="1.5rem"
-          h="1.5rem"
+          w="2rem"
+          h="2rem"
           bgColor="brand.secondary.100"
           borderRadius="6px"
           mr="0.75rem"
--- a/apps/studio/src/features/settings/SettingsSidenav/SettingsSidenav.tsx
+++ b/apps/studio/src/features/settings/SettingsSidenav/SettingsSidenav.tsx
@@ -56,6 +56,7 @@ export const SettingsSidenav = ({ onSidenavClose }: SettingsSidenavProps) => {
               borderColor="base.divider.medium"
               px="0.75rem"
               align="start"
+              w="100%"
             >
               {items.map((item) => (
                 <SettingsItem
--- a/apps/studio/src/features/settings/SettingsSidenav/components/SettingsItem.tsx
+++ b/apps/studio/src/features/settings/SettingsSidenav/components/SettingsItem.tsx
@@ -15,6 +15,10 @@ export const SettingsItem = ({ isActive, label, href }: SettingsItemProps) => {
       textDecoration="none"
       px="0.5rem"
       py="0.25rem"
+      w="100%"
+      _hover={{
+        bgColor: "interaction.muted.main.hover",
+      }}
     >
       <Text
         textStyle="subhead-2"


packages/components/src/types/site.ts
```diff
@@ -13,45 +13,54 @@ import { NotificationSchema } from "~/interfaces/internal/Notification"
...
   agencyName: Type.Optional(
     Type.String({
-      title: "Agency name",
-      description:
-        "The name of the agency, displayed in the copyright footer for non-Government websites.",
+      title: "Website is owned by",
+      description: "This isn't displayed anywhere on your site",
+      format: "disabled",
     }),
   ),
```
dcshzj, line 31: question(non-blocking): this one is a bit weird in the sense that it should be required if `isGovernment` is set to false, because that is when it is shown at the copyright text in the footer

But I know we don't really have the ability to show/hide fields based on the inputs of another, so as long as UI side is ok with this then I'm ok

apps/studio/src/pages/sites/[siteId]/settings/agency.tsx
```diff
@@ -37,6 +36,12 @@ const AgencySettingsPage: NextPageWithLayout = () => {
   })
   const trpcUtils = trpc.useUtils()
   const toast = useToast()
+  const agencySettingsSchema = Type.Pick(
+    SiteConfigSchema,
+    Type.Union([Type.Literal("siteName"), Type.Literal("agencyName")]),
+  )
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/settings/agency.tsx:42]: praise: I love this cleaner design tbh, still allows JSONForms to be the one doing the calculation and rendering

thought: I still think there is some value in having the components export some sort of function, say `getScopedSchema(scope)` that allows Studio to get the appropriate subschema to render. That function's implementation would be this `Type.Pick`, but at least the ownership is on the components rather than on Studio

apps/studio/src/pages/sites/[siteId]/settings/agency.tsx
```diff
@@ -37,6 +36,12 @@ const AgencySettingsPage: NextPageWithLayout = () => {
   })
   const trpcUtils = trpc.useUtils()
   const toast = useToast()
+  const agencySettingsSchema = Type.Pick(
+    SiteConfigSchema,
+    Type.Union([Type.Literal("siteName"), Type.Literal("agencyName")]),
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/settings/agency.tsx:41]: question: not sure if the `Type.Union` works but seems like we can just directly do `['siteName', 'agencyName']`?

apps/studio/src/pages/sites/[siteId]/settings/agency.tsx
```diff
@@ -37,6 +36,12 @@ const AgencySettingsPage: NextPageWithLayout = () => {
...
+  )
+  type AgencySettings = Static<typeof agencySettingsSchema>
+  const validateFn = ajv.compile<AgencySettings>(agencySettingsSchema)
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/settings/agency.tsx:44]: thought: I'm not sure how the persisting to DB part would be done later, but seems like we might need to use `Type.Mapped` rather than `Type.Pick`, and map it such that the `format` is set to `hidden` for the fields that we don't want to show here.

That way, the JSON saved to the backend is still the full JSON value, but the UI is just the 2 fields that you want to show

dcshzj (approved):
I'm ok with it as-is now but left some comments for your consideration

PR #1619, opened 2025-09-25
update limits for image and infocard descriptions

sehyunidaaa:
## Problem

Character limits for components are limiting with not enough user education, causing migration inefficiencies.

Closes [ISOM-2076]

## Solution

This PR focuses on loosening limitations for the two components:
- Individual image captions 
- Image gallery image captions
- Individual infocard descriptions

### DIFF
--- a/packages/components/src/interfaces/complex/Image.ts
+++ b/packages/components/src/interfaces/complex/Image.ts
@@ -46,7 +46,8 @@ export const ImageSchema = Type.Object(
     caption: Type.Optional(
       Type.String({
         title: "Caption",
-        maxLength: 250,
+        description:
+          "Describe the image or add attributions. To make sure your caption is readable, keep it under 250 characters.",
         format: "textarea",
       }),
     ),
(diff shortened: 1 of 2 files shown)


packages/components/src/interfaces/complex/ImageGallery.ts
```diff
@@ -9,7 +9,7 @@ export const IMAGE_GALLERY_TYPE = "imagegallery"
 const SingleImageSchema = Type.Object({
   src: ImageSrcSchema,
   alt: AltTextSchema,
-  caption: Type.Optional(Type.String({ maxLength: 250 })),
+  caption: Type.Optional(Type.String({})),
```
seaerchin, line 12:
```suggestion
  caption: Type.Optional(Type.String()),
```

seaerchin (approved):
lgtm, should we also consider relooking at limits on other sections of the codebase? (or is this strictly image related? then i'd suggest also looking at the homepage collection block, if applicable)

adriangohjw (commented):
should i add a story each for extra long text?

PR #1626, opened 2025-09-29
Enhance/remove inactive isomer admins

adriangohjw:
## Problem

past isomer admins aren't removed, which can look like an security issues to agency users

Reference: https://opengovproducts.slack.com/archives/C06R4DX966P/p1759130877982859

## Solution

**Breaking Changes**

  - Details ...

---

> [!NOTE]
> Includes isomer admins and migrators in inactive user selection and notifications, while excluding them from adminEmails in email content.
> 
> - **Backend**:
>   - `getInactiveUsers`: remove exclusion of `ISOMER_ADMINS_AND_MIGRATORS_EMAILS` so isomer admins/migrators are included in inactivity detection.
>   - `bulkSendAccountDeactivationWarningEmails`: include isomer admins/migrators when aggregating users to notify (removed email exclusion in query).
>   - `getSiteAndAdmins`: retain exclusion of `ISOMER_ADMINS_AND_MIGRATORS_EMAILS` from `adminEmails` (clarified comment).
> - **Tests** (`inactiveUsers.service.test.ts`):
>   - Update expectations to ensure isomer admins/migrators are selected as inactive and receive emails, but are not listed in `adminEmails`.
>   - Adjust assertions for email payloads and call counts accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d659e1c6c1288bcecb2fb15daebdff8748d30f32. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/studio/src/server/modules/user/__tests__/inactiveUsers.service.test.ts
+++ b/apps/studio/src/server/modules/user/__tests__/inactiveUsers.service.test.ts
@@ -374,14 +374,14 @@ describe("inactiveUsers.service", () => {
       expect(sendAccountDeactivationEmail).toHaveBeenCalledTimes(1)
     })
 
-    it("should not include isomer admins and migrators in the email", async () => {
+    it("should NOT include isomer admins and migrators in the email", async () => {
       // Arrange
       const userToDeactivate = await setupUserWrapper({
         siteId: site.id,
         createdDaysAgo: 91,
         lastLoginDaysAgo: null,
       })
-      await Promise.all(
+      const isomerAdminsAndMigrators = await Promise.all(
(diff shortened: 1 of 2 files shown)


dcshzj:
Hmm I'm thinking we don't need to automatically remove? We should instead do it via our offboarding procedures

Reason being I think it is possible that someone active doesn't actually log in to Studio (since we technically got no sites). Re-adding the access would be a bit more work.

adriangohjw:
@dcshzj yeah i think i can add that as a side quest in godmode. question: how do yall handle this in Classic? or there's no auto removal of inactive users?

dcshzj:
Ps butter fingers, closed it by accident

PR #1627, opened 2025-09-29
clean up isomer admin and migrator list

adriangohjw:
housekeeping

---

> [!NOTE]
> Move alexander, jan, and jinhui from ISOMER_ADMINS to PAST_ISOMER_MEMBERS.
> 
> - **Data updates** in `apps/studio/prisma/constants.ts`:
>   - `ISOMER_ADMINS`: remove `alexander`, `jan`, `jinhui`.
>   - `PAST_ISOMER_MEMBERS`: add `alexander`, `jan`, `jinhui`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 653cd44c5b059e5103babd2ad75192fb14eac3a3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/studio/prisma/constants.ts
+++ b/apps/studio/prisma/constants.ts
@@ -1,13 +1,10 @@
 export const ISOMER_ADMINS = [
-  "alexander",
-  "jan",
   "jiachin",
   "sehyun",
   "harish",
   "zhongjun",
   "adriangoh",
   "shazli",
-  "jinhui",
   "rachellin",
   "gautam",
 ]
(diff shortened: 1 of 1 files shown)


apps/studio/prisma/constants.ts
```diff
@@ -25,7 +22,13 @@ export const ISOMER_MIGRATORS = [
   "felicia",
 ]
 
-export const PAST_ISOMER_MEMBERS = ["kishore", "hanpu"] as const
+export const PAST_ISOMER_MEMBERS = [
```
harishv7, line 25: saw that it's used to determine user deletion, but do we actually need to maintain this list?

adriangohjw: It's used to display users under the Isomer Admin table in User management instead of agency users since we don't have a super admin role.

We won't need it if we remove them when they leave, but given that they are likely added to all sites, it's quite a hassle to manually remove them from all sites - will let the Cron job auto removed them after 90 days

PR #1655, opened 2025-10-14
feat(scheduled-publishing): ameliorate race condition around codebuild

gautammmanek:
## Problem

When testing #1639, came across some **major** issues with the schedule publishing; i.e. in that we **cannot** effectively rely on our build algorithm to deconflict concurrent builds for the same site.

Refer to the document https://www.notion.so/opengov/Dealing-with-codebuild-race-conditions-28c77dbba78880ad9ff8ee9454524bd9?source=copy_link for a clear explanation on what this issue is, and the proposed solution

This PR does the following (explained in the document):
1. Modify PublishPageResource to accept an argument startSitePublish, which determines whether publishing the page should also publish the site. The scheduled publishing queue should set this to **false**, since we no longer wish to publish the site every time we publish the page
2. The PublishPageResource also inserts a new codebuildjob, with a new status of PENDING (indicating that it's pending being picked up by the next site publish), and with a buildId of **null** (since a build hasn't been started for the site yet)
3. Inside the SchedulePublishQueue, we push a new job onto the SitePublishQueue for all resources which are scheduled at the same time. The key for the job is constructed from **scheduledAt** & **siteId**, which makes de-duplication easier
4. The site publish then sends a build to Codebuild **once**, and updates all the **pending rows** inside the codebuild table with the new buildId, for the same site

This effectively **decouples** the page publish and the site publish functionality; pages are published by the first queue and then forwarded to another **site publish** queue for publishing the entire site. It also makes error handling and retries much more predictable

Changelog:
1. Refactor SchedulePublishWorker to use **handleFailedJob** on failure, so it can be used for both queues
2. Use the webhook function updateCodebuildStatusAndSendEmails inside the SitePublishWorker, and allow an argument of **buildId** of null to be provided. This allows the sending of failure emails effectively for rows which are left PENDING in the codebase, if the site publish job fails
3. Add integration tests for publishing page, publishing site and pushing new SitePublish events on the queue

**Breaking Changes**

  - Includes expanding migrations which need to be run; making the buildId, startedAt optional, and adding a PENDING status to the BuildStatusType

### DIFF
--- a/apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts
+++ b/apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts
@@ -429,5 +429,26 @@ describe("webhook.router", async () => {
           )
         })
     })
+    it("does NOT throw an error if the build arn is not found", async () => {
+      // Arrange
+      const { site } = await setupCodeBuildJob({
+        userId: user.id,
+        arn: "build/test-id",
+        startedAt: FIXED_NOW,
+        isScheduled: true,
+        omitResourceId: true, // this will create a codebuild job without a resourceId, simulating a site publish
+      })
+      const caller = getCallerWithMockGrowthbook(session)
+
+      // Act & Assert
+      await expect(
+        caller.updateCodebuildWebhook({
+          projectName: "test-project",
+          siteId: site.id,
+          arn: "build/wrong-arn", // saved in the db
+          status: "SUCCEEDED",
+        }),
+      ).resolves.not.toThrow()
+    })
   })
 })


apps/studio/src/server/bullmq/queues/site-publish.ts
```diff
@@ -0,0 +1,98 @@
...
+      workerName: "SitePublishWorker",
+      retryLimit: WORKER_RETRY_LIMIT,
+      onFinalFailure: async ({ data: { siteId } }: Job<SitePublishJobData>) => {
+        const gb = await createGrowthBookContext()
+        await updateCodebuildStatusAndSendEmails(
+          logger,
+          gb,
+          null,
+          siteId,
+          "FAILED",
```
gautammmanek: Use enum

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -609,14 +627,8 @@ export const publishPageResource = async ({
...
-  // Step 2: Trigger a publish of the site
-  await publishSite(logger, {
-    siteId,
-    userId: user.id,
-    resourceId,
-    isScheduled,
-    addCodebuildJobRow,
-  })
+  // Step 2: Trigger a publish of the site, if startSitePublish is true
+  if (startSitePublish) await publishSite(logger, siteId)
```
seaerchin: question: why not just return `codeBuildId` here since the immediate next step is to grab `codeBuildId` from the codebase (and it's safe to assume this variable doesn't change)

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -595,6 +599,20 @@ export const publishPageResource = async ({
...
+      await tx
+        .insertInto("CodeBuildJobs")
+        .values({
+          resourceId,
+          siteId,
+          userId: user.id,
+          isScheduled,
+        })
+        .execute()
+    }
```
seaerchin: comment: this api feels very "fat" in the sense that alot of the control flow variable exposed to the caller seem to be what the method should, in and of itself, handle properly. 

eg: for this variable, i thought the conditional should be whether this is a scheduled publish (unless we also want to inform users of non-scheduled publish, but then that raises the question of what happens if i publish a great # of pages at once... but then in that case, then we don't need this variable `addCodebuildJobRow` anymore)

(edit: went to check the call sites) i think we should just shift this in since this is a feature flag and the caller shouldn't care about htis.

packages/redis/src/utils.ts
```diff
@@ -22,15 +22,21 @@ export const withRedlock = async <T>(
   resource_name: string,
   fn: () => Promise<T>,
   logger: pino.Logger<string>,
-  opts?: RedlockOptions,
+  opts: RedlockOptions = {
+    lock_ttl_ms: DEFAULT_REDLOCK_TTL_MS,
```
seaerchin: nit: we should remove this whole method entirely

PR #1657, opened 2025-10-14
chore: don't throw if no parent found for collection block

seaerchin:
## Problem                                                                      
                                                                              
The CollectionBlock component was throwing errors when encountering edge  cases like:                                                                  
  - Collection not found in the sitemap                                        
  - Collection exists but has no items                                         
                                                                              
This resulted in broken rendering and a poor user experience, especially     
in Studio where users might be configuring collections or working with       
empty/new collections.                                                       
                                                                              
## Solution                                                                     
                                                                              
Refactored error handling to use graceful degradation instead of throwing  errors:                                                                     
                                                                              
Changes to utility functions:                                                
 - `getCollectionParent()`: Returns null instead of throwing when collection    
  is not found                                                                
  - `getCollectionPages()`: Returns empty array [] instead of throwing when      
collection has no items  

## Notes
- not too sure if we want to display an error message on end sites (when teh collection is already deleted and `collectionParent` can't be found)
- ok with not showing nothing instead

### DIFF
--- a/apps/studio/docker-compose.yml
+++ b/apps/studio/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   postgres:
-    image: postgres:latest
+    image: postgres:15.10
     ports:
       - 5432:5432
     environment:
--- a/turbo.json
+++ b/turbo.json
@@ -41,7 +41,7 @@
     "services:setup": {
       "cache": false
     },
-    "setup:test": {},
+    "setup:test": { "cache": false },
     "storybook": {
       "cache": false
     },


packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionParent.ts
```diff
@@ -21,7 +21,7 @@ export const getCollectionParent = ({
     return collectionParent as IsomerCollectionPageSitemap
   }
 
-  throw new Error(
-    `CollectionBlock: No collection parent found for collection ID ${collectionId}`,
-  )
+  // NOTE: Signal an error to the caller that no parent could be found
+  // so that we can fail gracefully
```
adriangohjw: i don't quite understand this comment? 👀

packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
```diff
@@ -180,11 +185,25 @@ export const CollectionBlock = ({
...
+      />
+    )
   }
 
   const collectionParent = getCollectionParent({ site, collectionId })
 
+  if (!collectionParent) {
+    return (
+      <CollectionBlockSkeleton
+        title="This collection does not appear to be part of any site"
```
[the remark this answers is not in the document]
seaerchin: i updated this to return a `Fragment` instead, because this error state shoulnd't be shown to MOPs

packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionParent.ts
seaerchin: we return `null` here rather than throwing an `error`, so that the caller can handle this state by themselves

PR #1659, opened 2025-10-17
refactor(performance): fix slow endpoints and high RDS CPU

adriangohjw:
## Problem

| description | image |
| - | - |
| High RDS CPU usage with high DB load | Screenshot 2025-10-17 at 12 07 27 PM |
| Caused by one query with high latency > 2 seconds | Screenshot 2025-10-17 at 12 01 01 PM |
| Datadog shows top 3 paths with high p95 latencies over the past 1 day | Screenshot 2025-10-17 at 12 31 09 PM |

| Suspect that this is because `getResourcesWithFullPermalink` is doing a `Promise.all(...)` which runs multiple DB read queries. And that this method is in turn used by 3 other API endpoints |

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- update `getWithFullPermalink` to accept multiple resources so we can avoid `Promise.all(...)` and do one query instead
- others: remove unnecessary resource existence check for `resource.getWithFullPermalink` endpoint
- others: combine query to get results and total count into using Promise.all(...) -> note: this will actually has more load on RDS. However, im assuming the main culprit is `getWithFullPermalink` accepting multiple resources initially rather than this. Ideally with this, we can fix the slow endpoint instead

---

> [!NOTE]
> Batch `fullPermalink` resolution, parallelize search result/count queries, and update router endpoint to new array-based API for faster responses.
> 
> - **Resource Service**:
>   - Batch `getWithFullPermalink` to accept `resourceIds: string[]`, query via `IN`, and return an array; add empty-input early return.
>   - Replace per-item `Promise.all` in `getResourcesWithFullPermalink` with single batched query and map results.
>   - In `getSearchResults`, fetch paged resources and total count in parallel; use typed `countAll<number>()` and return structured `totalCount`.
>   - Type tweak: cast `GREATEST(Resource.updatedAt, Blob.updatedAt)` as `sql<Date | null>`.
> - **Router**:
>   - `getWithFullPermalink` endpoint now calls batched service (`resourceIds: [resourceId]`), handles empty results, and returns first item; removes prior existence check query.
> - **Search Utilities**:
>   - `getSearchRecentlyEdited` streamlined to use shared helpers without extra casts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fcea1c02600d6afedf4941055aa12517b55174b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -908,24 +913,20 @@ export const getSearchResults = async ({
...
-      .executeTakeFirstOrThrow()
-  ).total_count as number // needed to cast as the type can be `bigint`
+      .executeTakeFirstOrThrow(),
+  ])
 
   return {
-    totalCount,
     resources: await getResourcesWithFullPermalink({
-      resources: resourcesToReturn,
+      resources: resourcesToReturn as SearchResultResource[],
```
dcshzj: issue: hmm this type cast is quite iffy, cos `resourceToReturn` is almost correct just except for the `lastUpdatedAt` property. The issue should be fixed upstream under `getResourcesWithLastUpdatedAt`:

```ts
const getResourcesWithLastUpdatedAt = ({ siteId }: { siteId: number }) => {
  return db
    .selectFrom("Resource")
    .select([
      "Resource.id",
      "Resource.title",
      "Resource.type",
      "Resource.parentId",
      // To handle cases where either the resource or the blob is updated
      sql<Date | null>`GREATEST("Resource"."updatedAt", "Blob"."updatedAt")`.as(
        "lastUpdatedAt",
      ),
    ])
    .leftJoin("Blob", "Resource.draftBlobId", "Blob.id")
    .where("Resource.siteId", "=", siteId)
}
```

I'm not entirely sure if `Date | null` is correct here but just following the type definition inside `SearchResultResource`. Once that type argument is added, the type cast can be omitted.

adriangohjw: yeah true i agree

though i don't think i want to refactor this part in this PR since its unrelated to the performance improvement

dcshzj: Can we just make this change please? You're already touching this part of the code and it improves the overall codebase as a whole, so we don't need to open multiple PRs for the same thing.

If it's a major overhaul then yes I would agree, but in this case it's a very quick change.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -908,24 +913,20 @@ export const getSearchResults = async ({
...
-    .execute()) as SearchResultResource[]
-
-  const totalCount: number = (
-    await db
+  const [resourcesToReturn, totalCountResult] = await Promise.all([
+    orderedResources.offset(offset).limit(limit).execute(),
+    db
       .with("queriedResources", () => queriedResources)
       .selectFrom("queriedResources")
       .select(db.fn.countAll().as("total_count"))
```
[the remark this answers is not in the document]
adriangohjw: updated https://github.com/opengovsg/isomer/pull/1659/commits/a7b34d0ef6b32c7aa6918e981cce30a92913dde3 👍

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: @dcshzj done! https://github.com/opengovsg/isomer/pull/1659/files/26cf0e89fc36adf7e0a7ea58be011bfc70bcd8b7..798442898d8b2b5feaa12663d8ffe779caaa491a

dcshzj (approved):
I think CI got some unrelated issue, approving first

PR #1661, opened 2025-10-21
Long words can't fit into hero block

sehyunidaaa:
## Problem

- Long words (e.g., "Pronouncements") can't fit into the hero block variant
- While this is a single case, it signals that the `prose-display-xl` type style might be too big

Closes [ISOM-2114]

## Solution

Adjust typography token
- Adjusted `prose-display-xl`'s responsive behaviour to become smaller between the LG and XL breakpoints
- At this range, `prose-display-xl` will have the same size as `prose-display-lg`
- `prose-display-xl` is only used exclusively on Hero so far 

Modify word-break behaviour
- Removed `break-word` and using `text-balance` instead
- Added `wrap-break-word` to maintain word-break properties for older browsers that do not support `text-balance`

Adjust spacing of Hero Block variant 
- Adjusted right-padding to p-8 from p-12 to allow for more space
- Updated max-width of text container accordingly

Note: Haven't updated `preview-tw` yet, having some issues with build

### DIFF
--- a/apps/studio/public/assets/css/preview-tw.css
+++ b/apps/studio/public/assets/css/preview-tw.css
@@ -1,115 +1,5 @@
-*,
-::before,
-::after {
-  --tw-border-spacing-x: 0;
-  --tw-border-spacing-y: 0;
-  --tw-translate-x: 0;
-  --tw-translate-y: 0;
-  --tw-rotate: 0;
-  --tw-skew-x: 0;
-  --tw-skew-y: 0;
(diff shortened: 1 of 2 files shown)


adriangohjw (approved):
non-blocker, just wanna clarify: for the storybook, should we use the word "Pronouncements" instead? cos thats better reflective of the "worst case" scenario

engineering wise okay - if design okay then okay! 👍

PR #1668, opened 2025-10-23
fix(collection-links): prevent stale data on gsib

seaerchin:
## Problem
previously, we had a stale state issue in collection links due to the atom's state. to avoid this, this PR lifts state to a `useState` at the top level.

## Solution
1. shift state from `atom` to the `page` level.
2. JSONForms writes data back to the `state`, which is then fed down into the `EditLinkPreview`
3. add `useMemo` for less rerender
4. shift type to be on `collection/schemas`

## Tests
- [ ] open a collection 
- [ ] ensure that the collection has at least 1 page and 2 links
- [ ] open the page and make some edits to the category (and save)
- [ ] go to the other two links
- [ ] the category on the links should not be that of the page
- [ ] make some edits on the link itself 
- [ ] the preview should update to show these changes
- [ ] saving the changes should still work

---

> [!NOTE]
> Lift collection link state to page-level and refactor editor/preview and category control to use props (removing Jotai coupling), while centralizing types and schemas.
> 
> - **State management**:
>   - Lift collection link editing state to page-level `useState` in `pages/sites/[siteId]/links/[linkId]/index.tsx` and remove `linkAtom` from `atoms.ts` (retain only `linkRefAtom`).
>   - Pass `link` and `setLink` as props to `LinkEditorDrawer` and `EditCollectionLinkPreview`.
> - **Editor/Preview refactor**:
>   - `LinkEditorDrawer` now controlled via props, saves with TRPC mutation using current `link` state; supports raw JSON editor using provided `initialLinkState`.
>   - `EditCollectionLinkPreview` builds sitemap from props and queries; adds `useMemo` for derived values and uses current date for `lastModified`.
>   - `PreviewWithoutSitemap` sets a module-level `defaultLastModified` and uses it as the default.
> - **Form controls**:
>   - `JsonFormsCategoryControl` removes Jotai coupling; writes changes via `handleChange` only and uses category dropdown when feature-flagged.
> - **Schemas/Types**:
>   - Move `CollectionLinkProps` to `schemas/collection.ts` and add Zod schemas (`editLinkSchema`, `readLinkSchema`, etc.) with date parsing/formatting helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ecc0f1d251a113396773005e88c7d308cfd6c74d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
adriangohjw: nit: might not actually need memo since they are primitives operations

apps/studio/src/features/editing-experience/components/EditLinkPreview.tsx
```diff
@@ -38,43 +38,49 @@ export const EditCollectionLinkPreview = (): JSX.Element => {
...
-            ref,
-            summary: summary ?? "",
-            layout: "link",
-            permalink,
-            lastModified: new Date().toString(),
-            category,
-            tagged,
+            id: "collection",
+            permalink: parentPermalink,
+            lastModified: "02 May 2023",
```
dcshzj: nit: I think we should take the chance to change all the `lastModified` to use the `currentDate`. The dates used here seems very arbitrary.

seaerchin: sounds good, done

apps/studio/src/features/editing-experience/components/LinkEditorDrawer.tsx
```diff
@@ -162,31 +160,19 @@ const DrawerState = (
...
-  useEffect(() => {
-    setLinkAtom({
-      ...(content.page as CollectionLinkProps),
-      title,
-    })
-    setLinkRef((content.page as CollectionLinkProps).ref)
-  }, [content, setLinkAtom, setLinkRef, title])
 
   const { mutate, isPending } =
     trpc.collection.updateCollectionLink.useMutation({
```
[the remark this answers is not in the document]
seaerchin: to avoid scope creep - i think the intent of this pr is to fix the stale state bug, by using jsonforms. this is a collorary of this PR and i'll do it in a follow-up if you're amenable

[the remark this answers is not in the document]
dcshzj: Yeah that's fair, ok to do separately since I think the refactor would be quite complex

seaerchin:
only merging after confirming fix on gsib

PR #1671, opened 2025-10-23
chore(vica/askgov): add widget icons

seaerchin:
## Problem
The current integrations page is misisng the widget icons for feedback to the end users

## Solution
- Add the icons into the page; these were done on studio because it didn't feel like they were part of components. to get the styling, i looked at the component on askgov (w/ some help from adrian) then translated into `styles` (no chakra on iframe)
- for vica, i inspected our sample vica site and took the styling. 

## Videos

https://github.com/user-attachments/assets/fb379ba2-0e5a-421e-b68f-d16e02d7258a

### DIFF
--- a/apps/studio/src/components/Askgov/index.tsx
+++ b/apps/studio/src/components/Askgov/index.tsx
@@ -1,38 +1,42 @@
-import { Flex } from "@chakra-ui/react"
-
 export const AskgovWidget = () => {
   return (
-    <Flex
-      position="fixed"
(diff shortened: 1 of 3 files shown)


apps/studio/src/components/Askgov/index.tsx
adriangohjw: idk why the display looks a bit long-ish than round, but i think nevermind haha

adriangohjw (approved):

PR #1672, opened 2025-10-23
chore: allow markdown for link descrioption

seaerchin:
## Problem
we want to allow links in the description for forms, so that there can be useful help texts

## Solution
add a markdown renderer for the string control and specify styling for links manually.
 
## Photos
image

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTextControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTextControl.tsx
@@ -1,5 +1,5 @@
 import type { ControlProps, RankedTester } from "@jsonforms/core"
-import { Box, FormControl, useToken } from "@chakra-ui/react"
+import { Box, FormControl } from "@chakra-ui/react"
 import { isStringControl, rankWith } from "@jsonforms/core"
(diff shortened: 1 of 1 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTextControl.tsx
```diff
@@ -62,7 +69,25 @@ export function JsonFormsTextControl({
...
+                color: linkHoverColor,
+                textDecoration: "none",
+              },
+              _active: {
+                color: linkActiveColor,
+                textDecoration: "none",
+              },
+            },
+          }}
+        >
```
adriangohjw: can do some simple extraction of this into something reusable? IMO this is quite useful and can be used for other stuff

adriangohjw (approved):

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTextControl.tsx
seaerchin: i don't mind, but need to be careful about this, because it alters styles in-line (style has higher precedence than classname). 

i'll put as a `MarkdownLabel` for now w/ just the `a` tag styling if you're ok with that?

PR #1682, opened 2025-10-28
chore: fix styling

seaerchin:
## Problem
Need to add 1 more bounding `Box` so that the gap isn't large

## Solution
Add `Box`

## Photos
**Before**
image

**After**
image


apps/studio/src/pages/sites/[siteId]/settings/notification.tsx
```diff
@@ -109,19 +109,21 @@ const NotificationSettingsPage: NextPageWithLayout = () => {
...
-            handleChange={(data) => {
-              setState(data)
-              // NOTE: We have to set `isDismissed` here because
-              // we need to show the notification banner again when
-              // the user toggles it on
-              if (isEmpty(data)) setIsDismissed(false)
-            }}
-          />
+          <Box>
+            <Box mb="-0.5rem" />
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/settings/notification.tsx]: why need this to begin with ah?

PR #1692, opened 2025-10-31
Relax all other character limits for components

sehyunidaaa:
## Problem

Isomer initially relied on hard restrictions to shape user behaviour. This was the most scalable and fastest way when we were supporting agencies closely and our priority was growth above all. But this approach fell short. In a workflow where content came to Isomer post-approvals, strict limitations meant users had to modify their content last-minute to accommodate to Isomer.

At times, the restrictions would lead to less-than-ideal content. Under time constraints and stress, our users can't rewrite content to be succinct because this would come with extra approvals. Instead, they would shorten words or introduce acronyms, which is a net negative impact to the public.

We were resistant to "the proper way" for the longest time. But there's no quick, computational way around writing good content. User-centric intentions shape good content, not restrictions. We've accepted this as a platform and will be dedicating more time to thought leadership and guidance, rather than system-level limitations. This way, we can give space for our agency users to focus on their communications craft. 

Closes [ISOM-2144]

## Solution

All limits are relaxed, with the exception of:

### Exceptions

- **Hero title:** 100 characters
- **Hero description:** 300 characters
- **Page titles (all page layouts):** 250 characters
- **Page summary (all page layouts):** 500 characters
- **Notification title:** 150 characters
- **Notification description:** 300 characters

Also, Navbar and Button limits will stay, but we won't announce it.

### DIFF
--- a/apps/studio/src/schemas/folder.ts
+++ b/apps/studio/src/schemas/folder.ts
@@ -3,9 +3,9 @@ import { z } from "zod"
 import { generateBasePermalinkSchema } from "./common"
 import { offsetPaginationSchema } from "./pagination"
 
-export const MAX_FOLDER_TITLE_LENGTH = 100
-export const MAX_FOLDER_PERMALINK_LENGTH = 200
-export const MAX_FOLDER_DESCRIPTION_LENGTH = 300
+export const MAX_FOLDER_TITLE_LENGTH = 250
+export const MAX_FOLDER_PERMALINK_LENGTH = 500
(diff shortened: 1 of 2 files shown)


dcshzj (approved):
No issues with this! I realised that even if we have an arbitrarily large character limit, somehow it would become a target that editors will try to reach which becomes counter-intuitive.

Let's keep track of the previous limits somewhere so it can serve as a foundation for any site report card that we might want to do in the future!

PR #1702, opened 2025-11-05
fix(storybook): upgrade to v9

seaerchin:
## Problem
storybook has [an issue](https://github.com/storybookjs/storybook/issues/32301) with next + node22 

## Solution
1. upgraded storybook using `npx storybook@9 upgrade`
2. installed the swc add on
3. fix a bug in imports (we should import from `generatedEnums` and not `prisma/client`)
4. fix viewport parameters being outdated


adriangohjw (approved):
thanks sir!

looking forward to you upgrading it to [v10](https://storybook.js.org/blog/storybook-10/) hehe

dcshzj:
Need to confirm our sites still can build on CodeBuild, either this one or #1689 might be breaking

PR #1703, opened 2025-11-05
⚡ Performance: Fix treeshaking to exclude Typebox in bundle

adriangohjw:
## Issue

Messy imports causing tree shaking to not work correctly - and that Typebox included in the client bundle as a result

closes https://linear.app/ogp/issue/ISOM-2084/performance-improvements

## Solution

Better directory structuring to split constants, types, schemas etc. properly to ensure not side effects and better tree shaking

also remove barrel export - they are bad!

Reference: https://www.notion.so/opengov/performance-bundle-size-2a277dbba78880a18e80cc2f95b805ce?source=copy_link

---

> [!NOTE]
> Refactors the components library to use named, side‑effect‑free exports and literal component types, reorganizing constants/schemas/rendering to improve tree‑shaking and reduce bundle size; updates Studio to new block keys.
> 
> - **Components library refactor (tree‑shaking)**:
>   - Replace default exports with named exports; remove/avoid barrel exports across complex/native/internal components and templates.
>   - Move/centralize layout constants to `~/types/constants`; switch imports from `~/engine` to `~/types`.
>   - Use string literal component `type` values instead of exported TYPE constants; add dedicated constants modules (e.g., ChildrenPages, Infobar, ContactInformation, DynamicDataBanner).
>   - Reorganize rendering: split `renderLayout`, `renderPageContent`, and `renderComponent`; update `engine` to import from `~/types` and new next render entry.
>   - Extract list schema builders to `schemas/utils`; update native list schemas to use them.
>   - Tidy public exports: re‑export from `src/index.ts`; adjust `templates/next/components/index` to explicit named exports.
> - **Studio/editor updates**:
>   - Update default blocks, allowed block lists, and icon mappings to new literal keys (`imagegallery`, `collectionblock`, `contactinformation`, etc.).
> - **Utilities/types**:
>   - Shift sitemap/layout helpers to `~/types` usage; create `types/constants` for layouts.
> - **Misc**:
>   - Small typing/packaging cleanups (e.g., package.json types, migration script typings).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 212b1b66c21f372c0d75cce01fe229a95110af24. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -157,6 +157,7 @@
     "react-hook-form": "^7.52.1",
     "react-icons": "^5.2.0",
     "react-markdown": "^10.1.0",
+    "rehype-external-links": "^3.0.0",
     "sass": "^1.83.4",
     "superjson": "^2.2.1",
     "tiptap-text-direction": "^0.3.1",
--- a/apps/studio/src/components/MarkdownLabel.tsx
+++ b/apps/studio/src/components/MarkdownLabel.tsx
@@ -1,5 +1,6 @@
 import { Box, useToken } from "@chakra-ui/react"
 import Markdown from "react-markdown"
+import rehypeExternalLinks from "rehype-external-links"
 
 export const MarkdownLabel = ({ description }: { description?: string }) => {
   const [linkColor, linkHoverColor, linkActiveColor] = useToken("colors", [
@@ -25,7 +26,9 @@ export const MarkdownLabel = ({ description }: { description?: string }) => {
         },
       }}
     >
-      <Markdown>{description}</Markdown>
+      <Markdown rehypePlugins={[[rehypeExternalLinks, { target: "_blank" }]]}>
+        {description}
+      </Markdown>
     </Box>
   )
 }
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsColourPickerControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsColourPickerControl.tsx
@@ -11,7 +11,11 @@ import {
 } from "@chakra-ui/react"
 import { rankWith, schemaMatches } from "@jsonforms/core"
 import { useJsonForms, withJsonFormsControlProps } from "@jsonforms/react"
-import { FormLabel, Input } from "@opengovsg/design-system-react"
+import {
+  FormErrorMessage,
+  FormLabel,
+  Input,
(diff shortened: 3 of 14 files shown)


packages/components/src/templates/next/components/native/Prose/Prose.tsx
dcshzj: nit: should we use named export for Prose as well?

packages/components/src/constants/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from "./pageLayouts"
+export * from "./componentTypes"
+export * from "./childrenPages"
+export * from "./infobar"
+export * from "./dynamicDataBanner"
+export * from "./contactInformation"
```
dcshzj: question: having these constants exported from its own namespace feels quite weird and violates our principle of each component being isolated building blocks. I also noticed that many (if not all) of them can be derived some way or another from the schema itself.

Can these constants be eliminated, or otherwise stored within each individual component's folder?

dcshzj (commented):
Skimmed most of the changes, seems to be just a conversion of the default export into named export (with the exception of Prose which I have flagged out)

packages/components/src/templates/next/layouts/Collection/Collection.stories.tsx
```diff
@@ -4,10 +4,10 @@ import { expect, userEvent, within } from "storybook/test"
 
 import { withChromaticModes } from "@isomer/storybook-config"
 
-import type { IsomerSitemap } from "~/engine"
-import { type CollectionPageSchemaType } from "~/engine"
+import type { IsomerSitemap } from "~/types"
 import { generateSiteConfig } from "~/stories/helpers"
-import CollectionLayout from "./Collection"
+import { type CollectionPageSchemaType } from "~/types"
```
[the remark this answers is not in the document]
adriangohjw: updateed https://github.com/opengovsg/isomer/pull/1703/commits/88a9cc9f3fc41f9701fa909f55aacdc120c10646

packages/components/src/constants/index.ts
adriangohjw: @dcshzj yeah true, moved it as a constants files under the respective schema and importing from them directly https://github.com/opengovsg/isomer/pull/1703/commits/3c6a087091fe180ba1c4f2ba95dc9dde73d76e67

PR #1709, opened 2025-11-07
⚡ Performance: Replace RAC with react-aria hooks

adriangohjw:
## Issue

closes https://linear.app/ogp/issue/ISOM-2084/performance-improvements

- Using react-aria-components because they are simple out of the box
- Hoever, they aren't properly treeshaken and thus result in higher than expected bundle size
- Note: IMO this should not be done for normal codebases, but for ours which are performance-sensitive + these components are basic ones that aren't changed often, this switch to handrolling using react-aria makes sense

reference: https://www.notion.so/opengov/performance-bundle-size-2a277dbba78880a18e80cc2f95b805ce?source=copy_link

---

> [!NOTE]
> Refactors UI components to use react-aria hooks with native elements, removing react-aria-components and updating dependencies for smaller, more tree-shakable bundles.
> 
> - **Components (refactor to hooks + native elements)**
>   - `Button`, `IconButton`, `DownloadButton`: use `useButton`/`useFocusRing`, native `<button>`, support `className`, merged props/refs.
>   - `CheckboxGroup`/`Checkbox`: rebuilt with `useCheckboxGroup(State)`/`useCheckbox(Group)`, custom renderer, context, improved pressed/focus/invalid/disabled styling.
>   - `Filter`/`FilterDrawer`: replace RAC `Button/Label` with custom buttons using hooks; extracted small button components; same UI behavior.
>   - `Search`: replaced RAC `SearchField` with `useTextField` + native `<input type="search">`; adds clear via `IconButton`; removes internal `Input` component.
>   - `Navbar`: imports `FocusScope` from `@react-aria/focus`; mobile menu close now `useButton`.
>   - `Masthead`: splits mobile/desktop buttons, both using `useButton`/`useFocusRing`.
>   - `Pagination`: drop `composeRenderProps`; use `twMerge`; pass `isDisabled` through.
> - **Utils**
>   - Remove RAC compose helpers; add `mergeRefs`; keep `dataAttr`/`focusVisibleHighlight`.
> - **Dependencies**
>   - Remove `react-aria` and `react-aria-components`; add `@react-aria/*` and `@react-stately/*` packages; bump `@internationalized/date`/`number`; update `package.json` accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4992b1b8038d277728ad61859d93f9c626c728d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/components/src/templates/next/components/internal/Search.tsx
```diff
@@ -64,32 +51,67 @@ export function SearchField({
...
+        <input
+          {...mergeProps(inputProps, focusProps)}
+          ref={inputRef}
+          type="search"
+          className={twMerge(
+            inputStyles(),
             "bg-transparent [&::-webkit-search-cancel-button]:hidden",
+            classNames?.input,
           )}
           placeholder={placeholder}
```
seaerchin: [comment] from the react aria repo, we should apply an empty space string for safari lolol: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/s2/src/TextField.tsx#L191

packages/components/src/templates/next/components/internal/Masthead/Masthead.tsx
```diff
@@ -88,60 +92,90 @@ const RestrictedHeaderBarContent = ({ children }: PropsWithChildren) => {
...
 }
 
-const RestrictedHeaderBar = ({ children }: PropsWithChildren) => {
+const MobileMastheadButton = ({ children }: PropsWithChildren) => {
   const { isMastheadContentVisible, toggleMastheadContent } = useMasthead()
+  const buttonRef = useRef<HTMLButtonElement>(null)
+
+  const buttonProps: AriaButtonProps<"button"> = {
+    "aria-controls": MASTHEAD_CONTROL_ID,
+    "aria-describedby": "masthead-aria",
```
seaerchin: nit: can we make this a `const` at the top? not super concerned cos this won't escape this file but easier for us down the line

seaerchin (approved):
lgtm on my end, but some questions that i think might be worth considering.

most of what i see here is that we just remove the wrapper then we just call `useButton`; other bigger stuff like `Checkbox`, `SearchField` etc, i looked into the react aria impl but i don't think got any big issues ba

packages/components/src/templates/next/components/internal/Checkbox/Checkbox.tsx
```diff
@@ -68,36 +134,185 @@ const boxStyles = tv({
...
+    isDisabled,
+    onPress: () => {
+      if (!isDisabled) {
+        // Programmatically click the input to trigger its onChange handler
+        // This ensures grouped checkboxes use inputProps.onChange from useCheckboxGroupItem
+        // and standalone checkboxes use inputProps.onChange from useCheckbox
+        inputRef.current?.click()
+      }
+    },
+  })
```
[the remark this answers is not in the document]
adriangohjw: yeah it doesn't work for keyboard unfortunately - maybe is implementation issue but this way it works

PR #1716, opened 2025-11-11
tests(site-settings): add unit tests for new endpoints introduced

seaerchin:
## Problem
- we should have tests for various functionality in studio
- this PR adds those tests in

## Solution
1. add tests for audit log
2. add tests for permissions check 
3. add tests for misc functionality (extra properties, update search sg)

### DIFF
--- a/apps/studio/src/server/modules/site/__tests__/site.router.test.ts
+++ b/apps/studio/src/server/modules/site/__tests__/site.router.test.ts
@@ -388,7 +388,7 @@ describe("site.router", async () => {
       const result = unauthedCaller.updateSiteConfig({
         siteName: MOCK_SITE_NAME,
         logoUrl: MOCK_LOGO_URL,
-        url: "www.isomer.gov.sg",
+        url: "https://www.isomer.gov.sg",
         theme: "isomer-next",
         siteId: 1,
       })
@@ -406,7 +406,7 @@ describe("site.router", async () => {
       const result = caller.updateSiteConfig({
         siteName: MOCK_SITE_NAME,
         logoUrl: MOCK_LOGO_URL,
-        url: "www.isomer.gov.sg",
+        url: "https://www.isomer.gov.sg",
         theme: "isomer-next",
         siteId: site.id,
       })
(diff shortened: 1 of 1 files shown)


apps/studio/src/server/modules/site/__tests__/site.router.test.ts
```diff
@@ -340,6 +378,394 @@ describe("site.router", async () => {
...
+      // Assert
+      await expect(result).rejects.toThrowError(
+        new TRPCError({
+          code: "FORBIDDEN",
+          message:
+            "You do not have sufficient permissions to perform this action",
+        }),
+      )
+    })
+    it("should update the site config if the user is a site admin", async () => {
```
adriangohjw: nit: can we also have another test where someone is given publisher rights and still face 403?

apps/studio/src/server/modules/site/__tests__/site.router.test.ts
```diff
@@ -340,6 +378,394 @@ describe("site.router", async () => {
...
+
+      // Assert
+      expect(result).toEqual({
+        siteName: MOCK_SITE_NAME,
+        logoUrl: MOCK_LOGO_URL,
+        url: "www.isomer.gov.sg",
+        theme: "isomer-next",
+      })
+    })
+    it("should generate an audit log entry", async () => {
```
dcshzj: nit: we should assert this together with the previous test, meaning we combine the two tests

adriangohjw (approved):

apps/studio/src/server/modules/site/__tests__/site.router.test.ts
seaerchin: i actually did this cos it's clearer what's being asserted. for discussion going forward, should we do this or combine all the assertions for success case? 

i thought having various assertions for the invariants in teh success case made it clearer. i'm happy to change it to whichever the team prefers
@adriangohjw @dcshzj

seaerchin: discussed on slack: https://opengovproducts.slack.com/archives/C06R4DX966P/p1762933350084159?thread_ts=1762916421.142049&cid=C06R4DX966P; 

not doing this (sorry)

PR #1725, opened 2025-11-13
feat: add automated migration scripts for NDF

dcshzj:
## Problem

NDF is some special site that requires a significant automation work.

## Solution

**Breaking Changes**

**Features**:

- Add a new scripts package to handle all recurring automation work.
- Add script for migration NDF collection.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6279,7 +6279,9 @@
       }
     },
     "node_modules/@csstools/color-helpers": {
-      "version": "5.0.2",
+      "version": "5.1.0",
+      "resolved": "https://registry.npmjs.org/@csstools/color-helpers/-/color-helpers-5.1.0.tgz",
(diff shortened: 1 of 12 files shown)


adriangohjw:
@dcshzj this seems like a one-off script for NDF and no need to be merged so i won't review - do correct me if im wrong!

dcshzj:
> @dcshzj this seems like a one-off script for NDF and no need to be merged so i won't review - do correct me if im wrong!

Hmm conflicted, I created this PR cos we will run this for at least 6 times - July to December 2026, but also a bit pointless to review honestly. Can review if got time! Super not urgent.

adriangohjw (approved):
TBH didn't really review very deep, skimmed through it 

will be helpful if can include a README, or sample CSVs for the inputs for reference, in case someone needs to run this when you are not around

but non-blocker and good-to-have, so approving to clean up the PRs

thanks!

PR #1739, opened 2025-11-22
feat(scheduled-publishing): use cronjob for scheduled publishing

gautammmanek:
## Problem

Following the suggestions in https://www.notion.so/opengov/Scheduled-publishing-thoughts-2aa77dbba788805cbb5cefeedfe040ad?source=copy_link, this PR implements a **cronjob-style** approach towards scheduled publishing, whereby a standardized cronjob is run on a per-minute basis to determine whether there are any resources which require publishing, publish these resources, publish the corresponding site, and update the scheduledAt timestamps to set it to **null**.

Refer to https://www.notion.so/opengov/Scheduled-publishing-moving-to-a-cronjob-2b577dbba7888024b2cad28d28721e5c?source=copy_link for more details on what's going on in this PR

Briefly, this PR does the following
1. Adds a dedicated **pgboss** package which can be used for future cronjobs. As @adriangohjw mentioned, keeping it as a separate package is a lot more convenient for future usage in another place inside the codebase
2. Adds a migration to add the **scheduledBy column** to the DB, thereby ensuring the scheduler's id is captured
3. Deletion of the old redis package and scheduled publishing logic, since we aren't using the message queues anymore. **NOTE: environment variables etc will be removed in the next PR**, as part of a full clean-up once we commit to a specific logic
4. Addition of a dedicated **schedulePublishingJob**, which implements the core of the job logic. Refer to the notion page for details on how this works

**New dependencies**:

- `dependency` : Addition of **pg-boss**, which uses the existing postgres DB for managing the scheduled cronjob

---

> [!NOTE]
> Switches scheduled publishing to a PgBoss-driven per-minute cron, adds `Resource.scheduledBy`, and updates publishing/build flows to batch-link resources while removing BullMQ/Redis code.
> 
> - **Backend/Infra**:
>   - Introduces `@isomer/pgboss` package (PgBoss client, heartbeat util) and initializes cron jobs on server start.
>   - Adds `schedulePublishingJob` (runs every minute) to publish due resources, trigger site publishes, send failure emails, and clear `scheduledAt/By`.
>   - Removes BullMQ/Redis-based scheduling and related queues, utils, and tests.
> - **Database/Schema**:
>   - Adds `Resource.scheduledBy` (FK to `User`), updates Prisma/Kysely types.
>   - Makes `CodeBuildJobs.buildId` required.
> - **Publishing/Build pipeline**:
>   - `publishPageResource`/`publishSite` updated to support optional site publish and batch CodeBuild job rows (`resourceWithUserIds`), and to record scheduled builds.
>   - Refactors CodeBuild utils/types (non-optional `latestRunningBuild`, batch insert + supersede logic).
> - **API/Logic**:
>   - Page scheduling now records `scheduledBy`; cancel clears it. Page publish path updated to new signatures.
>   - Adds optional `SCHEDULED_PUBLISHING_HEARTBEAT_URL` env for cron heartbeats.
> - **Tooling/Misc**:
>   - Updates Turbo/dep graph to build `@isomer/pgboss`; drops `@isomer/redis`.
>   - Minor ESLint ignore and CSS utility additions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b91b1cb364713c1141fffcd6e5c6ca68f72c2384. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/studio/src/schemas/__tests__/webhook.test.ts
+++ b/apps/studio/src/schemas/__tests__/webhook.test.ts
@@ -29,17 +29,14 @@ vi.mock("~/features/mail/service", () => ({
 }))
 
 const createMockRequest = ({
-  siteId,
   arn,
   apiKey = "test-webhook-api-key",
 }: {
-  siteId: number
   arn: string
   apiKey?: string
 }) => {
   const body: z.input<typeof codeBuildWebhookSchema> = {
     projectName: "test-project",
-    siteId,
     arn,
     status: "SUCCEEDED",
   }
@@ -63,14 +60,13 @@ describe("webhook", () => {
     it("should process valid webhook payload", async () => {
       // Arrange
       const user = await setupUser(createTestUser())
-      const { site } = await setupCodeBuildJob({
+      await setupCodeBuildJob({
         userId: user.id,
         arn: "build/test-id",
         startedAt: new Date(),
         isScheduled: true,
       })
       const { req, res } = createMockRequest({
-        siteId: site.id,
         arn: "build/test-id",
       })
 
@@ -83,14 +79,13 @@ describe("webhook", () => {
     it("providing an incorrect API key causes a 401", async () => {
       // Arrange
       const user = await setupUser(createTestUser())
-      const { site } = await setupCodeBuildJob({
+      await setupCodeBuildJob({
         userId: user.id,
         arn: "build/test-id",
         startedAt: new Date(),
         isScheduled: true,
       })
       const { req, res } = createMockRequest({
-        siteId: site.id,
         arn: "build/test-id",
         apiKey: "wrong-api-key",
       })
@@ -104,13 +99,12 @@ describe("webhook", () => {
     it("requests missing an API key causes a 401", async () => {
       // Arrange
       const user = await setupUser(createTestUser())
-      const { site } = await setupCodeBuildJob({
+      await setupCodeBuildJob({
         userId: user.id,
         arn: "build/test-id",
         startedAt: new Date(),
       })
       const { req, res } = createMockRequest({
-        siteId: site.id,
         arn: "build/test-id",
         apiKey: "wrong-api-key",
       })
--- a/apps/studio/src/schemas/webhook.ts
+++ b/apps/studio/src/schemas/webhook.ts
@@ -24,7 +24,6 @@ export const buildIdFromArn = (arn: string) => {
 export const codeBuildWebhookSchema = z
   .object({
     projectName: z.string(),
-    siteId: z.coerce.number(),
     arn: z.string(),
     status: z.nativeEnum(BuildStatusType),
   })
--- a/apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts
+++ b/apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts
@@ -70,7 +70,7 @@ describe("webhook.router", async () => {
     })
     it("it should update the codebuildjobs table if the received build status is successful", async () => {
       // Arrange
-      const { site, page, codebuildJob } = await setupCodeBuildJob({
+      const { page, codebuildJob } = await setupCodeBuildJob({
         userId: user.id,
         arn: "build/test-id",
         startedAt: FIXED_NOW,
@@ -81,7 +81,6 @@ describe("webhook.router", async () => {
       // Act
       await caller.updateCodebuildWebhook({
         projectName: "test-project",
-        siteId: site.id,
         arn: "build/test-id", // saved in the db
         status: "SUCCEEDED",
       })
@@ -131,7 +130,6 @@ describe("webhook.router", async () => {
       // Act
       await caller.updateCodebuildWebhook({
         projectName: "test-project",
-        siteId: site.id,
         arn: "build/test-id", // saved in the db
         status: "SUCCEEDED",
       })
@@ -155,7 +153,7 @@ describe("webhook.router", async () => {
     })
     it("it should update the codebuildjobs table if the received build status is failed", async () => {
       // Arrange
-      const { site, page, codebuildJob } = await setupCodeBuildJob({
+      const { page, codebuildJob } = await setupCodeBuildJob({
         userId: user.id,
         arn: "build/test-id",
         startedAt: FIXED_NOW,
@@ -166,7 +164,6 @@ describe("webhook.router", async () => {
       // Act
       await caller.updateCodebuildWebhook({
         projectName: "test-project",
-        siteId: site.id,
         arn: "build/test-id", // saved in the db
         status: "FAILED",
       })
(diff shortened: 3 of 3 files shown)


apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -0,0 +1,196 @@
...
+ * Publishes all resources scheduled for publishing up to the current time,
+ * publishes their associated sites, and resets their scheduledAt fields.
+ */
+const schedulePublishJobHandler = async () => {
+  const scheduledAtCutoff = new Date()
+  const gb = await createGrowthBookContext()
+  const enableCodebuildJobs = gb.isOn(ENABLE_CODEBUILD_JOBS)
+  const enableEmailsForScheduledPublishes = gb.isOn(
+    ENABLE_EMAILS_FOR_SCHEDULED_PUBLISHES_FEATURE_KEY,
+  )
```
adriangohjw: nit: might be cleaner if we initialize this in the respective methods so we don't have to pass it in?

apps/studio/src/server/modules/aws/codebuild.service.ts
```diff
@@ -58,13 +59,12 @@ export const publishSite = async (
...
-  if (addCodebuildJobRow) {
+  if (codebuildJobArgs.addCodebuildJobRow) {
     await addCodeBuildAndMarkSupersededBuild({
       buildChanges: buildChangesWithStartedBuild,
       siteId,
-      userId,
-      isScheduled,
-      resourceId,
+      resourceWithUserIds: codebuildJobArgs.resourceWithUserIds,
+      isScheduled: codebuildJobArgs.isScheduled,
```
adriangohjw: super nit: can use lodash.pick and then spread it
```ts
...pick(codebuildJobArgs, ["resourceWithUserIds", "isScheduled"])
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -550,21 +536,21 @@ export const getResourceFullPermalink = async (
 
 interface PublishPageResourceArgs {
   logger: Logger<string>
-  user: User
+  userId: string
   siteId: number
   resourceId: string
+  sitePublishOptions:
+    | { enable: true; isScheduled: boolean; addCodebuildJobRow: boolean }
+    | { enable: false }
```
adriangohjw: question: why not make `sitePublishOptions` an optional prop then we don't need `enable`? we can just assume if prop is passed in, then it is enabled

apps/studio/src/server/modules/aws/codebuild.service.ts
```diff
@@ -8,21 +8,22 @@ import {
...
+      addCodebuildJobRow: true
+      isScheduled: boolean
+      resourceWithUserIds: { resourceId: string; userId: string }[]
+    }
+  | { addCodebuildJobRow: false }
+
+interface PublishSiteArgs {
+  siteId: number
+  codebuildJobArgs?: CodebuildJobArgs
+}
```
adriangohjw: question: since `codebuildJobArgs` is optional, why do we need `{ addCodebuildJobRow: false }`? can we assume that if the args is passed in, then its enabled?

adriangohjw:
bugbot run

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -0,0 +1,196 @@
...
+        `Failed to publish page for resource: ${resourceId}`,
+      )
+      if (enableEmailsForScheduledPublishes) {
+        const recipientEmail = await getUserEmailFromId(logger, scheduledBy)
+        if (recipientEmail)
+          await sendFailedPublishEmail({
+            recipientEmail,
+            isScheduled: true,
+            resource,
+          })
```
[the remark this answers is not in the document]
adriangohjw: valid! good job bro

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -0,0 +1,196 @@
...
+        ...(siteResourcesMap[siteId] ?? []),
+        { ...resource, scheduledBy },
+      ]
+    } catch (error) {
+      logger.error(
+        { error },
+        `Failed to publish page for resource: ${resourceId}`,
+      )
+      if (enableEmailsForScheduledPublishes) {
+        const recipientEmail = await getUserEmailFromId(logger, scheduledBy)
```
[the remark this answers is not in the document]
gautammmanek: Sounds good, let's do that instead

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
gautammmanek: Ah, small thing but I'd rather keep the FF hoisted at the top level to make the behaviour more obvious, at the cost of one extra function param!

apps/studio/src/server/modules/aws/codebuild.service.ts
gautammmanek: Good note, added, thanks

apps/studio/src/server/modules/aws/codebuild.service.ts
gautammmanek: Same as above, fixed this, thanks

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -599,7 +580,17 @@ export const publishPageResource = async ({
...
+        .selectFrom("User")
+        .selectAll()
+        .where("id", "=", userId)
+        .executeTakeFirstOrThrow(
+          () =>
+            new TRPCError({
+              code: "BAD_REQUEST",
+              message: "Please ensure that you are logged in!",
+            }),
+        ),
```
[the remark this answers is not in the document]
gautammmanek: Refactored a bit here; primarily to extract the getUserById to another method inside the user service. Still keeping the user fetching inside the publishPage since it doesn't require us to query for the user at all callsites (such as inside the publishPage API), and aliased joins with kysely require listing out all columns from both tables (and need to be extra-careful of overlaps between the two tables)

I think performance from doing a single PK lookup is negligible in this case

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -0,0 +1,190 @@
...
+      if (enableEmailsForScheduledPublishes) {
+        try {
+          await sendFailedPublishEmail({
+            recipientEmail: resource.email,
+            isScheduled: true,
+            resource,
+          })
+          logger.info(
+            `Sent failed publish email to ${resource.email} for resource: ${resourceId}`,
+          )
```
adriangohjw: nit: should be `logger.warning`? sorry i missed this out in the initial review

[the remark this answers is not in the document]
gautammmanek: Sounds good, let me add in a couple of tests for this as well

adriangohjw (approved):
LGTM, thank you!

gautammmanek:
### Merge activity

* **Dec 3, 2:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/1739).

PR #1745, opened 2025-11-24
feat(scheduled-publishing): add tests for schedulePublishingJob

gautammmanek:
## Problem

This PR adds basic integration testing for the cronjob, as well as the newly-introduced **pg-boss** package. Specifically, the following changes are made:
1. For the scheduled publishing cronjob, test that the core success and error handling logic works fine for the resource publishing and site publishing flows
2. For the pg-boss package, since this is an integration test, spin up a testing db at **5431** to run the tests against pg-boss. Assert that the setup is successful, and that the heartbeat functionality worked as expected when a HEARTBEAT_URL is provided
3. In order to setup the pg-boss package, add vitest as a dependency and add a vitest config

### DIFF
--- a/apps/studio/src/server/cron/jobs/__test__/schedulePublishingJob.test.ts
+++ b/apps/studio/src/server/cron/jobs/__test__/schedulePublishingJob.test.ts
@@ -247,6 +247,59 @@ describe("schedulePublishingJob", async () => {
         versionNum: 1,
       })
     })
+    it("a resource without userId inside scheduledBy is skipped and does not prevent other resources from being published", async () => {
+      // Arrange
+      const { site, page } = await setupPageResource({
+        resourceType: ResourceType.Page,
+        scheduledAt: FIXED_NOW,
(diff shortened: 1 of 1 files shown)


packages/pgboss/vitest.config.ts
```diff
@@ -0,0 +1,19 @@
...
+    alias: {
+      "~/": fileURLToPath(new URL("./src/", import.meta.url)),
+    },
+  },
+  test: {
+    globals: true,
+    exclude: [...configDefaults.exclude],
+
+    coverage: {
+      provider: "istanbul",
```
adriangohjw: also have to add devDependencies `@vitest/coverage-istanbul`?

gautammmanek:
### Merge activity

* **Dec 3, 2:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/1745).

PR #1593, opened 2025-09-14
⚡ Performance: Defer widget loading

adriangohjw:
## Problem

closes https://linear.app/ogp/issue/ISOM-2084/performance-improvements

reference:  https://opengovproducts.slack.com/archives/C06R4DX966P/p1762855731532589?thread_ts=1762765218.516989&cid=C06R4DX966P

TLDR: widget loading makes it be counted as part of TBT by wogaa

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [x] No - this PR is backwards compatible

**Improvements**:

- defer loading of widget such that its only triggered on interaction or after 3 seconds

---

> [!NOTE]
> Defer AskGov and VICA widget scripts using a new interaction-based loader hook and remove ScriptComponent usage from these widgets and their types.
> 
> - **Performance**:
>   - Add `useInteractionScriptLoader` hook to delay third-party script loading until first user interaction (scroll/click/touchstart) or after 3s.
>   - Update `AskgovWidget` and `VicaWidgetClient` to load scripts via the hook and render only their container elements.
> - **API/Types**:
>   - Remove `ScriptComponent` from `AskgovWidgetProps` and `Vica*` widget props; adjust interfaces to type aliases accordingly.
> - **Integration**:
>   - Update `renderApplicationScripts` to stop passing `ScriptComponent` to `VicaWidget`/`AskgovWidget`; retain `VicaStylesheet` handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ebdefe97f2d60d1c9f15c966c0c0f0875afaa7a4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1748, opened 2025-11-27
⚡ Performance: Defer Microsoft Clarity

adriangohjw:
## Problem

Similar to https://github.com/opengovsg/isomer/pull/1593 but for Microsoft Clarity

The tradeoff is to lose users who bounce immediately, but spoke to Rachel and she said its fine since we also do not use Clarity to track non-interactive bounce rate

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- Also updated to trigger script load on `keydown` and `mousemove` to load more eagerly without affecting performance

---

> [!NOTE]
> Defers Microsoft Clarity until user interaction/timeout via a client component using an interaction-based loader, adds mousemove/keydown triggers, and removes the ScriptComponent prop.
> 
> - **Analytics: `MicrosoftClarity`**
>   - Convert to client component; inject inline `window.clarity` queue and defer loading real script via `useInteractionScriptLoader`.
>   - Remove `ScriptComponent` prop and associated inline script usage.
> - **Hook: `useInteractionScriptLoader`**
>   - Add `mousemove` and `keydown` listeners to trigger load; retains timeout fallback.
> - **Integration**
>   - Update `renderApplicationScripts` to use `<MicrosoftClarity msClarityId={...} />` without `ScriptComponent`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e96c2e9edf9fceea4a93fdf0721e2c19d0165884. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/components/src/templates/next/components/internal/MicrosoftClarity/MicrosoftClarity.tsx
+++ b/packages/components/src/templates/next/components/internal/MicrosoftClarity/MicrosoftClarity.tsx
@@ -3,7 +3,7 @@
 import { useEffect } from "react"
 
 import type { MicrosoftClarityProps } from "~/interfaces"
-import { useInteractionScriptLoader } from "~/hooks/useInteractionScriptLoader" // your hook
+import { useInteractionScriptLoader } from "~/hooks/useInteractionScriptLoader"
 
 export const MicrosoftClarity = ({ msClarityId }: MicrosoftClarityProps) => {
   // Step 1: insert the tiny inline init function


packages/components/src/templates/next/components/internal/MicrosoftClarity/MicrosoftClarity.tsx
```diff
@@ -1,20 +1,36 @@
...
+    // to not render during static site generation on the server
+    if (typeof window === "undefined") return
+
+    // @ts-expect-error - Clarity is not typed
+    if (!window.clarity) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+      ;(window as any).clarity = function () {
+        // @ts-expect-error - Clarity is not typed
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, prefer-rest-params
+        ;(window.clarity.q = window.clarity.q || []).push(arguments)
```
dcshzj: question(non-blocking): do we need the semi-colon ah, or write this in two lines?

packages/components/src/templates/next/components/internal/MicrosoftClarity/MicrosoftClarity.tsx
```diff
@@ -1,20 +1,36 @@
...
-      dangerouslySetInnerHTML={{
-        __html: `(function(c,l,a,r,i,t,y){
-          c[a]=c[a]||function(){(c[a].q=c[a].q||[]).push(arguments)};
-          t=l.createElement(r);t.async=1;t.src="https://www.clarity.ms/tag/"+i;
-          y=l.getElementsByTagName(r)[0];y.parentNode.insertBefore(t,y);
-        })(window, document, "clarity", "script", "${msClarityId}");`,
-      }}
-    />
-  )
+  return null // this component does not render anything
```
dcshzj: nit: think no need the comment since quite self-explanatory

dcshzj (approved):

packages/components/src/templates/next/components/internal/MicrosoftClarity/MicrosoftClarity.tsx
```diff
@@ -1,20 +1,36 @@
...
 import type { MicrosoftClarityProps } from "~/interfaces"
+import { useInteractionScriptLoader } from "~/hooks/useInteractionScriptLoader" // your hook
+
+export const MicrosoftClarity = ({ msClarityId }: MicrosoftClarityProps) => {
+  // Step 1: insert the tiny inline init function
+  // If code calls window.clarity() before the script loads, it would fail because window.clarity doesn't exist yet.
+  // The function below (taken from the Clarity docs) implements a queue pattern:
+  // 1. Creates a placeholder window.clarity function that queues calls
+  // 2. Stores calls in window.clarity.q array
+  // 3. When the real Clarity script loads, it processes all queued calls
```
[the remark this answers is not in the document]
adriangohjw: from what i read, it does processes the queue though if `window.clarity(...)` is called before the script loads, then those won't be tracked

in our case where we don't have custom Clarity events, we can actually remove this, but thought to keep it as-is to ensure no unknown regression

packages/components/src/templates/next/components/internal/MicrosoftClarity/MicrosoftClarity.tsx
adriangohjw: oh prettier auto does it, if i remove it then `format` will fail haha

PR #1772, opened 2025-12-16
⚡ Performance: Font optimization

adriangohjw:
## Problem

Currently our sites suffer from [FOUT](https://fonts.google.com/knowledge/glossary/fout) (flash of unstyled text) which causes CLS in certain situation 

More: https://opengovproducts.slack.com/archives/C06R4DX966P/p1765337775902339
> **what**: Some sites are below 90 because they are hit by CLS (cumulative layout shift).
**cause**: google font taking sometime to load. So when it finally loads, it adjusted the location and height of certain text divs, causing it to shift and be penalized
**why it doesn't affect all sites:** only in specific combination of letters will it result in the new text style changing the height. if you refer to the screenshots, only the "description" changes but not other text div like Notification

Before and after font loads (notice the Searchbar shifted down)
Screenshot 2025-12-16 at 12 03 23 PM

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- use `next/font` which will 1) load and self host the font locally as _next static asset, 2) preconnect to the font
- have an option to NextConfig to not include font-face to prevent duplicate loading 

### start time of font loading
| comparison | screenshots | 
| - | - |
| before (1.88 seconds) | font-current-timestart |
| after (0.66 seconds) | font-after-timestart |

### timestamp of script load
Notice the timestamp where the font (turquoise colour) loads
| comparison | screenshots | 
| - | - |
| before | font-current-gstatic-timestamp |
| after | font-after-selfhost-timestamp |

### amount of CLS from manual observation
| comparison | screenshots | 
| - | - |
| before (~1000ms) | font-current-gstatic-1000ms |
| after (~90ms) | font-after-selfhost-90ms |

### Why not preload all subsets?
Because we mostly support English sites. Preloading too many subsets using `next/font` will fire too many preloading requests and backfires - notice how the font loading itself is stacked and not concurrent:
font-overpreloading-bad-example

---

> [!NOTE]
> Adds `createNextPreset` with an `includeFonts` option (default `NextPreset` includes fonts) and switches template to load Inter via `next/font/google`, updating Tailwind configs accordingly.
> 
> - **Components / Tailwind Preset**:
>   - Introduce `createNextPreset({ includeFonts })` with extracted `interFontFaces`; conditionally injects `@font-face`.
>   - Export `NextPreset` as `createNextPreset({ includeFonts: true })` and update `presets/index.ts` exports.
> - **Template**:
>   - Load Inter via `next/font/google` and apply `className` on `<html>`.
>   - Use `createNextPreset({ includeFonts: false })` in `tooling/template/tailwind.config.js` to avoid duplicate font loading.
> - **Configs**:
>   - Update imports/usages to named `NextPreset` and new `createNextPreset` in Tailwind configs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7137271892b3001d0de5096d388aadf325de2c09. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/components/src/presets/next/index.ts
+++ b/packages/components/src/presets/next/index.ts
@@ -81,7 +81,7 @@ export function createNextPreset({
   includeFonts = true,
 }: {
   includeFonts?: boolean
-}): Config {
+} = {}): Config {
   return {
     content: [],
     theme: {


packages/components/src/presets/next/index.ts
adriangohjw: FYI: mostly whitespace changes. review this with https://github.com/opengovsg/isomer/pull/1772/changes?w=1

seaerchin (approved):
i get the change but just clarifying:

1. now, we use `next/fonts` which preloads. this prevents the flash because the font is compiled into our initial assets and loaded after the initial skeleton
2. if this is the case, why do we need `includeFonts: true` as the font would already have been loaded by next?

PR #1812, opened 2026-02-06
chore: increase thresholds

seaerchin:
## Problem
we exceeded our threat thresholds


dcshzj (approved):
I think we should just remove the fail job if threshold is exceeded step, since we verified that there is no need to block the release

dcshzj (approved):

PR #1822, opened 2026-02-11
🐞 fix: set Wogaa as <head>

adriangohjw:
> [!WARNING]
> DO NOT DELETE, currently in use for a/b testing 

## Problem

Wogaa not tracking some numbers

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- move it explicitly into `head` in the layout

## Before & After Screenshots

**BEFORE**:

`/scripts/wogaa.js` is loaded in the `body` and then pushed to the `head`

Screenshot 2026-02-11 at 5 52 17 PM

**AFTER**:

`/scripts/wogaa.js` is found in the `head` directly

Screenshot 2026-02-11 at 5 51 24 PM

**AFTER V.S. BEFORE**

on how the event is actually fired on page load in the AFTER (1st part) and only fired on interaction in the CURRENT (2nd part)

https://github.com/user-attachments/assets/4a979abd-05f2-4406-b6f5-af6c8531b49b

## Tests

already a/b tested on some sites, ref: https://opengovproducts.slack.com/archives/C06R4DX966P/p1771475040957889?thread_ts=1770804345.533419&cid=C06R4DX966P

---

> [!NOTE]
> **Medium Risk**
> Changes how a third-party analytics script is injected and where it loads (head vs body), which can affect tracking behavior and potentially impact page load/performance if the script misbehaves.
> 
> **Overview**
> Ensures WOGAA loads from the document `<head>` instead of being injected via the body script pipeline, to improve analytics tracking reliability.
> 
> This introduces `RenderApplicationHeadScripts` (exported from `engine/index.ts`) and updates the Next.js template `layout.tsx` to render it inside `<head>`. WOGAA is removed from `RenderApplicationScripts`, and the `Wogaa` component/API is simplified to always emit a plain `<script>` tag (dropping the `ScriptComponent`/`next/script` approach) so the tag exists in the head at initial render.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 250cf64938e09fe0ebeceeace1f2718fba6bca8f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/components/src/engine/renderApplicationHeadScripts.tsx
+++ b/packages/components/src/engine/renderApplicationHeadScripts.tsx
@@ -10,10 +10,10 @@ export const RenderApplicationHeadScripts = ({
 }: RenderApplicationHeadScriptsProps) => {
   return (
     <>
-      {/* NOTE: we load in wogaa regardless of whether the site is  */}
-      {/* a government site as wogaa still requires the agency to register their site */}
-      {/* and wogaa is still gated behind techpass login. */}
-      {/* Additionally, wogaa will still load but not track metrics if the site  */}
+      {/* NOTE: we load in WOGAA regardless of whether the site is  */}
+      {/* a government site as WOGAA still requires the agency to register their site */}
+      {/* and WOGAA is still gated behind techpass login. */}
+      {/* Additionally, WOGAA will still load but not track metrics if the site  */}
       {/* is not registered, so no end impact to user */}
       <Wogaa environment={site.environment} />
     </>


tooling/template/app/layout.tsx
```diff
@@ -40,6 +43,14 @@ const RootLayout = ({ children }: { children: React.ReactNode }) => {
...
       className={inter.variable}
     >
+      <head>
+        <RenderApplicationHeadScripts
+          site={{
+            ...config.site,
+            environment: process.env.NEXT_PUBLIC_ISOMER_NEXT_ENVIRONMENT,
+          }}
+        />
+      </head>
```
seaerchin: i don't think we can do this leh - should probably be using `Head` (ref: https://nextjs.org/docs/pages/api-reference/components/head)

adriangohjw: @seaerchin no Head in app router

PR #1829, opened 2026-02-19
🪄 feat: add script to rebuild all AWS CodeBuild projects

adriangohjw:
## Problem

Need a way to trigger rebuilds for all AWS CodeBuild projects (e.g. after infrastructure or config changes) without starting each project manually.

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- New script `rebuildAllCodebuildProjects.ts`: lists all CodeBuild projects, starts a build for each with `--interval` (default 5s), supports `--dry-run`, `--start-at <index>`, and `--region` (default `ap-southeast-1`). Writes timestamped logs to `scripts/logs/`.

## Tests

1. From `apps/studio`: run `npx tsx scripts/rebuildAllCodebuildProjects.ts --dry-run` to list projects without starting builds.
2. With AWS credentials configured: run `npx tsx scripts/rebuildAllCodebuildProjects.ts` (or with `--start-at`, `--interval`, `--region`) and verify builds start and logs are written to `scripts/logs/`.

**New scripts**:

- `apps/studio/scripts/rebuildAllCodebuildProjects.ts` : Rebuild all AWS CodeBuild projects. Usage: `npx tsx scripts/rebuildAllCodebuildProjects.ts [--dry-run] [--start-at <index>] [--interval <seconds>] [--region <region>]`. Requires AWS auth (e.g. `aws sso login` and `AWS_PROFILE`).

---

> [!NOTE]
> **Low Risk**
> Low risk because it adds a standalone developer script and does not affect application runtime. Operational risk remains if run unintentionally, since it can trigger builds across many CodeBuild projects and incur cost/queue pressure.
> 
> **Overview**
> Adds a new `tsx` script (`apps/studio/scripts/rebuildAllCodebuildProjects.ts`) to **bulk-trigger AWS CodeBuild builds** by listing all projects, sorting them, and starting a build for each.
> 
> The script supports `--dry-run` project listing, `--start-at` resumability, `--interval` throttling between builds, `--region` selection, and writes timestamped run logs under `scripts/logs/` (when not in dry-run).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5acfa2ec63e1ebc017245c2ac97388d9a64a345d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


seaerchin (commented):

## Diff summary

Single new file [`apps/studio/scripts/rebuildAllCodebuildProjects.ts`](https://github.com/opengovsg/isomer/blob/script/rebuildAllCodebuildProjects/apps/studio/scripts/rebuildAllCodebuildProjects.ts) — a tsx CLI that:
1. Lists all CodeBuild projects in the configured region (default `ap-southeast-1`) via paginated `ListProjects`.
2. Sorts them alphabetically by `localeCompare`.
3. Slices from `--start-at <index>` onwards.
4. Calls `StartBuild` for each, sleeping `--interval <seconds>` (default 5) between calls.
5. Writes a timestamped log to `apps/studio/scripts/logs/codebuild-rebuild-<ISO>.log` with one line per build (`started` or `ERROR`).
6. Supports `--dry-run` to print the project list without starting any builds.

Custom argv parser (no commander/yargs); `main().catch(err => process.exit(1))` at the entry point.

---

## Review

### **issue:** No confirmation prompt before bulk-starting builds (confidence: medium)
`--dry-run` is opt-**in**. Run without it and the script immediately fans out a build across every CodeBuild project in the region, racking up cost / queue pressure with no chance to abort. For a destructive-by-volume operation, a `--yes` / `--confirm` flag (or an interactive y/N prompt when stdin is a TTY) is the safer default. As written, a typo or distraction could trigger many hundreds of builds.

### **issue:** `--start-at` index is volatile (confidence: medium)
The script lists projects fresh on every run and sorts alphabetically — meaning indices shift if any project is added/removed between runs. `--start-at 42` doesn't reliably resume from where the previous run stopped. Consider `--start-after <project-name>` (or write the last successfully-started project name to a small state file alongside the log).

### **issue (minor):** `parseInt` without NaN check (confidence: high)
`parseInt(getArg("start-at") ?? "0", 10)` returns `NaN` for `--start-at abc`; `arr.slice(NaN)` returns the full array (no immediate error). Likewise `--interval abc` makes `setTimeout(resolve, NaN * 1000)` fire immediately, defeating the throttle. A `Number.isFinite` check + helpful error message would prevent surprise. Same for `interval`.

### **suggestion:** No filtering of projects (confidence: low)
"All" really means all. A `--filter <regex>` or `--include <prefix>` would let operators target subsets (e.g. only `prod-*`). Useful when only a known cohort needs the rebuild.

### **suggestion:** logs directory not gitignored (confidence: low)
`mkdirSync(logsDir, { recursive: true })` creates `apps/studio/scripts/logs/`; nothing in `.gitignore` excludes it. If a contributor commits while a log file is sitting there, it'll get tracked. Add `apps/studio/scripts/logs/` to a gitignore.

### **suggestion:** custom arg parser doesn't support `--name=value` (confidence: low)
Only `--name value` is recognized. Minor UX nit; commander/yargs is already a transitive dep via other scripts and would simplify this.

### **note:** `localeCompare` without explicit locale (confidence: low)
Sort uses system locale. Pass `"en-US"` (or `undefined, { sensitivity: "base" }`) for determinism across machines.

### **note:** no retry on failed `StartBuild` (confidence: low)
Failures get logged but not retried. A `--retries N` flag (with exponential backoff) would help if AWS is throttling.

### **praise:** good defensive coding around list-pagination, dry-run path, ISO-timestamped logs, top-level `.catch(err => process.exit(1))` (confidence: high)
The build-by-build log lines (`[time] [index] project | buildId | started`) are easy to grep/diff. The dry-run helpfully prints the indexable list with a one-liner pointing at `--start-at`.

---

## Verdict

LGTM as a one-shot ops script. The lack of a confirmation prompt is the only thing I'd push for before merge — bulk-starting builds across a whole region from a single command without prompting is the kind of footgun that bites once and then you wish you'd added it. Everything else is nice-to-have.

adriangohjw:
@seaerchin im okay to risk-accept the issues since they are unlikely to be an issue in real-life. WDYT?

apps/studio/scripts/rebuildAllCodebuildProjects.ts
```diff
@@ -0,0 +1,148 @@
...
+      projects.push(...response.projects)
+    }
+    nextToken = response.nextToken
+  } while (nextToken)
+
+  return projects.sort((a, b) => a.localeCompare(b))
+}
+
+async function main(): Promise<void> {
+  const { region, dryRun, startAt, interval } = parseArgs()
```
dcshzj: suggestion: maybe can use `@inquirer/prompts` to `confirm()` with the user that the inputs are correct before proceeding with the script?

apps/studio/scripts/rebuildAllCodebuildProjects.ts
dcshzj, line 1: suggestions: can we:
1. Maybe put this script under `isomer-admin`, and
2. Add to the README.md file there to briefly describe what it is and how to run it?

Can also consider using `@inquirer/prompts` to prompt for the questions for the arguments that you need, so you can also provide explanations as an interactive prompt on what each argument does, without having to refer to the README.md file when running.

adriangohjw: ah yes, this PR was done awhile back before the isomer-admin stuff - will make the change, thanks!

dcshzj (approved):
LGTM

PR #1838, opened 2026-03-03
fix(studio): prevent moving folder into its own descendants

seaerchin:
## Problem

The `move` mutation in the resource router lacked validation to prevent moving a folder/collection into one of its own descendants. This vulnerability could be exploited to create circular references in the resource hierarchy, causing infinite loops when resolving resource paths and leading to denial of service (database CPU exhaustion, disk quota exceeded errors).

## Solution

Added a recursive CTE query to detect all descendants of a folder/collection before allowing the move operation. If the destination is found to be a descendant of the resource being moved, the operation is rejected with a `BAD_REQUEST` error.

**Breaking Changes**

**Features**:

- N/A

**Improvements**:

- N/A

**Bug Fixes**:

- Prevents moving a folder into its direct child (A → B, then moving A into B)
- Prevents moving a folder into any deeply nested descendant (A → B → C → D, then moving A into D)
- Returns clear error message: "Cannot move a folder into one of its descendants"

## Before & After Screenshots

**BEFORE**:

Moving folder A into its child B would succeed, creating a circular reference (A → B → A) that causes infinite loops in path resolution queries.

**AFTER**:

Attempting to move a folder into any of its descendants returns a 400 error with message "Cannot move a folder into one of its descendants".
image

**resource structure in db**
image

## Tests

3 new test cases added:
- `should return 400 if moving a folder into its direct child (prevents circular reference)`
- `should return 400 if moving a folder into a deeply nested descendant (prevents circular reference)`
- `should allow moving a folder to a sibling folder (not a descendant)` - ensures valid moves still work

Run tests with:
```bash
npm run test:unit -- --testNamePattern="should return 400 if moving a folder into|should allow moving a folder to a sibling"
```

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

### DIFF
--- a/apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
+++ b/apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
@@ -1609,6 +1609,39 @@ describe("resource.router", async () => {
       expect(auditSpy).toHaveBeenCalled()
     })
 
+    it("should return 400 if moving a RootPage into its descendant (prevents circular reference)", async () => {
+      // Arrange
+      const auditSpy = vitest.spyOn(auditService, "logResourceEvent")
+      const { page: rootPage, site } = await setupPageResource({
+        resourceType: "RootPage",
+      })
+      const { folder } = await setupFolder({
+        siteId: site.id,
+        permalink: "child-folder",
+        parentId: rootPage.id,
+      })
+      await setupAdminPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+
+      // Act - try to move RootPage into its child folder (would create cycle)
+      const result = caller.move({
+        siteId: site.id,
+        movedResourceId: rootPage.id,
+        destinationResourceId: folder.id,
+      })
+
+      // Assert
+      expect(auditSpy).not.toHaveBeenCalled()
+      await expect(result).rejects.toThrowError(
+        new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Cannot move a folder into one of its descendants",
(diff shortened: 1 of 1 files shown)


apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -428,6 +428,44 @@ export const resourceRouter = router({
               })
             }
 
+            if (
+              toMove.type === "Folder" ||
+              toMove.type === "Collection" ||
+              toMove.type === "RootPage"
```
adriangohjw: nit: can use ResourceType enum?

seaerchin:
> btw i thought about it and that this doesn't prevents more circular descendants from being created, but if there's an existed one aldy, it will still cause an issue with the other recursion query we have. Im okay though as i think its safe to assume there's no existing ones since they would have tripped the alarms if there are. But just thought to point out! haha

sorry @adriangohjw abit stupid but can i clarify how this does not prevent new circular descendants from being created?

adriangohjw:
> > btw i thought about it and that this doesn't prevents more circular descendants from being created, but if there's an existed one aldy, it will still cause an issue with the other recursion query we have. Im okay though as i think its safe to assume there's no existing ones since they would have tripped the alarms if there are. But just thought to point out! haha
> 
> sorry @adriangohjw abit stupid but can i clarify how this does not prevent new circular descendants from being created?
> 
> 

@seaerchin sorry typo, I meant this **only** prevents new ones from being created 🤦  psps for the confusion

PR #1844, opened 2026-03-03
Define display-xs class to improve type hierarchy on Article layout

sehyunidaaa:
#### Current typography setup
- Isomer Next has 4 display type tokens: `display-xl` to `display-sm`.
- `display-xl` is reserved for certain hero banner design's titles.
- On other layouts, we use `display-lg` for large titles. 

#### But the Article layout has a different pattern 
- However, the Article page layout is an exception - it uses `display-md` for titles. 
- This is because government agencies have press releases or news articles that have very long titles; simplifying titles is challenging due to agencies often wanting to capture as many details in the title.
- If we use `display-lg` for titles on Article pages, this will be very jarring and hard to scan.
- But, because Prose heading styles are fixed for every layout, this means that Section headings that also use `display-md` will be as big as the Article's title. This lack of hierarchy is a problem.

Closes [ISOM-2210]

## Solution
- Defined a `display-xs` class, which uses the old `display-sm` sizing. The old `display-md` class has now become `display-sm`. 
- As a result,`display-md` is only used for the Article page title. 
- This impacts all components that were previously using `display-md`.

**Breaking Changes**

  - Details ...

## Disclaimers

> ⚠️  From a designer's POV, I'm fully aware that this solution is not ideal for Design System governance, but it solves the problem the most cleanly at this stage without introducing too much debt. I've experimented with a full typography refresh, but it's unwarranted now and will cause a lot of visible change to sites (that we don't need). I'll be reviewing tokens and breakpoints at a future date to reduce Design Ops debt. 

> ❓ Why not style headings differently for every layout? - We've not explored this as a product and this creates the risk of components/layouts that are very hard to maintain. 

## Tests

- On Chromatic, review that Article page layout has better hierarchy.
- On Chromatic, review all diffs captured to ensure that no components changed unintentionally.

### DIFF
--- a/packages/components/src/presets/next/typography.ts
+++ b/packages/components/src/presets/next/typography.ts
@@ -17,12 +17,12 @@ export const isomerTypography = plugin(({ addComponents, theme }) => {
     ".prose-display-md": {
       fontWeight: theme("fontWeight.semibold"),
       letterSpacing: "-0.022em",
-      "@apply text-[2rem] lg:text-[2.438rem] leading-[1.2]": {},
+      "@apply text-[1.75rem] lg:text-[2.25rem] leading-[1.2]": {},
     },
     ".prose-display-sm": {
       fontWeight: theme("fontWeight.semibold"),
       letterSpacing: "-0.022em",
-      "@apply text-[1.75rem] lg:text-[1.938rem] leading-[1.2]": {},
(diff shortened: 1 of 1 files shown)


adriangohjw (approved):
need to update preview-tw.css? anyway approving since its mostly design related stuff!

PR #1849, opened 2026-03-03
🧹 Refactor/remove barrel exports

adriangohjw:
## Problem

Barrel export is bad for bundle size. We remove it as much as possible to reduce footgun

closes https://linear.app/ogp/issue/ISOM-2217/refactor-remove-barrel-export

## Solution

Import not from barrel export files whenever possible

**Breaking Changes**

  - Details ...

---

> [!NOTE]
> **Low Risk**
> Primarily rewires imports away from barrel `index` files to direct paths, with minimal behavioral change; main risk is missed/incorrect import paths causing build-time failures.
> 
> **Overview**
> Refactors components and engine code to **stop importing from barrel exports** (e.g. `~/utils`, `~/types`, `~/interfaces`) and instead import from **direct module paths** like `~/utils/isExternalUrl`, `~/utils/tailwind`, `~/types/schema`, and `~/interfaces/internal/Link`.
> 
> Removes a couple of local barrel re-exports (`Link/utils/index.ts`, `internal/utils/index.ts`) and updates call sites (e.g. `GoogleTagManagerPreload`, `Link`) to import helpers directly, reducing bundling/tree-shaking footguns.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a215d59537ee5d5d92dfc60f0e03e12a753389bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/components/src/templates/next/components/internal/Link/utils/index.ts
+++ b/packages/components/src/templates/next/components/internal/Link/utils/index.ts
@@ -1,1 +1,0 @@
-export { getReactNodeText } from "./getTextContent"
--- a/packages/components/src/templates/next/components/internal/utils/index.ts
+++ b/packages/components/src/templates/next/components/internal/utils/index.ts
@@ -1,1 +1,0 @@
-export * from "./PreloadHelper"


adriangohjw:
> clarification: my understanding is that treeshaking is impossible for `components` because we import it as a single entrypoint (`RenderEngine -> layouts -> components`), which means that the bundler, on build, has to keep all its of `RenderEngine`'s dependencies.
> 
> because of this, wouldn't barrel exports be ok since removing it wouldn't lead to a decrease in bundle size and keeping it in is an improvement in dx?
> 
> cmiiw cos atm this is my understanding

@seaerchin you are correct! this PR doesn’t actually reduce the bundle size right now. However, we previously observed that messy barrel exports, or barrels that import heavier modules, can cause unintended files to be pulled into the bundle. For example, importing a component through a barrel file instead of directly could end up bundling things like typebox.

Because our current setup has nested barrel exports, it can also make Webpack’s treeshaking less reliable and reduce its effectiveness. So the main goal of this PR is to prevent these kinds of unexpected footguns in the future.

I agree there is a slight drop in DX, but most of the imports remain fairly straightforward. In my view, as long as Studio can still consume a barrel export file, the DX trade-off is acceptable.

WDYT?

PR #1854, opened 2026-03-04
chore: revert unsafe-eval CSP restriction to development only

seaerchin:
## Problem

PR #1847 changed the Content Security Policy (CSP) to only allow `'unsafe-eval'` in development mode. However, this caused issues because some dependencies or features require `'unsafe-eval'` in production as well.

## Solution

Reverted the conditional `'unsafe-eval'` back to always being present in the CSP `script-src` directive, regardless of environment.

**Breaking Changes**

**Bug Fixes**:

- Reverted CSP change that was causing script execution issues by conditionally removing `'unsafe-eval'` in production

## Tests

Verify that the application works correctly in both development and production environments without CSP violations.

```bash
npm run build && npm run start
```

### DIFF
--- a/apps/studio/next.config.mjs
+++ b/apps/studio/next.config.mjs
@@ -4,6 +4,8 @@
  */
 const { env } = await import("./src/env.mjs")
 
+// NOTE: Keep the `unsafe-eval` for `script-src` as the removal
+// led to nextjs crashing on start
 /*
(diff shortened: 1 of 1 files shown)


adriangohjw (approved):
can add a comment somewhere at the top that this is needed cos of nextjs or something? to prevent future repeat of such issue haha thanks!

PR #1870, opened 2026-03-10
chore: only notify channel on production releases

dcshzj:
## Problem

The Slack channel gets pinged for every deploy, including staging and UAT, which can be quite noisy.

## Solution

**Breaking Changes**

**Improvements**:

- Perform the Slack notification only if the deployment is for the production environment.


adriangohjw (approved):
nice thanks!

PR #1930, opened 2026-03-26
fix: remove fallback for original behaviour

dcshzj:
## Problem

Original behaviour was an error but apparently it worked for people.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Restore original logic


dcshzj:
Replaced with #1932

PR #1932, opened 2026-03-27
feat: enhance collection index page with configurable sort, thumbnail, and date controls

dcshzj:
## Problem

Collection index pages had limited customisation options. Site editors could not control sort order via a single unified field, toggle thumbnail/date visibility, or set a fallback image when items lacked thumbnails. Additionally, the `Site.name` column was not kept in sync when updating site config, and `getSiteNameAndCodeBuildId` did not prefer the canonical `config.siteName` value.

## Solution

**Breaking Changes**

**Features**:

- Added a new unified `sortOrder` field (e.g. `date-desc`, `title-asc`, `category-desc`) to replace the separate `defaultSortBy` + `defaultSortDirection` fields (old fields kept for backwards compatibility)
- Added `showThumbnail` toggle — when enabled, items without images fall back to the site logo with `object-contain` styling
- Added `showDate` toggle to control date visibility on collection items
- Added `image` field to collection page schema for custom thumbnails
- Exposed tag category management to Isomer admin users in the collection editor drawer
- Replaced the tabbed collection editor drawer UI with a streamlined single-panel layout including a back button and info box

**Improvements**:

- `Site.name` column is now kept in sync when `updateSiteConfig` is called (`site.router.ts`)
- `getSiteNameAndCodeBuildId` now prefers `config.siteName` over `Site.name`, falling back to `Site.name` when `siteName` is absent (`site.service.ts`)
- Collection editor metadata drawer now excludes fields already covered by the new collection index editing experience
- Updated `CollectionCard` and `BlogCard` to support `isFallbackImage` prop for `object-contain` vs `object-cover` styling
- Sitemap and publishing pipeline now propagate the `sortOrder` field

**Bug Fixes**:

- Fixed collection index thumbnail logic (reverted previous revert)
- Removed fallback for original sort behaviour

## Before & After Screenshots

N/A — UI changes are in the collection editor drawer and collection page rendering.

## Tests

- Added new test suite `getCollectionItems.test.ts` covering image fallback to site logo behaviour
- Added new test suite `sortCollectionItems.test.ts` with `sortOrder`-based sorting assertions alongside existing `sortBy`/`sortDirection` tests
- Updated `site.router.test.ts` to verify `Site.name` column is updated when `updateSiteConfig` is called

**New scripts**:

None

**New dependencies**:

None

**New dev dependencies**:

None

### DIFF
--- a/packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
+++ b/packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
@@ -53,71 +53,73 @@ export const getCollectionItems = ({
     return []
   }
 
-  const items = currSitemap.children.flatMap((child) =>
-    getSitemapAsArray(child),
-  )
-
-  const transformedItems = items
+  const items = currSitemap.children
+    .flatMap((child) => getSitemapAsArray(child))
     .filter(
       (item) =>
         item.layout === "file" ||
         item.layout === "link" ||
         item.layout === "article",
     )
-    .map((item) => {
-      const date =
-        showDate !== false && item.date !== undefined && item.date !== ""
-          ? getParsedDate(item.date)
-          : undefined
-      const hasOriginalImage = !!item.image?.src
-      const image =
-        showThumbnail !== false
-          ? hasOriginalImage
-            ? item.image
-            : {
-                src: site.logoUrl,
-                alt: `${site.siteName} site logo`,
-              }
-          : undefined
 
-      const baseItem = {
-        type: "collectionCard" as const,
-        id: item.permalink,
-        date,
-        lastModified: item.lastModified,
-        category: item.category || CATEGORY_OTHERS,
-        title: item.title,
-        description: item.summary,
-        image,
-        isFallbackImage: showThumbnail !== false && !hasOriginalImage,
-        site,
-        tags:
-          tagCategories && item.tagged
-            ? getTagsFromTagged(item.tagged, tagCategories)
-            : item.tags,
-      }
+  const isAnyItemHaveImage = items.some((item) => !!item.image?.src)
+  // If showThumbnail is not explicitly set, show the thumbnail if any of the
+  // items have an image
+  const shouldShowThumbnail = showThumbnail ?? isAnyItemHaveImage
 
-      if (item.layout === "file") {
-        return {
-          ...baseItem,
-          variant: "file",
-          url: item.ref,
-          fileDetails: item.fileDetails,
-        }
-      } else if (item.layout === "link") {
-        return {
-          ...baseItem,
-          variant: "link",
-          url: item.ref,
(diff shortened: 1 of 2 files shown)


packages/components/src/templates/next/layouts/Collection/utils/sortCollectionItems.ts
```diff
@@ -262,19 +275,34 @@ const sortCollectionItemsByCategory = ({
...
   items,
+  sortOrder,
   sortBy,
   sortDirection,
 }: SortCollectionItemsProps): AllCardProps[] => {
-  switch (sortBy) {
+  const derivedSortBy = sortOrder ? (sortOrder.split("-")[0] as SortBy) : sortBy
+  const derivedSortDirection = sortOrder
+    ? (sortOrder.split("-")[1] as SortDirection)
+    : sortDirection
```
adriangohjw: non-blocking nit: maybe something like this will be more readable than splitting it twice and manually getting 1st and 2nd item?

```ts
const [orderBy, orderDirection] = sortOrder?.split("-") ?? []

const derivedSortBy = orderBy ?? sortBy
const derivedSortDirection = (orderDirection as SortDirection) ?? sortDirection
```

packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
```diff
@@ -48,63 +53,77 @@ export const getCollectionItems = ({
...
+  const isAnyItemHaveImage = items.some((item) => !!item.image?.src)
+  // If showThumbnail is not explicitly set, show the thumbnail if any of the
+  // items have an image
+  const shouldShowThumbnail = showThumbnail ?? isAnyItemHaveImage
+
+  const transformedItems = items.map((item) => {
+    const date =
+      showDate !== false && item.date !== undefined && item.date !== ""
+        ? getParsedDate(item.date)
+        : undefined
```
adriangohjw: when `showDate` is false, the date field is set to undefined on every item before sorting, which means if a user configures `showDate: false` with `sortOrder: "date-desc"`, the sort function sees all items as having no date and falls back to alphabetical title sorting instead. is this intended?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -3,8 +3,6 @@ import type {
   IsomerSchema,
 } from "@opengovsg/isomer-components"
 import {
-  COLLECTION_PAGE_DEFAULT_SORT_BY,
-  COLLECTION_PAGE_DEFAULT_SORT_DIRECTION,
```
adriangohjw: minor ask: since no longer imported, we should remove both of their `export` from `packages/components/src/types/page.ts`

adriangohjw (commented):
thanks! refer to slight comments below

PR #1936, opened 2026-03-27
feat(studio): allow adding and deleting child pages block on index pages

seaerchin:
## Problem

Closes ISOM-2244

Previously, the children pages block was automatically added to index pages and could not be deleted. Users had no control over whether to display the child pages block or how to manage it.

## Solution

This PR gives users full control over the children pages block on index pages:

1. **Allow deletion of children pages block** - Removed the restriction that prevented users from deleting the children pages block
2. **Allow manual addition of children pages block** - Added "Child pages" to the block selector for index pages so users can add it when needed
3. **Prevent duplicate blocks** - Disabled the "Child pages" option in the block selector when one already exists on the page
4. **Removed auto-insertion behavior** - Removed the `useEffect` that automatically inserted the children pages block on index pages

**Breaking Changes**

Note: New index pages created via folder creation will still have the children pages block auto-added (via `createFolderIndexPage`), but existing pages without the block will no longer have it forcefully inserted.

**Features**:

- Users can now add a "Child pages" block to index pages via the block selector
- Users can now delete the children pages block from index pages
- Added preview image for the children pages block in the block selector

**Improvements**:

- Updated block description to "Automatically display all child pages in this folder"
- Block selector shows "Child pages" option as disabled when one already exists (prevents duplicates)
- Cleaned up unused CSS classes in preview-tw.css

**Bug Fixes**:

- None

## Before & After Screenshots

**BEFORE**:

- Children pages block could not be deleted (no trash icon)
- Children pages block was auto-inserted on page load
- No way to manually add children pages block

**AFTER**:

- Children pages block can be deleted like any other block
- Children pages block can be manually added from block selector
- Block selector shows disabled state when block already exists

## Tests

New Storybook stories added:
- `AddBlockWithChildrenPagesDisabled` - Tests that the child pages option is disabled when one already exists
- `AddBlockWithChildrenPagesEnabled` - Tests that the child pages option is enabled when none exists

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

### DIFF
--- a/packages/components/src/templates/next/render/renderPageContent.ts
+++ b/packages/components/src/templates/next/render/renderPageContent.ts
@@ -21,13 +21,8 @@ export const renderPageContent = ({
   ...rest
 }: RenderPageContentParams) => {
   // Filter out hidden childrenpages blocks
-  const visibleContent = content.filter(
-    (component) =>
-      !(
-        component.type === "childrenpages" &&
-        "isHidden" in component &&
-        component.isHidden
-      ),
+  const visibleContent = content.filter((component) =>
+    component.type === "childrenpages" ? !component.isHidden : true,
   )
 
   // Find index of first component with image


apps/studio/src/components/PageEditor/ComponentSelector.tsx
```diff
@@ -239,12 +248,18 @@ function ComponentSelector() {
             <BlockList>
               {section.types.map((type) => {
                 const blockMeta = BLOCK_TO_META[type]
+                const isDisabled = savedPageState.content.some(
+                  (block) =>
+                    block.type === "childrenpages" && type === "childrenpages",
+                )
```
dcshzj: suggestion: can extract this to a separate function? Seems like selectively disabling blocks based on what is already on the page is quite useful (e.g. no more than 10 images on the page)

Can ignore if it's premature to do now

apps/studio/src/contexts/EditorDrawerContext.tsx
```diff
@@ -73,53 +59,6 @@ export function EditorDrawerProvider({
...
-          status: "error",
-          title: "Failed to update blocks",
-          description: error.message,
-          ...BRIEF_TOAST_SETTINGS,
-        })
-      },
-    },
-  )
-
-  useEffect(() => {
```
dcshzj: question(non-blocking): just want to confirm my understanding of this, we remove this auto-add cos now we allow users to add the children pages block themselves and no longer want to do this automatically for them?

packages/components/src/templates/next/render/renderPageContent.ts
```diff
@@ -20,14 +20,24 @@ export const renderPageContent = ({
   content,
   ...rest
 }: RenderPageContentParams) => {
+  // Filter out hidden childrenpages blocks
+  const visibleContent = content.filter(
+    (component) =>
+      !(
+        component.type === "childrenpages" &&
+        "isHidden" in component &&
+        component.isHidden
```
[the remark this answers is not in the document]
seaerchin: it was to avoid type errors but updated for clarity!

PR #1960, opened 2026-04-01
Improve favicon upload limits and format guidance

adriangohjw:
## Problem

Favicon configuration in site settings used the same MIME allowlist as generic images and a 50 kB cap, while product copy referenced formats (e.g. `.ico`) that were not consistently reflected in validation. Upload hints in Studio always expressed limits in megabytes, which was misleading when the real cap was under 1 MB.

## Solution

Introduced `formatFileSizeLimit` in Studio so maximum upload size copy uses **KB below 1 MB** and **MB from 1 MB upward**, with stable rounding. Centralised favicon upload rules in `packages/components` via `FAVICON_ACCEPTED_MIME_TYPE_MAPPING` and updated `LogoSettingsSchema` to use that mapping, align the description with supported types, and lower the favicon `maxSizeInBytes` to **20 kB**.

**Breaking Changes**

  - Details ...

Studio users uploading a new favicon may hit stricter type and size checks than before (narrower than the previous generic image list, and a lower byte cap). Existing published favicon URLs are unchanged until someone replaces the asset in Studio.

**Features**:

- Favicon field uses a dedicated accepted-type mapping (`.png` / `.svg` first in UI order, plus `.jpg` / `.jpeg` / `.webp`) and a 20 kB maximum in schema.

**Improvements**:

- File attachment and image meta controls show human-readable max size labels via `formatFileSizeLimit` (KB vs MB) instead of dividing by 1 MB only.

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

N/A — copy and validation limit change only; no layout change.

**AFTER**:

N/A

## Tests

- [ ] In **Studio**, open a page editor flow that includes a **file attachment** control and confirm the “Maximum file size” line matches the configured limit (KB if under 1 MB, MB at or above 1 MB).
- [ ] In **Studio**, open a block or field that uses the **meta image** control and confirm the “Maximum file size” line uses the same KB/MB formatting.
- [ ] In **Studio** site settings (logo / favicon), confirm the favicon field description mentions **.png** or **.svg** (and related accepted types), and that uploading a favicon larger than **20 kB** or a disallowed type (e.g. **.gif** or **.bmp**) is rejected as expected.
- [ ] Confirm uploading a small **.png** or **.svg** favicon within the limit still succeeds.

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it tightens favicon validation (new MIME allowlist and lower 20kB size cap), which may block previously acceptable uploads. UI text-only changes for file size display are low risk but touch common upload components.
> 
> **Overview**
> Studio upload controls now display maximum file size using a new `formatFileSizeLimit` helper (KB for limits under 1 MB, otherwise MB) instead of always showing MB.
> 
> Favicon validation is narrowed and centralized via `FAVICON_ACCEPTED_MIME_TYPE_MAPPING`, and `LogoSettingsSchema` updates the favicon description and reduces `maxSizeInBytes` from 50 kB to 20 kB, enforcing stricter type/size checks on new favicon uploads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 900ed9a1334f1d5b2e4d536c4267dc7ec04dfa3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/components/src/constants/image.ts
```diff
@@ -8,3 +8,14 @@ export const IMAGE_ACCEPTED_MIME_TYPE_MAPPING: Record<string, string> = {
   ".bmp": "image/bmp",
   ".webp": "image/webp",
 }
+
+/** Subset of {@link IMAGE_ACCEPTED_MIME_TYPE_MAPPING};
+ * key order is used for UI lists in Studio (.png and .svg first).
+ * .jpg, .jpeg, and .webp are also accepted. */
+export const FAVICON_ACCEPTED_MIME_TYPE_MAPPING: Record<string, string> = {
```
dcshzj, line 15: nit: should we do a `pick` instead from the superset?

packages/components/src/types/site.ts
```diff
@@ -69,10 +72,10 @@ export const LogoSettingsSchema = Type.Object({
     Type.String({
       title: "Favicon",
       description:
-        "This appears on a browser tab to help people recognise your site. We recommend a minimum size of 24px by 24px, in .ico, .svg, or .png format.",
+        "This appears on a browser tab to help people recognise your site. We recommend a minimum size of 24px by 24px, in .png or .svg format.",
```
dcshzj, line 71: nit: can we document as an ADR that we are dropping support for .ico files?

If I understand correctly it's due to performance but might need to elaborate a bit cos I'm not sure how a 50kb .ico file impacts performance. I'm ok with dropping anyway since online sources seem to suggest going for .svg/.png files.

dcshzj (approved):

packages/components/src/constants/image.ts
adriangohjw: yeah thought about it but we want them in a specific order for this use case, so all things considered, i feel like it will be more readable and direct to duplicate

packages/components/src/types/site.ts
adriangohjw: done https://www.notion.so/opengov/Architectural-Decision-Records-ADR-8af42b8a723d4efd9643868b4f075b57?p=33e77dbba788804388d1f949a4655814&pm=s

PR #1962, opened 2026-04-01
feat: automatically append file type and size to link text on file upload

### LINKED ISSUE #1911: Add file type and size automatically
## Summary

When a file is uploaded and shown as a download link, the visible title should include **file type** and **file size** in brackets so reviewers know what they’re downloading without opening the file.

## Background

Migrators currently add `[type, size]` to link text by hand. That’s repetitive, easy to get wrong, and slows migration work.

## Problem

Manually filling file type and size for every uploaded file is high effort and inconsistent. Users reviewing pages should see type and size **automatically** in the link label.

## Approach

Compute and store **file type** and **file size** **at upload time** (calculate once when the file is saved), then use that data when rendering the link title.  

Note: image and document uploads are currently separate from the link/text field, so some integration or flow work is needed to attach metadata to the right link or field.

## Goal

On upload, derive **MIME/extension → human-readable type** (e.g. PDF) and **byte size → readable size** (e.g. 280 KB) and persist them so the UI can show a consistent label such as:  
**Example:** `Download National Day Rally Speech [PDF, 280 KB] ↗`

## Acceptance criteria

1. For an uploaded file used as a download link, the **link text** includes type and size after the title, in square brackets:  
   **Example:** `Download National Day Rally Speech [PDF, 280 KB] ↗`  
   (Exact wording of the leading phrase may follow existing product copy; the bracketed part is the requirement.)

2. **File type** and **size** are determined **when the file is uploaded** and stored or associated with the asset so they do not need to be re-derived on every page load.

3. **File type** is shown in a predictable, user-facing form (e.g. PDF, DOCX, XLSX—not raw MIME strings unless that’s the only option).

4. **File size** uses a standard human-readable format (appropriate unit: KB/MB as usual for the product; consistent rounding/precision with other file displays).

5. Behaviour applies **without** migrators manually editing the title for type/size.

6. If type or size is **unavailable** for an asset, behaviour is defined (e.g. omit missing part, or show “—”) and documented so it’s not blocking.

## Out of scope (unless product says otherwise)

Note: the external-link icon (↗) is automatically rendered by the components. This issue focuses on appending the file type and file size into the prose editor content

brendancjz:
## Problem

Closes #1911

Closes https://linear.app/ogp/issue/ISOM-1907/add-file-type-and-size-automatically

When editors upload a file via the prose link editor, the link text gave no indication of the file type or size. Users reading the published page had no way to know what they were about to download.

## Solution

When a file is uploaded through the link editor modal, the link text is automatically suffixed with the file's type and size in the format ` [TYPE, size]` (e.g. ` [PDF, 1.00 MB]`). The suffix is hidden from the editor input so editors can write natural link text, and is stripped + re-appended on save to avoid stacking duplicates when a file is re-uploaded.

**Breaking Changes**

  - Details ...

**Features**:

- Automatically appends a ` [TYPE, size]` suffix to file link text when a file is uploaded via the link editor (e.g. `Download form [PDF, 1.00 MB]`)
- Suffix is only shown on the published page; the link text input in the editor displays the clean text without the suffix

**Improvements**:

- `FileAttachment` component now accepts an optional `onUploadedFile` callback, invoked after a successful upload with the raw `File` object
- Extracted and co-located file-upload metadata utilities (`buildFileUploadMetaSuffix`, `stripFileUploadMetaSuffix`, `toDisplayType`) under the `useLinkEditorFileMetaSuffix` hook folder for better discoverability
- `stripFileUploadMetaSuffix` validates bracket content against known display types and size format before stripping, preventing false positives on user-written brackets

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

**AFTER**:

https://github.com/user-attachments/assets/fa2932e9-436b-496b-a22e-d53c37acacc3

## Tests

- [ ] Open a page in the prose editor and select some text to add a link
- [ ] In the link editor modal, switch to the "File" link type and upload a file (e.g. a PDF)
- [ ] Verify that the link text input shows the clean text without any suffix while the modal is open
- [ ] Save the link and confirm the published/preview output appends ` [TYPE, size]` to the link text (e.g. `Download form [PDF, 1.00 MB]`)
- [ ] Re-open the link editor for the same link and re-upload a different file; verify the suffix updates correctly and does not stack duplicates
- [ ] Manually type `[PDF, 1.00 MB]` into the link text and save — confirm it is preserved and not stripped (user-written brackets with valid-looking content should be left alone if they were not auto-generated)

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- `@testing-library/react`, `happy-dom` — added to pnpm workspace config to support unit tests for the new hook


brendancjz:
> @brendancjz hi sorry, i was busy and haven't gotten a chance to review this

Hi hi @adriangohjw, any updates for this? :)

apps/studio/src/hooks/useLinkEditorFileMetaSuffix/useLinkEditorFileMetaSuffix.ts
```diff
@@ -0,0 +1,84 @@
...
+  linkText,
+  linkHref,
+  fileMetaSuffix,
+}: {
+  showLinkText: boolean
+  linkText: string
+  linkHref: string
+  fileMetaSuffix: string | null
+}): string {
+  if (!showLinkText || getLinkHrefType(linkHref) !== LINK_TYPES.File)
```
seaerchin: question: do you think this would also be similarly useful for images? i'm thinking we add it there too and also help our migrators out. wdyt @adriangohjw @brendancjz

seaerchin (commented):
happy to approve, but seeing if @adriangohjw has any comments since he created the issue!

apps/studio/src/hooks/useLinkEditorFileMetaSuffix/useLinkEditorFileMetaSuffix.ts
adriangohjw: @seaerchin how does it work if its for images ah? cos users cannot upload images as a file upload

seaerchin: we can call into it via `FileAttachment` - it also accepts a `onUploadedFile` props.

at present, we call a `buildFinalLinkTextForSave` in our `LinkEditorModal` + `onUploadedFile` from our hook - we need to shift these either 1 more level up or update the hook to do more

seaerchin (commented):

## Diff summary

Adds a UX where file uploads in the prose link editor automatically append `[TYPE, SIZE]` to the displayed link text (e.g. `Annual Report [PDF, 1.20 MB]`).

**Code flow**
- New util [`apps/studio/src/utils/fileUploadMetaSuffix.ts`](https://github.com/opengovsg/isomer/blob/feat/add-file-type-size-suffix/apps/studio/src/utils/fileUploadMetaSuffix.ts):
  - `EXTENSION_TO_DISPLAY_SUFFIX` derived from `FILE_UPLOAD_ACCEPTED_MIME_TYPE_MAPPING` (`.pdf` → `PDF`, etc.).
  - `DISPLAY_TYPE_SUFFIX` joins display strings into a regex alternation, **sorted by length desc** so XLSX matches before XLS.
  - `FILE_UPLOAD_META_SUFFIX_RE` strips `\s[TYPE]`, `\s[TYPE, SIZE]`, or `\s[SIZE]` at end of string.
  - `buildFileUploadMetaSuffix(file)` returns ` [TYPE, SIZE]` (omits parts that aren't available).
- New hook [`apps/studio/src/hooks/useLinkEditorFileMetaSuffix.ts`](https://github.com/opengovsg/isomer/blob/feat/add-file-type-size-suffix/apps/studio/src/hooks/useLinkEditorFileMetaSuffix.ts): on open, strips suffix from incoming text and snapshots it in a `useRef`; on upload, replaces the ref with the new file's suffix; on save, calls `resolveLinkTextForSave` which re-appends only if `linkHref` is a file link.
- [`FileAttachment.tsx`](https://github.com/opengovsg/isomer/blob/feat/add-file-type-size-suffix/apps/studio/src/components/PageEditor/FileAttachment.tsx): new optional `onUploadedFile?: (file: File) => void` callback fired in the `onSuccess` of `getPresignedPutUrl`.
- [`LinkEditorModal.tsx`](https://github.com/opengovsg/isomer/blob/feat/add-file-type-size-suffix/apps/studio/src/components/PageEditor/LinkEditorModal.tsx): wires the hook — uses `strippedLinkTextForForm` as the form's `defaultValue`, passes `onUploadedFile` through to `ModalLinkEditor` → `FileAttachment`, and calls `buildFinalLinkTextForSave` in the form submit handler.
- Tests in [`fileUploadMetaSuffix.test.ts`](https://github.com/opengovsg/isomer/blob/feat/add-file-type-size-suffix/apps/studio/src/utils/__tests__/fileUploadMetaSuffix.test.ts) for util only.

---

## Review

### **issue:** Silent loss of suffix-shaped text on non-file links (confidence: medium)
`useLinkEditorFileMetaSuffix` strips the suffix from `initialLinkText` **unconditionally** on open, but `resolveLinkTextForSave` only re-appends it when `getLinkHrefType(linkHref) === LINK_TYPES.File`. So if a user has a **page link** with text like `Click here [PDF, 1.00 MB]` (rare but possible — e.g. it described a downloadable PDF that's now on a different page), opening and saving the modal silently drops the bracketed portion.

Two ways to fix:
1. Pass `initialLinkHref` into the hook and only strip when it's a file link.
2. Re-append the snapshotted suffix even on non-file links (treat the suffix as opaque text, only update it on actual upload).

Option 2 is closer to a minimal fix.

### **issue (minor):** Test coverage gap on accepted extensions (confidence: medium)
[`fileUploadMetaSuffix.test.ts`](https://github.com/opengovsg/isomer/blob/feat/add-file-type-size-suffix/apps/studio/src/utils/__tests__/fileUploadMetaSuffix.test.ts) covers PDF/XLSX/XLS. The accepted set is pdf/xls/xlsx/csv/tsv/doc/docx. Adding a parametric `it.each` over all extensions would (a) catch any future extension-to-display-suffix mapping bug and (b) cover the strip/round-trip for each. The hook itself has no tests.

### **suggestion:** `fileMetaSuffixRef` reset on save loses state if save fails (confidence: low)
```ts
const final = resolveLinkTextForSave(...)
fileMetaSuffixRef.current = null
return final
```
If `onSave` throws or the mutation fails, the suffix is gone. In practice the modal is re-mounted on retry, so this is unlikely to bite — but if a future change adds in-place retry, the suffix would be missing. Resetting on a confirmed success (or simply not resetting) would be more defensive.

### **note:** loose numeric pattern in size regex (confidence: low)
`[\d.]+` matches `...`, `1.2.3`, etc. Used only for stripping previously-generated suffixes, where `formatBytes` always emits well-formed `12.34 MB`. Not exploitable, but `\d+(?:\.\d+)?` would be more correct.

### **praise:** XLSX-before-XLS alternation sort (confidence: high)
`.sort((a, b) => b.length - a.length)` ensures longer extensions match first in the regex alternation, and the test explicitly covers `File [XLSX, 1.00 MB]` to lock this behaviour in. Nicely thought-through.

### **praise:** edge case with `formatBytes` returning undefined (confidence: high)
`buildFileUploadMetaSuffix` correctly handles `formatBytes(0)` (returns `undefined`) by falling through to the `parts.length === 0 → ""` path. No malformed `[PDF, undefined]` output.

---

LGTM with minor nits. The non-file-link-strip case (#1) is the only behavioural concern worth fixing before merge; the rest are coverage / defensive notes.

adriangohjw (commented):
**THIS IS AN AUTOMATED REVIEW BY CLAUDE**

## Summary

Auto-appends a `[TYPE, SIZE]` suffix to link text after a file upload in the prose link editor.

**Code flow**
- `apps/studio/src/utils/fileUploadMetaSuffix.ts` — `buildFileUploadMetaSuffix(file)` produces ` [PDF, 1.00 MB]`; `stripFileUploadMetaSuffix(text)` removes it via a regex built from the accepted-extensions list, with alternatives sorted longest-first so `XLSX` matches before `XLS` (`fileUploadMetaSuffix.ts:13-14`).
- `apps/studio/src/hooks/useLinkEditorFileMetaSuffix.ts` — on mount, snapshots the suffix off `initialLinkText` into a `useRef` and feeds the **stripped** value into the form as `defaultValues.linkText` (`useLinkEditorFileMetaSuffix.ts:44-53`). `onUploadedFile` overwrites the ref with the new file's suffix. `buildFinalLinkTextForSave` re-appends only if the link resolves to a file via `getLinkHrefType` (`useLinkEditorFileMetaSuffix.ts:13-29`).
- `FileAttachment.tsx:63` — new `onUploadedFile` callback fires inside `uploadFile.onSuccess`.
- `LinkEditorModal.tsx:84-117` — wires the hook into the form's default value and submit handler. `onUploadedFile` is threaded through `ModalLinkEditor` (`LinkEditorModal.tsx:162`) → `FileAttachment`.
- `packages/components/src/index.ts:12` — re-exports `formatBytes`.

**Overall**: Approach is reasonable. Tests cover the util surface but not the hook, which is where the actual product behavior lives. Two correctness issues below.

---

## Findings

### **issue:** Silent data loss when text shaped like a suffix is on a non-file link
`useLinkEditorFileMetaSuffix.ts:44` strips the suffix from `initialLinkText` unconditionally, but `useLinkEditorFileMetaSuffix.ts:24-25` only re-appends it when `getLinkHrefType(linkHref) === LINK_TYPES.File`. So a **page / external / email** link whose text happens to be `Annual Report [PDF, 1.20 MB]` (e.g. a historical reference to a file that has since moved, or someone describing a download by-name) will silently lose the bracketed portion the next time the modal is opened and saved — even if the user changed nothing.

Fix options:
1. Pass `initialLinkHref` into the hook and only strip when it resolves to a file link.
2. Always re-append the snapshotted suffix unless the user actually edited the text (treat the trailing bracket as opaque text whenever no fresh upload has occurred).

Option 1 is the smaller change and preserves the "no stacking on re-upload" guarantee.

Confidence: high

### **issue:** Same data-loss path when `showLinkText` is `false`
`useLinkEditorFileMetaSuffix.ts:44` strips unconditionally, but `useLinkEditorFileMetaSuffix.ts:24` early-returns `linkText` (i.e. the stripped form value) when `!showLinkText`. `onUploadedFile` is also disabled in that mode (`useLinkEditorFileMetaSuffix.ts:81`), so the ref is never repopulated. Net effect: opening a `showLinkText=false` file-link modal and saving drops any existing suffix. Easiest fix is to skip stripping entirely when `!showLinkText` (the linkText isn't user-editable in that mode anyway, so re-using the original is the safe default).

Confidence: medium

### **issue:** Hook has no tests
`useLinkEditorFileMetaSuffix.ts` is where the product behavior lives — the round-trip of stripping on open, snapshotting, swapping on re-upload, and re-appending on save. Util tests in `fileUploadMetaSuffix.test.ts` don't exercise any of that. At minimum I'd want tests for: (a) file link round-trip preserves the suffix when nothing changes, (b) re-upload replaces the suffix instead of stacking, (c) the non-file-link case from the first issue above, (d) `showLinkText=false` behavior. `renderHook` + a fake `File` should be enough — no DOM needed.

Confidence: high

### **suggestion (non-blocking):** Reset of `fileMetaSuffixRef` happens before save is confirmed
`useLinkEditorFileMetaSuffix.ts:73` clears `fileMetaSuffixRef.current` inside `buildFinalLinkTextForSave`, which runs synchronously before `onSave`. If `onSave` ever becomes async / retry-able in place, the suffix is gone for the retry. Today the modal closes-and-remounts on success and the path is fine; flagging because the reset isn't free defensively and could be removed (the ref dies with the modal anyway). 

Confidence: low

### **nitpick:** `[\d.]+` is looser than `formatBytes` ever emits
`fileUploadMetaSuffix.ts:17` — `[\d.]+` happily matches `...` or `1.2.3`. The only producer is `formatBytes`, which emits `12.34`-style. `\d+(?:\.\d+)?` would be more correct. Not exploitable; just looser than it needs to be.

Confidence: low

### **nitpick:** `useRef` initializer runs every render but only the first value is kept
`useLinkEditorFileMetaSuffix.ts:47-53` — the ternary that computes the initial suffix runs on every render even though `useRef` only stores the first value. Cheap so doesn't matter performance-wise, but the dead recomputation is also a subtle correctness footgun if anyone ever changes `initialLinkText` to a derived/changing value. Wrapping in a lazy form (`useRef(null)` + a one-shot `useEffect` or `useState` initializer) makes the "computed once" intent explicit. 

Confidence: low

### **praise:** XLSX-before-XLS ordering plus a test that locks it in
`fileUploadMetaSuffix.ts:13-15` sorts the alternation by length desc, and `fileUploadMetaSuffix.test.ts:53-60` explicitly tests `File [XLSX, 1.00 MB]`. Same trap would apply to `DOCX` vs `DOC` — currently safe by the same sort, and worth a sibling test (`it.each`) over all seven accepted extensions to prevent regressions if anyone refactors the regex.

Confidence: high

### **praise:** `formatBytes(0) === undefined` handled gracefully
`fileUploadMetaSuffix.ts:45-54` accumulates parts and short-circuits to `""` when both are missing, so a zero-byte file produces no malformed `[PDF, undefined]` output. Nice.

Confidence: high

---

LGTM modulo the two `issue:` items — the non-file-link strip is the only one I'd block merge on.

adriangohjw (commented):
## Summary

- Auto-appends a `[TYPE, SIZE]` suffix (e.g. ` [PDF, 280.00 KB]`) to prose link text after a file upload in the TipTap link editor.
- **Code flow:** `buildFileUploadMetaSuffix` / `stripFileUploadMetaSuffix` utils produce and validate suffixes (`buildFileUploadMetaSuffix.ts`, `stripFileUploadMetaSuffix.ts`); `useLinkEditorFileMetaSuffix` strips on open into form state, snapshots the suffix in `useState`, refreshes it via `FileAttachment` → `onUploadedFile`, and re-appends on save (`useLinkEditorFileMetaSuffix.ts:18-37`, `LinkEditorModal.tsx:84-117`, `FileAttachment.tsx:63`).
- **Overall:** The validation-based strip (replacing the earlier regex approach) is clearer and well unit-tested. Two correctness gaps around when stripping happens vs when re-appending happens are still worth fixing before merge; the hook itself remains untested.

---

## Findings

### **issue:** Silent data loss when suffix-shaped text is on a non-file link
`useLinkEditorFileMetaSuffix.ts:L18` strips the suffix from `initialLinkText` unconditionally, but `useLinkEditorFileMetaSuffix.ts:L33-L35` only re-appends when `getLinkHrefType(linkHref) === LINK_TYPES.File`. A page/external/email link whose visible text happens to end with ` [PDF, 1.00 MB]` (e.g. describing a download, or legacy content migrated from another CMS) will lose the bracketed portion the next time the modal is opened and saved — even if the user changed nothing. If the user switches a file link to a page link inside the same edit session, the same thing happens.

Fix: pass `initialLinkHref` (or `linkHref` at open) into the hook and only strip/snapshot when the link is already a file link; or always re-append the snapshotted suffix unless a fresh upload occurred.

Confidence: high

### **issue:** Same data-loss path when `showLinkText` is `false`
`useLinkEditorFileMetaSuffix.ts:L18` still strips unconditionally, but `useLinkEditorFileMetaSuffix.ts:L33` returns bare `linkText` when `!showLinkText`, and `onUploadedFile` is disabled (`useLinkEditorFileMetaSuffix.ts:L44`). Any caller that passes real `linkText` with a suffix and `showLinkText={false}` (today `BaseLinkControl.tsx:L147-L148` hardcodes `"Link"`, so no immediate prod hit) will drop the suffix on save. Safest fix: skip stripping entirely when `!showLinkText`.

Confidence: medium

### **issue:** Re-upload can stack suffixes when user text follows the old suffix
The PR notes the display edge case (`Download [PDF, 280.00 KB] extra` won't hide the suffix in the input — covered by `stripFileUploadMetaSuffix.test.ts:L88-L96`). On save the hook is worse: if stripping failed at open, `fileMetaSuffix` stays `""` (`useLinkEditorFileMetaSuffix.ts:L20-L22`), so a re-upload appends a fresh suffix to the full string → `… [PDF, 280.00 KB] extra [PDF, 1.00 MB]`. Consider detecting a partial/old suffix in `buildFinalLinkTextForSave` (strip what you can, then append) or blocking re-append when the trailing bracket isn't at end-of-string.

Confidence: medium

### **issue:** Hook has no tests
Util tests are solid (16 passing), but `useLinkEditorFileMetaSuffix.ts` owns the product behaviour: strip-on-open → snapshot → swap-on-upload → re-append-on-save, plus the non-file and `showLinkText=false` paths above. `renderHook` + a mocked `File` would cover this without DOM.

Confidence: high

### **suggestion (non-blocking):** `setFileMetaSuffix("")` runs before save is confirmed
`useLinkEditorFileMetaSuffix.ts:L36` clears suffix state inside `buildFinalLinkTextForSave`, which runs synchronously before `onSave` / modal close. Fine today because the modal unmounts on success, but if save ever becomes retry-in-place the suffix is gone. The state dies with the modal anyway — dropping the reset is simpler.

Confidence: low

### **suggestion (non-blocking):** Parametric coverage for all accepted extensions
`buildFileUploadMetaSuffix.test.ts` covers PDF/XLS/XLSX; accepted set is also `.csv`, `.tsv`, `.doc`, `.docx` (`constants.ts:L12-L21`). An `it.each` over all seven would lock the `toDisplayType` mapping and round-trip strip behaviour (especially `DOCX` vs `DOC`).

Confidence: medium

### **note:** `isFormattedSize` is looser than `formatBytes` output
`stripFileUploadMetaSuffix.ts:L15-L22` accepts any non-negative `Number(num)` with a unit, so manually typed `[PDF, 100 KB]` (no decimals) would strip even though `formatBytes` always emits `100.00 KB`. Only affects stripping of user-typed bracket text, not generation.

Confidence: low

### **praise:** Validation-based strip is easier to reason about than regex
Moving from alternation-sorted regex to `isFileUploadMetaSuffix` + `toDisplayType` (`stripFileUploadMetaSuffix.ts:L30-L48`) makes the accepted-shape contract explicit, and the strip tests cover the important false-positive cases (trailing non-suffix brackets, mid-string brackets, missing leading space).

Confidence: high

### **praise:** Upload callback wiring is minimal and correctly timed
`onUploadedFile?.(file)` fires inside `uploadFile` `onSuccess` (`FileAttachment.tsx:L62-L63`), after the upload succeeds and before `setHref`, so the suffix reflects the file that actually landed.

Confidence: high

adriangohjw (approved):

adriangohjw:
@brendancjz thanks for your work on this! i did some refactoring but it wouldn't have been posisble without the bulk of the work you've done

PR #1965, opened 2026-04-02
chore: migrate asset deletion to godmode admin panel

### LINKED ISSUE #1914: Self-serve S3 removal + CloudFront invalidation (Isomer Admin / God mode)
### Problem

Engineering oncall is often asked to **manually remove objects from S3** and **invalidate them in CloudFront**. This should be something the **Isomer Admin** team can do themselves from **Studio**, under **God mode**, without a deploy or AWS console runbook.

### Proposed solution

Add a **self-serve** flow in **Studio God mode** that:

1. Accepts **one path at a time** (no batch in v1).
2. Runs the same operations as the current manual process: **delete from the assets S3 bucket** and **create a CloudFront invalidation** for that path.

**Reference implementation (today’s manual script):** [`tooling/s3/remove/remove.sh`](tooling/s3/remove/remove.sh) — dry run / delete loop against the bucket, then `aws cloudfront create-invalidation` with the path(s). Product logic should mirror this (bucket, distribution, path semantics), with secrets/config handled appropriately for the app environment.

### Acceptance criteria

- **Single item input** — User submits **one** asset path per action (no multi-row / bulk upload in v1).
- **Input format** — User enters the object key in the form **`/SITE_ID/UUID`** only (leading slash, site id segment, UUID segment). **Do not** require or parse full URLs, query strings, or file extensions; this avoids confusion with **URL encoding** and duplicate “same file, different encoding” issues. If the user pastes more than needed, validation should normalize or reject clearly (product decision), but the **canonical accepted shape** is `/SITE_ID/UUID`.
- **Validation** — The **`/SITE_ID/UUID`** shape is enforced in **both the frontend and the backend** (never trust client-only checks).
- **Loading / completion** — After submit, the UI stays in a **loading / in-progress** state until **both** steps finish: **S3 removal** and **CloudFront invalidation** (these can take a noticeable amount of time; no premature “success” after only one step).
- **UI / polish** — Visual polish is **not** the priority for v1. Reuse the **same OGP design system and patterns** as other God mode features; functional correctness and safe ops matter more than bespoke UI.

### Notes / constraints

- CloudFront invalidation is **priced per path**; keeping **one path per action** aligns with controlled cost and matches the “one item at a time” requirement.
- Backend should use the same **bucket** and **distribution** semantics as production assets removal today (see script); wire IDs via config/secrets as appropriate for Studio’s environment.

seaerchin:
**NOTE: UNTESTED ON STAGING FOR STUDIO - only the slack version was tested**
- still missing the ability to post to slack too

## Problem

Asset deletion was previously done automatically during page saves, which led to potential issues with orphaned assets and lack of visibility into what was being deleted. Additionally, the deletion logic was scattered across editor drawers and used a soft-delete approach that wasn't being cleaned up properly.

## Solution

Migrated the asset deletion functionality from automatic inline deletion to a dedicated godmode admin panel, giving Isomer admins full control and visibility over asset cleanup operations.

**Breaking Changes**

  - The `deleteAssets` tRPC mutation now requires admin privileges (CORE or MIGRATORS role) instead of standard user permissions
  - The `deleteAssets` schema no longer accepts `siteId` and `resourceId` parameters - it now only accepts `fileKeys`
  - Assets are no longer automatically deleted when users replace images in the editor drawers

**Features**:

- New godmode page at `/godmode/delete-s3-assets` for admins to manually delete S3 assets
- Added CloudFront invalidation after asset deletion to ensure CDN cache is cleared
- Deletion results now return invalidation IDs for tracking

**Improvements**:

- Changed asset soft-delete tagging from `ISOMER_STATUS=DELETED` to `deletedAt=<timestamp>` for better tracking of deletion time
- Added `@aws-sdk/client-ssm` to fetch CloudFront distribution ID from AWS Parameter Store
- Simplified editor drawers by removing automatic asset deletion logic

**Bug Fixes**:

- N/A

## Before & After Screenshots

N/A (Admin tooling change)

## Tests

**New scripts**:

- N/A

**New dependencies**:

- `@aws-sdk/client-ssm` : Used to fetch CloudFront distribution ID from AWS Parameter Store for cache invalidation

**New dev dependencies**:

- N/A


seaerchin:
## Automated PR summary (by Claude)

This PR migrates the `asset.deleteAssets` tRPC mutation from a per-user, per-resource flow invoked implicitly by editor drawers, into a manual godmode admin operation, and bolts CloudFront cache invalidation on top.

### Files changed (excluding `package-lock.json`)

| File | Change |
| --- | --- |
| [`apps/studio/package.json`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/package.json) | Adds `@aws-sdk/client-cloudfront` and `@aws-sdk/client-ssm`. |
| [`apps/studio/src/schemas/asset.ts`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/schemas/asset.ts) | `deleteAssetsSchema` drops `siteId` + `resourceId`; only `fileKeys` remains. |
| [`apps/studio/src/server/modules/asset/asset.router.ts`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/server/modules/asset/asset.router.ts) | `deleteAssets` now gates on `validateUserIsIsomerAdmin([Core, Migrator])`, replaces the per-site authorization with a regex `^\d+\/[a-f0-9-]{36}\/.+$`, no longer throws on partial failures, and triggers `invalidateAssetsBySiteIds`. |
| [`apps/studio/src/server/modules/asset/asset.service.ts`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/server/modules/asset/asset.service.ts) | Removes `doAllFileKeysBelongToSite`, adds `invalidateAssetsBySiteIds` (CloudFront `CreateInvalidation` over `/{siteId}/*`) and `getDistributionId` (SSM `GetParameter` with a process-global cache). |
| [`apps/studio/src/server/modules/asset/__tests__/asset.service.test.ts`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/server/modules/asset/__tests__/asset.service.test.ts) | Drops the cross-site IDOR test suite for `doAllFileKeysBelongToSite`; nothing replaces it. |
| [`apps/studio/src/lib/s3.ts`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/lib/s3.ts) | `deleteFile` now writes a `deletedAt=<unix-ms>` tag instead of `ISOMER_STATUS=DELETED`. |
| [`apps/studio/src/pages/godmode/delete-s3-assets.tsx`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/pages/godmode/delete-s3-assets.tsx) | New godmode page: textarea of newline-separated keys, calls the mutation, surfaces invalidation id in a toast. |
| [`apps/studio/src/pages/godmode/index.tsx`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/pages/godmode/index.tsx) | Adds the new page to the godmode hub. |
| [`apps/studio/src/features/editing-experience/components/Drawer/ComplexEditorStateDrawer.tsx`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/features/editing-experience/components/Drawer/ComplexEditorStateDrawer.tsx) and [`HeroEditorDrawer.tsx`](https://github.com/opengovsg/isomer/blob/297e5c41e249e19617c55c50e4fce165bd835731/apps/studio/src/features/editing-experience/components/Drawer/HeroEditorDrawer.tsx) | Stop calling `deleteAssets` on save; previously-replaced asset keys are now leaked in S3 until godmode cleanup. |

### Codeflow

1. Admin opens `/godmode/delete-s3-assets`, pastes one fileKey per line.
2. Page derives `assetsToDelete` (split, trim, filter), calls `trpc.asset.deleteAssets.useMutation` with just `{ fileKeys }`.
3. Server checks `validateUserIsIsomerAdmin` for `Core | Migrator`, then validates each key against a structural regex.
4. Each key is soft-deleted via `markFileAsDeleted` → `deleteFile` (PutObjectTagging with `deletedAt=<ts>`), wrapped in `Promise.allSettled` so one failure does not abort the others.
5. SiteIds are extracted from successful keys (split on `/`, take index 0), deduped, and passed to `invalidateAssetsBySiteIds`, which fetches the CloudFront distribution id from SSM (`/cloudfront/assets-distribution-id`, cached forever in-process) and issues a single `CreateInvalidation` over `/{siteId}/*`.
6. The handler returns `{ invalidatedSites: { success, invalidationId? } }`; the page toasts the id.

Tagging note: the previous on-drawer-save deletion path is fully removed — no caller now invokes `deleteAssets` outside godmode. This is the intended migration but means orphan accumulation between cleanups is now a real concern.

seaerchin (commented):

PR #1965 — `chore: migrate asset deletion to godmode admin panel`
Reviewed at SHA: `297e5c41e249e19617c55c50e4fce165bd835731`

Scope of review focused on: authorization boundary, blast radius / reversibility, audit logging, race conditions / IDOR, CloudFront and SSM client hygiene, schema / shape changes, dead-code/UX issues.

---

### `issue (security, blocking, confidence: high)` Client-side admin gate is a no-op — `useIsUserIsomerAdmin` return value is misused

In `apps/studio/src/pages/godmode/delete-s3-assets.tsx`:

```tsx
const isUserIsomerAdmin = useIsUserIsomerAdmin({ ... }) // returns { isAdmin, isLoading }
...
if (!isUserIsomerAdmin) { /* toast + redirect */ }
if (!isUserIsomerAdmin) { return null }
```

`useIsUserIsomerAdmin` returns an object `{ isAdmin, isLoading }` (see `apps/studio/src/hooks/useIsUserIsomerAdmin.ts`). The truthiness check on the object is *always true*, so the redirect never fires and the page renders for any authenticated user. The page also calls `deleteAssetsMutation.mutate(...)` regardless. The server-side `validateUserIsIsomerAdmin` on the tRPC mutation is the real gate, but the UI is misleading: a non-admin can land on `/godmode/delete-s3-assets`, paste keys, and submit — they only see the rejection toast after a server round-trip. Compare with `apps/studio/src/pages/godmode/index.tsx`, which correctly destructures `{ isAdmin: isUserIsomerAdmin, isLoading }`. Fix: destructure the same way and gate on `!isLoading && !isAdmin`.

### `issue (security, blocking, confidence: high)` No audit log for a destructive admin operation

Other admin / config-mutating routes log via `logConfigEvent` / `logPublishEvent` (see `apps/studio/src/server/modules/site/site.router.ts` and `apps/studio/src/server/modules/audit/audit.service.ts`). A godmode S3 deletion is strictly more consequential than a footer edit, yet `deleteAssets` only emits `ctx.logger.error` when something *fails*. There is no record of:
- which admin invoked the deletion,
- the input `fileKeys` set,
- which keys actually succeeded,
- the CloudFront invalidation id.

Recommend adding a structured audit log entry on entry (intended keys + actor) and on exit (success/failure counts, invalidation id), so a forensic trail exists if an admin paste-deletes the wrong keys.

### `issue (security, confidence: high)` Site-prefix authorization removed without compensating control

The PR deletes `doAllFileKeysBelongToSite` and the schema's `siteId` constraint. The new validation is just a regex `^\d+\/[a-f0-9-]{36}\/.+$`. That's structural, not authorization — a Migrator-role admin can delete *any site's* assets in one paste, including sites they otherwise have no permission over. For godmode this may be deliberate, but it warrants:
- explicit acknowledgement in the page copy / a confirm dialog enumerating the unique siteIds before submission,
- per-key logging so a misclick is recoverable.

The S3 soft-delete tag (`deletedAt=<ts>`) presumably feeds an external lifecycle policy, but I cannot find that policy in this repo, so the *real* irreversibility window is opaque from this PR alone.

### `issue (security, confidence: high)` Tag rename `ISOMER_STATUS` → `deletedAt` may silently break external S3 lifecycle rules

`apps/studio/src/lib/s3.ts:88` now strips/sets `deletedAt` instead of `ISOMER_STATUS`. The PR description frames this as "better tracking of deletion time", but if any S3 bucket lifecycle policy / Athena query / external job expects `ISOMER_STATUS=DELETED` to identify tombstones, this rename converts a *soft* delete into something that may *never* be reaped, or worse, may stop being protected. Two follow-ups:

1. The stale comment in `apps/studio/src/server/modules/asset/asset.service.ts:179` still says "the ISOMER_STATUS tag (and any other object tags) carry over without us having to set them explicitly" — please update.
2. Confirm that the lifecycle rule (likely in IaC outside this repo) is updated in lockstep, and that already-tagged-with-`ISOMER_STATUS` historical objects either get re-tagged or remain handled.

### `issue (security, confidence: medium)` Re-delete clobbers original `deletedAt` timestamp

`s3.ts` filters out the existing `deletedAt` tag and writes a fresh `Date.now().toString()`. If an admin re-runs deletion on a key that was already tombstoned, the original deletion time is overwritten, which defeats the stated "better tracking of deletion time" goal and may extend the lifecycle rule's grace window. Consider preserving the earliest `deletedAt`, or no-op if the tag already exists.

### `issue (security, confidence: medium)` `Date.now()`-only `CallerReference` — collision risk under burst use

In `apps/studio/src/server/modules/asset/asset.service.ts`:

```ts
CallerReference: `delete-assets-${Date.now()}`,
```

CloudFront uses `CallerReference` as an idempotency key. Two admins (or one admin clicking twice) within the same millisecond will collide and the second invalidation will be rejected silently — but the response is currently swallowed inside try/catch and only the first id propagates back. Add `randomUUID()` (already imported in this file) or include the user id.

### `issue (security, confidence: medium)` CloudFront/SSM clients constructed without `region` and without re-using the env validator

```ts
const cloudfrontClient = new CloudFrontClient({})
const ssmClient = new SSMClient({})
```

This matches the Copilot bot's note. `apps/studio/src/lib/s3.ts` already wires `NEXT_PUBLIC_S3_REGION` from `env.mjs`. Same pattern should apply here so prod behavior is independent of `AWS_REGION` env propagation. Also: SSM parameter name `/cloudfront/assets-distribution-id` is hard-coded — at minimum thread it through `env.mjs` so it's a typed config knob and trivially mockable in tests.

### `issue (resilience, confidence: medium)` `cachedDistributionId` is process-global with no TTL or invalidation

If the SSM parameter is rotated, every running studio pod will keep the stale id until next deploy. CloudFront distribution ids rarely change — but this is the *one* knob on the destructive endpoint. Either drop the cache (one extra SSM call per invocation is fine), or add a TTL.

### `issue (correctness, confidence: high)` `Promise.allSettled` swallows per-key failure context

`asset.router.ts` logs only the *count* of failures, then proceeds to invalidate CloudFront on the *successful* siteIds. The user-facing toast says "Successfully deleted assets" even when some failed. Two fixes:
- log (and ideally return) the failed keys plus their underlying error messages so the operator can retry the exact subset,
- reword the toast to "N of M deleted" when partial.

### `issue (correctness, confidence: medium)` Returned shape is inconsistent with `Promise.all` semantics

The handler returns `{ invalidatedSites: { success, invalidationId?, error? } }`. The page renders `data.invalidatedSites.invalidationId` unconditionally — when zero deletes succeed, `siteIds=[]` triggers the early-return `{ success: true }` (no `invalidationId`), and the toast prints `invalidation id: undefined`. Either short-circuit on the page or always populate `invalidationId`.

### `issue (testing, confidence: high)` Removed `doAllFileKeysBelongToSite` tests, but no replacement coverage for the new regex / godmode auth path

The deleted tests in `apps/studio/src/server/modules/asset/__tests__/asset.service.test.ts` covered cross-site IDOR. The replacement is a regex that, for example, accepts `1/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa/../../etc/passwd` (the `/` is in the filename slot — `.+$` allows it). S3 keys are not paths, so the realistic risk is "operator deletes the wrong thing", not RCE — but tests for `^/`, `..`, and empty filename would lock the regex's intent. Also no router test for the `validateUserIsIsomerAdmin` gate; add one mirroring `whitelist.router.test.ts`.

### `nitpick (confidence: high)` Misleading hint in the UI copy

The page tells admins "If there are trailing parameters like `?original=true`, **delete** them before submitting". A more defensive UX is to strip them server-side (or in `assetsToDelete = ...` derivation) — copy-pasted browser URLs will silently be rejected by the regex otherwise, with no per-line feedback.

### `nitpick (confidence: medium)` `sitesToInvalidate` double-wraps `Set` → `Array.from` twice

```ts
const sitesToInvalidate = Array.from(new Set(siteIds))
const paths = Array.from(sitesToInvalidate).map(...)
```

Drop the second `Array.from` — `sitesToInvalidate` is already an array.

### `praise (confidence: high)`

- Centralizing the destructive op in godmode, removing the implicit-on-save deletion path in editor drawers, is a clear net safety win.
- Server-side `validateUserIsIsomerAdmin` is the right primary gate.
- CloudFront invalidation as part of the same flow avoids a class of cache-staleness bugs.

---

Summary: the *server* authorization story is correct, but the **client gate is broken**, there is **no audit log on a destructive admin endpoint**, and the **`ISOMER_STATUS` → `deletedAt` tag rename** has external-system implications that need verification before this lands. The CloudFront / SSM client setup needs a `region` and a real `CallerReference`.

PR #1967, opened 2026-04-03
Feat/create logging package

### LINKED ISSUE #1917: Extract logging (Pino) from Studio into a standalone package
> [!NOTE]
> Similar to https://github.com/opengovsg/isomer/issues/1916

## Summary

Request-scoped and base logging today lives in **`apps/studio`**—notably `src/lib/logger.ts` (Pino singleton, custom syslog-style levels, `pino-pretty` in dev/test, formatters bound to Studio `env`, child loggers enriched with path, client IP, request id, and Datadog trace headers). Call sites such as **tRPC middleware** (`src/server/trpc.ts`) depend on this module directly.

That **couples shared logging conventions** (levels, JSON shape, transports) to **Studio’s env schema, Next.js request types, and IP helpers**, which makes it awkward for other packages (e.g. workers using **`@isomer/pgboss`**, which already accepts a `pino` logger) to **reuse the same logger factory and field contract** without importing Studio.

We should refactor logging into a **dedicated package** under `packages/`—similar in spirit to [**Open Government Products’ starter-kit `packages/logging`**](https://github.com/opengovsg/starter-kit/tree/main/packages/logging) and structurally aligned with **`packages/pgboss/`** (`@isomer/pgboss`): clear `package.json` exports, build/test/lint in the package, and a **small public API** that apps compose with their own env and HTTP context.

## Problem / motivation

- **Studio-specific wiring**: Logger reads `~/env.mjs`, uses `NextApiRequest`, `~/utils/getClientIp`, and Studio env fields (e.g. `NEXT_PUBLIC_APP_ENV`) inside formatters/bindings.
- **Reuse friction**: Other runtimes (CLI, workers, future apps) either duplicate Pino setup or depend on Studio internals for “the” Isomer log line shape.
- **Inconsistent injection**: Some code paths take a `pino.Logger` parameter (e.g. pgboss client); the “canonical” factory still lives only in Studio.
- **Versioning / ownership**: Log format and dependencies (`pino`, `pino-pretty`) are not owned as a first-class library boundary.

## Proposed direction

1. **Add `packages/logging`** (name TBD, e.g. `@isomer/logging`) with:
   - Factory function(s), e.g. `createLogger(options)` / `createChildLogger(parent, context)`, **without** importing Studio `env` or Next types.
   - **Injected config**: log level, app env label, optional transport (or presets for “development pretty” vs “production stdout”).
   - Optional **adapters** in Studio (thin layer): map `NextApiRequest` → `{ clientIp, traceId }` and pass into the package (keeps the package framework-agnostic).
2. **Studio** imports the package and passes only plain data (strings, optional headers map) into child loggers; tRPC middleware keeps the same behaviour with updated imports.
3. **Align with reference patterns**: mirror the **small, focused** layout of [starter-kit `packages/logging`](https://github.com/opengovsg/starter-kit/tree/main/packages/logging) (src + package metadata + tsconfig); use **`@isomer/pgboss`** as the monorepo precedent for build/export conventions.
4. **Optional later step**: publish from a **standalone repository** if we need independent semver/releases (same discussion as other extracted packages).

## Acceptance criteria (draft)

- [ ] New package exposes a **documented API** for base + child loggers and preserves (or consciously versions) **field names** used in production (level format, timestamp, bindings such as env).
- [ ] **No dependency** from `packages/logging` on `apps/studio` (no `~/env`, no Next types in core).
- [ ] Studio builds and tests pass; tRPC logging middleware uses the package.
- [ ] **Single place** declares `pino` / `pino-pretty` versions for the shared package (consumers align or use peer deps—decide and document).
- [ ] README or short doc: how to use from Studio vs from a worker (pgboss-style `logger` argument).

## Open questions

- Should **Datadog / tracing** fields stay in an optional “observability” adapter vs. core bindings?
- **Peer dependency** on `pino` vs. bundling a supported major—what matches starter-kit and our worker packages best?

## References (internal)

- Current implementation: `apps/studio/src/lib/logger.ts`
- Example consumer: `apps/studio/src/server/trpc.ts` (logger middleware)
- Related pattern (inject logger): `packages/pgboss/src/client.ts`
- [opengovsg/starter-kit — `packages/logging`](https://github.com/opengovsg/starter-kit/tree/main/packages/logging)

brendancjz:
## Problem

Wrote my understanding of the problem and solution here: [doc](https://docs.google.com/document/d/1wj20u_r6wMDdC2-4Nxoxi0LSp18BynsuFTVA0Pu2N5I/edit?usp=sharing)

Closes #1917 

## Solution

- Moved the pino logger from app studio into standalone logging package
- Added new API createRootLogger and createChildLogger so that app studio can call upon the logger
- Abstract app specific internals from logger implementation out its a thin adaptor in app/studio
- Replace pino Logger usage in pgboss with new standalone logger usage
- Duplicate package configs from pgboss package into new logging package

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots
Logging structure looks the same as before. I repeat the use cases of:
- Creating a new folder 
- In a page, I updated the page header title and create a new text block

**BEFORE**:
image

image

**AFTER**:
image
image

## Tests

- [ ] Step 1
- [ ] Step 2
- [ ] Step 3

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


packages/logging/src/logging.test.ts
adriangohjw: ask: can we add the Arrange/Act/Assert pattern as per the other part of the codebase? thanks!

packages/logging/src/logging.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  afterEach(() => {
+    if (logFile !== undefined && fs.existsSync(logFile)) {
+      fs.rmSync(logFile)
+      logFile = undefined
+    }
+  })
+
+  it("createRootLogger writes syslog-shaped JSON with env binding", () => {
+    logFile = path.join(os.tmpdir(), `isomer-log-${Date.now()}.ndjson`)
```
adriangohjw: non-blocking question: any reason we using `.ndjson`? i would think `.log` would be cleaner

adriangohjw (commented):
issue:
- formatter is throwing error on the new package, can run `npm run format:fix`? thanks!
- apps/studio turbo.json `tasks.generate.dependsOn` should also include the new package, just like how it included `@isomer/pgboss#build`

ask:
- can add a turbo.json for the logging package?
- can we also move `dd-trace` into the new package, like [starter kit](https://github.com/opengovsg/starter-kit/blob/main/packages/logging/src/tracer.ts)?

question:
- currently the implementation isn't too big of an issue but seems like we opted to switch out from a PinoLogger approach. Any reason for that? IMO was thinking we should not switch up the implementation because 1) more room for error, esp. since this PR is about porting it over to a new repo for maintainability, and 2) it strays further from the [starter kit approach](https://github.com/opengovsg/starter-kit/blob/main/packages/logging/src/logger.ts) which can make maintainability slightly harder

thanks!

brendancjz:
> @brendancjz FYI (no action needed): i took over to do some cleanup (for pgboss package as well) and WIP. marking as draft first

Hey @adriangohjw, any action on my side?

seaerchin:
## Automated PR summary (Claude)

**Head:** `9390ca1` — `feat/create-logging-package`

### What this PR does

Extracts the studio pino logger into a new `@isomer/logging` workspace package, lets `@isomer/pgboss` consume the `Logger` type from there instead of carrying its own pino, and moves the dd-trace setup into a `@isomer/logging/tracer` subpath that's lazily imported only when `DD_SERVICE` is set.

### Codeflow (concrete)

1. App boot — [`apps/studio/src/instrumentation.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/instrumentation.ts) checks `NEXT_RUNTIME === "nodejs"`, then if `DD_SERVICE` is defined dynamically imports `@isomer/logging/tracer` and calls `initTracer({ service })`. The previous `apps/studio/src/server/modules/tracer.ts` module is deleted.
2. Per-request logger — tRPC middleware in [`apps/studio/src/server/trpc.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd86603 2f8c1d40503/apps/studio/src/server/trpc.ts) calls `createLogger({ path, req })` from [`apps/studio/src/lib/logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/lib/logger.ts).
3. Studio adapter — `createLogger` reads `env.NODE_ENV` / `env.NEXT_PUBLIC_APP_ENV`, derives `clientIp` via `getClientIp(req)`, derives `traceId` from the `x-datadog-trace-id` header, and forwards to `createBaseLogger` from `@isomer/logging`.
4. Package core — [`packages/logging/src/logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts) memoizes a single root pino instance (`PinoLogger.getInstance`) and returns a child binding `path`, `id` (nanoid), `clientIp`, `trace_id`. Custom syslog levels live in [`packages/logging/src/levels.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/levels.ts). Levels are uppercased; timestamp is ISO. Pretty output in dev/test, stdout otherwise.
5. pgboss — [`packages/pgboss/src/client.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/src/client.ts) now imports only `Logger` type from `@isomer/logging` (no own pino).

### Files touched

**New (logging package)**
- [packages/logging/src/logger.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts)
- [packages/logging/src/index.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/index.ts)
- [packages/logging/src/env.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/env.ts)
- [packages/logging/src/levels.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/levels.ts)
- [packages/logging/src/tracer.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/tracer.ts)
- [packages/logging/src/logging.test.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logging.test.ts)
- [packages/logging/README.md](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/README.md)
- [packages/logging/package.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/package.json)
- [packages/logging/turbo.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/turbo.json)
- [packages/logging/tsconfig.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/tsconfig.json)
- [packages/logging/vitest.config.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/vitest.config.ts)
- [packages/logging/.oxlintrc.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/.oxlintrc.json)

**Studio**
- [apps/studio/src/lib/logger.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/lib/logger.ts) (rewritten as adapter)
- [apps/studio/src/instrumentation.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/instrumentation.ts)
- [apps/studio/src/server/trpc.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/trpc.ts)
- [apps/studio/src/server/cron/index.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/cron/index.ts)
- [apps/studio/src/server/cron/jobs/deactivateInactiveUsersJob.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/cron/jobs/deactivateInactiveUsersJob.ts)
- [apps/studio/src/server/cron/jobs/schedulePublishingJob.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/cron/jobs/schedulePublishingJob.ts)
- [apps/studio/src/server/cron/jobs/sendAccountDeactivationWarningEmailsJob.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/cron/jobs/sendAccountDeactivationWarningEmailsJob.ts)
- [apps/studio/src/server/modules/searchsg/searchsg.service.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/modules/searchsg/searchsg.service.ts)
- [apps/studio/src/server/modules/user/inactiveUsers.service.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/modules/user/inactiveUsers.service.ts)
- [apps/studio/src/features/mail/service.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/features/mail/service.ts)
- [apps/studio/src/lib/mail.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/lib/mail.ts)
- [apps/studio/src/pages/api/sign-in/singpass/jwks.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/pages/api/sign-in/singpass/jwks.ts)
- [apps/studio/src/server/modules/tracer.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/modules/tracer.ts) (deleted)
- [apps/studio/next.config.mjs](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/next.config.mjs)
- [apps/studio/package.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/package.json)
- [apps/studio/turbo.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/turbo.json)

**pgboss**
- [packages/pgboss/src/client.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/src/client.ts)
- [packages/pgboss/src/utils.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/src/utils.ts)
- [packages/pgboss/src/__test__/client.test.ts](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/src/__test__/client.test.ts)
- [packages/pgboss/package.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/package.json)
- [packages/pgboss/tsconfig.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/tsconfig.json)
- [packages/pgboss/turbo.json](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/pgboss/turbo.json)
- removed: `packages/pgboss/tsconfig.build.json`, `packages/pgboss/tsconfig.module.build.json`

seaerchin (commented):

Scoped to: PII leakage, structured-log redaction, log injection, correlation IDs. Findings use [Conventional Comments](https://conventionalcomments.org/).

---

**issue (security, blocking): unredacted `rawInput` is logged on every tRPC call**

[`apps/studio/src/server/trpc.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/trpc.ts#L60-L86) calls `logger.info({ ..., rawInput, userId }, ...)` on the success path and the same on error. tRPC `rawInput` is the full mutation body — emails, draft page contents (which on a gov CMS can include PII collected from forms), session-bound identifiers, etc. Pino has a `redact` option which is **not configured** on the root in [`packages/logging/src/logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts#L42-L77). Suggested:

1. Configure `redact: { paths: [...], censor: "[REDACTED]" }` on the root with sensible defaults (`*.password`, `*.email`, `*.nric`, `req.headers.authorization`, `req.headers.cookie`).
2. Drop `rawInput` from the success-path `info` log; keep only on the error path; gate behind a known-key allowlist if any input must be retained.

Confidence: **high** — this is the largest user-facing risk in the PR.

---

**issue (security): log injection via `path` and `error.message` interpolated into the message string**

[`trpc.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/server/trpc.ts#L72) builds the pino `msg` as `` `[${type}]: ${path} - ${durationInMs}ms - OK` ``. Pino JSON-encodes structured fields, but the message string is preserved verbatim — a `\n{"level":"INFO",...}` injected via a crafted procedure name or `error.message` produces a forged second NDJSON line that downstream parsers (Datadog, etc.) will treat as a real log event. Mitigations: validate `path` against a tRPC route whitelist regex (it should be already, but enforce server-side) and strip CR/LF from any string interpolated into log messages.

Confidence: **medium** (depends on whether tRPC's path is already constrained).

---

**issue (security/PII): raw `clientIp` is stamped on every request log**

[`apps/studio/src/lib/logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/lib/logger.ts#L24-L32) attaches the client IP to every line. Under PDPA, IP combined with `userId` (also logged) is personal data. Suggested: hash or truncate by default (`/24` for v4, `/64` for v6), with raw IP only when `LOG_RAW_CLIENT_IP=true`. Also: document this PII surface in the package README.

Confidence: **medium-high**.

---

**issue (correctness): `id` is not a request correlation id for module-scoped loggers**

In [`packages/logging/src/logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts#L80-L93), `id: nanoid()` is generated inside `PinoLogger.logger`. Files like [`apps/studio/src/lib/mail.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/lib/mail.ts#L11) and [`singpass/jwks.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/pages/api/sign-in/singpass/jwks.ts#L7) call `createLogger` once at module load — so their `id` is fixed for the process lifetime, defeating the correlation purpose. The tRPC middleware case is fine, but service-layer logs cannot be correlated to a request. Suggested: pass `id` as an option (request-scoped) and have services accept `ctx.logger` rather than instantiating their own.

Confidence: **high**.

---

**issue (correctness): dual `trace_id` source can disagree with dd-trace's auto-injection**

[`tracer.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/tracer.ts#L11) sets `logInjection: true`, so dd-trace adds its own `dd.trace_id`/`dd.span_id` to every pino line. Meanwhile [`apps/studio/src/lib/logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/lib/logger.ts#L8-L17) reads `x-datadog-trace-id` header into a separate `trace_id` field. The two will disagree (different sampling, different propagation). Suggested: drop `datadogTraceIdFromRequest` and rely on dd-trace's `logInjection`, or pull from `tracer.scope().active()?.context().toTraceId()` for a single source.

Confidence: **medium**.

---

**issue (correctness): root logger options are silently locked to the first caller**

[`PinoLogger.getInstance`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts#L37-L41) caches the root regardless of subsequent `nodeEnv`/`appEnvLabel`/`logLevel`/`destination`. The presence of `resetPinoLoggerRootForTests` confirms the design tension. In Next.js dev hot reload (where module-level state can be retained on `globalThis` across reloads) divergent options are silently ignored. At minimum, throw on mismatched options.

Confidence: **high**.

---

**issue (api): README documents `createRootLogger` + `createChildLogger` but only `createBaseLogger` exists**

[`packages/logging/README.md`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/README.md) describes a two-function API matching the starter-kit pattern, but [`logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts#L100-L109) exports a single combined `createBaseLogger`. Either split the API (cleaner — root once, children many times), or rewrite the README. This will mislead any future consumer outside studio.

Confidence: **high**.

---

**issue (correctness): `env` field collision between binding and dd-trace `logInjection`**

The pino `bindings` formatter sets `env: appEnvLabel`. dd-trace `logInjection: true` also sets `env` (from `tracer.init({ env })`). Pino field precedence will determine the winner and it will not be obvious. Suggested: rename to `app_env` or `deployment` to avoid the collision.

Confidence: **medium**.

---

**nit (consistency): package reads `process.env.PINO_LOG_LEVEL` despite README claiming env-free**

[`logger.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logger.ts#L57) reads the env directly, contradicting the README's "callers pass plain config". Either drop it or document.

Confidence: **low**.

---

**nit (packaging): `"type": "module"` with raw `.ts` exports works only inside Next**

[`packages/logging/package.json`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/package.json) exports `./src/index.ts`. `apps/studio` adds it to `transpilePackages` in [`next.config.mjs`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/next.config.mjs#L164-L171), so this works for studio. But any non-Next consumer (CLI script, future worker) will fail. Either restore a real CJS+ESM build like the original pgboss had, or accept that this package is studio-internal.

Confidence: **medium**.

---

**praise**: lazy import of `@isomer/logging/tracer` in [`instrumentation.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/apps/studio/src/instrumentation.ts#L8-L14) gated on `DD_SERVICE` is a clean improvement over importing `tracer` for side effects.

**praise**: deduplicating pino out of `@isomer/pgboss` removes a real version-skew risk.

---

### Test coverage suggestions

[`packages/logging/src/logging.test.ts`](https://github.com/opengovsg/isomer/blob/9390ca198110da5d46feccdd866032f8c1d40503/packages/logging/src/logging.test.ts) is good for shape, but missing:
- A redaction test (would force the redaction config to exist).
- A test that two calls with different `appEnvLabel` raise/warn rather than silently keeping the first.
- A test that `\r\n` in `clientIp`/`path` does not break NDJSON parsing.

adriangohjw (commented):
## Review summary

Extracts the studio pino setup and dd-trace init into a new internal `@isomer/logging` package, and rewires `@isomer/pgboss` + studio call sites to import `Logger`/`BaseLogger`/`pino` from there. After the latest refactor commits, the port is intentionally minimal: `packages/logging/src/logger.ts:1-86` is a near-verbatim copy of the previous `apps/studio/src/lib/logger.ts`, the studio adapter (`apps/studio/src/lib/logger.ts:15-27`) only maps `NextApiRequest` → plain options, and `packages/logging/src/tracer.ts:5-19` keeps the original dd-trace init but now lazy-loaded from `apps/studio/src/instrumentation.ts:8-13` when `DD_SERVICE` is set. Packages are consumed via source TS through `transpilePackages` in `apps/studio/next.config.mjs:174-175`, so there's no build step.

Overall the structural move is clean and the previous round of cleanup commits removed the bulk of the over-engineered surface. The one finding worth pausing on is the dd-trace behavior change — flagged below.

---

**issue (correctness, blocking-ish):** `apps/studio/src/instrumentation.ts:8-13` — this changes dd-trace from "always initialized in nodejs runtime" to "only initialized when `DD_SERVICE` env var is set". On `main`, `apps/studio/src/server/modules/tracer.ts` ran `tracer.init({ service: "isomer-next", ... })` unconditionally. After this PR, if the deployment env (ECS task def / EC2 user data) does not export `DD_SERVICE`, dd-trace silently no-ops: no runtime metrics, no `logInjection`, no profiling. I grepped this repo and `DD_SERVICE` is not referenced anywhere outside the new code — it must be set in `isomer-next-infra` for prod/staging/uat. Two things to confirm before merge:
1. `DD_SERVICE` is actually exported in every deployed environment that previously had tracing (prod, staging, uat).
2. The value matches the previous hardcoded `"isomer-next"`, otherwise the DD service identifier (and any dashboards/monitors keyed off it) shifts.

If either is uncertain, safer fallback is `service: process.env.DD_SERVICE ?? "isomer-next"` so the default behavior is preserved.

Confidence: **high** that the behavior changed; **medium** on whether it breaks anything (depends on infra).

---

**issue (maintainability):** `packages/logging/src/env.ts:4-21` — the package validates `NEXT_PUBLIC_APP_ENV` / `NEXT_PUBLIC_APP_VERSION` itself, parallel to `apps/studio/src/env.mjs`. Two consequences:
- The `NEXT_PUBLIC_` prefix is a Next.js convention. Hardcoding it into a "shared" package quietly couples the package to studio. If you ever consume `@isomer/logging` from a non-Next.js entrypoint (a standalone worker, CLI), these env vars won't be set and the tracer silently defaults to `"development"` / `"0.0.0"`.
- The `skipValidation` heuristic (`npm_lifecycle_event === "lint"`) is brittle — `typecheck`, `vitest`, e2e, etc. all silently skip into defaults instead of validating.

Lower-friction option: have `initTracer` accept `env` and `version` as arguments (the caller already has validated env via `~/env.mjs`), and delete `packages/logging/src/env.ts`. That makes the package a pure adapter and removes the duplicate validation.

Confidence: **medium-high** — it works today but it's a future footgun once a second consumer shows up.

---

**issue (footgun):** `packages/logging/src/logger.ts:27-31` — `PinoLogger.instance` is memoized on the first call's `nodeEnv` + `appEnvLabel`, and subsequent calls silently ignore those args. In studio this is fine (single set of values), but the API shape invites mistakes: `createBaseLogger({ nodeEnv: "test", ... })` followed by `createBaseLogger({ nodeEnv: "production", ... })` quietly uses test settings forever. Either:
- move `nodeEnv`/`appEnvLabel` to a separate `initRootLogger(...)` called once, with `createBaseLogger` only taking child bindings, or
- throw if the second-and-later calls disagree with the cached instance.

This is exactly the "two-function API" the README at `packages/logging/README.md` hints at via "process-wide singleton", and arguably what the starter-kit shape (createRootLogger + createChildLogger) avoids.

Confidence: **medium** — not breaking, but the singleton-with-args shape is the kind of thing that bites later.

---

**nit:** `apps/studio/src/lib/logger.ts:19-25` — `req.headers["x-datadog-trace-id"]` is typed `string | string[] | undefined`. The `Array.isArray(traceId) ? traceId[0] : traceId` picks the first occurrence. Behavior matches the previous (`req?.headers[...]` passed straight through, which would have stamped an array literal into the binding), so this is strictly better. Just worth noting the choice — if a request is somehow proxied with two of these headers it's an anomaly worth logging rather than silently picking one.

Confidence: **low**.

---

**nit:** `packages/logging/src/logger.ts:48` — `level: process.env.PINO_LOG_LEVEL || "info"` keeps a direct `process.env` read in the package, despite the README ("This package does **not** import application env modules"). Either accept a `level` option from the caller (consistent with the rest of the API), or document the exception in the README.

Confidence: **low**.

---

**nit:** `packages/logging/src/index.ts:2` re-exports `pino` as a named export. Studio doesn't use it; only `packages/pgboss/src/__test__/client.test.ts:5` does (to construct a silent test logger). Fine, but if the goal is "studio + pgboss don't depend on pino directly", consider exposing a `createSilentLogger()` helper from the package instead, so even tests don't reach for pino primitives. Take it or leave it.

Confidence: **low**.

---

**praise:** Lazy `await import("@isomer/logging/tracer")` gated on `DD_SERVICE` (`instrumentation.ts:8-13`) is a clean improvement over importing `tracer` purely for side effects.

**praise:** Dropping the dual CJS/ESM build for pgboss + logging in favor of `transpilePackages` + raw TS exports (`packages/logging/package.json:6-9`, `packages/pgboss/package.json:6-9`) is the right call for studio-internal packages — much less moving machinery.

**praise:** Removing `pino` / `pino-pretty` from `apps/studio` and `@isomer/pgboss` direct deps and centralizing them in `@isomer/logging` removes a real version-skew risk.

---

### Note on the earlier Claude-authored review on this PR

A prior automated review on this PR raised a long list of concerns (rawInput logging, clientIp PII, log injection via `path`, `id` not being request-scoped for module-level loggers, `env` field collision with dd-trace `logInjection`, redaction, etc.). Re-reading those against the **current** diff:

- Most of them target `apps/studio/src/server/trpc.ts` and module-load uses of `createBaseLogger` in `mail.ts` / `jwks.ts` — **none of those files are touched by this PR**, and the package is a faithful port of the prior behavior. Those concerns are valid but pre-existing and out of scope here.
- The "two functions documented, one exported" and "test for redaction" findings referenced files (`levels.ts`, `logging.test.ts`, `createLogger` API) that the latest cleanup commits (`3c5edbb`, `bb71994`, `e3e8353`) removed/renamed. They no longer apply to the current head.

Recommend opening a follow-up issue specifically for the trpc `rawInput` redaction / IP-PII handling rather than blocking this port on it.

adriangohjw:
## pino: ^9.6.0 → ^10.3.1

**Verdict:** SAFE

| Version | Breaking change | Impact | Where in codebase | Migration hint |
| --- | --- | --- | --- | --- |
| v10.0.0 | Dropped Node.js 18 support | Safe | — | Project already requires Node `>=22` (`package.json` engines, `.nvmrc` v22.18.0) |
| v10.0.0 | Safer type signature for `censor` function (PR #2308) | Safe | — | No `censor`/`redact` usage in `packages/logging`, `packages/pgboss`, or `apps/studio` |
| v10.1.0 | Redaction implementation migrated to `@pinojs/redact` | Safe | — | No `redact` option configured anywhere in the codebase |
| v10.2.0–10.3.1 | None documented | Safe | — | — |

**Notes:**
- Only pino usage is in `packages/logging/src/logger.ts` — uses `pino()` factory, `pino.destination(1)`, `.child()`, `customLevels`, `useOnlyCustomLevels`, `timestamp`, `formatters`. None affected by v10 changes.
- Type imports (`Logger`, `BaseLogger`, `DestinationStream`) are unchanged across v10.
- Bonus: `pino-pretty@13.1.3` (already in lockfile) declares `pino: ^10.1.0` as a peer — this upgrade aligns the peer constraint.

Sources:
- [Pino Releases](https://github.com/pinojs/pino/releases)
- [Issue #2317: Clarify v10 breaking changes](https://github.com/pinojs/pino/issues/2317)
- [Pino v10.0.0 release notes](https://github.com/pinojs/pino/releases/tag/v10.0.0)

adriangohjw (approved):

adriangohjw:
@brendancjz thanks for your work on this! i did some refactoring because i realized the risk will be lower to do a as-is port over and leave out the refactoring. appreciate your help and research into this!

brendancjz:
> @brendancjz thanks for your work on this! i did some refactoring because i realized the risk will be lower to do a as-is port over and leave out the refactoring. appreciate your help and research into this!

Hey @adriangohjw, thank you so much for pushing this through the finish line :) Sorry, I have been busy lately. If there's any issues you need help with, please let me know! I would love to continue contributing~

PR #1980, opened 2026-04-08
feat: add ScheduledJobs table for scheduled publishing

seaerchin:
## Problem

The system currently lacks a dedicated table to track scheduled jobs for resources. This makes it difficult to manage and track operations like scheduled publishing and document pushes.

## Solution

Added a new `ScheduledJobs` table with a corresponding `ScheduledJobType` enum to track scheduled operations on resources.

**Breaking Changes**

**Features**:

- New `ScheduledJobs` table with the following columns:
  - `id`: Auto-incrementing BigInt primary key
  - `type`: Enum (`PublishResource`, `PushDocument`) to categorize the job type
  - `resourceId`: Foreign key reference to the `Resource` table
  - `scheduledAt`: Timestamp for when the job should execute
  - `scheduledBy`: Foreign key reference to the `User` who scheduled the job
  - `createdAt` / `updatedAt`: Audit timestamps
- Added relations from `Resource` and `User` models to `ScheduledJobs`

**Improvements**:

- N/A

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

N/A - Database schema change only

**AFTER**:

N/A - Database schema change only

## Tests

- Run the database migration and verify the table is created correctly
- Verify foreign key constraints work as expected

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A


adriangohjw (commented):
we can search by `select * from "pgboss".job`? 👀

seaerchin (commented):

---

### **issue:** ScheduledJobs_id_idx is redundant with the primary key. [Confidence: HIGH]

id BIGSERIAL ... PRIMARY KEY (id) already creates a unique B-tree index on id. A second index on the same column is dead weight. Drop both the migration index and the @@index([id]) line in the Prisma schema.

---

### **issue:** Missing indexes on the columns this table will actually be queried by. [Confidence: HIGH]

A worker picking up due jobs will run WHERE type = ? AND scheduledAt <= now() ORDER BY scheduledAt LIMIT N. Without an index on (scheduledAt) (or composite (type, scheduledAt)), this query becomes a sequential scan as the table grows. Also consider (resourceId) for finding jobs related to a resource before deleting it.

---

### **issue:** No status / lifecycle column. [Confidence: HIGH]

The table cannot represent "this job has run / is running / failed." Without a status, you cannot distinguish processed vs unprocessed rows, avoid double-execution if two workers race, or diagnose failures. Recommend at minimum a status enum plus lastRunAt and error columns. Or document an explicit "we delete jobs after they run" decision.

---

### **issue:** updatedAt won't auto-update via raw Kysely writes. [Confidence: MEDIUM]

Migration sets DEFAULT CURRENT_TIMESTAMP but no ON UPDATE trigger. Prisma emulates @updatedAt at the application layer; Kysely does not. Updates via Kysely will leave updatedAt stale. Add a Postgres BEFORE UPDATE trigger.

---

### **issue:** No deduplication constraint. [Confidence: MEDIUM]

Two PublishResource jobs can be scheduled for the same resourceId at the same scheduledAt. Consider a partial unique index on (resourceId, type) once a status column lands.

---

### **issue:** ON DELETE RESTRICT on resourceId constrains resource deletion. [Confidence: MEDIUM]

A resource with scheduled jobs cannot be deleted without cleaning up jobs first. Sensible, but the resource-deletion flow will need to handle this. Worth a follow-up TODO.

---

### **suggestion:** Singular table name (ScheduledJob) is more conventional. [Confidence: LOW]

Most models in schema.prisma are singular (Resource, User, Site). CodeBuildJobs and now ScheduledJobs are the plural exceptions.

---

### **suggestion:** No deletedAt for soft delete. [Confidence: LOW]

Other tables in this repo follow a soft-delete pattern. If jobs are append-only audit-style, add deletedAt for consistency.

---

### **note:** Schema-only, no application wiring. [Confidence: HIGH]

Reviewers should look at the consumer (feat/egazette-protoo / #1949) together to confirm that the access patterns there match what is indexable here.

seaerchin:

Schema-only change. No application logic.

**New enum** — `ScheduledJobType { PublishResource, PushDocument }`

**New table — `ScheduledJobs`**
- `id BIGSERIAL` PK
- `type ScheduledJobType`
- `resourceId BIGINT` — FK to `Resource.id`, `ON DELETE RESTRICT ON UPDATE CASCADE`
- `scheduledAt TIMESTAMP(3)`
- `scheduledBy TEXT` — FK to `User.id`, `ON DELETE RESTRICT ON UPDATE CASCADE`
- `createdAt`, `updatedAt` (`DEFAULT CURRENT_TIMESTAMP`, no DB-level `ON UPDATE` trigger)
- `CREATE INDEX ScheduledJobs_id_idx ON ScheduledJobs(id)` (note: redundant with the PK)

**Prisma model**
- `model ScheduledJobs` with relations `resource` to `Resource`, `scheduler` to `User`
- Reverse relations added on `Resource` and `User`
- `@@index([id])` mirrors the redundant SQL index

**Generated types** — `ScheduledJobs` added to Kysely `DB`, `Selectable<ScheduledJobs>` exported.

**Stats:** +68 / -0 across 5 files.

adriangohjw:
@seaerchin seem like we dont need this anymore since we are doing multi-tables?

seaerchin:
@adriangohjw - i have another PR up oops, going to close this

PR #1988, opened 2026-04-12
Feat/add duplicate page

### LINKED ISSUE #1915: Studio: Duplicate Page
## Problem

In **apps/studio**, there is **no way to duplicate a page**. Users who want a similar layout or content must recreate pages manually, which is slow and error-prone.

## Scope

- **In scope:** Duplication for **Pages** only (the primary page entity users edit in Studio).
- **Out of scope:** Collections, standalone files, links, folders, and other non-page resources. Do not extend this behaviour to those types in this issue unless explicitly specified later.

## Proposed behaviour

Provide an action (e.g. from the dropdown for the page) that creates a **new page** copied from the selected page, with content and structure preserved where the product supports it today.

## Acceptance criteria

1. **Draft by default**  
   The duplicated page is created in **draft** state (not published), so it can be reviewed before going live.

2. **Pages only**  
   Duplication is available only for **Pages**, not for collections, files, links, folders, or other resource types.

3. **Independent file/image assets**  
   For every **image and file** referenced by the duplicated page’s content:
   - The system **creates a new copy** of each asset (new backing record / storage association as appropriate for the product).
   - Each copy gets a **different UUID** from the original (and from other duplicates), so assets remain **tied to the specific page (resource)** they belong to, not shared site-wide by accident.

4. **Deletion safety**  
   Deleting the original page (or its assets) **must not** remove assets that belong only to the duplicate, and vice versa—behaviour consistent with “one resource per page” file semantics.

5. **Consistency**  
   The duplicated page’s content must reference the **new** asset UUIDs, not the originals.

## Technical / product notes

- Files are modelled as **bound to a specific resource (page)**; reusing the same file UUID across two pages would break that invariant and could cause **cascade deletes** or **wrong ownership** when one page is removed.
- Implementation should **walk page content** (whatever structure Studio uses: blocks, JSON, etc.) and **duplicate** each referenced file/image, then **rewrite references** to the new IDs.

## Open questions (optional)

- Naming convention for the duplicate (e.g. “Copy of …”, suffix, or prompt).
- Whether navigation/menu entries that point to the page should be duplicated (often **no** for v1).

brendancjz:
## Problem

Closes #1915 

Closes https://linear.app/ogp/issue/ISOM-2082/duplicate-pages

## Solution

1. Create "duplicate page" button in ResourceTableMenu
2. Logic will get the current page's content, do some validations, deep clone, traverse page content to get file asset keys, copy object in s3 bucket with new ids, rewrite the new page content with updated asset keys
3. Show toast successful message and go to the new duplicated page.

Disclaimer:
I used cursor to assist me in this MR. Used deepwiki to familiarise myself with the various files and main workflow. Then devised a solution plan with cursor, ironing out the solution. I wrote a rough draft and cursor helped to finalised the code. Each line of code has been reviewed carefully based on my understanding of the codebase thus far.

**Breaking Changes**

  - Details ...

**Features**:

- User can duplicate a page

**Improvements**:

- Abstracted common logic in `createPage` and `duplicatePage` to `insertPageWithBlob`.

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

No duplicate page 

**AFTER**:

https://github.com/user-attachments/assets/27298bc5-f7a6-4427-a658-963bf1a3a084

## Tests

- [ ] Step 1
- [ ] Step 2
- [ ] Step 3

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


apps/studio/src/lib/s3.ts
```diff
@@ -42,6 +43,25 @@ export const generateSignedPutUrl = async ({
...
+  Bucket: string
+}) => {
+  await storage.send(
+    new CopyObjectCommand({
+      Bucket,
+      Key: destinationKey,
+      CopySource: `${Bucket}/${sourceKey}`,
+    }),
+  )
+  // TODO: Check if after copying the object we need to remove object tagging like soft-deletion tag
```
brendancjz, line 68: question: if the current page has soft-deleted assets, should the duplicated page also have those same soft-deleted assets? Or should the new duplicated page be free of soft-deleted assets?

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
```diff
@@ -20,6 +24,8 @@ import {
 } from "../../atoms"
 
 interface ResourceTableMenuProps {
+  siteId: number
+  tableScopeResourceId?: number
```
brendancjz, line 26: note: this `tableScopeResourceId` is the `id` of the table and it is used so that `onSuccess` can invalidate the table, causing the table to refetch the updated new duplicated page. I followed its typing from upstream components.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -645,6 +624,149 @@ export const pageRouter = router({
...
+          throw new TRPCError({
+            code: "INTERNAL_SERVER_ERROR",
+            message:
+              "Could not copy one or more files for this page. Try again later.",
+            cause,
+          })
+        }
+      }
+
+      rewriteAssetFileKeysInValue(newContent, siteId, oldToNew)
```
brendancjz, line 671: note: both `rewriteAssetFileKeysInValue` and `collectUniqueAssetFileKeys` traverse through the page content object to get the file keys, one to find them and the other to rewrite them with the new key. Seems possible to optimise this but currently can't think of a way because we need the old file keys to copy the assets in s3 bucket.

seaerchin (commented):
preliminary review only - just from reading this PR, it seems like we're doing alot of work that might potentially make this very slow for end users

if i'm understanding this correctly, what we do in this PR is to copy the page blob then traverse the content searching for assets and copy these assets to a new key. not saying that we can't do this but can i ask why we want to copy the asset rather than just reusing the original asset as is? 

this is what notion does actually and it serves quite a few benefits here:

1. whatever metadata on the original asset is retained. this is important for things like delete markers, malware scan setatus
2. it takes very little time to do this (it's just a copy in our db)
3. we don't have to maintain so much code (less maint debt)

again, not saying that we can't do this - curious to find your reasoning!

adriangohjw:
> preliminary review only - just from reading this PR, it seems like we're doing alot of work that might potentially make this very slow for end users
> 
> if i'm understanding this correctly, what we do in this PR is to copy the page blob then traverse the content searching for assets and copy these assets to a new key. not saying that we can't do this but can i ask why we want to copy the asset rather than just reusing the original asset as is? 
> 
> this is what notion does actually and it serves quite a few benefits here:
> 
> 1. whatever metadata on the original asset is retained. this is important for things like delete markers, malware scan setatus
> 2. it takes very little time to do this (it's just a copy in our db)
> 3. we don't have to maintain so much code (less maint debt)
> 
> again, not saying that we can't do this - curious to find your reasoning!

@seaerchin oh actually I listed that requirement. IMO the issue with reusing the same asset is that, let's say we want to delete assets of a resource when we delete it, we cannot confident say that the same asset isn't used in other pages. Traversing through the blob isn't ideal performance-wise, but I think it's okay-ish for the worst case scenario too

seaerchin:
> > preliminary review only - just from reading this PR, it seems like we're doing alot of work that might potentially make this very slow for end users
> > if i'm understanding this correctly, what we do in this PR is to copy the page blob then traverse the content searching for assets and copy these assets to a new key. not saying that we can't do this but can i ask why we want to copy the asset rather than just reusing the original asset as is?
> > this is what notion does actually and it serves quite a few benefits here:
> > 
> > 1. whatever metadata on the original asset is retained. this is important for things like delete markers, malware scan setatus
> > 2. it takes very little time to do this (it's just a copy in our db)
> > 3. we don't have to maintain so much code (less maint debt)
> > 
> > again, not saying that we can't do this - curious to find your reasoning!
> 
> @seaerchin oh actually I listed that requirement. IMO the issue with reusing the same asset is that, let's say we want to delete assets of a resource when we delete it, we cannot confident say that the same asset isn't used in other pages. Traversing through the blob isn't ideal performance-wise, but I think it's okay-ish for the worst case scenario too

on this issue - when we want to remove an asset, do we want to remove the logical pointer (ie, the link) or mark the underlying asset as deleted ah? 

i'm thinking of hte former, but my reasoning is from my pov - the 2 pages are logically disjoint (it's just a clone, so content shouldn't be related) so deletion of 1 asset (link removal) from page A shouldn't impact page B (it's logical clone). 

my take is that actually deleting the asset (ie, marking s3 asset as deleted) can be slightly harder, because this isn't on the hot path and should only be done when there's an emergency (eg: confidential info published ahead of time) whereas i think duplicating etc would be done more often and having the asset be maintained as a pointer is more useful for our use case!

might be useful to get design in on this cos ultimately we're designing for hte user's workflow!

adriangohjw:
@brendancjz i discussed with @seaerchin and the team

we will proceed to make a copy of any asset (likely using `aws s3 cp` / `CopyObjectCommand`). So it means `/SITE_ID/UUID_123/hello.png` is copied to `/SITE_ID/UUID_456/hello.png`

reason being this is more similar to the exiting UX mental model of end users (until we build a media library)

there might be a performance hit, esp. for pages with tonnes of images (for example: those having an ImageGallery). We accept that trade-off, though we can mitigate by doing concurrent copying on the server side to reduce wait time -> this also means we likely need some form of loader to signal to users we are in the midst of duplicating the page to manage their expectations

also, we have to take into consideration to traverse through the schema to ensure all images are duplicated, even when we add new components in the future - ensuring it's not too hardcoded or anything that it will break

---

on the UX aspect of things, instead of duplicating it on clicking "duplicate" and appending a "copy" to the title and slug, could we instead have a modal to let the user update the title/slug before confirming the duplication? this is because in many use cases, users do this to duplicate earlier versions of a file to a new version. E.g. `News of the month: 2026-04` to `News of the month: 2026-05` -> auto appending `copy` such that it is `News of the month: 2026-04 copy` requires them to take additional step to change it

cc @sehyunidaaa

sehyunidaaa:
Adding on with a screenshot of the modal. Let's modify the copy for both the title and button to clearly signal to the users what they are confirming. 

Please ignore the parts where it says "Additional description".

image


Adding a note on the wait time, maybe we can first test it with a very asset-heavy page and evaluate whether the loading needs a dedicated interaction or not. Please ping me again before we merge this~

brendancjz:
> @brendancjz i discussed with @seaerchin and the team
> 
> we will proceed to make a copy of any asset (likely using `aws s3 cp` / `CopyObjectCommand`). So it means `/SITE_ID/UUID_123/hello.png` is copied to `/SITE_ID/UUID_456/hello.png`
> 
> reason being this is more similar to the exiting UX mental model of end users (until we build a media library)
> 
> there might be a performance hit, esp. for pages with tonnes of images (for example: those having an ImageGallery). We accept that trade-off, though we can mitigate by doing concurrent copying on the server side to reduce wait time -> this also means we likely need some form of loader to signal to users we are in the midst of duplicating the page to manage their expectations
> 
> also, we have to take into consideration to traverse through the schema to ensure all images are duplicated, even when we add new components in the future - ensuring it's not too hardcoded or anything that it will break
> 
> on the UX aspect of things, instead of duplicating it on clicking "duplicate" and appending a "copy" to the title and slug, could we instead have a modal to let the user update the title/slug before confirming the duplication? this is because in many use cases, users do this to duplicate earlier versions of a file to a new version. E.g. `News of the month: 2026-04` to `News of the month: 2026-05` -> auto appending `copy` such that it is `News of the month: 2026-04 copy` requires them to take additional step to change it
> 
> cc @sehyunidaaa

Hey @adriangohjw, thanks for the update, really appreciate how you and the team are making progress on this issue. On my side, I will internalise the messages and give an update on this issue end of the week.

brendancjz:
Updated with the following changes:
- Added duplicate page modal that allows user to specific new title and slug before duplicating page.
- Validation ensures perma-link is unique, but title can be non unique.
- Upon successful confirmation, duplicate page process is done synchronously and prevent user action until completion
- Successful/failed duplication will lead to an appropriate toast message. Newly duplicated page will not automatically be opened, the site's table will be invalidated and refetch to show the newly duplicated page.


https://github.com/user-attachments/assets/81c35c5a-2907-4c03-8fe2-5bcfe44aed8c



Consideration
- I tried to implement an asynchronous duplicate page operation where the user can continue any action while the operation is handled in the background. But this added significant complexities:
- What happens if user goes out of the dashboard page? Proper toast message handling outside of dashboard page component.
- What happens if user duplicates the page again while its already duplicating? 
- Conclusion is that I will prioritise completing the feature synchronously first and if there's use case to support async, I will expand on that.

Hi @adriangohjw, here's my update! It's not completely cleaned up yet but looking to get a sensing if the direction is okay so far.

seaerchin:

End-to-end "Duplicate page" feature.

**Server**
- New `duplicatePage` tRPC mutation in `page.router.ts`
- New helpers:
  - `helpers/duplicatePagePreconditions.ts` — page-exists + read/create perm + permalink-uniqueness checks
  - `helpers/duplicatePageContent.ts` — `tryParseSiteAssetFileKey`, `collectUniqueAssetFileKeys`, `buildNewAssetFileKeyForSite`, `rewriteAssetFileKeysInValue` (longest-key-first replacement)
  - `helpers/insertPageWithBlob.ts` — shared `insertPageBlobResourceAndAudit`, also adopted by `createPage`
- New `copyObjectInBucket` in `apps/studio/src/lib/s3.ts` (uses `CopyObjectCommand`; default tag-copy)
- AJV-validates the rewritten blob before insert
- `schemas/page.ts` exports `pageTitleSchema` / `permalinkSchema` / new `duplicatePageSchema`

**Flow**
1. Validate preconditions
2. Load source page content; `cloneDeep`
3. Collect unique site asset keys; copy each into a fresh UUID folder via `Promise.all`
4. Rewrite blob references to point at new keys
5. AJV validate
6. Transaction: insert blob + resource + audit log

**Studio UI**
- New `DuplicatePageModal` (form, full-permalink preview, conflict-as-form-error handling, progress toast)
- jotai `duplicatePageModalAtom` keyed by `pageId` + `tableScopeResourceId`
- `ResourceTableMenu` gains a "Duplicate page" item; modal mounted on both site root and folder pages

**Tests**
- Unit: `duplicatePageContent.test.ts`
- Integration: `page.router.test.ts` `duplicatePage` block, plus `s3.copyObjectInBucket` mocked in `asset.router.test.ts`

**Stats:** +1130 / -46, 17 files.

seaerchin (commented):

---

### **praise:** Strong test coverage and helper extraction. **[Confidence: HIGH]**

`duplicatePageContent.test.ts` covers leading-slash paths, placeholder rejection, dedup, longest-first rewrite, and fresh-UUID generation. The helper extraction (`insertPageWithBlob`, `duplicatePagePreconditions`, `duplicatePageContent`) is clean and the `createPage` refactor to share `insertPageBlobResourceAndAudit` is a real win.

---

### **issue:** S3 copy + DB insert are not transactional — orphan files on failure. **[Confidence: HIGH]**

The flow is (1) validate, (2) `Promise.all(copyTasks)` copying every asset to a new key in S3, (3) AJV-validate the rewritten blob, (4) DB transaction insert.

If step 3 or 4 fails, the S3 copies remain orphaned with no compensating delete. Mitigations: track copied keys and `markFileAsDeleted` them in a `try/catch` covering steps 3+4; or insert with old refs first then copy + rewrite + update blob; or schedule copies in a worker after the resource exists.

---

### **issue:** Soft-deleted files get copied along with their lifecycle tag. **[Confidence: MEDIUM]**

`copyObjectInBucket` uses default `TaggingDirective` (`COPY`). Your TODO already flags this. If a page references an asset that was soft-deleted (still readable but tagged for cleanup), the duplicate's copy carries the same tag and will be removed by your S3 lifecycle policy — so the duplicate ends up with broken assets. Set `TaggingDirective: "REPLACE"` with an empty tag set, or remove the soft-delete tag explicitly after copy.

---

### **issue:** `cloneDeep` from lodash on large blobs is slow — prefer `structuredClone`. **[Confidence: MEDIUM]**

```ts
const newContent = cloneDeep(fullPage.content) as IsomerSchema
```

Native `structuredClone` (Node ≥ 17) handles JSON-shaped data faster and more correctly.

---

### **issue:** `rewriteAssetFileKeysInValue` mutates strings in place via substring match. **[Confidence: MEDIUM]**

The rewriter does naive string-substitution to find old keys inside arbitrary string values (e.g. ``). Sorting longest-first reduces overlap risk, but: (1) if a page coincidentally contains a string matching an old key in plain text, it will be rewritten; (2) the function mutates `value` in place without the name making that clear.

A stricter alternative: only rewrite values exactly `"/key"` or `"key"` and surface HTML asset references through structured fields.

---

### **issue:** Permalink uniqueness check has TOCTOU race. **[Confidence: LOW]**

`assertDuplicatePagePreconditions` queries for an existing permalink, then *outside* the transaction we copy S3 + insert. The DB insert already handles `uniqueViolation` → `CONFLICT` correctly, so the worst case is wasted S3 copy. The current code does both checks (precondition + insert constraint), which is redundant — drop one.

---

### **issue:** Permission check for root pages passes `[null]` to bulk validator. **[Confidence: MEDIUM]**

```ts
await bulkValidateUserPermissionsForResources({
  siteId, action: "create", userId,
  resourceIds: [parentKey],   // can be null
})
```

When the source page lives at root, `parentKey` is `null` and `resourceIds: [null]` is passed. Behaviour depends on the validator. Worth a test pinning root-level duplication; or pass `[]` and rely on site-level validation.

---

### **suggestion:** Loading toast for fast operations is noisy. **[Confidence: LOW]**

For pages with no/few assets, the duplicate completes in ~200ms and the user sees a "Duplicating page..." toast flash before the success toast. Consider showing the loading toast only after an 800ms delay.

---

### **suggestion:** `tableScopeResourceId` deserves a one-line comment. **[Confidence: LOW]**

`DuplicatePageModal` reads `modal.tableScopeResourceId` for invalidation. A short comment in `atoms.ts` explaining what this represents (the current folder/site-root showing in the table) helps the next reader.

---

### **chore:** `vi` imported twice in `page.router.test.ts`. **[Confidence: HIGH]**

Quick dedupe.

---

### **note:** `createPage` outer catch is now redundant. **[Confidence: HIGH]**

After the refactor, `insertPageBlobResourceAndAudit` rethrows `uniqueViolation` as `CONFLICT`. The outer `db.transaction(...).catch(...)` in `createPage` does the same. Harmless but worth deleting.

---

### **issue (security):** `tryParseSiteAssetFileKey` doesn't reject `..` in filename. **[Confidence: LOW]**

The regex check on `uuidPart` is correct, but the `filePart` is unconstrained. S3 treats keys opaquely so this isn't an immediate vulnerability, but rejecting traversal-like substrings (`..`, leading `/`) is cheap defence in depth.

---

### **praise:** Clean separation of concerns. **[Confidence: HIGH]**

Each helper does one job and is testable in isolation. `tryParseSiteAssetFileKey` in particular is precise — explicit checks, correct UUID regex, easy to extend.

brendancjz:
> **THIS IS A AUTOMATED REVIEW BY CLAUDE**
> 
> Reviewed against this branch's diff. Findings below use [Conventional Comments](https://conventionalcomments.org/), each with a confidence rating.
> 
> ### **praise:** Strong test coverage and helper extraction. **[Confidence: HIGH]**
> `duplicatePageContent.test.ts` covers leading-slash paths, placeholder rejection, dedup, longest-first rewrite, and fresh-UUID generation. The helper extraction (`insertPageWithBlob`, `duplicatePagePreconditions`, `duplicatePageContent`) is clean and the `createPage` refactor to share `insertPageBlobResourceAndAudit` is a real win.
> 
> ### **issue:** S3 copy + DB insert are not transactional — orphan files on failure. **[Confidence: HIGH]**
> The flow is (1) validate, (2) `Promise.all(copyTasks)` copying every asset to a new key in S3, (3) AJV-validate the rewritten blob, (4) DB transaction insert.
> 
> If step 3 or 4 fails, the S3 copies remain orphaned with no compensating delete. Mitigations: track copied keys and `markFileAsDeleted` them in a `try/catch` covering steps 3+4; or insert with old refs first then copy + rewrite + update blob; or schedule copies in a worker after the resource exists.
> 
> ### **issue:** Soft-deleted files get copied along with their lifecycle tag. **[Confidence: MEDIUM]**
> `copyObjectInBucket` uses default `TaggingDirective` (`COPY`). Your TODO already flags this. If a page references an asset that was soft-deleted (still readable but tagged for cleanup), the duplicate's copy carries the same tag and will be removed by your S3 lifecycle policy — so the duplicate ends up with broken assets. Set `TaggingDirective: "REPLACE"` with an empty tag set, or remove the soft-delete tag explicitly after copy.
> 
> ### **issue:** `cloneDeep` from lodash on large blobs is slow — prefer `structuredClone`. **[Confidence: MEDIUM]**
> ```ts
> const newContent = cloneDeep(fullPage.content) as IsomerSchema
> ```
> 
> Native `structuredClone` (Node ≥ 17) handles JSON-shaped data faster and more correctly.
> 
> ### **issue:** `rewriteAssetFileKeysInValue` mutates strings in place via substring match. **[Confidence: MEDIUM]**
> The rewriter does naive string-substitution to find old keys inside arbitrary string values (e.g. ``). Sorting longest-first reduces overlap risk, but: (1) if a page coincidentally contains a string matching an old key in plain text, it will be rewritten; (2) the function mutates `value` in place without the name making that clear.
> 
> A stricter alternative: only rewrite values exactly `"/key"` or `"key"` and surface HTML asset references through structured fields.
> 
> ### **issue:** Permalink uniqueness check has TOCTOU race. **[Confidence: LOW]**
> `assertDuplicatePagePreconditions` queries for an existing permalink, then _outside_ the transaction we copy S3 + insert. The DB insert already handles `uniqueViolation` → `CONFLICT` correctly, so the worst case is wasted S3 copy. The current code does both checks (precondition + insert constraint), which is redundant — drop one.
> 
> ### **issue:** Permission check for root pages passes `[null]` to bulk validator. **[Confidence: MEDIUM]**
> ```ts
> await bulkValidateUserPermissionsForResources({
>   siteId, action: "create", userId,
>   resourceIds: [parentKey],   // can be null
> })
> ```
> 
> When the source page lives at root, `parentKey` is `null` and `resourceIds: [null]` is passed. Behaviour depends on the validator. Worth a test pinning root-level duplication; or pass `[]` and rely on site-level validation.
> 
> ### **suggestion:** Loading toast for fast operations is noisy. **[Confidence: LOW]**
> For pages with no/few assets, the duplicate completes in ~200ms and the user sees a "Duplicating page..." toast flash before the success toast. Consider showing the loading toast only after an 800ms delay.
> 
> ### **suggestion:** `tableScopeResourceId` deserves a one-line comment. **[Confidence: LOW]**
> `DuplicatePageModal` reads `modal.tableScopeResourceId` for invalidation. A short comment in `atoms.ts` explaining what this represents (the current folder/site-root showing in the table) helps the next reader.
> 
> ### **chore:** `vi` imported twice in `page.router.test.ts`. **[Confidence: HIGH]**
> Quick dedupe.
> 
> ### **note:** `createPage` outer catch is now redundant. **[Confidence: HIGH]**
> After the refactor, `insertPageBlobResourceAndAudit` rethrows `uniqueViolation` as `CONFLICT`. The outer `db.transaction(...).catch(...)` in `createPage` does the same. Harmless but worth deleting.
> 
> ### **issue (security):** `tryParseSiteAssetFileKey` doesn't reject `..` in filename. **[Confidence: LOW]**
> The regex check on `uuidPart` is correct, but the `filePart` is unconstrained. S3 treats keys opaquely so this isn't an immediate vulnerability, but rejecting traversal-like substrings (`..`, leading `/`) is cheap defence in depth.
> 
> ### **praise:** Clean separation of concerns. **[Confidence: HIGH]**
> Each helper does one job and is testable in isolation. `tryParseSiteAssetFileKey` in particular is precise — explicit checks, correct UUID regex, easy to extend.

Morning @seaerchin, cool notes from claude! Should I try to resolve all these issues/suggestions?

seaerchin:
> > **THIS IS A AUTOMATED REVIEW BY CLAUDE**
> > Reviewed against this branch's diff. Findings below use [Conventional Comments](https://conventionalcomments.org/), each with a confidence rating.
> > ### **praise:** Strong test coverage and helper extraction. **[Confidence: HIGH]**
> > `duplicatePageContent.test.ts` covers leading-slash paths, placeholder rejection, dedup, longest-first rewrite, and fresh-UUID generation. The helper extraction (`insertPageWithBlob`, `duplicatePagePreconditions`, `duplicatePageContent`) is clean and the `createPage` refactor to share `insertPageBlobResourceAndAudit` is a real win.
> > ### **issue:** S3 copy + DB insert are not transactional — orphan files on failure. **[Confidence: HIGH]**
> > The flow is (1) validate, (2) `Promise.all(copyTasks)` copying every asset to a new key in S3, (3) AJV-validate the rewritten blob, (4) DB transaction insert.
> > If step 3 or 4 fails, the S3 copies remain orphaned with no compensating delete. Mitigations: track copied keys and `markFileAsDeleted` them in a `try/catch` covering steps 3+4; or insert with old refs first then copy + rewrite + update blob; or schedule copies in a worker after the resource exists.
> > ### **issue:** Soft-deleted files get copied along with their lifecycle tag. **[Confidence: MEDIUM]**
> > `copyObjectInBucket` uses default `TaggingDirective` (`COPY`). Your TODO already flags this. If a page references an asset that was soft-deleted (still readable but tagged for cleanup), the duplicate's copy carries the same tag and will be removed by your S3 lifecycle policy — so the duplicate ends up with broken assets. Set `TaggingDirective: "REPLACE"` with an empty tag set, or remove the soft-delete tag explicitly after copy.
> > ### **issue:** `cloneDeep` from lodash on large blobs is slow — prefer `structuredClone`. **[Confidence: MEDIUM]**
> > ```ts
> > const newContent = cloneDeep(fullPage.content) as IsomerSchema
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Native `structuredClone` (Node ≥ 17) handles JSON-shaped data faster and more correctly.
> > ### **issue:** `rewriteAssetFileKeysInValue` mutates strings in place via substring match. **[Confidence: MEDIUM]**
> > The rewriter does naive string-substitution to find old keys inside arbitrary string values (e.g. ``). Sorting longest-first reduces overlap risk, but: (1) if a page coincidentally contains a string matching an old key in plain text, it will be rewritten; (2) the function mutates `value` in place without the name making that clear.
> > A stricter alternative: only rewrite values exactly `"/key"` or `"key"` and surface HTML asset references through structured fields.
> > ### **issue:** Permalink uniqueness check has TOCTOU race. **[Confidence: LOW]**
> > `assertDuplicatePagePreconditions` queries for an existing permalink, then _outside_ the transaction we copy S3 + insert. The DB insert already handles `uniqueViolation` → `CONFLICT` correctly, so the worst case is wasted S3 copy. The current code does both checks (precondition + insert constraint), which is redundant — drop one.
> > ### **issue:** Permission check for root pages passes `[null]` to bulk validator. **[Confidence: MEDIUM]**
> > ```ts
> > await bulkValidateUserPermissionsForResources({
> >   siteId, action: "create", userId,
> >   resourceIds: [parentKey],   // can be null
> > })
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When the source page lives at root, `parentKey` is `null` and `resourceIds: [null]` is passed. Behaviour depends on the validator. Worth a test pinning root-level duplication; or pass `[]` and rely on site-level validation.
> > ### **suggestion:** Loading toast for fast operations is noisy. **[Confidence: LOW]**
> > For pages with no/few assets, the duplicate completes in ~200ms and the user sees a "Duplicating page..." toast flash before the success toast. Consider showing the loading toast only after an 800ms delay.
> > ### **suggestion:** `tableScopeResourceId` deserves a one-line comment. **[Confidence: LOW]**
> > `DuplicatePageModal` reads `modal.tableScopeResourceId` for invalidation. A short comment in `atoms.ts` explaining what this represents (the current folder/site-root showing in the table) helps the next reader.
> > ### **chore:** `vi` imported twice in `page.router.test.ts`. **[Confidence: HIGH]**
> > Quick dedupe.
> > ### **note:** `createPage` outer catch is now redundant. **[Confidence: HIGH]**
> > After the refactor, `insertPageBlobResourceAndAudit` rethrows `uniqueViolation` as `CONFLICT`. The outer `db.transaction(...).catch(...)` in `createPage` does the same. Harmless but worth deleting.
> > ### **issue (security):** `tryParseSiteAssetFileKey` doesn't reject `..` in filename. **[Confidence: LOW]**
> > The regex check on `uuidPart` is correct, but the `filePart` is unconstrained. S3 treats keys opaquely so this isn't an immediate vulnerability, but rejecting traversal-like substrings (`..`, leading `/`) is cheap defence in depth.
> > ### **praise:** Clean separation of concerns. **[Confidence: HIGH]**
> > Each helper does one job and is testable in isolation. `tryParseSiteAssetFileKey` in particular is precise — explicit checks, correct UUID regex, easy to extend.
> 
> Morning @seaerchin, cool notes from claude! Should I try to resolve all these issues/suggestions?

heya, sorry for the spam... was testing out a claude workflow to prevent stale PRs for too long; i think just use your discretion - not everything is worth resolving. i do think the reviewer (me) should take some time out to check the suggestions as otherwise it becomes very easy to delegate all responsibility to the PR writer so apologies on this

eg: issue: S3 copy + DB insert are not transactional — orphan files on failure. [Confidence: HIGH]
for me, at least, i honestly don't really care if there are orphaned s3 files. the reason is that the cost of storage is low and this is unlikely to ever rise to a huge amount and the file itself is gated behind a UUID prefix, meaning that even if the contents are sensitive, it's not practically guessable anyway. f

additionally, this is quite unlikely to impact studio users since they can just retry if it fails - the retry mechanism is transparent to them (just clicking a button) so i think it's not worth the effort to fix and we can just pay a small price. 

i do think `issue: cloneDeep from lodash on large blobs is slow — prefer structuredClone. [Confidence: MEDIUM]` is worth fixing - we store data as a structured json blob. some of these can get really large, especially on the searchable table. i did [some googlefu](https://benchmarklab.azurewebsites.net/Benchmarks/Show/24343/0/lodash-clonedeep-vs-structuredclone-vs-json-parse-deep) and apparently this is correct so might be worth doing

seaerchin:
heya just an update - adrian is actually out on leave for now, so unfortunately this might be abit blocked till he comes back

adriangohjw (commented):
**Senior-engineer review of #1988 (Duplicate page feature).** Conventional Comments with confidence.

## Summary

Adds a "Duplicate page" mutation + modal. Flow lives in `apps/studio/src/server/modules/page/page.router.ts:626` `duplicatePage` — preconditions → load source content via `getFullPageById` → collect asset S3 keys → `Promise.all` copy each to a fresh UUID → rewrite blob → AJV validate → insert in a transaction via the shared helper `insertPageBlobResourceAndAudit` at `apps/studio/src/server/modules/page/helpers/insertPageWithBlob.ts`. The `createPage` refactor onto the same helper is a real cleanup.

The helper trio (`duplicatePageContent`, `duplicatePagePreconditions`, `insertPageWithBlob`) is clean and well-unit-tested.

Main concerns are around the async S3 flow (unbounded parallelism, soft-delete tag propagation), the silent draft-vs-published behaviour, and a couple of UX edges in the modal.

---

## Findings

**issue:** `apps/studio/src/server/modules/page/page.router.ts:660` — `Promise.all(copyTasks)` fires every asset copy in parallel with no concurrency cap. The earlier discussion on this PR explicitly called out image-heavy pages (e.g. `ImageGallery`) as an accepted slow path, but unbounded parallelism is a different failure mode from "slow": for a page with dozens of assets, you can exhaust the SDK's keep-alive pool and S3 will start throttling. Bound with `p-limit` (e.g. 8) or a manual queue. Side benefit: the rejection handler currently surfaces only the first failure while the rest keep running and create orphans.

Confidence: high

---

**issue:** `apps/studio/src/lib/s3.ts:65` — `copyObjectInBucket` uses the default `TaggingDirective` (`COPY`), and the in-file TODO already flags this. If the source asset carries the `deletedAt` soft-delete tag (still GET-able but slated for lifecycle removal), the duplicate inherits the tag and S3 will delete the copy on its next sweep, leaving the duplicated page with broken images. Pass `TaggingDirective: "REPLACE"` with an empty `Tagging: ""`. Worth resolving before merge — the TODO is a known landmine.

Confidence: high

---

**issue:** `apps/studio/src/server/modules/page/page.router.ts:649` — `getFullPageById` preferentially returns the draft blob (`resource.service.ts:137`). If a user opened a published page, edited it (saving a draft), then clicks "Duplicate page" from the dashboard table, the duplicate silently inherits the WIP draft rather than what's live. That's surprising — the user invoked duplicate from the resource list, not from inside the editor. Either (a) document the intent on the mutation, or (b) explicitly fetch the published version (fall through to draft only when no published version exists).

Confidence: medium

---

**issue:** `apps/studio/src/server/modules/page/page.router.ts:651` — `cloneDeep(fullPage.content)` on a multi-MB blob is notably slower than the native `structuredClone` (Node ≥ 17). For pages with large nested content (common on collection landing pages, image galleries), this is a free win and the only existing dep on lodash for this call goes away.

Confidence: high

---

**issue:** `apps/studio/src/server/modules/page/helpers/duplicatePageContent.ts:131` — the fallback `s.replaceAll(\`/${oldKey}\`, \`/${newKey}\`)` does substring substitution across arbitrary string values. Sorting longest-first reduces overlap inside the asset-key set, but it does nothing about plain prose. A user writing literal text like `"see /3/<uuid>/photo.png in the staging bucket"` in a paragraph block will have that string rewritten on duplicate. Unlikely but possible. Stricter alternative: only rewrite when `tryParseSiteAssetFileKey` matches the whole string, and surface HTML-embedded image refs through structured fields rather than substring scanning.

Related: the function signature says "Mutates `value` in place" but the top-level `visit(value)` discards the return — if `value` were a primitive string (it never is in practice) the call would be a no-op. Either return the rewritten value or document that the input must be an object/array.

Confidence: medium

---

**issue:** `apps/studio/src/server/modules/page/helpers/duplicatePagePreconditions.ts:23` — when the resource exists but `type !== ResourceType.Page` (e.g. `CollectionPage`), the error is `"Page not found"`. The resource was found; it's just the wrong type. The integration test at `page.router.test.ts:1929` even asserts this misleading message. The UI gates the menu item on `type === ResourceType.Page`, so this only surfaces via direct API calls — but the error is still wrong. Suggest `"Only top-level pages can be duplicated"` or similar.

Confidence: medium

---

**issue:** `apps/studio/src/features/dashboard/components/DuplicatePageModal/DuplicatePageModal.tsx:142` — `setDuplicatePageModal(...)` in `ResourceTableMenu.tsx:84` can replace the atom while a previous duplicate is in flight. The previous mutation's `onSuccess` will then close the modal that's now showing a different page's form, and the invalidation will target `tableScopeResourceId` from the prior trigger — fine — but the second user click feels like it cancels the first. Consider disabling the menu item while `isPending` on the mutation, or guarding `setDuplicatePageModal` if a duplicate is already running.

Confidence: low

---

**suggestion:** `apps/studio/src/server/modules/page/page.router.ts:585` — after refactoring to `insertPageBlobResourceAndAudit` (which already rethrows `uniqueViolation` as `CONFLICT`), the outer `.catch` on the `createPage` transaction at line 605–620 is partially redundant: only the FK-violation branch (Site not found) is still doing useful work. Trim or document why both layers exist.

Confidence: high

---

**suggestion:** `apps/studio/src/schemas/page.ts:21,30` — `pageTitleSchema` and `permalinkSchema` were promoted from `const` to `export const`, but nothing outside this file consumes them (the new `duplicatePageSchema` is defined in the same file). Revert to non-exported unless there's another consumer planned.

Confidence: high

---

**suggestion (UX, non-blocking):** `apps/studio/src/features/dashboard/components/DuplicatePageModal/DuplicatePageModal.tsx:145` — the loading toast fires immediately. For zero-asset pages the duplicate completes in ~200ms and the user sees the toast flash before the success toast. Gate the toast behind an 800ms delay, or only show it once `assetKeys.length > some_threshold`.

Confidence: low

---

**note:** the prior bot review flagged "`vi` imported twice in `page.router.test.ts`" — that's incorrect (single import at line 26). It also flagged passing `[parentKey]` (which can be `null`) to `bulkValidateUserPermissionsForResources` as suspicious — that's actually the established pattern (see `createPage` at `page.router.ts:561`, which does the exact same thing for root-level inserts), so no action needed there.

Confidence: high

---

**praise:** the helper extraction — `insertPageBlobResourceAndAudit`, `assertDuplicatePagePreconditions`, the parse/collect/rewrite trio in `duplicatePageContent.ts` — is genuinely good. The unit tests pin the right invariants (leading-slash handling, dedup, longest-first rewrite, fresh UUID generation) and the `createPage` ↔ `duplicatePage` shared insert path is a real reduction in surface area.

Confidence: high

adriangohjw:
LOL ignore the "Senior-engineer review" - it's the prompt in the skills LOL 💀

adriangohjw:
note to self: not only do we have to duplicate the images, we also have to duplicate every isomer-user-content in the prose blocks etc.

seaerchin:
heya @brendancjz appreciate your contribution to Isomer - can i check if this is still actively developed?

adriangohjw:
@brendancjz thanks for the work but after internal discussion, the team decides to rebuild this with some product considerations we have in mind! nevertheless, this PR is a great starting point for us in terms of the exploration process, so thank you regardless! :D

brendancjz:
> @brendancjz thanks for the work but after internal discussion, the team decides to rebuild this with some product considerations we have in mind! nevertheless, this PR is a great starting point for us in terms of the exploration process, so thank you regardless! :D

Hey @adriangohjw, apologies for the absence recently. No worries, thanks for the opportunity to help contribute! I'll continue to help out when I'm have some time. Cheers!

PR #1995, opened 2026-04-14
👮‍♀️ feat: Anti-scam (GOIS) site notification and optional union form control

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 14 | +360 | -29 |
| 🧪 Test | 4 | +104 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Sites need a standardized **Government Official Impersonation Scam (GOIS)** disclaimer banner option alongside the existing custom notification banner, without operators retyping fixed copy. The notification schema and Studio form must support choosing between **custom** copy and the **anti-scam** preset, while remaining compatible with existing saved notification data.

## Solution

**Breaking Changes**

  - Details ...

Existing notifications without a `type` field are still treated as custom banners: `isSiteNotificationActive` treats missing `type` like legacy data and requires a title for the banner to be active. Anti-scam notifications use `type: "antiscam"` and do not require a user-defined title. `getNotification` skips legacy array-to-prose migration for `antiscam` payloads.

**Features**:

- **Anti-scam site notification**: `NotificationSchema` is a union of **Custom notification** (optional `type: "custom"`, title + optional content) and **Anti-scam disclaimer** (`type: "antiscam"`). The Next.js `Notification` component renders fixed title and body from `antiScamSiteNotification` (aligned with the AntiScam disclaimer pattern, including ScamShield link).
- **`isSiteNotificationActive`**: Central helper to decide whether a saved notification should show in the layout and whether Studio should treat the form as having an active banner; exported from `@opengovsg/isomer-components`.
- **`JsonFormsOptionalCombinatorControl`**: New JsonForms renderer for **optional** `anyOf` / `oneOf` unions of **object** branches—a master switch to enable or clear the combinator while preserving prior input when toggled—registered in the Studio form builder.

**Improvements**:

- **Skeleton** uses `isSiteNotificationActive` instead of only checking `site.notification?.title`, so anti-scam banners render without a stored title.
- **Studio notification settings** initializes form state and enables Save only when the notification is active (custom with title, or anti-scam).
- **Storybook**: Story coverage for the anti-scam notification variant.

**Bug Fixes**:

- None called out.

## Before & After Screenshots

https://github.com/user-attachments/assets/7d2d2249-b871-4085-84c9-1de8729e5c27

## Tests

Automated: `isSiteNotificationActive` unit tests in `packages/components/src/utils/__tests__/isSiteNotificationActive.test.ts`.

- [ ] In **Studio**, open a site → **Settings** → **Notification banner**. Confirm the form shows a **type** choice between custom and anti-scam (union / combinator UI).
- [ ] Select **Anti-scam disclaimer**, save, then open the **published or preview** site. Confirm the banner shows the fixed GOIS title and ScamShield copy (and behaves like other notifications for dismiss, if applicable).
- [ ] Select **Custom notification**, set title (and optional body), save, and confirm the **custom** banner appears as before.
- [ ] For a site that **already had** a saved custom notification from before this change (no `type` field), confirm the banner still appears when a title is present and saving still works.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new union-typed site notification config and updates both Studio and Next rendering logic, which could affect persisted notification data and when banners show. Risk is moderated by backward-compat checks for legacy notifications and targeted tests for `isSiteNotificationActive`.
> 
> **Overview**
> Adds an **anti-scam (GOIS) preset** as a new `type: "antiscam"` branch alongside the existing custom notification, including fixed title/body copy and updated Storybook coverage.
> 
> Updates notification activation and rendering to be **type-aware** via new `isSiteNotificationActive` (used by Next `Skeleton` and Studio notification settings init), and adjusts server `getNotification` to skip legacy array-to-prose migration for `antiscam` payloads.
> 
> Introduces `JsonFormsOptionalCombinatorControl` (registered in Studio’s form builder) to render optional `anyOf`/`oneOf` object unions with a master toggle that clears/restores the union value, and tweaks `JsonFormsCombinatorControl` spacing for radio-format variants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08b54a2d2e0fbb48dfe98f6c348fd04ac8f59b24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seaerchin (commented):

## Diff summary

**Schema reshape** ([`Notification.ts`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/packages/components/src/interfaces/internal/Notification.ts)): `NotificationSchema` becomes `Type.Union([NotificationCustomSchema, NotificationAntiScamSchema])` — custom branch (optional `type: "custom"`, required `title`, optional `content`) plus antiscam branch (required `type: "antiscam"`, no other fields). Adds `SiteNotificationConfig` and `NotificationTitleContentConfig` types.

**Render branch** ([`Notification.tsx`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/packages/components/src/templates/next/components/internal/Notification/Notification.tsx)): If `type === "antiscam"`, renders fixed title + body from new [`antiScamSiteNotification.ts`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/packages/components/src/templates/next/components/internal/Notification/antiScamSiteNotification.ts) constant (with ScamShield link). Otherwise renders the existing custom-banner path unchanged.

**Active-banner gate** ([`isSiteNotificationActive.ts`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/packages/components/src/utils/isSiteNotificationActive.ts)): central helper used by both [`Skeleton.tsx`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/packages/components/src/templates/next/layouts/Skeleton/Skeleton.tsx) (which previously short-circuited on `site.notification?.title`) and [`pages/sites/[siteId]/settings/notification.tsx`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/apps/studio/src/pages/sites/%5BsiteId%5D/settings/notification.tsx) (form init).

**Server normalization** ([`site.service.ts`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/apps/studio/src/server/modules/site/site.service.ts)): `getNotification` returns antiscam payloads as-is, skipping the legacy array→prose migration.

**Studio form control** ([`JsonFormsOptionalCombinatorControl.tsx`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsOptionalCombinatorControl.tsx)): new ranked renderer (rank 4 — beats AnyOf/OneOf at 3) that provides a master Switch to toggle the union on/off; preserves the prior data via a snapshot when toggled off so re-enabling restores it. Tester runs only on optional `anyOf`/`oneOf` of object-variant unions.

**Coverage**: unit tests for `isSiteNotificationActive` (legacy, custom, antiscam, empty-title); Storybook stories for the antiscam variant; MSW handler for getNotification antiscam mock.

---

## Review

### **praise:** Backward-compat for legacy notification data is solid (confidence: high)
`isSiteNotificationActive` handles the no-`type` legacy case via `if (!notification.type) return !!notification.title`, and `Notification.tsx`'s `if (type === "antiscam") else { ... }` branch correctly covers both `"custom"` and `undefined`. The TypeBox optional `Type.Literal("custom", { default: "custom" })` also gives a parse-time default for new writes.

### **issue (minor):** `_exhaustiveCheck` returns its assigned value, not a boolean (confidence: high)
```ts
default:
  const _exhaustiveCheck: never = notification.type
  return _exhaustiveCheck
```
At runtime, if `notification.type` is ever an unknown variant (downcast from `unknown`, hand-edited DB row, future schema), the assignment evaluates fine and the function returns that string. The signature is `boolean`. Prefer `throw new Error(\`Unknown notification type: ${_exhaustiveCheck}\`)` or `return false`.

### **question:** antiscam path skips array-migration unconditionally (confidence: medium)
[`site.service.ts`](https://github.com/opengovsg/isomer/blob/feat/gois-notification/apps/studio/src/server/modules/site/site.service.ts) returns `result` directly when `notification.type === "antiscam"`. If a legacy record somehow has `type: "antiscam"` plus a stray `content` array (malformed write, manual DB edit), it would leak through to clients. Is the antiscam payload normalized on write to enforce that no `content` is persisted? A one-line confirmation in the PR description would help.

### **suggestion:** `JsonFormsOptionalCombinatorControl.isChecked` doesn't track external data changes (confidence: medium)
`useState(!isEmpty(data))` initializes once. If the underlying JsonForms data is reset externally (server reload, parent form reset), `isChecked` will desync from data presence — the toggle could read "off" while data is non-empty, or vice versa. A `useEffect(() => setIsChecked(!isEmpty(data)), [data])`, or deriving the toggle state from `data` instead of state, closes the gap. Likely OK for the current form lifecycle, just brittle.

### **note:** snapshot-on-toggle ordering is fine but worth a one-line comment (confidence: high)
```ts
if (isChecked) {
  setDataSnapshot(data)
  handleChange(path, undefined)
}
setIsChecked((prev) => !prev)
```
React 18 batches state updates within event handlers, so the snapshot capture and the external dispatch happen together. No race, but the dependency on the `data` closure being the latest prop is non-obvious for the next reader.

### **suggestion:** consider a shared `assertNever` helper (confidence: low)
The `_exhaustiveCheck: never` pattern is showing up in multiple places. A one-line shared utility (`assertNever(x: never): never { throw ... }`) cleans this up everywhere and avoids the boolean-vs-runtime mismatch above.

### **praise:** test coverage on `isSiteNotificationActive` is thorough (confidence: high)
All three branches (legacy, custom, antiscam) plus empty-title for both legacy and custom — six cases in 68 lines.

---

LGTM with minor nits. The exhaustive-check return-value issue is the only one I'd ask be fixed before merge; the rest are notes / suggestions.

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -93,21 +93,26 @@ export const getNotification = async (
...
 
-  // NOTE: Handle old array format
-  // Add in the `prose` wrapper
-  if (Array.isArray(result.notification.content)) {
+  if (notification.type === "antiscam") {
+    return result
+  }
+
+  // NOTE: Handle old array format — add in the `prose` wrapper
+  if (Array.isArray(notification.content)) {
```
dcshzj, line 145: question(non-blocking): not directly related to this PR but do we still need this? I think would be worth doing a one-time migration if there are any, then we can remove this branch to simplify the code

packages/components/src/templates/next/components/internal/Notification/antiScamSiteNotification.ts
```diff
@@ -0,0 +1,30 @@
+import type { NotificationTitleContentConfig } from "~/interfaces/internal/Notification"
+
+/** Standard anti-scam GOIS (Government Official Impersonation Scam) disclaimer, aligned with AntiScamDisclaimerBanner. */
+export const antiScamSiteNotification = {
```
dcshzj, line 4: question(non-blocking): hmm any value if we make the current AntiScamDisclaimerBanner component render contents from this object? Seems doable but not sure if the aria labels are required

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsOptionalCombinatorControl.tsx
```diff
@@ -0,0 +1,198 @@
...
+  isControl,
+  isOneOfControl,
+  or,
+  rankWith,
+  resolveSchema,
+  schemaMatches,
+} from "@jsonforms/core"
+import { JsonFormsDispatch } from "@jsonforms/react"
+import { Switch } from "@opengovsg/design-system-react"
+import isEmpty from "lodash/isEmpty"
```
dcshzj: minor nit: lodash-es?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsOptionalCombinatorControl.tsx
dcshzj, line 1: question: hmm just to understand this new control a bit, it is to add the toggle to activate or deactivate the notification banner?

Is this the same as the "Set a thumbnail image" that we have on index pages?

apps/studio/src/server/modules/site/site.service.ts
adriangohjw: Not too sure tbh so I didn't change this haha

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsOptionalCombinatorControl.tsx
adriangohjw: Oh yes, this was prior to that, good catch

adriangohjw:
be gone for now

PR #2002, opened 2026-04-16
🔖 feat(studio): Refresh collection index drawer and dashboard copy

adriangohjw:
## Problem

Collection index pages in Studio needed a clearer editing layout and copy when the **new collection editing experience** is enabled: the root drawer should surface collection-specific actions separately from generic “fixed blocks,” and the dashboard should describe collections consistently with that experience.

Closes https://linear.app/ogp/issue/ISOM-2283/refresh-collection-index-drawer-and-dashboard-copy

## Solution

This PR refreshes the collection index **root drawer** (behind the GrowthBook flag), tightens **BaseBlock** behaviour for this flow, and updates the **dashboard** subtitle for collection resources when the flag is on. **Storybook** coverage is expanded for the new experience, including Isomer Admin vs non-admin cases.

**Breaking Changes**

  - Details ...

**Features**:

- **Manage Collection** section in `RootStateDrawer` when layout is `collection` and the new collection editing experience flag is enabled: heading, helper text, and fixed actions grouped separately from the generic “Fixed blocks” pattern.
- **Collection display** fixed action (replaces the previous single “Collection settings” entry for this flow): opens the collection editor; updated label and description; supports larger icon via `iconProps` and vertical spacing via `containerProps` on `BaseBlock`.
- **Filters** fixed action, shown only to **Isomer Admin** users (Core or Migrator roles): label, description, and slider icon. _(Click handler is still a placeholder pending implementation.)_
- Dashboard **collection** row subtitle uses `useNewCollectionEditingExperience()`: when the flag is on, copy is “Manage the Collection’s layout, filters, and sorting.”

**Improvements**:

- **`BaseBlock`**: discriminated props for optional `icon` vs `icon` + `iconProps`; early `return null` when `isHidden` instead of rendering a disabled-looking card; return type `null` instead of `false`; removed unused “hidden” styling branches since hidden blocks no longer render here.
- **`getIndexPageSubtitle`**: object parameter for `type` and `isNewCollectionEditingExperienceEnabled` for clearer call sites.
- **Storybook**: `CollectionPage` story asserting new subtitle when the flag is enabled; `EditCollectionIndexPage` stories split for non-admin vs Isomer Admin, plus display and filters interaction plays where applicable.

**Bug Fixes**:

- Code quality: address “useless conditional” findings in the drawer/block flow where appropriate.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Enable the **new collection editing experience** GrowthBook flag for your test user/site (same flag used by `useNewCollectionEditingExperience`).
- [ ] In the **dashboard**, open a site that has a **collection** resource: confirm the collection row subtitle reads **“Manage the Collection’s layout, filters, and sorting.”** with the flag on; with the flag off, confirm the previous subtitle still appears.
- [ ] Open a **collection index page** in the Studio editor with the flag on: confirm a **“Manage Collection”** section with helper text, then **Collection display** (and icon/spacing consistent with the updated layout).
- [ ] As a user **without** Isomer Admin (Core/Migrator): confirm the **Filters** block is **not** shown.
- [ ] As an **Isomer Admin** (Core or Migrator): confirm the **Filters** block **is** shown; clicking it currently only hits a placeholder handler (no full filters UI yet—verify no crash).
- [ ] Click **Collection display** and confirm the drawer transitions to the **collection** editor state as before.
- [ ] With the flag **off**, confirm collection index editing still uses the previous **“Fixed blocks”** drawer layout and behaviour.

### DIFF
--- a/apps/studio/src/features/editing-experience/components/Block/BaseBlock.tsx
+++ b/apps/studio/src/features/editing-experience/components/Block/BaseBlock.tsx
@@ -9,6 +9,7 @@ export interface BaseBlockProps {
   dragHandle?: React.ReactNode
   label: string
   description?: string
+  variant?: "horizontal" | "vertical"
   containerProps?: StackProps
   onClick?: () => void
   draggableProps?: DraggableProvidedDragHandleProps | null
@@ -23,6 +24,7 @@ export const BaseBlock = ({
   dragHandle,
   label,
   description,
+  variant = "horizontal",
   draggableProps,
   containerProps,
   onClick,
@@ -92,10 +94,10 @@ export const BaseBlock = ({
           borderColor: "utility.feedback.critical",
         }}
         bg={isHidden ? "gray.100" : "white"}
-        py="0.75rem"
-        px="0.75rem"
+        py={variant === "vertical" ? "1.25rem" : "0.75rem"}
+        px={variant === "vertical" ? "1.25rem" : "0.75rem"}
         flexDirection="row"
-        align="center"
+        align={variant === "vertical" ? "flex-start" : "center"}
         textAlign="start"
         onClick={onClick}
         opacity={isHidden ? 0.7 : 1}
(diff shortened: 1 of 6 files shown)


apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx
```diff
@@ -105,16 +117,40 @@ const FixedBlock = () => {
...
+          }}
+          label="Collection display"
+          description="Customise the Collection’s Summary, Layout, Sorting logic, and Thumbnail."
+          containerProps={containerProps}
+          icon={BiCog}
+          iconProps={iconProps}
+        />
+        {isUserIsomerAdmin && (
+          <BaseBlock
+            onClick={() => console.log("to implement")}
```
adriangohjw, line 145: do in subsequent PR https://github.com/opengovsg/isomer/pull/2003

apps/studio/src/features/editing-experience/components/Block/BaseBlock.tsx
```diff
@@ -16,10 +15,20 @@ export interface BaseBlockProps {
...
+} & (
+  | {
+      icon: IconType
+      iconProps?: IconProps
+    }
+  | {
+      icon?: undefined
+      iconProps?: never
+    }
+)
```
seaerchin, line 27: nit: should we use typefest's `RequireAllOrNone`? https://github.com/sindresorhus/type-fest/blob/main/source/require-all-or-none.d.ts

apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx
```diff
@@ -417,179 +455,203 @@ export default function RootStateDrawer() {
...
-                  <Text textStyle="subhead-1">Fixed blocks</Text>
+                  <Text textStyle="subhead-1">Manage Collection</Text>
                   <Text textStyle="caption-2" color="base.content.medium">
-                    These are built into the layout, so you can't delete them.
+                    Modify the Collection’s look and feel or manage filters.
                   </Text>
                 </VStack>
 
                 <FixedBlock />
               </VStack>
```
seaerchin, line 470: didn't read too closely, but i assumed the else branch is existing production code and only this is new

seaerchin (approved):
lgtm! minor nit to swap over to using `RequireAllOrNone` but otherwise looks good

apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx
adriangohjw: yup!

apps/studio/src/features/editing-experience/components/Block/BaseBlock.tsx
adriangohjw: i think its not a complete drop-in replacement because our current type allows three combinations:
- neither
- icon only
- icon + iconProps

`RequireAllOrNone<T, 'icon' | 'iconProps'>` would only allow neither or both, disallowing the icon-only case

PR #2003, opened 2026-04-16
🔖 Feat/implement collection filters drawer

adriangohjw:
## Problem

This PR is stacked on **`feat/change-collection-indexpage-look`**. It completes the collection index editing flow by letting users manage **collection filters** (tag categories and their options) from the collection editor drawer, with validation that prevents ambiguous duplicate labels and UX that matches the new drawer experience.

Closes https://linear.app/ogp/issue/ISOM-2284/implement-collection-filters-drawer

## Solution

**Base branch**: `feat/change-collection-indexpage-look` (all changes below are relative to that branch.)

**Breaking Changes**

**Features**:

- **Collection drawer modes**: `CollectionEditorState` now distinguishes `display` vs `filter`. The drawer header shows **Collection display** or **Manage filters** accordingly; the form schema includes or excludes `tagCategories` / `tags` based on mode and **Isomer admin** access (non-admins still do not see filter fields).
- **JsonForms tag category controls**: Dedicated renderers for `tag-categories` and `tag-category-options` formats—reorderable lists, add/remove, delete **confirmation modals**, optional **required filter** toggle (`isRequired`) for new items, subtitles and trailing actions on list rows.
- **Duplicate label validation**: Custom AJV keyword `uniqueItemPropertiesIgnoreCase` enforces **trimmed, case-insensitive** uniqueness on filter names and option names, aligned with UI duplicate detection helpers. Save stays disabled when builder errors exist (including schema/validation errors).
- **Schema updates** (`packages/components`): Tag category schema documents `isRequired` as optional for legacy rows; arrays declare the new uniqueness keyword for Studio validation.

**Improvements**:

- **`DraggableDrawerButton` / `JsonFormsArrayControl`**: Richer list item API (`listItemTrailing`, `listItemSubtitle`, `listItemContentProps`, drag-handle vs label button prop split), **`emptyState`** for empty arrays, **`belowDescription`** for inline errors, **`mapNewArrayItem`** for controlled defaults when adding rows.
- **Stories**: Adds **`EditCollectionIndexPageNewExperience.stories`** for filter management and interactions; trims redundant **`EditCollectionIndexPage.stories`** coverage superseded by the new experience.

**Bug Fixes**:

- Minor layout fix on array control padding (`p` → `py`) and description spacing behavior.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Check out this branch (or use the PR preview) with **`feat/change-collection-indexpage-look`** as the merge base; run Studio as usual.
- [ ] Open a **Collection** index page in the **new** editing experience; open the collection editor drawer from the UI entry point used on that branch.
- [ ] In **Collection display** mode: confirm title/infobox copy and that **filter fields are hidden** for a **non–Isomer-admin** user (or equivalent test account).
- [ ] As an **Isomer admin** (Core or Migrator): switch to **Manage filters**; add a filter and options, reorder items, toggle **required** on a new filter, and **Save**—reload or re-open and confirm data persists.
- [ ] Attempt **duplicate filter names** and **duplicate option names** that differ only by case or surrounding spaces; confirm **validation errors** appear and **Save** remains disabled until resolved.
- [ ] Delete a filter or option and confirm **confirmation modals** behave correctly and removals persist after save.

**New scripts**:

- None

**New dependencies**:

- None (lockfile-only metadata normalization for an existing package entry)

**New dev dependencies**:

- None

---

> [!NOTE]
> **Medium Risk**
> Adds a new collection filter management flow and custom AJV validation, which affects how collection metadata is edited/saved and can block saves when rules are violated. Risk is moderate due to new UI interactions (nested drawers, delete modals) and schema/validation changes around `tagCategories`/options.
> 
> **Overview**
> Adds a new **collection drawer split**: `collectionEditor` now has `type: "display" | "filter"`, with the root drawer exposing a *Filters* entry (Isomer admins only) and the collection drawer switching header/copy and scoping the form schema to include/exclude `tagCategories`/`tags` accordingly.
> 
> Implements **filter/option management UI** via new JsonForms renderers for `tag-categories` and `tag-category-options`, including reorderable list rows with trailing actions, subtitles, custom empty states, and delete confirmation modals.
> 
> Introduces **case-insensitive duplicate label validation** by registering a custom AJV keyword `uniqueItemPropertiesIgnoreCase`, adding helpers/tests to surface array-level duplicate errors in the UI, and updating the shared collection page schema to annotate filters/options with this constraint (plus an optional `isRequired` flag for backward-compatible required filters).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0e340dc3712c4ca74d35d5c41a9fbaaac275020. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx
```diff
@@ -57,10 +58,32 @@ export default function CollectionEditorStateDrawer(): JSX.Element {
...
+        ? {
+            exclude: ["tagCategories", "tags"],
+          }
+        : {
+            include: ["tagCategories", "tags"],
+          }
+    }
+    return {
+      exclude: ["tagCategories", "tags"],
+    }
```
adriangohjw, line 79: fyi: for now still gatekeep from non-isomeradmins until we dogfood and ready to roll out.

apps/studio/src/stories/Page/EditPage/EditCollectionIndexPage.stories.tsx
adriangohjw, line 1: moved it to a standalone file

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
```diff
@@ -1,26 +1,210 @@
...
+        belowDescription={
+          hasDuplicateFilterNameError ? (
+            <Infobox width="100%" size="sm" variant="error" mt="0.5rem">
+              <Text textStyle="body-2">
+                Each filter must have a unique name. Names are not
+                case-sensitive, so rename the duplicate before saving changes.
+              </Text>
+            </Infobox>
+          ) : undefined
+        }
```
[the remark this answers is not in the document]
adriangohjw: yeah i havent thought of a way to do this yet, without custom ajv :(

apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx
```diff
@@ -57,10 +58,32 @@ export default function CollectionEditorStateDrawer(): JSX.Element {
...
+      return drawerStateType === "display"
+        ? {
+            exclude: ["tagCategories", "tags"],
+          }
+        : {
+            include: ["tagCategories", "tags"],
+          }
+    }
+    return {
+      exclude: ["tagCategories", "tags"],
```
seaerchin, line 78: question: should we collapse this condition? when `drawerType === "display" && isUserIsomerAdmin`, we also return `{ exclude: [...] }`

apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx
```diff
@@ -179,7 +205,10 @@ const SaveButton = ({
       w="100%"
       isLoading={isLoading}
       isDisabled={!isEmpty(errors)}
-      onClick={onClick}
+      onClick={() => {
+        if (!isEmpty(errors)) return
```
seaerchin: question: this is just defensive programming rgiht cos we do have a `isDisabled={!isEmpty(errors)}` on top

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/DraggableTagButton.tsx
```diff
@@ -86,46 +112,86 @@ const DraggableDrawerButton = forwardRef<DraggableDrawerButtonProps, "div">(
               mr="-6px"
```
seaerchin, line 102: question: do you think we should refactor this? this component is getting rather bloated with its props and rather than just piling on everythign onto it, we might want to consider a composition pattern where this bigger component takes a `children` prop and the other components here are defined explicitly. then, the caller decides what children it wants.

seaerchin (commented):
preliminary review - just making sure i understand what this PR is doing.

we are adding the filter here on frontend and delegating the work of rendering (and the logic) to JSON forms here? additionally, we add in extra validation for `label` to make sure that it's unique via `ajvKeywords` to make sure that no 2 categories can have the same label? 

some things that i felt is worth flagging and we might want to invest extra effort into finding a different pattern - some of the files here (`DraggableDrawerButton`, some of the JSON forms renderers) pile on the props. this is one of those slippery slope things where it looks quick and easy to do, but after awhile, the code does 101928374x different things and becomes resistant to change (without piling on yet another prop)

do you think it's worht investing some extra effort into a different pattern or just accepting extra duplication in our codebase? fwiw, i think using the `children` pattern here or a different (and slightly altered) `JSONArrayControl` would be good enough.

apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx
adriangohjw: i think no need for now, since more explicit and readable like this (IMO) - also will update this once we roll out, so this code will be cleaned up in coming weeks

apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx
adriangohjw: yup

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/DraggableTagButton.tsx
adriangohjw: im neutral haha i had the same thought but after thinking i feel like it's still okay and given that it's unclear if this is gonna be used (or expanded on) in the near future, i can't justify the upside of refactoring it

adriangohjw:
> we are adding the filter here on frontend and delegating the work of rendering (and the logic) to JSON forms here? additionally, we add in extra validation for `label` to make sure that it's unique via `ajvKeywords` to make sure that no 2 categories can have the same label?

yup. the unfortunate thing is that:
- we cannot rely on json schema for the "no case-insensitive duplicates" because every filter/option actually comes is an object with uuid so it's always unique
- `ajvKeywords` built-in `uniqueItemProperties` only support case-sensitive, not case-insensitive.

Thus we have to write our own `uniqueItemPropertiesIgnoreCase` (which TBH hard carried by Cursor and i don't know what are the footguns, but blast area is limited to this 2 use cases + there's sufficient test case for me to feel okay with it) - I can't think of another solution 

> some things that i felt is worth flagging and we might want to invest extra effort into finding a different pattern - some of the files here (`DraggableDrawerButton`, some of the JSON forms renderers) pile on the props. this is one of those slippery slope things where it looks quick and easy to do, but after awhile, the code does 101928374x different things and becomes resistant to change (without piling on yet another prop)

replied here: https://github.com/opengovsg/isomer/pull/2003#discussion_r3123680361

> do you think it's worht investing some extra effort into a different pattern or just accepting extra duplication in our codebase? fwiw, i think using the `children` pattern here or a different (and slightly altered) `JSONArrayControl` would be good enough.

yeah i thought about it but i feel like for these cases, there's too much deviation from the default use case, e.g. requires custom modal (with its own backend API call) + error message when there's duplicated labels

I opted for a middle ground where i abstract the `JSONArrayControl` core logic out to use, but i can't think of a better way to reuse more stuff without having too many "if-else"

seaerchin (commented):

## Diff summary

This PR builds the "Manage filters" drawer for collection pages (Isomer-admin gated), reusing the existing collection editor drawer and splitting it into `display`/`filter` modes via a new `type` discriminator on `CollectionEditorState`. New AJV keyword `uniqueItemPropertiesIgnoreCase` enforces trim + case-insensitive uniqueness on `label` for both filter rows and option rows. `JsonFormsArrayControl` is generalised into a reusable `JsonFormsArrayControlView` (icon, subtitle, trailing menu, error caption, custom empty state, `mapNewArrayItem`), and two thin wrappers — `JsonFormsTagCategoryControl` (tag-categories) and the new `JsonFormsTagCategoryOptionsControl` (tag-category-options) — render the row UI with delete-confirmation modals.

- [apps/studio/src/utils/ajvKeywords/uniqueItemPropertiesIgnoreCase.ts](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer/apps/studio/src/utils/ajvKeywords/uniqueItemPropertiesIgnoreCase.ts)
- [apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer/apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx)
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx)
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx)
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx)
- [packages/components/src/types/page.ts](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer/packages/components/src/types/page.ts)

## Review

**praise:** Nice extraction of `JsonFormsArrayControlView` with explicit, well-typed extension points (`renderListItemTrailing`, `renderListItemSubtitle`, `getListItemHasError`, `mapNewArrayItem`, `emptyState`). The inline comment on why `isRequired`'s default is set in `mapNewArrayItem` rather than via JSON Schema `default` (to avoid AJV `useDefaults` retroactively mutating legacy rows) is exactly the kind of context that pays off later. (confidence: high)

**praise:** Test coverage for `indicesWithDuplicateLabels` and `uniqueItemPropertiesIgnoreCase` is thorough — covers empty/whitespace/missing labels, trim+case-fold equivalence, multi-key schemas, non-adjacent duplicates, and the no-mutation invariant. (confidence: high)

**issue (security/auth):** The Isomer-admin gate on filters is only enforced client-side. `CollectionEditorStateDrawer` excludes `tagCategories`/`tags` from the *scoped* schema for non-admins, but `validatedPageProcedure` in `apps/studio/src/server/modules/page/page.router.ts` validates the incoming blob against the full unscoped `IsomerSchema` and there is no role check on `updatePageBlob`. A non-admin with collection write permission can still POST `tagCategories` directly via the tRPC client. The other admin-only renderer (`JsonFormsTagCategoryOptionsControl`) returning `null` for non-admins is purely cosmetic for the same reason. Consider gating server-side (reject `tagCategories`/`tags` mutations from non-admins, or only allow these properties to change when the actor has the admin role). (confidence: high)

**issue (a11y):** In `JsonFormsTagCategoryControl`, every filter row's actions menu uses the same `aria-label="Filter actions"`. Multiple identical labels in the same list make screen-reader navigation ambiguous. The options control already does this correctly (`Option ${index + 1} actions`). Mirror that pattern, or include the filter label (e.g. `Actions for "${label}"`). (confidence: high)

**issue (correctness):** `addItem(path, mapNewArrayItem ? mapNewArrayItem(createDefaultValue(...)) : ...)` in `JsonFormsArrayControl.tsx` is evaluated during render rather than on click. JsonForms' `addItem(path, value)` returns the click handler so a frozen value is fine, but `createDefaultValue(schema, rootSchema)` runs on every render. Hoist into a `useCallback` or compute inside the click handler — robust to future non-pure mappers and avoids extra work. (confidence: medium)

**issue (validation gap):** `Type.String({ maxLength: 70, title: "Filter name" })` has no `minLength`. The uniqueness keyword (by design) ignores empty labels, so two filters with `label: ""` pass both type and uniqueness checks and the UI looks valid. Add `minLength: 1` (and trim before validation) for both filter and option `label`. (confidence: high)

**issue (UX/data integrity):** `DeleteOptionModal` body still contains `"This option is being used in XX items"` with a `// TODO: replace XX with usage count from backend` left in user-visible copy. Either wire this up before merging or replace the copy with something accurate; shipping literal "XX" to admins erodes trust in the warning. (confidence: high)

**suggestion:** `normalizedKey` in the AJV keyword and `normalizedLabelKey` in `indicesWithDuplicateLabels` are the same logic copy-pasted. Extract to a shared util so the "definition of duplicate" can't drift between client UI and AJV. Add a co-located test for combining-mark / Turkish-I edge cases (`"İ".toLowerCase()` vs `"i"`) so any future locale change is caught. (confidence: medium)

**suggestion:** `hasUniqueItemPropertiesError` checks both `uniqueItemProperties` and the new `uniqueItemPropertiesIgnoreCase`, but only the latter is actually registered on `ajv`. Drop the `uniqueItemProperties` branch (dead code) or add a comment that it's parity-for-future-use if the upstream `ajv-keywords` package is wired in later. (confidence: medium)

**question:** When two filter labels are duplicates (case-insensitive) and the user deletes one row via the menu, the other duplicate becomes unique and the array-level `belowDescription` error banner should clear. Does JsonForms' `onChange` re-run AJV on the `removeItems` mutation, or does the banner linger until the next keystroke? Worth a manual check. (confidence: medium)

**note:** `EditCollectionIndexPage.stories.tsx` deletes four feature-flagged stories and re-creates them in a new `EditCollectionIndexPageNewExperience.stories.tsx`. The rename loses git blame on the play functions. Not blocking. (confidence: low)

seaerchin (commented):
## Summary

This PR implements the "Manage filters" drawer for collection pages (Isomer-admin gated), introducing:

- **Drawer mode split**: `CollectionEditorState` now has `type: "display" | "filter"`, routing admins to either collection display settings or filter management
- **Custom AJV keyword**: `uniqueItemPropertiesIgnoreCase` for case-insensitive duplicate detection on filter/option labels
- **New JsonForms renderers**: `JsonFormsTagCategoryControl` and `JsonFormsTagCategoryOptionsControl` with delete confirmation modals, reorderable rows, and duplicate-detection error banners
- **Schema changes**: `packages/components/src/types/page.ts` adds `isRequired` flag, `TRIMMED_NON_EMPTY_STRING_REGEX` pattern, and the new uniqueness keyword

Key flow: `RootStateDrawer.tsx:127-130` → `CollectionEditorStateDrawer.tsx:47-67` (schema scoping) → `JsonFormsTagCategoryControl.tsx` (filter list) → `JsonFormsTagCategoryOptionsControl.tsx` (options list).

---

## Findings

**praise:** The extraction of `JsonFormsArrayControlView` with explicit, well-typed extension points (`renderListItemTrailing`, `renderListItemSubtitle`, `getListItemHasError`, `mapNewArrayItem`, `emptyState`) is clean. The inline comment explaining why `isRequired`'s default is set in `mapNewArrayItem` rather than via JSON Schema `default` (to avoid AJV `useDefaults` mutating legacy rows) is exactly the kind of context that pays off later.

Confidence: high

---

**praise:** Test coverage for `indicesWithDuplicateLabels` and `uniqueItemPropertiesIgnoreCase` is thorough—covers empty/whitespace/missing labels, trim+case-fold equivalence, multi-key schemas, non-adjacent duplicates, and the no-mutation invariant.

Confidence: high

---

**issue (security/auth):** `apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx:54-67`

The Isomer-admin gate on filters is only enforced client-side. `CollectionEditorStateDrawer` excludes `tagCategories`/`tags` from the scoped schema for non-admins, and `JsonFormsTagCategoryOptionsControl` returns `null` for non-admins—but these are purely cosmetic. The server-side `validatedPageProcedure` in `apps/studio/src/server/modules/page/page.router.ts:74-100` validates the incoming blob against the full unscoped `IsomerSchema` and there is no role check on `updatePageBlob`. A non-admin with collection write permission can still POST `tagCategories` directly via the tRPC client.

Consider gating server-side: reject `tagCategories`/`tags` mutations from non-admins, or only allow these properties to change when the actor has the admin role. (This was also flagged by Cursor's automated review.)

Confidence: high

---

**issue (UX/data integrity):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx:70-72`

`DeleteOptionModal` body contains `"This option is being used in XX items"` with a `// TODO: replace XX with usage count from backend` comment left in user-visible copy. Either wire this up before merging or replace the copy with something accurate (e.g., "This option may be in use by collection items"). Shipping literal "XX" to admins erodes trust in the warning.

Confidence: high

---

**suggestion:** `apps/studio/src/utils/ajvKeywords/uniqueItemPropertiesIgnoreCase.ts:4-7` and `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/utils/indicesWithDuplicateLabels.ts:3-5`

`normalizedKey` in the AJV keyword and `normalizedLabelKey` in `indicesWithDuplicateLabels` are the same logic copy-pasted. Extract to a shared util so the "definition of duplicate" can't drift between client UI and AJV validation.

Confidence: medium

---

**suggestion:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/utils/hasUniqueItemPropertiesError.ts:24-27`

The function currently only checks for `uniqueItemPropertiesIgnoreCase`. The name `hasUniqueItemPropertiesError` (without "IgnoreCase") suggests it might also check for the upstream `ajv-keywords` `uniqueItemProperties` keyword in the future, but that keyword isn't registered. Either rename to `hasUniqueItemPropertiesIgnoreCaseError` to match what it does, or add a comment clarifying the scope.

Confidence: medium

---

**question:** When two filter labels are duplicates (case-insensitive) and the user deletes one row via the menu, the other duplicate becomes unique and the array-level `belowDescription` error banner should clear. Does JsonForms' `onChange` re-run AJV on the `removeItems` mutation immediately, or does the banner linger until the next keystroke? Worth a manual verification.

Confidence: medium

---

**note:** Storybook stories in `EditCollectionIndexPage.stories.tsx` were partially migrated to a new `EditCollectionIndexPageNewExperience.stories.tsx` file. This loses git blame on the play functions but isn't blocking.

Confidence: low

---

Overall: Good feature implementation with solid test coverage. The main actionable items are (1) the server-side auth gap for `tagCategories` mutations and (2) the placeholder "XX items" text in the delete modal.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/DraggableTagButton.tsx
```diff
@@ -24,6 +27,15 @@ interface DraggableDrawerButtonProps extends OwnPropsOfMasterListItem {
...
   isError: boolean
+  listItemIcon?: IconType
+  /** Merged into padding on the drag handle and label button (e.g. tighter `py`). */
+  listItemContentProps?: BoxProps
+  /** Rendered after the label area, typically flush right (e.g. row actions menu). */
+  listItemTrailing?: ReactNode
+  /** Caption under the row title (stacked with 0.25rem gap). */
+  listItemSubtitle?: ReactNode
+  /** When `isError` is true, replaces the default “Fix issues before saving” line. */
+  listItemErrorCaption?: string
```
seaerchin, line 28: comment: these are only used in one place and specific enough that i think the correct pattern would be to create a new component called `DraggableTagButton` or something then we pass it in at top level to `JSONFormsArrayControlView` and the default `JSONFormsArrayControl` just uses the default `DraggableDrawerButton`

adriangohjw:
> **issue (security/auth):** `apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx:54-67`
> 
> The Isomer-admin gate on filters is only enforced client-side. `CollectionEditorStateDrawer` excludes `tagCategories`/`tags` from the scoped schema for non-admins, and `JsonFormsTagCategoryOptionsControl` returns `null` for non-admins—but these are purely cosmetic. The server-side `validatedPageProcedure` in `apps/studio/src/server/modules/page/page.router.ts:74-100` validates the incoming blob against the full unscoped `IsomerSchema` and there is no role check on `updatePageBlob`. A non-admin with collection write permission can still POST `tagCategories` directly via the tRPC client.
> 
> Consider gating server-side: reject `tagCategories`/`tags` mutations from non-admins, or only allow these properties to change when the actor has the admin role. (This was also flagged by Cursor's automated review.)
> 
> Confidence: high

not an issue - currently gated for rollout purposes, not security purposes

> **issue (UX/data integrity):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx:70-72`
> 
> `DeleteOptionModal` body contains `"This option is being used in XX items"` with a `// TODO: replace XX with usage count from backend` comment left in user-visible copy. Either wire this up before merging or replace the copy with something accurate (e.g., "This option may be in use by collection items"). Shipping literal "XX" to admins erodes trust in the warning.
> 
> Confidence: high

in subsequent PR

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -140,7 +169,7 @@ function ComplexEditorNestedDrawer({
   )
 }
 
-function JsonFormsArrayControl({
+export function JsonFormsArrayControlView({
```
seaerchin, line 154: comment: this is used only for tags and tag categories and 5 of hte new props are passed to draggable drawer button

this suggests that we should split out draggable drawer button and pass it as a prop rather than doing prop drilling.

packages/components/src/types/page.ts
```diff
@@ -9,40 +9,95 @@ import {
...
+        description:
+          "Every item must have at least one option selected from this filter.",
+      }),
+    ),
+  }),
+  Type.Object({
+    options: Type.Array(
+      Type.Object({
+        label: Type.String({
+          maxLength: 70,
```
seaerchin: question: i thought we were removing the maxLength?

seaerchin (commented):
looks ok logic wise, but i think some concerns regarding prop drilling and expanding the interface on JsonFormsArrayControl. 

wondering if we should instead create a new component and reuse that for this feature so that we keep changes isolated unless the changes are used elsewhere

packages/components/src/types/page.ts
adriangohjw: updated! https://github.com/opengovsg/isomer/pull/2003/commits/879c6c8825efa4f27a21cad0502dff94664ea20c

i didnt update for the existing one since they gonna be removed after this feature is live anyway

packages/components/src/types/page.ts
```diff
@@ -9,40 +9,93 @@ import {
...
+            pattern: "cannot be empty or have leading/trailing spaces",
+          },
+        }),
+        id: TagOptionUuidSchema,
+      }),
+      {
+        title: "Options",
+        description:
+          "Collection filter will display options in this order. Only options that are in use will appear on the Preview.",
+        addItemLabel: "Add option",
```
dcshzj: issue: hmm I think there would be some contractual violations if we add these additional properties into the schema. The principle is that the schema generated by TypeBox should adhere to JsonSchema 7, which downstream consumers (ajv and JSONForms) would expect to follow.

We should continue to use the `format` to influence how Studio renders the form input but it is not for the schema to decide since it is not part of the validation.

packages/components/src/types/page.ts
```diff
@@ -9,40 +9,93 @@ import {
...
+        format: "tag-category-options",
+        /**
+         * Studio AJV: duplicate option names (case-insensitive, trim) fail validation in JsonForms.
+         * @see {@link ../../../../apps/studio/src/utils/ajv.ts}
+         */
+        uniqueItemPropertiesIgnoreCase: ["label"],
```
dcshzj: issue: same case for this also, it isn't a components requirement that the labels should not be duplicate using case insensitive matching, but rather a construct that Studio is providing

adriangohjw: > it isn't a components requirement that the labels should not be duplicate using case insensitive matching

hmm IMO it makes sense for the schema to enforce this since duplicates aren't allowed. This is not so much of a studio thing but whats "legal" as a schema

packages/components/src/types/page.ts
adriangohjw: okay to shift this to studio.

initial thought is that it will be too many customization on studio end if we put this into studio, and since "format" is already a custom prop, i think the line to decide when is a custom format acceptable and when is it not, becomes a bit blurred

dcshzj: The `format` prop is part of the specifications: https://json-schema.org/draft-07/schema hence we default to that to influence Studio's behaviour.

The idea is that we should conform to the specifications when generating it on the components side via TypeBox, so that if it were to be used by other people, any validator would be able to produce a JSON that can be properly handled by our components package. Studio via ajv and JSONForms is one such consumer, and `format` is only read by JSONForms to influence the display on the frontend whereas ajv ignores it. As such, we cannot rely on `format` being a determiner of how the exact string should be, that being handled by `pattern` which we have currently.

On Studio side if there are common patterns like the draggable block, then yeah that can be a component that is relied upon by the various formats.

packages/components/src/types/page.ts
dcshzj: In that case can we explore runtime validations instead? Studio will validate that duplicates are not allowed, but components assume the worst and try to recover from such potential duplicates, say like using the first label it encounters and merging the rest to it.

My concern is still that this is non-standard so we still violate the principle that our schemas are JsonSchema 7 compliant. Effectively since it is non-standard we can't say for sure that components receive the inputs in a valid non-duplicated format.

adriangohjw: understand the rationale behind jsonschema compliant though TBH not fully convinced if there's really a need for it to be "used by other people" and our primary use case is ajv

runtime validation works, tho IMO it still feel like a weaker validation as compared to using schema. that said, more than happy to move it to studio - in fact, i actually spent quite some time figuring out how to make it part of schema instead of relying on studio validation because of said reason

adriangohjw: updated https://github.com/opengovsg/isomer/pull/2003/commits/0731fecf1bfe36b4c19531fcbd576412d80b21ab

packages/components/src/types/page.ts
adriangohjw: updated https://github.com/opengovsg/isomer/pull/2003/commits/0731fecf1bfe36b4c19531fcbd576412d80b21ab

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/DraggableTagButton.tsx
seaerchin, line 1: question: saw your previous comment saying this is under-used. agree w/ it in principle, but i think tacking it on is probably worse than just creating a new component straight up. 

my thinking is that it's probably cheap to create a scoped component for each part here then we compose them together:

```ts
<DraggableTagWrapper>
    <DraggableTagListItem> blah blah </DraggableTagListItem>
</DraggableTagWrapper>
```

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -32,15 +33,25 @@ import {
...
 )
 
+export type JsonFormsArrayControlProps = ArrayLayoutProps & {
+  /** When the array is empty, replaces the default placeholder inside the empty-state container. */
+  emptyState?: ReactNode
+  /** Custom row renderer. Receives per-row draggable wiring and metadata; must render and forward the ref. */
+  renderListItem?: (props: DraggableArrayItemRenderProps) => ReactNode
+  /** Override for the "Add item" button label; takes precedence over any schema-derived value. */
+  addItemLabel?: string
+}
```
seaerchin, line 53: comment: feel abit iffy because our JSONForms components are not supposed to accept UI arguments. We render a fixed component - the view is dictated by Studio, not by the component side. I feel like the correct abstraction here should probably be another component specially for the tag categories/tags rather than this.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -239,95 +259,116 @@ function JsonFormsArrayControl({
...
+                          uischema: childUiSchema,
+                          childLabelProp: undefined,
+                          translations: {},
+                          setSelectedIndex,
+                        }
+                        return renderListItem ? (
+                          renderListItem(rowProps)
+                        ) : (
+                          <DraggableTagButton {...rowProps} />
+                        )
```
seaerchin, line 354: comment: i think the intent here is to conditionally swap the behaviour - but this is entirely dictated by the caller, not by this component. 

i think the correct approach is to have a view wrapper and the caller then just passes in the children. we can do this by letting the View component specify a `children`

seaerchin (commented):
from my understanding, the PR works as follows:

we use a state variable at `RootStateDrawer` level to dictate the top level view; for individual tags/tag categories, we use the draggable to allow drag/dropping. from glancing through, these are constructs that are already present in our codebase so i assumed they were correct by default.

on deletion, we set a state variable to the id of the item to be deleted, then trigger the deletion modal. on confirmation, we then set the item to `null`, which seems fine imo.

my hesistation is with the UI patterns in this PR that i feel don't fit w/ our codebase's philosophy. this is important to me because it impacts future work downstream, especially by AI agents where they are likely to pick-up on this pattern and further amplify it. if possible, my ask is that we can split out the view layer from the JSONforms side. ideally, the initial call from Jsonforms to our component **should not** contain any view related logic. this should be wrapped within the component or exposed as child components which are plug and play by our top level components

karrui:
Both tag controls (`JsonFormsTagCategoryControl`, `JsonFormsTagCategoryOptionsControl`) duplicate near-identical row rendering, which is leaking out as repeated/single-caller slots on the generic `DraggableTagButton`. Two cleanups:

**1. Extract the trailing actions menu** — it's structurally identical in both controls; only the noun, aria-label, item label and delete handler differ.

```tsx
function RowActionsMenu({ noun, index, isDisabled, onDelete }: {
  noun: string
  index: number
  isDisabled: boolean
  onDelete: () => void
}) {
  return (
    <Menu isLazy>
      <MenuButton
        as={IconButton}
        icon={<BiDotsHorizontalRounded fontSize="1.5rem" />}
        {...ROW_ACTIONS_MENU_BUTTON_PROPS}
        isDisabled={isDisabled}
        aria-label={`${upperFirst(noun)} ${index + 1} actions`}
        onClick={(e) => e.stopPropagation()}
      />
      <Portal>
        <MenuList>
          <MenuItem
            colorScheme="critical"
            icon={<BiTrash fontSize="1rem" />}
            isDisabled={isDisabled}
            onClick={(e) => {
              e.stopPropagation()
              onDelete()
            }}
          >
            Delete {noun}
          </MenuItem>
        </MenuList>
      </Portal>
    </Menu>
  )
}

// usage in each control's renderListItem
listItemTrailing={
  <RowActionsMenu
    noun="filter"
    index={rowProps.index}
    isDisabled={isRemoveItemDisabled}
    onDelete={() => setDeleteTarget({ index: rowProps.index, label: ... })}
  />
}
```

**2. Consolidate the tag-specific row decoration.** `listItemContentProps={{ py: "0.5rem" }}` is repeated in both controls — it can't move to `DraggableTagButton`'s default since the generic array control depends on the existing `isError`-based padding. Pull it, plus the single-caller `listItemIcon` / `listItemSubtitle`, into a single shared tag-row helper used by both controls. That keeps `DraggableTagButton` as generic chrome and puts the tag-specific bits in one place instead of as scattered slots.

adriangohjw:
Fixed in 694c72b20.

1. Extracted `TagRowActionsMenu` into a new shared component (parameterised by `noun`, `index`, `isDisabled`, `onDelete`) — both controls now use it instead of the duplicated inline `<Menu>` JSX.
2. Added `TagDraggableButton` (named export on `DraggableTagButton.tsx`) that wraps the generic button with `listItemContentProps={{ py: '0.5rem' }}` baked in — both controls switch to it, removing the repeated prop.

seaerchin (approved):

PR #2005, opened 2026-04-20
🔖 Feat/implement collection filters drawer (backend)

adriangohjw:
## Problem

This PR stacks on [`feat/collection-filters-drawer`](https://github.com/opengovsg/isomer/tree/feat/collection-filters-drawer) and adds backend support plus Studio UI so editors can see **how many collection child resources still reference tag options** before confirming deletion of a filter category or option.

Closes https://linear.app/ogp/issue/ISOM-2286/implement-collection-tag-filters-drawer-backend

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Added `collection.countTagOptionsUsage` (tRPC query) with input schema `countTagOptionsUsageSchema` (`siteId`, `pageId` for the collection index page, `tagOptionIds`). It counts collection pages and collection links under the parent collection whose **draft or published** blob has `page.tagged` overlapping any of the given option IDs (SQL via `jsonb_array_elements_text` and `= ANY(...)`).
- Wired **delete confirmation modals** in `JsonFormsTagCategoryControl` and `JsonFormsTagCategoryOptionsControl` to load that count (Suspense + skeleton + error boundary fallback) so copy can warn when a category (all its option IDs) or a single option is still in use.

**Improvements**:

- Refactored tag category/option types in the JsonForms controls for clearer option ID typing and consistency with the schema.
- Extended `collection.router` tests, integration seed helpers, and MSW collection handlers; updated `EditCollectionIndexPageNewExperience` stories for the new delete flows.

**Bug Fixes**:

- _(none)_

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] In Studio, open a site’s **collection index** page that uses the new filters / collection editor experience (same flow as the parent branch).
- [ ] Ensure at least one **collection page or link** under that collection has `tagged` values that include a known tag option id (create or edit content so tags are saved on draft or published content as appropriate).
- [ ] Open the UI to **delete an entire tag category**; confirm the modal shows a **usage count** that matches how many child resources reference any option in that category, and that the count loads without a hard error.
- [ ] Open the UI to **delete a single tag option**; confirm the modal shows the **usage count** for that option only.
- [ ] (Optional) Repeat with a collection that has **no** usages and confirm counts show **0** where expected.

**New scripts**:

- _(none)_

**New dependencies**:

- _(none)_

**New dev dependencies**:

- _(none)_

---

> [!NOTE]
> **Medium Risk**
> Adds a new protected tRPC query backed by custom SQL over JSON blobs; incorrect query logic or invalidation could lead to wrong counts or performance issues on large collections. UI changes are localized but introduce Suspense/error-boundary loading paths in critical editor flows.
> 
> **Overview**
> Adds a new `collection.countTagOptionsUsage` tRPC query (validated by `countTagOptionsUsageSchema`) to count how many **child collection pages/links** reference a set of tag option UUIDs in either their *draft or published* blob.
> 
> Updates the collection filters UI so “Delete filter/option” confirmation modals fetch and display this live usage count (with Suspense + skeleton + error fallback), and adjusts deletion behavior to skip the modal for unsaved rows with no persisted `id`. Related cache invalidations, integration tests/seed helpers, MSW handlers, and Storybook plays are updated to cover the new endpoint and flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bdae1134bea8f99561a4a01f76a5beacb50977b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seaerchin (commented):

## Diff summary

- New zod input schema [`countTagOptionsUsageSchema`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/schemas/collection.ts) — `siteId`, `pageId` (collection index page id), `tagOptionIds: z.string().uuid().array().max(100)`. No `min(1)` on the array (empty arrays are accepted and short-circuit to `{count: 0}` in the resolver).
- New tRPC query [`collection.countTagOptionsUsage`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/server/modules/collection/collection.router.ts) wired as `protectedProcedure`. Flow:
  1. `bulkValidateUserPermissionsForResources({ siteId, action: "read", userId })` — site-wide read gate (no `resourceIds`), same pattern as the sibling `getMetadata` / `readCollectionLink` procedures.
  2. Lookup the index page by `(id=pageId, siteId, type=IndexPage)` to derive `parentId` (the parent collection); 404 if missing.
  3. `getSiteResourceById({ siteId, resourceId: parentId, type: Collection })` to confirm parent is a Collection; 404 otherwise.
  4. De-dupe `tagOptionIds`; early-return `{count:0}` for empty input.
  5. Kysely SQL: `Resource r LEFT JOIN Blob draftBlob ON r.draftBlobId LEFT JOIN Version v ON r.publishedVersionId LEFT JOIN Blob publishedBlob ON v.blobId`, filtered by `r.parentId=parentId AND r.siteId AND r.type IN (CollectionPage, CollectionLink)`, with an `EXISTS … jsonb_array_elements_text(COALESCE(content->'page'->'tagged','[]')) WHERE tag = ANY(ARRAY[…]::text[])` predicate over draft OR published blob, returning `count(*)::int`.
- Studio drawers ([`ComplexEditorStateDrawer`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/features/editing-experience/components/Drawer/ComplexEditorStateDrawer.tsx), [`RootStateDrawer`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx), [`LinkEditorDrawer`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/features/editing-experience/components/Drawer/LinkEditorDrawer.tsx)) invalidate `collection.countTagOptionsUsage` on mutation success when type is `CollectionPage` (drawer mutations) or unconditionally (link mutation).
- Delete-modal UIs ([`JsonFormsTagCategoryControl`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx), [`JsonFormsTagCategoryOptionsControl`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx)) call `useSuspenseQuery` inside `Suspense + ErrorBoundary`; rows without a persisted `id` skip the modal entirely.
- Integration tests ([`collection.router.test.ts`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts)) and seed helpers ([`seed/index.ts`](https://github.com/opengovsg/isomer/blob/feat/collection-filters-drawer-backend/apps/studio/tests/integration/helpers/seed/index.ts)) cover auth (401/403), validation (>100 ids), 404 paths (missing page / non-IndexPage / cross-site), zero/one/two matches, draft-only vs published-only, dedupe across blobs, empty input.

## Review

**praise:** Permission gate, ownership pinning (siteId on both the IndexPage lookup and the child query), parameterised `sql` template, and the `[…new Set(ids)]` + early-return for empty input are all clean. SQL uses Kysely's tagged template — no injection surface. (confidence: high)

**praise:** Test coverage is thorough — 401/403/BAD_REQUEST/NOT_FOUND, cross-site isolation, draft-vs-published-vs-both, dedupe per resource, multi-id query. (confidence: high)

**issue (non-blocking):** the count query joins `Blob` twice and runs two `jsonb_array_elements_text` subqueries per child row. There is no functional/expression index on `content->'page'->'tagged'`, so this scans every child blob's content for the parent. For collections with thousands of pages this can get slow. Consider (a) a GIN index on the tagged path, or (b) rewriting the predicate to `(draftBlob.content->'page'->'tagged') ?| array_arg` (jsonb `?|` operator on a text array), which is index-friendly and avoids the `EXISTS … unnest` pattern. Worth a benchmark on the largest production collection before shipping. (confidence: medium)

**issue (non-blocking):** the second guard calls `getSiteResourceById({ resourceId: parentId, type: Collection })` purely to assert "yes, it's a Collection", but `parentId` was already returned from a row where `id=pageId AND siteId AND type=IndexPage`. By construction the parent is the collection (the `IndexPage` row exists only as a child of a Collection in this codebase). This is a redundant DB round-trip per call; either drop it or hoist into a single join. (confidence: medium)

**issue:** `tagOptionIds` is validated as `z.string().uuid()` but the SQL casts each id with `::text` and compares against `jsonb_array_elements_text(...)` (also text). This works today, but the schema and SQL representation disagree: if a future migration ever stores tagged ids as UUID-typed jsonb (or array), the comparison silently breaks. Consider `::uuid` casts in SQL (and a UUID array) for type symmetry with the validator, or document why text comparison is intentional. (confidence: low)

**issue (non-blocking):** empty-array input is accepted and yields `{count:0}` from a server round-trip. Cheap, but the suspense query in `JsonFormsTagCategoryControl` will fire a network call when a user opens "Delete filter" on a category with zero persisted options. Either `.min(1)` on the schema and gate the modal client-side, or short-circuit on the client when `tagOptionIds.length === 0`. (confidence: medium)

**issue (non-blocking):** `closeDeleteModal` invalidates `collection.countTagOptionsUsage` even when the user just *cancels* the modal without doing any mutation. Stale-cache risk is essentially nil here, so this just causes an extra network round-trip when the next modal opens. Move the invalidation into `handleConfirmDelete` only (after the actual remove). Same comment for `JsonFormsTagCategoryOptionsControl.tsx`. (confidence: medium)

**issue (non-blocking):** `handleDeleteFilterMenuItemClick` reads `cat?.options` and maps `o.id` without filtering by uuid shape. Because `compact()` only strips empty strings, a malformed/non-uuid id from local state will hit the server and produce a `BAD_REQUEST` (caught by `ErrorBoundary`, rendered as `—`). Functional, but the user sees no warning. Worth filtering to `isUuid(id)` client-side or surfacing a more useful fallback. (confidence: low)

**question:** Filter management is described in the PR body as editor-facing (modals for category/option deletion in the index editor). The auth gate is `bulkValidate(action: "read")` — i.e. any user with site read permission can call it. That is *probably* fine because counts are not sensitive, but the calling UI is gated by `useIsUserIsomerAdmin`. Is "read on site" the intended server-side gate, or should this match the UI gate (Isomer admin)? Consistency between client gating and tRPC gating matters. (confidence: medium)

**nitpick:** `countTagOptionsUsageSchema` — trailing `// This is an arbitrary limit to prevent abuse, might have to adjust based on` comment is truncated. Either complete or delete. (confidence: high)

**nitpick:** `optionIdsAsSqlArray = sql.join(ids.map(id => sql\`${id}::text\`), sql\`, \`)` followed by `sql\`ARRAY[${...}]::text[]\`` is awkward. Kysely's `sql` template accepts arrays directly: `sql\`${uniqueTagOptionIds}::text[]\`` (or use the `Kysely.parameters` helper) generates the same parameterised array literal. (confidence: low)

**nitpick:** No audit-log entry. Reasonable here — this is a read-only count query, and other read procedures in `collection.router.ts` (e.g. `getMetadata`, `readCollectionLink`) also don't audit. Confirming this is the established pattern. (confidence: high)

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -311,6 +313,94 @@ export const collectionRouter = router({
...
+        userId: ctx.user.id,
+      })
+
+      const indexPage = await db
+        .selectFrom("Resource")
+        .where("id", "=", String(pageId))
+        .where("siteId", "=", siteId)
+        .where("type", "=", ResourceType.IndexPage)
+        .select(["parentId"])
+        .executeTakeFirst()
```
seaerchin: nit: should we do a `takeFirstOrThrow()` that distinguishes between no indexPage and no parentId? tbh very minor so idc if we do or not

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
```diff
@@ -123,12 +183,50 @@ function JsonFormsTagCategoriesArrayLayoutInner(props: ArrayLayoutProps) {
...
+
+    const tagId = cat?.id?.trim()
+
+    // No id means the filter is new and never saved — nothing references it in the DB,
+    // so we remove the row immediately instead of opening the usage warning modal.
+    if (!tagId) {
+      if (!removeItems || isRemoveItemDisabled) return
+      removeItems(path, [index])()
+      return
+    }
```
seaerchin: comment: i don't think this path is required. we generate the uuid on the fly on the frontend. you can confirm this by checking the schema; we do this via jsonformsuuidcontrol so this code chunk can be removed

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
```diff
@@ -123,12 +183,50 @@ function JsonFormsTagCategoriesArrayLayoutInner(props: ArrayLayoutProps) {
...
+      index,
+      label: cat?.label?.trim() ?? "",
+      tagOptionIds: compact(map(cat?.options, (o) => o.id?.trim())),
+    })
+  }
+
+  const closeDeleteModal = () => {
+    void utils.collection.countTagOptionsUsage.invalidate()
+    setDeleteTarget(null)
+  }
```
seaerchin: issue: we should probably shift hte `countTagOptionsUsage.invalidate` call to `handleConfirmDelete` - this is because it renders on modal open, which makes the loading very obvious. we should keep this to a minimum and avoid if not required when the user just exits the modal normally without confirming

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx
```diff
@@ -26,29 +28,64 @@ import {
...
+  tagOptionId: CollectionTagOptionId
+}) => {
+  const [{ count }] = trpc.collection.countTagOptionsUsage.useSuspenseQuery({
+    siteId,
+    pageId,
+    tagOptionIds: [tagOptionId],
+  })
+
+  return <>{count ?? "—"}</>
+}
```
seaerchin: comment: i think this might be duplicated - should we extract to a shared component?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
```diff
@@ -66,11 +102,35 @@ function DeleteFilterModal({
...
+                      }
+                    >
+                      <TagCategoryUsageCount
+                        siteId={siteId}
+                        pageId={pageId}
+                        tagOptionIds={tagOptionIds}
+                      />
+                    </Suspense>
+                  </ErrorBoundary>{" "}
+                  items.
```
seaerchin: nit: can we use item(s) instead? cos there might only be one item using

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
```diff
@@ -25,30 +26,65 @@ import {
...
+  const [{ count }] = trpc.collection.countTagOptionsUsage.useSuspenseQuery({
+    siteId,
+    pageId,
+    tagOptionIds,
+  })
+
+  return <>{count ?? "—"}</>
+}
+
 function DeleteFilterModal({
```
seaerchin: comment: not super opioniated here re: modal reuse but it appears largely similar between teh option and the category. can share if easy but ok leaving as is since there are some diffs

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -311,6 +313,94 @@ export const collectionRouter = router({
...
+        .where("r.siteId", "=", siteId)
+        .where("r.type", "in", [
+          ResourceType.CollectionPage,
+          ResourceType.CollectionLink,
+        ])
+        // Match child resources whose page.tagged JSON array overlaps the queried
+        // option ids. Postgres has no jsonb && jsonb overlap; unnest to text and use ANY.
+        // Draft or published blob alone is enough; one row per resource still counts once.
+        .where(
+          sql<boolean>`(
```
seaerchin, line 390: question: did we try this on prod? off the top of my head, i know nna eventse/ite coursefinder got quite alot of tag options. 

this looks efficient due to using exists + any to short circuit, but just wanted to be safe since collections are meant to hold large amount of items

seaerchin (approved):
looks mostly ok; some code duplication but seems fine. dropping as a comment first to double check worst case on prod but no issues w/ approving in principle

apps/studio/src/server/modules/collection/collection.router.ts
seaerchin: ran a query against prod, details below: 

collectionId: 204932
number of children: 5528
largest tag category size: 2052

afterwards, the largest tag category size quickly tapers off (see attached csv); i think ~5k children/~100 items should be enough for most cases

[collection tag details.csv](https://github.com/user-attachments/files/27486666/collection.tag.details.csv)

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
adriangohjw: yeah i thoguth about it but its only used in 2 places so i think good to keep separate for now

apps/studio/src/server/modules/collection/collection.router.ts
adriangohjw: thanks for benchmarking this!

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
adriangohjw: updated https://github.com/opengovsg/isomer/pull/2005/changes/8f6773178d11574e28b676a8bb61410c4721eaa8

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx
adriangohjw: i think okay for now, same as Modal - while many similarities, i think no harm keeping separate since its only duplicated in 2 places. If its used in more, then i think worth refactorign

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/2005/changes/8f6773178d11574e28b676a8bb61410c4721eaa8

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx
adriangohjw: oh this is to improve the UX, because i want to skip opening the confirmation modal if the user click on "Create new option" but did not click into the option itself

adriangohjw (commented):
## Summary

This PR adds a `collection.countTagOptionsUsage` tRPC query and wires it into the "Delete filter / Delete option" confirmation modals so editors see how many child pages/links still reference a tag option before deleting it. Backend lives in `apps/studio/src/server/modules/collection/collection.router.ts:317-411` (input schema at `apps/studio/src/schemas/collection.ts:117`), UI in `JsonFormsTagCategoryControl.tsx:56-166` and `JsonFormsTagCategoryOptionsControl.tsx:57-162`. Save mutations in the drawer files invalidate the new query so the modal count refreshes after a save. Tests cover authn/authz, validation limits, lookup failures, dedup across draft+published, and multi-option matching.

Overall this is solid: permission check matches the existing site-wide convention, the SQL is correctly parameterized, child rows are scoped by `siteId` + `parentId` + type to prevent IDOR via `pageId`, and the count(*) over a Resource-rooted query with EXISTS subqueries avoids double-counting when both draft and published blobs match. Findings below are mostly maintainability and edge-case correctness.

## Findings

**issue:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx:223-228` and `JsonFormsTagCategoryOptionsControl.tsx:219-224` — `handleConfirmDelete` calls `void utils.collection.countTagOptionsUsage.invalidate()` *after* `removeItems(...)`. `removeItems` only mutates local JsonForms form state — no DB write happens until the user saves the index page — so this invalidation flushes the cache for a state that hasn't changed yet, then any other open modal will re-fetch the same value it already had. The save-side invalidation (in `ComplexEditorStateDrawer`, `RootStateDrawer`, `LinkEditorDrawer`) is what's actually load-bearing. Remove the in-modal invalidation, or move it to the index-page save success handler if you want freshness there too.

Confidence: high

**issue:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx:196-217` — the modal's `tagOptionIds` comes from the *current local form state* (`compact(map(cat?.options, (o) => o.id?.trim()))`), not from what's actually persisted in the DB. So if an editor opens a persisted filter, deletes some or all of its options in the form (but hasn't saved), then clicks "Delete filter", the count shown reflects only the option ids still in local state — not the ones the DB still references. In the extreme case where all original options were removed locally, `tagOptionIds = []`, the backend short-circuits to `{count: 0}`, and the modal cheerfully says "0 items" even though the DB still has usages.

Two options: (a) accept this and document that the count reflects "what you're about to delete from this session," or (b) re-fetch the persisted category from the server (or pass the category id and resolve options server-side) before counting. Worth at least a comment so the next reader doesn't think the count is authoritative against the DB.

Confidence: medium

**suggestion (performance):** `apps/studio/src/server/modules/collection/collection.router.ts:389-406` — each row evaluates two `EXISTS` + `jsonb_array_elements_text` subqueries with no supporting index. For typical collections (tens to low-hundreds of pages) this is fine, but a large collection with thousands of child resources will scan and unnest every blob on every modal open. Two cheap mitigations if you observe slowness: (1) `WHERE "draftBlob"."content"->'page'->'tagged' IS NOT NULL OR "publishedBlob"."content"->'page'->'tagged' IS NOT NULL` pre-filter to skip empty cases; (2) a functional GIN index on `("content"->'page'->'tagged') jsonb_path_ops` and switch the predicate to `?| array[...]` (jsonb has-any-key) which is index-friendly. Not blocking — the modal is on-demand, not hot-path — but worth a follow-up if collections grow.

Confidence: medium

**question:** `apps/studio/src/server/modules/collection/collection.router.ts:389-406` — `jsonb_array_elements_text` errors if `page.tagged` exists but isn't a JSON array (e.g., a string or object). `COALESCE(..., '[]'::jsonb)` only handles the NULL/missing case. Is there any data path (legacy blobs, partial migrations, ad-hoc imports) where `tagged` could be set to a non-array? If yes, one bad blob anywhere in the collection makes the whole count fail, and the modal silently shows `—` via the error boundary with no log trail to find the offending resource. Consider wrapping in `CASE WHEN jsonb_typeof(...) = 'array' THEN ... ELSE '[]'::jsonb END` for defensive safety.

Confidence: low

**suggestion (maintainability):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx:56-72` and `JsonFormsTagCategoryOptionsControl.tsx:57-73` — `TagCategoryUsageCount` and `TagOptionUsageCount` are nearly identical (one takes a list, one wraps a single id into a list), and the surrounding `ErrorBoundary` + `Suspense` + `Skeleton` wrapper is duplicated verbatim across the two modals. Worth extracting a single `<TagOptionsUsageCount siteId pageId tagOptionIds />` plus a `<TagOptionsUsageCountWithFallback ... />` wrapper to deduplicate the ~30 lines and keep the fallback dimensions consistent.

Confidence: medium

**nitpick:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx:71` and `JsonFormsTagCategoryOptionsControl.tsx:72` — `count ?? "—"` is dead code: the tRPC return type is `{ count: number }`, so `count` is never nullish here. Either drop the fallback or change the return type to allow `null`/`undefined` if you actually want a server-side "unknown" state.

Confidence: high

**nitpick:** `apps/studio/src/schemas/collection.ts:120-127` — the comment block has both a JSDoc `/** ... */` and a `//` comment for the same `max(100)` line. Consolidate into one.

Confidence: high

**praise:** `apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts:1586-1941` — test coverage is thorough: distinct cases for unauthed, unauthorized site, index page belonging to another site, index page with no parent collection, draft-only / published-only / both-blobs (with the "count once" assertion at L794), and multi-option overlap. The named `setupCollectionWithIndexPage` helper and `collectionPageBlobContent` factory keep the test bodies readable.

Confidence: high

**note:** `apps/studio/src/server/modules/collection/collection.router.ts:320-324` — permission check uses `bulkValidateUserPermissionsForResources` without `resourceIds`, which (per the comment in `permissions.service.ts:88-90`) collapses to site-wide read. That's the existing convention across this router (e.g. `getMetadata`, `list`, `getCollectionTags`), so this is consistent — flagging only so reviewers know the new endpoint inherits the site-wide TODO and isn't introducing a new gap.

Confidence: high

adriangohjw:
@claude remove the 70 character limit in TagCategorySchema

adriangohjw:
@claude yes, remove maxLength: 70 from  option's label too

PR #2008, opened 2026-04-21
🔖 Feat/implement collection filters drawer (toast)

adriangohjw:
## Problem

After saving from the collection editor drawer, the success toast always used the same generic copy. Editors saving **Manage filters** versus **Collection display** could not tell from the toast which kind of change had been persisted, or what “publish” meant in context.

Closes https://linear.app/ogp/issue/ISOM-2287/implement-collection-tag-filters-drawer-toast

## Solution

This branch stacks on **`feat/collection-filters-drawer`** and only adjusts save feedback plus Storybook coverage for it.

**Breaking Changes**

  - Details ...

**Features**:

- **`CollectionEditorStateDrawer`**: On successful save, the toast title depends on drawer mode—**filter** uses copy about publishing so others can use the new filter options; **display** (and non–collection-editor fallback) uses copy about publishing so others can see collection display updates. Replaces the shared `CHANGES_SAVED_PLEASE_PUBLISH_MESSAGE` string for this path.

**Improvements**:

- **`EditCollectionIndexPageNewExperience.stories`**: Adds `CollectionDisplaySaveToast` and `ManageFiltersSaveToast` stories with `play` functions that save from the drawer and assert the matching toast text (using `waitFor` for async toast rendering).

**Bug Fixes**:

- None in this PR.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Run Storybook for Studio and open **Page → EditCollectionIndexPageNewExperience** (or the story group that contains these stories). Run **`CollectionDisplaySaveToast`** and confirm the toast shows: *Collection display saved. Remember to publish…*
- [ ] Run **`ManageFiltersSaveToast`** and confirm the toast shows: *Filter saved. Remember to publish…*
- [ ] (Optional) In Studio with the new collection editing experience, save from **Collection display** and from **Manage filters** and confirm the same two messages appear.


seaerchin (commented):

## Diff summary

- [`CollectionEditorStateDrawer.tsx`](https://github.com/opengovsg/isomer/blob/feat/collection-editor-toast/apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx)
  - Drops the shared `CHANGES_SAVED_PLEASE_PUBLISH_MESSAGE` import for this drawer.
  - Hoists the existing `drawerStateType` `useMemo` (`drawerState.state !== "collectionEditor"` → `"display"`, otherwise `drawerState.type`) above the `updatePageBlob` mutation so it can be referenced from `onSuccess`.
  - In `onSuccess`, after the three `invalidate()` calls, the `toast({ status: "success", ... })` `title` is now branched: `"filter"` → "Filter saved. Remember to publish the changes so that other users can use the new filter options.", else → "Collection display saved. Remember to publish the changes so that other users can see your updates."
- [`EditCollectionIndexPageNewExperience.stories.tsx`](https://github.com/opengovsg/isomer/blob/feat/collection-editor-toast/apps/studio/src/stories/Page/EditPage/EditCollectionIndexPageNewExperience.stories.tsx)
  - Adds `waitFor` to the `storybook/test` import.
  - Adds `CollectionDisplaySaveToast` (clicks "Collection display" → "Save changes", asserts display copy via `withinPortals` + 5s `waitFor`) and `ManageFiltersSaveToast` (uses `playOpenManageFilters` + admin MSW handler, asserts filter copy). Both rely on the existing local `withinPortals` helper.

## Review

**nitpick (non-blocking):** The toast title is computed at `onSuccess` time but reads `drawerStateType` from the closure created at mutation kickoff. Since `handleSaveChanges` does not close the drawer until `onSuccess` resolves, this is fine today — but it is a footgun if anyone later changes the order (e.g. closes the drawer optimistically, or reuses this mutation across drawer modes). Consider passing the mode as `mutate` variable / via `meta` so the toast is sourced from the variables that were dispatched. (confidence: medium)

**suggestion (non-blocking):** Hard-coding the two long sentences inline regresses the discoverability that `CHANGES_SAVED_PLEASE_PUBLISH_MESSAGE` gave. Consider extracting `COLLECTION_DISPLAY_SAVED_MESSAGE` / `FILTER_SAVED_MESSAGE` (or a `getCollectionEditorSaveMessage(type)`) into a constants file and importing in both the drawer and the storybook assertions so the regex in `play` and the actual rendered string cannot drift. Today the regex literally re-spells the sentence. (confidence: high)

**issue (non-blocking):** Fallback copy semantics. The `else` branch fires "Collection display saved…" whenever `drawerState.state !== "collectionEditor"` OR `drawerState.type !== "filter"`. If this drawer ever mounts in a non-`collectionEditor` state (race on close, or a future `drawerState.type` other than `display`/`filter`), users will see "Collection display saved…" for something that may not be a display change. Consider stricter branching: `drawerStateType === "filter" ? filterCopy : drawerStateType === "display" ? displayCopy : CHANGES_SAVED_PLEASE_PUBLISH_MESSAGE` to keep the generic fallback truthful. (confidence: medium)

**question:** Toast deduplication / double-fire. `BRIEF_TOAST_SETTINGS` likely doesn't carry an `id`, so a fast double-save (click "Save changes" twice while the first request is in flight) would queue two stacked toasts. `useToast` from `@opengovsg/design-system-react` supports `id` for de-dupe — was this considered, or is the save button already disabled via `isPending` for the entire window? (Looks like it is via `isLoading={isPending}` on `SaveButton`, so likely fine — flagging for confirmation.) (confidence: low)

**nitpick (non-blocking):** Accessibility / `aria-live`. The new copy is significantly longer ("Remember to publish the changes so that other users can use the new filter options.") and the brief toast settings will likely auto-dismiss before a screen reader finishes announcing. Consider verifying the duration in `BRIEF_TOAST_SETTINGS` accommodates the longer string, or shortening copy to e.g. "Filter saved. Publish to apply." (confidence: low)

**nitpick (non-blocking):** Story hard-coded `timeout: 5000`. Storybook's default `waitFor` timeout (1s) is usually enough for an MSW success path; 5s masks regressions where save becomes slow. Prefer asserting on the toast appearance with the default timeout and only bump if flakes appear in CI. (confidence: low)

**praise:** Good call hoisting `drawerStateType` above the mutation rather than re-deriving inside `onSuccess`, and adding play-function coverage for both branches via `withinPortals` (Chakra portals are a known footgun in Storybook tests). (confidence: high)

**nitpick (non-blocking):** Test coverage gap. The two new stories assert the happy path but no negative assertion is made (e.g. that "Collection display saved" does NOT appear in `ManageFiltersSaveToast`, and vice versa). A quick `expect(queryByText(/Collection display saved/)).toBeNull()` in the filter story would catch a future bug where both toasts fire or the wrong branch is taken. (confidence: medium)

seaerchin (approved):

adriangohjw:
@claude pull from main and deconflct

PR #2009, opened 2026-04-21
fix(studio): add default value to collection variant radio buttons

seaerchin:
## Problem

When opening the collection drawer, the radio buttons for selecting the collection layout variant (1-column vs 2-column) did not have a default selection. This could lead to a confusing user experience where no option appears selected initially.

## Solution

Added a `defaultValue` prop to the `Radio.RadioGroup` component in `JsonFormsCollectionVariantControl`, setting it to `COLLECTION_VARIANT_OPTIONS.Collection` (1-column layout). This ensures users see the 1-column option selected by default when the drawer opens without a pre-existing value.

**Breaking Changes**

**Bug Fixes**:

- Fixed radio button group not showing a default selection when no value is present

## Before & After Screenshots

**BEFORE**:
No radio button selected by default when opening collection drawer

**AFTER**:
1-column layout radio button is selected by default

## Tests

- Open the collection drawer for a collection page
- Verify that the 1-column layout option is selected by default
- Verify that selecting a different option still works correctly
- Verify that existing collections with a saved variant still display the correct selection

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionVariantControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCollectionVariantControl.tsx
@@ -39,6 +39,7 @@ function JsonFormsCollectionVariantControl({
             handleChange(path, value)
           }}
           value={data as string}
+          defaultValue={COLLECTION_VARIANT_OPTIONS.Collection}
         >
           <Radio
             value={COLLECTION_VARIANT_OPTIONS.Collection}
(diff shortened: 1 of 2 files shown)


adriangohjw (approved):
question: its defined in the schema, but i asssume this is for backward compactibility right for existing blobs?

seaerchin:
> question: its defined in the schema, but i asssume this is for backward compactibility right for existing blobs?

not exactly for existing blobs. right now when we create a new collection index page, this isn't set so the radio button won't be set. we could either do it on the backend when we create or on here (just unifying the view). 

i've since added to the backend too when we initially create the index page

PR #2010, opened 2026-04-21
🔖 Feat: category management (UI)

adriangohjw:
## Problem

Editors need to manage **category options** for collection index pages (used by collection items) in the new collection index editing experience, with Studio validation, drawer scoping, and accessibility consistent with the rest of the form builder. Collection items should be able to select a **category id** aligned with those options when the feature is rolled out.

Closes https://linear.app/ogp/issue/ISOM-2288/category-management-ui

## Solution

**Base branch:** `feat/collection-filters-drawer`. The GitHub diff for this PR is the commits on `feat/collection-category-management` **on top of that base** (stacked PR). The filters-drawer work lives in the base branch / [#2003](https://github.com/opengovsg/isomer/pull/2003); this PR adds **category options** and **category id** selection on top.

**Breaking Changes**

  - Details ...

**Features**:

- **Page model**: `CategoriesSchema` with optional `categoryOptions` on the collection page (backward compatible; documented path to required + migration). Optional `categoryId` on collection items for alignment with category options.
- **JsonForms**: `JsonFormsCategoryOptionsControl` for managing category options (add/remove, duplicate-label validation, delete confirmation, drawer integration). `JsonFormsCategoryIdControl` for choosing `categoryId`, gated by GrowthBook `category-id-dropdown`.
- **Studio wiring**: Form builder renderers, `CollectionEditorStateDrawer` scoping for `categoryOptions`, Storybook stories for category options flows.
- **Accessibility / UX**: `aria-label` improvements on drawer-related UI; save affordances where option counts are shown.

**Bug Fixes**:

- `CollectionEditorStateDrawer` include/exclude logic accounts for `categoryOptions` so user admin drawer state stays consistent.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] In Studio, open a site with the **new collection index** editing experience and open a **collection index** page.
- [ ] Open the **collection editor drawer**, go to **Manage filters**, and exercise **category options**: add and remove options, try duplicate names (case-insensitive), confirm validation and delete confirmation.
- [ ] With the `category-id-dropdown` feature enabled in GrowthBook, edit a **collection item** and confirm the **category** field uses the new id dropdown and persists after save and reload.
- [ ] Save the page; reload the editor and confirm **category options** and (if used) **category id** persist as expected.
- [ ] (Optional) Run Storybook for `EditCollectionIndexPageNewExperience` and walk through the new **category options** stories.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

---

> [!NOTE]
> **Medium Risk**
> Adds new collection schema fields and JsonForms renderers that affect how collection metadata is validated and edited; issues could block saving or introduce backward-compatibility quirks with existing page blobs. Feature-flagging reduces blast radius, but the new category options UI and validation paths are non-trivial.
> 
> **Overview**
> Adds **category management UI** to the new collection index editing experience by introducing editable `categoryOptions` on collection pages (with duplicate-name validation and delete confirmation) and wiring a new `JsonFormsCategoryOptionsControl` renderer into the form builder.
> 
> Extends the page model to include optional `categoryOptions` (collection page) and optional `categoryId` (collection items), and adds a feature-flagged `JsonFormsCategoryIdControl` (`category-id-dropdown`) for selecting the new id-based category field.
> 
> Updates drawer scoping so `categoryOptions` only appears in the Filters drawer (not Collection display), improves accessibility via new drawer save `aria-label`s, and enhances array controls to allow suppressing empty placeholders (`emptyState={null}`), with Storybook flows added for the new category options editor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd824c8147b9a9be900958935eeb4e865605c1a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,536 @@
...
+        <ModalCloseButton size="lg" />
+
+        <ModalBody>
+          <VStack align="stretch" spacing="1.5rem">
+            <Infobox width="100%" size="md" variant="warning">
+              <Text textStyle="body-2">
+                {/* TODO: replace XX with usage count from backend */}
+                This option is being used in XX items. To undo this change, you
+                will need to create and re-assign this option to all items.
+              </Text>
```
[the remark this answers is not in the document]
adriangohjw: nah, intentional, will have new PR to add it

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,536 @@
...
+              onClick={onConfirm}
+            >
+              Delete option
+            </Button>
+          </HStack>
+        </ModalFooter>
+      </ModalContent>
+    </Modal>
+  )
+}
```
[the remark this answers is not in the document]
adriangohjw: yeah nah, we are gonna use differnt endpoint. its okay to duplicate too, no biggies

seaerchin (commented):

## Diff summary

This PR (stacked on `feat/collection-filters-drawer`) adds a "Category options" editor to the Filters drawer plus a feature-flagged dropdown for collection items.

- [packages/components/src/types/page.ts](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/packages/components/src/types/page.ts): adds `CategoriesSchema` (optional `categoryOptions: { label, id }[]` with `format: "category-options"`, `uniqueItemPropertiesIgnoreCase: ["label"]`, `default: []`) and adds optional `categoryId` (`format: "category-id"`) to `categorySchemaObject`. Composes `CategoriesSchema` into `CollectionPagePageSchema`.
- [apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx): scopes `categoryOptions` into the Filters drawer and out of Display; adds `aria-label` on the save button.
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx) (new, 543 lines): "summary row" UI on the drawer that opens an absolute-positioned editor pane (`CategoryOptionsExpandedEditor`) wrapping `JsonFormsArrayControlView`. Handles duplicate-label highlighting (case-insensitive, via `indicesWithDuplicateLabels`), AJV `uniqueItemProperties` errors, blank-label gating for "Save changes", and a confirm modal `DeleteCategoryOptionModal`.
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx) (new): GrowthBook-gated `SingleSelect` for `categoryId`; currently backed by a hard-coded empty placeholder list with a TODO to wire up data.
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx): `emptyState` now accepts `null` to fully suppress the empty container.
- [apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryControl.tsx): passes `emptyState={null}`.
- [apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/FormBuilder.tsx), [renderers/controls/index.ts](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/index.ts), [constants/formBuilder.ts](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/constants/formBuilder.ts): wire up renderers and rankings.
- [apps/studio/src/lib/growthbook.ts](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/lib/growthbook.ts): adds `CATEGORY_ID_DROPDOWN_FEATURE_KEY = "category-id-dropdown"`.
- [apps/studio/src/stories/Page/EditPage/EditCollectionIndexPageNewExperience.stories.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/stories/Page/EditPage/EditCollectionIndexPageNewExperience.stories.tsx): six new Storybook stories covering open / add / row menu / delete modal / save+count.

## Review

**issue (blocking):** [JsonFormsCategoryIdControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx) ships a control that is wired into the renderer registry but is backed by `PLACEHOLDER_CATEGORY_OPTIONS: string[] = []`. When the GrowthBook flag is on for a site, an editor lands on a `SingleSelect` with zero options and no clear messaging — they can't pick anything, and the field is `isClearable={true}` so it persists empty. Unless this lands behind a flag that is off everywhere except staging, please either (a) keep the renderer unregistered until the data wiring lands, or (b) source items from the parent collection's `categoryOptions` blob (which exists in the same PR) so the dropdown has real values from day one. The TODO acknowledges this; the registration shouldn't precede it. (confidence: high)

**issue (non-blocking):** [JsonFormsCategoryOptionsControl.tsx](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx) duplicates `DeleteCategoryOptionModal`, the duplicate-label banner, the `useMemo` over `core?.data` + `path`, and the row-actions menu from `JsonFormsTagCategoryOptionsControl.tsx` and `JsonFormsTagCategoryControl.tsx`. The two flows are intentionally similar; please extract `DeleteOptionModal`, the duplicate-label banner block, and a `useDuplicateLabelIndices(path)` hook into shared modules now while the two are still in sync. The longer this lives as a copy, the harder the inevitable copy drift becomes. (confidence: high)

**issue (non-blocking):** `hasBlankOptionLabel` only blocks closing the expanded editor; it does not surface in the parent drawer the way the duplicate-label warning does. So a user can blank a label, click "Save category options" (disabled), tap the back arrow (also disabled), get stuck, and have no obvious cue beyond the per-row error. Consider mirroring the parent-row "Duplicate option names must be fixed before saving." treatment for blank labels, or auto-trimming/dropping blank rows on close. (confidence: medium)

**issue (non-blocking):** `DeleteCategoryOptionModal` shows hard-coded copy `"This option is being used in XX items."`. Editors will see the literal string `XX`. The TODO is fine for a stub, but the user-visible copy needs to read sensibly even before the count is wired up (e.g. "Items currently using this option will lose it"). (confidence: high)

**praise:** The schema notes and `CollectionEditorStateDrawer` scope change are well thought out — keeping `categoryOptions` optional with `default: []` and gating Display vs Filters drawer means existing blobs keep loading, and the path to making this required after migration is documented inline. (confidence: high)

**issue (non-blocking):** [page.ts](https://github.com/opengovsg/isomer/blob/feat/collection-category-management/packages/components/src/types/page.ts) — option `id` is generated as `format: "uuid"`, but there's no client-side guard ensuring new rows actually receive a uuid when they are added through `JsonFormsArrayControlView` (`mapNewArrayItem` is not passed in `CategoryOptionsExpandedEditor`). If JsonForms creates rows with `id: undefined` (or empty string), AJV will fail validation and silently block save. Please verify add-row produces a valid uuid (or pass `mapNewArrayItem={(item) => ({ ...item, id: crypto.randomUUID() })}` like `JsonFormsTagCategoryControl` does for tag options). (confidence: high)

**nitpick:** Copy uses a curly apostrophe (`don't`). Existing code in the repo uses straight apostrophes consistently; mixing them shows up in diffs and grep. (confidence: medium)

**nitpick:** `data === 0 ? "No option" : ...` reads awkwardly; "No options" is more natural. (confidence: low)

**question:** The summary row uses `<Box as="button">` to open the editor and lives inside an `aria-invalid` `HStack` styled like an input row. There's no list virtualization concern here (single row), but for assistive tech, the disabled drag handle (`pointerEvents="none"`, `aria-hidden`) plus the visible-but-decorative menu button next to it may be confusing — only the central area is keyboard-focusable. Was a single tab stop intentional, or should the row itself be the button (i.e. drop the visual menu icon since the real menu lives inside the expanded editor)? (confidence: medium)

**issue (non-blocking):** test coverage is Storybook-only. No unit tests for the new control, the duplicate-blocking logic, or the schema (`CategoriesSchema` composition into `CollectionPagePageSchema`). Given this PR introduces validation gates that block saves, at least `indicesWithDuplicateLabels` integration with the new `categoryOptions` shape and the `cannotLeaveExpandedCategoryOptions` predicate would be worth a Vitest. (confidence: medium)

**issue (non-blocking):** the Storybook `playFillThreeCategoryOptionNames` helper relies on `getByPlaceholderText(/Option name/i)` returning a single match in an editor that may have three rows expanded — if that selector is ambiguous it'll throw. Please confirm the play function actually clicks each row before typing (the for loop reads like it does, but a quick run-through would be reassuring). (confidence: low)

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -328,27 +331,28 @@ export function JsonFormsArrayControlView({
...
+                        <Text
+                          textStyle="subhead-1"
+                          textColor="base.content.default"
+                          textAlign="center"
+                        >
+                          Items you add will appear here
+                        </Text>
+                      )}
+                    </Flex>
+                  ))}
```
seaerchin: style: prefer `!emptyState && ...` means we can eliminate the inner conditional too cos `emptyState` will be defined

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -46,8 +46,11 @@ export const jsonFormsArrayControlTester: RankedTester = rankWith(
 
 export type JsonFormsArrayControlProps = ArrayLayoutProps & {
   listItemIcon?: IconType
-  /** When the array is empty, replaces the default placeholder inside the empty-state container. */
-  emptyState?: ReactNode
+  /**
+   * When the array is empty: replaces the default placeholder inside the empty-state container.
+   * Pass `null` to show no placeholder (no padded empty area).
+   */
+  emptyState?: ReactNode | null
```
seaerchin: nit: i think `ReactNode` already allows `null`

Image

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
```diff
@@ -0,0 +1,72 @@
...
+  path,
+  label,
+}: JsonFormsCategoryIdControlProps) {
+  return (
+    <SingleSelect
+      value={data}
+      name={label}
+      items={PLACEHOLDER_CATEGORY_OPTIONS.map((category) => ({
+        label: category,
+        value: category,
```
seaerchin, line 39: question: how will we do the migration from strings -> uuid? assuming that we're going to be running offhours but then we'd also need to do a code change here so that it uses uuid as the value

adriangohjw (commented):
## Review

A separate Claude review is already on this PR (by `@seaerchin`); this is a second pass focused on issues that one didn't surface, plus quick confirmation of overlap.

### Summary

This PR (stacked on `feat/collection-filters-drawer`) adds a default-Category options editor to the Filters drawer and a feature-flagged `categoryId` dropdown for collection items.

- `packages/components/src/types/page.ts:104-140` — adds `CategoriesSchema` (`categoryOptions: { label, id }[]`, `format: "category-options"`, `default: []`, `uniqueItemPropertiesIgnoreCase: ["label"]`). Adds optional `categoryId` (`format: "category-id"`) on `categorySchemaObject`. Composes `CategoriesSchema` into `CollectionPagePageSchema`.
- `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx` — new 543-line control: summary row that opens an absolute-positioned expanded editor wrapping `JsonFormsArrayControlView`, with duplicate-label highlighting, blank-label gating, and a delete-confirm modal.
- `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx` — GrowthBook-gated `SingleSelect` for `categoryId`, currently backed by `PLACEHOLDER_CATEGORY_OPTIONS: string[] = []`.
- `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx` — refactored to two `data === 0` branches (`!emptyState` / `emptyState`).
- `CollectionEditorStateDrawer.tsx:67-80` — scopes `categoryOptions` into Filters drawer, out of Display.
- Stories, renderer registry, GrowthBook key.

### Findings

**issue (blocking):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx:25,37-40` — `PLACEHOLDER_CATEGORY_OPTIONS` is hardcoded to `[]`. The renderer is registered, and the control is shown the moment GrowthBook's `category-id-dropdown.enabledSites` lists the current site. On a flagged site the editor sees a `SingleSelect` with **zero items** and `isClearable={true}` — they can clear the value but cannot pick anything. If staging is on the flag list, this is broken end-to-end. Either keep the tester unregistered until the data source lands, or wire it now to the parent collection's `categoryOptions` (which exists in the same PR). Mentioned in the prior Claude review — still worth treating as a hard blocker. Confidence: high.

**issue:** `packages/components/src/types/page.ts:117-126` — `categoryOptions[*].label` is `Type.String({ title: "Option name" })` with no `pattern`/`minLength`/`maxLength`. By contrast, `TagCategorySchema` option labels use `pattern: TRIMMED_NON_EMPTY_STRING_REGEX` and `maxLength: 70`. The new control's `hasBlankOptionLabel` only blocks the in-editor *Save changes* button; the parent drawer's `Save changes` is gated by `useBuilderErrors`, which derives from AJV. With no schema-level constraint, blank/whitespace labels (or 1000-char labels) are valid as far as AJV is concerned — so any path that bypasses the inner editor (initial state, programmatic mutation, future callers) can persist them. Align the schema with the tag-option label constraints. Confidence: high.

**issue (non-blocking):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx` is not gated by `useIsUserIsomerAdmin`, but the parallel `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx:258-268` wraps its inner component in an admin check. Today this is fine because `CollectionEditorStateDrawer` scopes `categoryOptions` out for non-admins (`CollectionEditorStateDrawer.tsx:67-80`), but the asymmetry breaks defense-in-depth — anything that ever registers this control on another schema/path drops the gate silently. Mirror the `JsonFormsTagCategoryOptionsControl` wrapper. Confidence: medium.

**issue (non-blocking):** PR description vs. code drift. The body says: *"`JsonFormsArrayControl` allows `emptyState` to be `null` to suppress the empty placeholder when not wanted."* But after `803254bd1 refactor(JsonFormsArrayControl): simplify emptyState handling`, the implementation is now `data === 0 && !emptyState` for the default branch — `null` is falsy, so `!null === true`, meaning `emptyState={null}` renders the default placeholder, not nothing. The "Improvements" bullet should be deleted (or the suppress capability re-added) before merge, otherwise reviewers and future callers will rely on a behaviour that no longer exists. Confidence: high.

**issue (non-blocking):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx:74-79` — the delete modal copy is the literal string *"This option is being used in XX items."* with `XX` un-substituted (TODO + same hard-coded text duplicated from `JsonFormsTagCategoryOptionsControl.tsx:69-73`). Editors will read `XX` verbatim. Either ship neutral copy ("Items currently using this option will lose it") or block this UI until the usage count is wired. Confidence: high.

**issue (non-blocking):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx:55` — `useQueryParse(collectionItemSchema)` is called unconditionally at the top of the component. `collectionItemSchema` requires `siteId: z.coerce.number()`, so if the tester ever matches a `format: "category-id"` field outside the collection-item route (storybook context without router, future reuse) the component throws on mount. Either move the parse inside the flag-enabled branch or guard with a safe-parse. Confidence: medium.

**issue (non-blocking):** Test coverage is Storybook-only. No Vitest for `indicesWithDuplicateLabels` against the new `categoryOptions` shape, the `cannotLeaveExpandedCategoryOptions` predicate, or `CategoriesSchema` composition into `CollectionPagePageSchema`. Given the PR adds validation gates that block save, at least one unit test asserting blank/duplicate labels block and a valid set passes would be worth adding. Confidence: medium.

**issue (non-blocking) — duplication:** `DeleteCategoryOptionModal` (`JsonFormsCategoryOptionsControl.tsx:50-112`) is structurally identical to `DeleteOptionModal` in `JsonFormsTagCategoryOptionsControl.tsx:44-106`, and the duplicate-label banner block is copied three times in this file plus once in `JsonFormsTagCategoryOptionsControl`. Extract `DeleteOptionModal`, the duplicate-label banner, and `useDuplicateLabelIndices(path)` now while the two are still in sync — copy drift gets harder to unwind every commit. Confirms the existing review. Confidence: high.

**question:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx:267-339` — when the expanded editor is open, the renderer returns an `position="absolute"` overlay relative to the *outer* `Box position="relative"`. That outer box is the summary row, which sits inside the drawer's scrollable container. Have you verified the overlay actually covers the drawer (not just the summary row)? `JsonFormsArrayControlView`'s `ComplexEditorNestedDrawer` uses the same pattern, so probably yes, but worth a quick check in the deployed preview. Confidence: low.

**nitpick:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx:479` — `"No option" / "1 option" / "N options"`; `"No options"` reads more naturally for the empty case. Also `JsonFormsCategoryOptionsControl.tsx:165` uses a curly apostrophe (`can’t`); the rest of the codebase mixes both but most files use straight `'`. Confidence: low.

**praise:** The schema-level back-compat reasoning is clearly documented inline (`page.ts:104-114`, `155-165`), and gating `categoryOptions` to the Filters drawer via `getScopedSchema` keeps existing blobs loading. Easy to read intent and the migration path. Confidence: high.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -328,7 +330,26 @@ export function JsonFormsArrayControlView({
                 spacing={0}
                 ref={innerRef}
               >
-                {data === 0 && (
+                {data === 0 && !emptyState && (
```
seaerchin: dumb question: why does `data === 0` work here? not super clear to me tbh

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
seaerchin, line 1: comment: i think this file has alot of duplicated code; understand that we want to dup to be future proof against changes but wdyt about extracting out similar UI components then reusing those?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,559 @@
...
+                w="2.125rem"
+                minH="2.125rem"
+                minW="2.125rem"
+                p="0.25rem"
+                display="flex"
+                alignItems="center"
+                justifyContent="center"
+                isDisabled={isRemoveItemDisabled}
+                aria-label={`Option ${index + 1} actions`}
+                onClick={(e) => e.stopPropagation()}
```
seaerchin: question: why do we need `stopPropagation`?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,559 @@
...
+          </Button>
+        </Box>
+      </VStack>
+    )
+  }
+
+  return (
+    <Box position="relative" w="full">
+      <VStack spacing={0} align="stretch" w="full">
+        {hasDuplicateOptionNameError ? (
```
seaerchin: style: prefer &&

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
seaerchin: comment: this file seems very convoluted - i'd like to see the JsonFormsArrayLayout extracted then common components reused 

as it stands right now, what i'm having trouble w is understanding how each component ties in together cos it feels like i'm reading all the concerns lumped into a single file

adriangohjw: okay i did it upstream!

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
adriangohjw: In JsonForms, `data` in `ArrayLayoutProps` is the **count** of array items (a number), not the array itself — that's how JsonForms passes array length to layout renderers. So `data === 0` just means the array is empty.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
adriangohjw: The row itself has a click handler (via `setSelectedIndex`) that opens the edit drawer on click. Without `stopPropagation`, clicking the dots menu button would bubble up and also fire the row click, opening the drawer at the same time as the menu. This prevents that double-fire.

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,444 @@
...
+              <DuplicateLabelError noun="option" />
+            )
+          }
+          getListItemHasError={(index) => duplicateOptionIndices.has(index)}
+          renderListItemErrorCaption={(index) =>
+            duplicateOptionIndices.has(index)
+              ? "An option with this name already exists."
+              : undefined
+          }
+          emptyState={
```
seaerchin: question: not sure if passing a prop like this is ideal vs just using ternaries, but not opinionated; should we make a decision here then codify it?

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,444 @@
...
+              }}
+              align="stretch"
+              overflow="hidden"
+            >
+              {duplicateOptionIndices.size > 0 && (
+                <Box
+                  aria-hidden
+                  bg="utility.feedback.critical"
+                  width="6px"
+                  mr="-6px"
```
seaerchin, line 374: question: do we need the negative margin? feels like it's an anti-pattern which we should avoid if possible, but given that this is just padding, don't feel very strongly about this

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -0,0 +1,444 @@
...
+                  py="0.5rem"
+                >
+                  <Icon
+                    as={BiGridVertical}
+                    fontSize="1.5rem"
+                    color="interaction.support.disabled"
+                    aria-hidden
+                  />
+                </Flex>
+                <Box
```
seaerchin: question: should we just use a `button` here? can use a plain `chakra.button`

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
```diff
@@ -44,8 +46,31 @@ export const jsonFormsArrayControlTester: RankedTester = rankWith(
...
+  getListItemHasError?: (index: number) => boolean
+  /** When the row is in error, replaces the default "Fix issues before saving" caption. */
+  renderListItemErrorCaption?: (index: number) => string | undefined
+  /** Rendered after the schema description and before the draggable list. */
+  belowDescription?: ReactNode
+  /**
+   * Applied to the value from `createDefaultValue` when adding an array item.
+   * Use when schema `default` would be wrong under AJV `useDefaults` (e.g. legacy rows).
+   */
+  mapNewArrayItem?: (item: unknown) => unknown
```
seaerchin: comment: it feels like alot of these are forwarded - as mentioned in previous PR, if we can avoid this prop drilling that would be great (perhaps by having `DraggableTagButton` as a specialised component that does more?)

feels abit iffy to me because this is a JSONForms component that suddenly is concerned with view logic which shouldn't be the case

seaerchin:
Scope: 14 files, +858/−40, stacked on feat/collection-filters-drawer. Sources read: CLAUDE.md, CONTRIBUTING.md, packages/components/CLAUDE.md, and all 7 isomer-conventions entries.

HARD VIOLATIONS
None that cleanly break a stated rule.

JUDGEMENT CALLS
1. JsonFormsArrayControl.tsx:107-152 — prop overload (react-new-component-over-prop-overload.md). The headline concern; arguable, leans fixable, not blocking. The PR adds 8 new props at once: listItemIcon, listItemContentProps, renderListItemTrailing, renderListItemSubtitle, getListItemHasError, renderListItemErrorCaption, belowDescription, mapNewArrayItem — exactly the "flag soup / render-prop accretion" signal the convention names. BUT it survives the convention's litmus test: the View still owns substantial shared machinery — drag/drop reorder, add/remove plumbing, nested-drawer editing, AJV error wiring. Per the "Shells with real shared machinery" carve-out, these are single-caller slots on a legitimate shared shell. Flagging per the rule: renderListItemTrailing/renderListItemSubtitle/renderListItemErrorCaption each have only the new CategoryOptions consumer; if a third divergent consumer appears, push decoration into consumers via the existing renderListItem seam rather than growing more slots. mapNewArrayItem is the cleanest addition (replaced an addItem-wrapping hack in JsonFormsTagCategoryControl — net simplification).

2. JsonFormsCategoryOptionsControl.tsx (444 lines) — no direct Storybook, but covered. No violation: the "every public component needs a story" rule applies to packages/components, not Studio's features/. The control is exercised via six new play-stories in EditCollectionIndexPageNewExperience.stories.tsx:326-419.

3. Forms — useZodForm convention does not apply. JSONForms schema-driven controls; state flows through JSONForms, not react-hook-form. The useState usages (expandedOpen, deleteTarget at JsonFormsCategoryOptionsControl.tsx:392,516) are UI-only — explicitly sanctioned by react-forms-usezodform-over-usestate.md.

adriangohjw:
@claude review the above unaddressed comments. for each of them, lemme know what you think and if you think its worth fixing, propose a fix

adriangohjw:
@claude 

1. skip
2. add pattern but not max length
3. fix it
4. fix it
5. fix it
6. no
7. fix anything worth doing
8. no
9. okay after fixing the above

adriangohjw:
lmao claude tagged the wrong issue/PR hahaha 🤡

adriangohjw:
@claude actually your `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/__tests__/categoryOptionsSchema.test.ts` is quite pointless and noisy - remove it

seaerchin (approved):

PR #2011, opened 2026-04-21
🔖 Feat: category management (backend)

adriangohjw:
## Problem

This PR stacks on [`feat/collection-category-management`](https://github.com/opengovsg/isomer/tree/feat/collection-category-management) and adds the server APIs and Studio wiring needed so collection **category options** are loaded from the parent collection index and **usage counts** are accurate when editors delete an option.

Closes https://linear.app/ogp/issue/ISOM-2289/category-management-backend

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Added `page.getCategoryOptions`, which reads `categoryOptions` from the **published** blob of the collection’s index page (same parent as the current collection page/link), validates each entry as `{ id, label }`, and returns a safe array for the editor.
- Added `collection.getCategoryOptionUsageCount`, which counts collection child **pages** and **links** whose draft **or** published blob sets `page.categoryId` to the given option id (used in the delete-option confirmation copy).
- Wired `JsonFormsCategoryIdControl` to `getCategoryOptions` (with `Suspense` + skeleton) instead of placeholder items, still behind the existing GrowthBook gate for enabled sites.
- Updated `JsonFormsCategoryOptionsControl` so the delete modal shows a live usage count (with inline skeleton and error-boundary fallback) and small accessibility improvements on the modal header.

**Improvements**:

- Invalidate `page.getCategoryOptions` after collection metadata save, page metadata save, and publish so the UI stays in sync with the index page blob.
- Exported `CollectionPageCategoryOption` from `@opengovsg/isomer-components` for shared typing between server and components.
- Expanded router and MSW coverage, and refreshed related Storybook stories for the new queries.

**Bug Fixes**:

- `getCategoryOptions` returns `NOT_FOUND` when the resource has no `parentId`, instead of querying with a missing collection context.

## Before & After Screenshots

**BEFORE**:

N/A — behaviour and copy changes are limited to existing collection editing flows (category id select and delete category option modal).

**AFTER**:

N/A — optional: capture the delete modal showing a numeric usage count instead of the previous placeholder.

## Tests

- [ ] Check out this branch with `feat/collection-category-management` as the merge base, run Studio, and open a **collection** that has category options configured on its **index** page (published).
- [ ] Open a **collection page** or **collection link** under that collection whose schema includes **Category ID** (GrowthBook-enabled site). Confirm the dropdown lists the same options (labels) as on the index, and saving a selection persists after refresh.
- [ ] On the collection **index** editor, open **category options**, pick an option that is assigned on at least one child page/link, open **delete**, and confirm the infobox shows the correct **usage count** (and a brief skeleton while loading).
- [ ] Publish or save collection settings from a drawer that touches the index metadata, then reopen a child page editor and confirm category options still load (cache invalidation).

---

> [!NOTE]
> **Medium Risk**
> Adds new protected tRPC queries and SQL that read published index-page JSON and count categoryId usage across draft/published blobs, which can affect editor correctness and query performance. UI wiring changes are mostly cache invalidation and Suspense rendering but rely on the new endpoints behaving correctly.
> 
> **Overview**
> Enables **real category option management for collections** by adding two new protected tRPC queries: `page.getCategoryOptions` (loads and type-filters `categoryOptions` from the *published* collection index blob) and `collection.getCategoryOptionUsageCount` (counts collection child pages/links whose *draft or published* blob matches a given `page.categoryId`).
> 
> Updates Studio to use these APIs: the `category-id` dropdown now fetches options via Suspense (no more placeholder data), the category-option delete modal shows a live usage count with Skeleton/ErrorBoundary fallback, and relevant mutations (save metadata/collection settings, publish, update link) now invalidate `getCategoryOptions`/usage-count caches so the UI stays in sync. Tests, MSW handlers, and Storybook plays are extended to cover the new queries and modal a11y labels.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a34267f86f9d50aa3ec83d2a63d34a2b68089cc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seaerchin (commented):

## Diff summary

This PR (stacked on `feat/collection-category-management`) adds backend support for collection category options. The flow:

- New tRPC query [`page.getCategoryOptions`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/server/modules/page/page.router.ts) reads `categoryOptions` from the **published** blob of the resource's parent's `IndexPage`, runtime-filters with `isCollectionPageCategoryOption`, returns `{ id, label }[]`. Returns `NOT_FOUND` if `parentId` is missing.
- New tRPC query [`collection.getCategoryOptionUsageCount`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/server/modules/collection/collection.router.ts) takes the index `pageId` + `categoryId`, walks to the parent `Collection`, then counts `CollectionPage`/`CollectionLink` whose **draft or published** blob has `page.categoryId == categoryId` via raw SQL JSON path with `nullif(trim(...), '')`.
- New zod schema [`getCategoryOptionUsageCountSchema`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/schemas/collection.ts) (`siteId`/`pageId` ≥ 1, `categoryId` UUID).
- UI wiring: [`JsonFormsCategoryIdControl`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx) now Suspense-fetches options; [`JsonFormsCategoryOptionsControl`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx) shows live usage count in the delete modal (Suspense + ErrorBoundary), short-circuits delete for unsaved (no-id) rows, and adds VisuallyHidden dialog labels.
- Cache invalidation added in [`CollectionEditorStateDrawer`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx), [`MetadataEditorStateDrawer`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/features/editing-experience/components/Drawer/MetadataEditorStateDrawer.tsx), [`PublishButton`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/features/editing-experience/components/PublishButton.tsx), [`LinkEditorDrawer`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/features/editing-experience/components/Drawer/LinkEditorDrawer.tsx).
- Type export [`CollectionPageCategoryOption`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/packages/components/src/types/page.ts), MSW handlers, Storybook updates, plus thorough router tests for both queries (auth/permission/404/count semantics).

## Review

**praise: comprehensive test coverage** (confidence: high)
The router tests cover unauthed/forbidden/not-found and all the count edge cases (zero, single draft, multiple, published-only). The "published-only" case in particular is good because it demonstrates draft-only-published items are still counted.

**praise: short-circuit for unsaved options** (confidence: high)
`handleDeleteOptionMenuItemClick` skipping the modal for rows without an `id` avoids a UUID-validation 400 on `getCategoryOptionUsageCount` and is the right UX. Nice.

**issue (security/correctness):** `getCategoryOptions` does not verify the resource is a child of a collection. (confidence: high)
[`page.router.ts`](https://github.com/opengovsg/isomer/blob/feat/collection-category-management-backend/apps/studio/src/server/modules/page/page.router.ts) only checks that the input `pageId` has *some* `parentId` and then queries for a sibling `IndexPage` under that parent. The parent could be a Folder (not a Collection), in which case the inner join on `IndexPage` is empty and you correctly return `[]` — but the error message says "Collection page not found" when `parentId` is null, which is misleading for non-collection resources (e.g., `RootPage`). More importantly, the query returns `categoryOptions: []` silently for any folder child instead of erroring, masking misuse. Consider also asserting `parent.type === Collection` and surfacing a clearer error otherwise.

**issue (correctness):** `getCategoryOptions` only reads the *published* index blob. (confidence: high)
The `innerJoin` on `Version`/`publishedVersionId` means an unpublished (or never-published) collection index returns an empty list — so the dropdown silently shows zero options even though the editor sees draft `categoryOptions` configured on the index page. Given the usage-count query reads draft *or* published (correctly), the asymmetry is surprising. Either document this as intentional ("only published options can be assigned to children") or coalesce draft/published like `getCategories` already does. This contradicts the PR description which says "reads `categoryOptions` from the published blob" — confirm intent.

**issue (correctness):** `getCategoryOptions` permission check is site-wide, not resource-scoped. (confidence: medium)
`bulkValidateUserPermissionsForResources` is called without `resourceIds`, so it falls back to site-level permission. That's consistent with `getCategories`, but combined with the missing parent-type check above it means a user with read access to a site can pass an arbitrary `pageId` from that site (e.g., a folder index) and probe `categoryOptions`. Low impact (data is non-sensitive and visible elsewhere), but worth tightening for defense-in-depth.

**issue (correctness):** `getCategoryOptionUsageCount` accepts any `IndexPage` on the site, but does not check the index is under a `Collection`. (confidence: medium)
The first SELECT filters `type = IndexPage` (good), then `getSiteResourceById(parentId, type: Collection)` enforces the parent is a Collection (good). However, an `IndexPage` of a Folder would fail the second check with "Collection not found" rather than "Collection index page not found" — minor UX nit, but the test suite doesn't cover this case. Add a test for "index page exists but parent is a Folder".

**nitpick (perf):** N+1 not present, but the count query joins the entire collection on every keystroke. (confidence: low)
Each open of the delete modal triggers a join across `Resource × Blob × Version × Blob` for every collection child. For collections with hundreds of items this is fine (single query), but the JSON path predicate cannot use a btree index. If category usage counts get called frequently, consider a GIN index on `Blob.content jsonb_path_ops` or extracting `categoryId` to a generated column. Not blocking.

**nitpick (style):** raw SQL in `getCategoryOptionUsageCount` — injection-safe but reformattable. (confidence: high)
The `categoryId` is interpolated via the kysely `sql` tagged template, so it's parameterized — no injection risk. The double-`nullif(trim(...), '')` is correct (handles `null`, `''`, and whitespace-only values). However, this could be expressed with kysely's expression builder (`eb.or([eb('draftBlob.content...', '=', categoryId), ...])`) to avoid the raw-SQL escape hatch and stay consistent with the rest of the file. Minor.

**nitpick:** input zod schema `pageId: z.number().min(1)` doesn't enforce integer. (confidence: low)
`getCategoryOptionUsageCountSchema` — `1.5` would pass validation, then `String(pageId)` becomes `"1.5"`, which compares against the (text) Resource id and silently misses. Add `.int()`. Same applies to `basePageSchema`, but that's pre-existing.

**nitpick (UI):** `pageId: pageId || linkId || -1` is a footgun. (confidence: medium)
In `JsonFormsCategoryIdControl.tsx`, if neither `pageId` nor `linkId` is present the suspense query fires with `pageId: -1`, which fails zod's `.min(1)` and surfaces a 400 inside Suspense. If this branch should be unreachable, prefer `invariant(pageId ?? linkId)` or skip rendering. Otherwise an `enabled: !!pageId || !!linkId` would be cleaner — but that requires `useQuery` not `useSuspenseQuery`.

**question:** no audit log entries for either new query. (confidence: high)
Both endpoints are reads, so this is correct — no audit log needed. Just confirming intent (the PR title's "category management" might suggest write paths; the actual writes happen via existing `updatePageBlob`).

**question:** cache invalidation on `LinkEditorDrawer`. (confidence: medium)
Only `getCategoryOptionUsageCount` is invalidated after `updateCollectionLink`, not `getCategoryOptions`. If updating a link can change the collection index's `categoryOptions`… it can't (links don't edit the index). So the asymmetry is intentional — but worth a one-line comment, since the other three call sites invalidate both.

---

Overall: clean read-only additions with good test coverage. The two issues worth addressing before merge are (1) the published-only blob read for `getCategoryOptions` (likely a bug given the PR description), and (2) tightening the parent-type assertion / error messaging in `getCategoryOptions`. Everything else is nitpick territory.

adriangohjw (commented):
## Summary

This PR adds two new tRPC queries (`page.getCategoryOptions`, `collection.getCategoryOptionUsageCount`) and wires the Studio editor to use them — replacing the placeholder category dropdown and the hard-coded "XX items" copy in the delete-option modal. Cache invalidations on metadata/collection saves, publishes, and link updates keep the new caches in sync.

- New backend reads live in `apps/studio/src/server/modules/page/page.router.ts:209-253` and `apps/studio/src/server/modules/collection/collection.router.ts:530-588`.
- The dropdown change is at `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx:23-50` (Suspense-wrapped at L70-72).
- The delete modal change is at `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx:61-78` and L116-138, with the menu-click short-circuit at L198-219.

Overall the diff is well-tested and small. SQL is properly parameterized via kysely tagged templates and UUID-validated, permissions are gated through `bulkValidateUserPermissionsForResources`. A couple of correctness/UX issues worth resolving before merge.

---

## Findings

**issue (correctness):** `apps/studio/src/server/modules/page/page.router.ts:233-244` — `getCategoryOptions` only reads the **published** index blob (it uses `innerJoin` on `Version`/`publishedVersionId`). If the collection index has never been published — or the editor just edited `categoryOptions` on the index and hasn't published yet — child page/link editors will silently see an empty dropdown even though the index has options configured. This is asymmetric with `getCategoryOptionUsageCount`, which already correctly coalesces `draft || published`. Either coalesce here too (mirror the pattern in `getCategories` at L184-200) or explicitly document this as intentional ("children can only pick from published options"). The PR description states this is the intended behavior, but I'd push back: editors typically expect newly-added options to be selectable before publish, and there's no UI affordance explaining the delay.

Confidence: high

**issue (UX/correctness):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx:33` — `pageId: pageId || linkId || -1` will pass `-1` into the query when both are falsy, which fails the `z.number().min(1)` validator and surfaces a `BAD_REQUEST` *inside* Suspense. There is no `ErrorBoundary` wrapping `SuspendableJsonFormsCategoryIdSelect` (unlike the count modal in `JsonFormsCategoryOptionsControl.tsx:117`), so the failure will bubble up and crash the surrounding form. If this branch is supposed to be unreachable, use `invariant(pageId ?? linkId)` or an explicit early return. Otherwise wrap in an `ErrorBoundary` for parity with the modal.

Confidence: medium

**issue (correctness / defense-in-depth):** `apps/studio/src/server/modules/page/page.router.ts:218-231` — `getCategoryOptions` does not assert that the queried resource is a child of a collection. The first query has no `type` filter, so any resource with a `parentId` will pass. The sibling-`IndexPage` join would then return a folder's `IndexPage` (`ResourceType.IndexPage` is shared by folders and collections per `prisma/schema.prisma:111-119`), whose blob lacks `categoryOptions`, and `categoryOptions: []` is returned silently. Worse, the error message at L229 says "Collection page not found" even when the resource exists but is e.g. a `RootPage`. Tighten by checking the parent resource is `ResourceType.Collection` and emitting clearer errors. Mostly a misuse/observability concern — not a data leak, since the published `categoryOptions` are visible to any site-read user via the index page anyway.

Confidence: high

**issue (test coverage gap):** `apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts:427-454` — the "page is not a collection index" test only exercises a `CollectionPage`, not a folder's `IndexPage`. Since `ResourceType.IndexPage` is overloaded between folders and collections, the more interesting case is "index page exists but parent is a Folder" — that should hit the second `getSiteResourceById(..., type: Collection)` guard at `collection.router.ts:555-565` and return "Collection not found". Worth a test to lock in that behavior.

Confidence: high

**issue (correctness):** `apps/studio/src/features/editing-experience/components/Drawer/CollectionEditorStateDrawer.tsx:51`, `MetadataEditorStateDrawer.tsx:53`, `PublishButton.tsx:51` — `utils.page.getCategories.invalidate()` was changed from `.invalidate({ pageId, siteId })` to a no-arg call, which now invalidates *all* `getCategories` queries across the React Query cache (every page, every site). For `getCategoryOptions`/`getCategoryOptionUsageCount` this is fine because they're new, but the side-effect on `getCategories` is broader than the previous behavior. If the intent was to also invalidate sibling pages (e.g. because category labels on one page surface in another's dropdown), confirm — otherwise revert that one call to scoped invalidation.

Confidence: medium

**nitpick (schema):** `apps/studio/src/schemas/collection.ts:95-99` — `pageId: z.number().min(1)` does not enforce `int()`. A request with `pageId: 1.5` would validate, then `String(pageId)` becomes `"1.5"` which is compared against the (text) `Resource.id` column and silently returns no rows. Add `.int()`. Same applies to `basePageSchema` in `schemas/page.ts:41-44`, but that's pre-existing.

Confidence: low

**nitpick (style):** `apps/studio/src/server/modules/collection/collection.router.ts:578-583` — raw `sql` template for the categoryId predicate. It is injection-safe (kysely parameterizes `${categoryId}`, and UUID is enforced upstream), and the double-`nullif(trim(...))` is correct for `null`/`""`/whitespace handling. But this could be expressed with kysely's expression builder (`eb.or([eb('draftBlob.content ...', '=', categoryId), ...])`) to stay consistent with the rest of the router. Optional.

Confidence: low

**praise:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx:198-219` — short-circuiting the delete modal for rows that don't yet have an `id` is the right UX call (and dodges the UUID-validation 400 you'd otherwise hit). Clear comment explaining why, too.

Confidence: high

**praise:** `apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts:571-612` — the "count an item that only has categoryId on the published blob" test is a nice touch — it explicitly asserts `draftBlobId === null` before checking the count, which locks in the draft-OR-published semantics rather than just incidentally passing.

Confidence: high

**question:** `apps/studio/src/features/editing-experience/components/Drawer/LinkEditorDrawer.tsx:180` — only `getCategoryOptionUsageCount` is invalidated after `updateCollectionLink`, not `getCategoryOptions`. That's correct (links can't modify the index's `categoryOptions`) but the asymmetry vs. the other three call sites is unobvious. Either add a one-line comment, or invalidate both for consistency. Non-blocking.

Confidence: medium

adriangohjw:
@claude /pr-review

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/DraggableTagButton.tsx
```diff
@@ -55,10 +55,14 @@ const DraggableTagButton = forwardRef<DraggableTagButtonProps, "div">(
   ) => {
     const mergedDragHandleProps: BoxProps = {
       py: isError ? "0.75rem" : "1.25rem",
+      pl: "0.5rem",
```
seaerchin: issue: i think we should delete this component and put the changes into the other `DraggableTagButton` located at `form-builder/components`

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
```diff
@@ -18,25 +21,27 @@ interface JsonFormsCategoryIdControlProps extends ControlProps {
...
   return (
     <SingleSelect
       value={data}
       name={label}
-      items={PLACEHOLDER_CATEGORY_OPTIONS.map((category) => ({
-        label: category,
-        value: category,
+      items={categoryOptions.map(({ id, label: optionLabel }) => ({
+        label: optionLabel,
+        value: id,
```
seaerchin: 

apps/studio/src/schemas/collection.ts
```diff
@@ -94,6 +94,12 @@ export const getCollectionTagsSchema = z
     { message: "Exactly one of resourceId or collectionId must be provided" },
   )
 
+export const getCategoryOptionUsageCountSchema = z.object({
+  siteId: z.number().min(1, { message: "Site ID must be at least 1" }),
+  pageId: z.number().int().min(1, { message: "Page ID must be at least 1" }),
```
seaerchin: question: how come `pageId` needs the `.int` but not `siteId`? no issue on my end but just abit strange

apps/studio/src/schemas/collection.ts
```diff
@@ -94,6 +94,12 @@ export const getCollectionTagsSchema = z
...
+  categoryId: z.string().uuid(),
```
seaerchin: comment: i think `uuid` is deprecated and we can just do `z.uuid()`

apps/studio/src/server/modules/collection/collection.service.ts
```diff
@@ -52,3 +56,84 @@ export const createCollectionIndexJson = (title: string) => {
...
 }
+
+export const getCollectionTagsForResource = async ({
+  resourceId,
+  collectionId,
+  siteId,
+}: {
+  resourceId?: number
+  collectionId?: number
+  siteId: number
```
seaerchin: question: we should use a discriminated union to type this properly. as of right now, passing in solely `siteId` is a valid argument

apps/studio/src/server/modules/collection/collection.service.ts
```diff
@@ -52,3 +56,84 @@ export const createCollectionIndexJson = (title: string) => {
...
+          .where("siteId", "=", siteId)
+          .select("parentId"),
+      ),
+    )
+    .select("id")
+    .executeTakeFirst()
+
+  if (!indexPage) {
+    return []
+  }
```
seaerchin: comment: it feels like the missing index page is an error during the migration process - wondering if we should automatically create one for them rather than accept the failure and return an empty array?

apps/studio/src/server/modules/collection/collection.service.ts
```diff
@@ -52,3 +56,84 @@ export const createCollectionIndexJson = (title: string) => {
...
+
+  const { draftBlobId, publishedVersionId } = await db
```
seaerchin: performance: should we collapse these 3 queries into a single one using left joins? now we do 3 round trips (in the worst case) to the db

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -536,88 +536,80 @@ export const collectionRouter = router({
...
-      if (!indexPage) {
-        return []
-      }
+  /**
+   * Counts collection pages/links whose draft **or** published blob has `page.categoryId` equal to the
+   * given id. `pageId` is the collection index page resource id (Studio URL `pageId`).
+   */
+  getCategoryOptionUsageCount: protectedProcedure
+    .input(getCategoryOptionUsageCountSchema)
+    .query(async ({ ctx, input: { siteId, pageId, categoryId } }) => {
```
seaerchin: nit: can we rename this to `indexPageId`?

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -536,88 +536,80 @@ export const collectionRouter = router({
...
+            siteId,
+            resourceId: parentId,
+            type: ResourceType.Collection,
+          })
+        : null
+      if (!collection) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Collection not found",
+        })
```
seaerchin: comment: should we simplify this into a single query and return the same error message in both cases where index page is missing and collection is missing to our user? i don't think the difference is noticeable to them and imo single query here seems simpler to read + better for perf

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -536,88 +536,80 @@ export const collectionRouter = router({
...
+            ),
+            eb(
+              sql`nullif(trim("publishedBlob"."content"->'page'->>'categoryId'), '')`,
+              "=",
+              categoryId,
+            ),
+          ]),
         )
-      }
+        .select(sql<number>`cast(count(*) as int)`.as("count"))
```
seaerchin: question: this is queried from the index page for the usage of teh category id but it assumes that both draftBlob and publishedBlob are interchangeable, which isn't the assumption that's made elsewhere. 

should we return the published count only since that's what's visible on the end user website and update the text to maek this clear?

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -80,3 +85,68 @@ export const createFolderIndexPage = (title: string) => {
...
 }
+
+function isCollectionPageCategoryOption(
+  value: unknown,
+): value is CollectionPageCategoryOption {
+  if (!value || typeof value !== "object") {
+    return false
+  }
+  const rec = value as Record<string, unknown>
+  return typeof rec.id === "string" && typeof rec.label === "string"
```
seaerchin: question: should we validate that the `rec.id` is a `uuid`

seaerchin:
idk why got duplicate review sadge

seaerchin (commented):
## Summary

Adds two protected tRPC queries — `page.getCategoryOptions` (reads `categoryOptions` from the *published* collection-index blob; see `apps/studio/src/server/modules/page/page.service.ts:99-152`) and `collection.getCategoryOptionUsageCount` (counts collection children whose draft *or* published blob's `page.categoryId` matches; see `apps/studio/src/server/modules/collection/collection.router.ts:546-613`) — and wires them into the category-id dropdown and the delete-option modal. Also extracts the existing `getCollectionTags` body into `collection.service.ts:getCollectionTagsForResource` and broadens cache invalidation in `MetadataEditorStateDrawer`, `PublishButton`, and `LinkEditorDrawer`.

Codeflow on the read path: `JsonFormsCategoryIdControl` parses `siteId` + `pageId|linkId` from the route (`JsonFormsCategoryIdControl.tsx:60-89`) → `page.getCategoryOptions` resolves the parent collection's published index and JSON-paths into `page->categoryOptions`, with a runtime type guard (`page.service.ts:89-97`). On delete: `JsonFormsCategoryOptionsControl` resolves the option's `id` from JsonForms state (`JsonFormsCategoryOptionsControl.tsx:135-144`) → `collection.getCategoryOptionUsageCount` left-joins draft + published blobs and counts matches via `nullif(trim(...->>'categoryId'), '')`.

Overall: design is sound, parameterised SQL is safe, permission/auth checks are in place, and the test coverage on the router is reasonable. A few correctness and resilience issues below.

---

### Findings

**issue (bug):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx:135-144` — `resolveTarget` falls back to `categoryId: ""` when the option has no persisted `id` yet (new option, hidden UUID control hasn't mounted). The delete modal still opens, `CategoryOptionUsageCount` fires `getCategoryOptionUsageCount` with `categoryId: ""`, the Zod `.uuid()` check in `getCategoryOptionUsageCountSchema` rejects it, and the ErrorBoundary shows "—" instead of a real count.

The sibling `JsonFormsTagCategoryOptionsControl.tsx:44-51` handles exactly this case by detecting "no persisted id" and calling `removeItems` directly without a modal — same logic belongs here. Today the only thing masking the bug is the storybook play function that explicitly opens-and-closes each row to mount the UUID control; a real user could trip it.

Confidence: high

---

**issue (resilience):** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx:78-89` — the suspending `<SuspendableJsonFormsCategoryIdSelect>` is wrapped in `<Suspense>` but not in an `ErrorBoundary`. If `page.getCategoryOptions` rejects (transient network blip, NOT_FOUND when the resource has no `parentId` per `page.service.ts:114-119`, BAD_REQUEST when the parent isn't a Collection, etc.), the throw bubbles past the dropdown and takes down the entire JsonForms editor tree.

`JsonFormsCategoryOptionsControl.tsx:283-301` does this correctly — Suspense *inside* an ErrorBoundary with a "—" fallback. Mirror the pattern here so the form degrades to "dropdown unavailable" instead of crashing the editor.

Confidence: high

---

**issue (test gap):** `apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts:1638-1719` — the new `getCategoryOptionUsageCount` suite asserts counts for `CollectionPage` resources (draft, published-only, multiple) but never sets up a `CollectionLink` whose blob carries `page.categoryId`. The router query explicitly includes both `ResourceType.CollectionPage` and `ResourceType.CollectionLink` (`collection.router.ts:591-594`), so a future refactor that drops `CollectionLink` from the filter would pass tests but silently undercount. Worth adding one link-only fixture row to lock the contract.

Confidence: high

---

**suggestion (consistency):** `apps/studio/src/server/modules/collection/collection.router.ts:569-582` — the new endpoint collapses "index page has no `parentId`" and "parent isn't a Collection" into a single `"Collection not found"` error. The neighbouring `countTagOptionsUsage` at lines 341-358 keeps them distinct, and the most recent commit on this branch (`0b33108bc "fix: restore distinct error message for null parentId in countTagOptionsUsage"`) was specifically about *preserving* that distinction after a refactor. Either both endpoints should keep the distinct messages or the rationale for diverging here should be in a comment.

Confidence: medium

---

**suggestion (architecture):** Per `apps/studio/src/server/CLAUDE.md` ("Routers wire input → service"), this PR already extracted `getCollectionTagsForResource` into `collection.service.ts:60-101` — but the new `getCategoryOptionUsageCount` keeps the DB query, parent-collection lookup, and aggregation SQL inline in the router callback (`collection.router.ts:546-613`). Moving the body into something like `countCategoryOptionUsage` in `collection.service.ts` would match the same refactor pattern this PR is otherwise applying and keep the router as wiring-only.

Confidence: medium

---

**nitpick:** `apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx:81` — Suspense fallback is `<Skeleton />` with no `height`/`width`. Depending on parent layout this can collapse to ~0px or cause noticeable layout shift when the `SingleSelect` mounts. The sibling control sets explicit dimensions (`JsonFormsCategoryOptionsControl.tsx:286-292`). Match the SingleSelect's resting height (e.g. `height="2.5rem"`).

Confidence: medium

---

**note:** `apps/studio/src/server/modules/page/page.service.ts:106-126` — `getCategoryOptionsForPage` checks that the page's parent is a `Collection` but doesn't check the page's own type. Today every resource sitting under a `Collection` is a `CollectionPage` / `CollectionLink` / `IndexPage`, so this is fine; but a defensive `where("r.type", "in", [...])` would make the intent explicit and fail fast if the resource model grows new types.

Confidence: low

---

**praise:** Runtime `isCollectionPageCategoryOption` filter in `page.service.ts:89-97` is the right call — the Kysely `sql<CollectionPageCategoryOption[] | null>` cast is a *type* assertion only, and blob content can drift from the schema. Filtering bad entries instead of throwing is the more forgiving behaviour for editors.

apps/studio/src/schemas/collection.ts
adriangohjw: oh ~vibe coded~ agentic engineered haha will remove to be consistent

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/DraggableTagButton.tsx
adriangohjw: moved it here https://github.com/opengovsg/isomer/pull/2449

apps/studio/src/server/modules/collection/collection.router.ts
adriangohjw: For a deletion guard (“can this category be safely deleted?”), counting both draft and published is intentionally conservative and correct. If you only counted published and someone deleted a category that a draft was still using, that draft would silently reference a deleted category — a data integrity bug. The current OR across draft and published blobs is the right behaviour here; the comment in the docstring already documents it explicitly. The reviewer’s concern about consistency with other parts of the codebase is noted, but the semantics of this endpoint are genuinely different from a “what does the user see?” read.

apps/studio/src/server/modules/collection/collection.router.ts
adriangohjw: udpated in https://github.com/opengovsg/isomer/pull/2011/commits/43595a6b5aa327143bb65682c5c850d48259ceed

apps/studio/src/server/modules/collection/collection.service.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/2011/commits/43595a6b5aa327143bb65682c5c850d48259ceed

apps/studio/src/server/modules/collection/collection.service.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/2011/commits/43595a6b5aa327143bb65682c5c850d48259ceed

apps/studio/src/server/modules/collection/collection.service.ts
adriangohjw: Already handled at the UI layer — `IndexpageRow.tsx` catches the NOT_FOUND and calls createIndexPage reactively, so returning [] is the correct fallback for the read path.

apps/studio/src/server/modules/collection/collection.router.ts
adriangohjw: updated in https://github.com/opengovsg/isomer/pull/2011/changes/190ac641869a41f87060a4197231e5b4cdd69db3

apps/studio/src/features/editing-experience/components/Drawer/MetadataEditorStateDrawer.tsx
```diff
@@ -48,9 +48,13 @@ export default function MetadataEditorStateDrawer(): JSX.Element {
...
-      await utils.page.readPageAndBlob.invalidate({ pageId, siteId })
-      await utils.page.readPage.invalidate({ pageId, siteId })
-      await utils.page.getCategories.invalidate({ pageId, siteId })
+      await Promise.all([
+        utils.page.readPageAndBlob.invalidate({ pageId, siteId }),
+        utils.page.readPage.invalidate({ pageId, siteId }),
+        utils.page.getCategories.invalidate(),
+        utils.page.getCategoryOptions.invalidate(),
+        utils.collection.getCategoryOptionUsageCount.invalidate(),
+      ])
```
dcshzj: nit: should be `void` to be consistent

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -527,97 +527,81 @@ export const collectionRouter = router({
...
+        siteId,
+        action: "read",
+        userId: ctx.user.id,
+      })
 
-      const { draftBlobId, publishedVersionId } = await db
-        .selectFrom("Resource")
-        .where("id", "=", String(indexPage.id))
-        .select(["draftBlobId", "publishedVersionId"])
+      const { id: collectionId } = await db
```
dcshzj: issue: should move this into the service instead of leaving in the router

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -244,12 +292,31 @@ function CategoryOptionsExpandedEditor({
...
+                    />
+                  }
+                >
+                  <CategoryOptionUsageCount
+                    siteId={siteId}
+                    indexPageId={pageId}
+                    categoryId={deleteTarget.categoryId}
+                  />
+                </Suspense>
+              </ErrorBoundary>{" "}
```
dcshzj: comment: I think this suspense boundary is a little weird, a cleaner one would be that the entire paragraph is put behind the suspense such that the appearance of the entire modal loading is clearer to the user rather than just the count

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
```diff
@@ -244,12 +292,31 @@ function CategoryOptionsExpandedEditor({
...
+              items. To undo this change, you will need to create and re-assign
```
dcshzj: nit: also should handle the singular and plural `items`

apps/studio/src/features/editing-experience/components/PublishButton.tsx
```diff
@@ -47,12 +47,16 @@ const SuspendablePublishButton = ({
 
   const { mutate, isPending } = trpc.page.publishPage.useMutation({
     onSettled: async () => {
-      await utils.page.readPage.refetch({ pageId, siteId })
-      await utils.page.getCategories.invalidate({ pageId, siteId })
-      await utils.site.getLocalisedSitemap.invalidate({
-        resourceId: pageId,
-        siteId,
-      })
+      await Promise.all([
```
dcshzj: nit: same here, should be void also

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
```diff
@@ -18,25 +22,27 @@ interface JsonFormsCategoryIdControlProps extends ControlProps {
...
       }))}
       isClearable={true} // TODO: change to false after migration where it's no longer optional
```
dcshzj, line 47: question(non-blocking): for completeness can I check if our default article page on Studio already defines the category in this new location? Or are we intending to do that after the migration is complete

Actually separately should we pin this to the feature flag? So it can no longer be clearable once the feature is activated

apps/studio/src/schemas/collection.ts
```diff
@@ -83,17 +83,20 @@ export const createCollectionSchema = z.object({
...
-    siteId: z.number().min(1),
-  })
-  .refine(
-    (data) =>
-      (data.resourceId !== undefined) !== (data.collectionId !== undefined),
-    { message: "Exactly one of resourceId or collectionId must be provided" },
+  .object({ siteId: z.number().min(1) })
+  .and(
+    z.union([
+      z.object({ resourceId: z.number().min(1) }),
```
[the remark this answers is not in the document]
dcshzj: Might need to double check that this is ok, cos I recall @seaerchin added this for a specific reason

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
adriangohjw: the process will be dogfooding -> migration -> remove feature flag + live

because once the migration is done, technically we cannot revert due to 2 sources of truth of the category (one string + one UUID)

the current feature flag is to block non-isomer admins from seeing it until then, so will clean this up together before going live

or do you mean reference the FF here so that we won't forget to remove?

apps/studio/src/server/modules/collection/collection.router.ts
adriangohjw: done 👍

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
adriangohjw: done 👍

apps/studio/src/features/editing-experience/components/PublishButton.tsx
adriangohjw: done 👍

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryOptionsControl.tsx
adriangohjw: will wait to fix post-dogfooding sessions together with the other design/UX feedback

apps/studio/src/features/editing-experience/components/Drawer/MetadataEditorStateDrawer.tsx
adriangohjw: done 👍

apps/studio/src/schemas/collection.ts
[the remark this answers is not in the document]
adriangohjw: reverted this for now

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsTagCategoryOptionsControl.tsx
```diff
@@ -41,10 +41,19 @@ const JsonFormsTagCategoryOptionsArrayLayoutInner = (
...
       | { label?: string; id?: string }
       | undefined
+    const tagId = item?.id?.trim()
+
+    // New item without a persisted id — remove immediately, no modal needed.
+    if (!tagId) {
+      if (!removeItems || isRemoveItemDisabled) return
+      removeItems(path, [index])()
+      return
+    }
```
seaerchin, line 51: nit: can combine but this is minor imo

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
dcshzj: Hmm was thinking because the migration will happen separately from our app deployment, so there can be a period of time in which we have migrated and this `isClearable` should be set to false. Meaning once we migrate, we should start to enforce that they are not able to clear this field to make sure that it is required.

Alternatively is to make this true/false be determined by the actual shape of the blob, but I'm ok with just tying it with the feature flag cos the implementation is simpler.

seaerchin (approved):
no issues on my end; just to make sure my read of this PR is correct:

1. the frontend changes are mainly for fetching data via the endpoint/using the new component
2. backend changes are to get the tags/usage counts of the category so that we can show the delete modal

for the collection tags, we are parsing via the `page.tagCategories` and if can't find, return a blank array. on this point, can we just run a quick query to see which collections:

1. are missing an index page 
2. are missing the `tagCategories` and using `tags`

i'm assuming we're migrating the `category` to the `categoryId` format down the line so no issues on my end, but remember to update the import script from github -> studio otherwise this will have correctness issues down the line where we stick to the old format.

we need to change components so that Collections layout will use the new format correctly; i did a search and our components still use `category` at present

adriangohjw:
@seaerchin okay i add as linear tickets to keep track https://linear.app/ogp/issue/ISOM-2299/pre-launch-checks

PR #2012, opened 2026-04-22
fix(studio): update page types to match new schema

seaerchin:
## Problem

The `packages/components/src/types/schema.ts` file was modified with stricter type checking, causing **29 TypeScript errors** across **12 files** in the `apps/studio` package where page objects didn't match the updated schema definitions.

Key schema changes:
- `ContentPageHeaderSchema` now requires `showThumbnail: boolean` (with default `false`)
- `LinkRefPageSchema` requires `ref` and `category` as required fields; uses `description` not `summary`
- `DatabasePagePageSchema` requires `contentPageHeader` (with `showThumbnail`) and `database` properties
- `IndexPagePageSchema` uses `ContentPagePageSchema` structure (no `title` or `lastModified` in page object)

## Solution

Updated all affected files to conform to the new schema requirements.

**Breaking Changes**

  - `contentPageHeader` now requires `showThumbnail: boolean` field
  - Database page structure changed: `title` moved to `database.title`, uses `contentPageHeader` instead of top-level properties
  - Index page structure simplified: removed `title` and `lastModified` from page object
  - Link page uses `description` instead of `summary`; `tags` removed from page object

**Improvements**:

- Core service files (`page.service.ts`, `collection.service.ts`, `collection.router.ts`) updated to generate correct page structures
- Preview components (`EditLinkPreview.tsx`, `CollectionEditorStateDrawer.tsx`, `Preview.tsx`) fixed for type narrowing issues with union types
- Test fixtures updated across `page.router.test.ts`, `resource.service.test.ts`, and `tests/integration/helpers/seed/index.ts`

## Before & After Screenshots

N/A - Type-only changes, no visual changes

## Tests

Run `npm run typecheck` to verify all TypeScript errors are resolved.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

### DIFF
--- a/apps/studio/src/features/editing-experience/components/preview/Preview.tsx
+++ b/apps/studio/src/features/editing-experience/components/preview/Preview.tsx
@@ -5,25 +5,28 @@ import { trpc } from "~/utils/trpc"
 import type { PreviewProps } from "./PreviewWithCustomSitemap"
 import PreviewWithCustomSitemap from "./PreviewWithCustomSitemap"
 
-function SuspendablePreview({
-  siteId,
-  resourceId,
-  ...rest
-}: Omit<PreviewProps, "siteMap"> & { resourceId: number }) {
+// TypeScript's Omit collapses discriminated unions. This assertion function
+// narrows `Omit<PreviewProps, "siteMap"> & Pick<PreviewProps, "siteMap">` to PreviewProps,
+// which is structurally equivalent but TypeScript can't infer automatically.
+function assertIsPreviewProps(
+  _props: Omit<PreviewProps, "siteMap"> & Pick<PreviewProps, "siteMap">,
+): asserts _props is PreviewProps {
+  // Structural equivalence - no runtime check needed
+}
+
+function SuspendablePreview(
+  props: Omit<PreviewProps, "siteMap"> & { resourceId: number },
+) {
+  const { resourceId, siteId } = props
   const [siteMap] = trpc.site.getLocalisedSitemap.useSuspenseQuery({
     siteId,
     resourceId,
   })
 
-  // Type narrowing issue: PreviewProps is a union and can't narrow from rest params
-  return (
-    <PreviewWithCustomSitemap
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      {...(rest as any)}
-      siteMap={siteMap}
-      siteId={siteId}
-    />
-  )
+  const { resourceId: _, ...rest } = props
+  const previewProps = { ...rest, siteMap }
+  assertIsPreviewProps(previewProps)
+  return <PreviewWithCustomSitemap {...previewProps} />
 }
 
 const Preview = withSuspense(
--- a/apps/studio/src/features/editing-experience/components/preview/Preview.tsx
+++ b/apps/studio/src/features/editing-experience/components/preview/Preview.tsx
@@ -15,8 +15,14 @@ function SuspendablePreview({
     resourceId,
   })
 
+  // Type narrowing issue: PreviewProps is a union and can't narrow from rest params
   return (
-    <PreviewWithCustomSitemap {...rest} siteMap={siteMap} siteId={siteId} />
+    <PreviewWithCustomSitemap
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      {...(rest as any)}
+      siteMap={siteMap}
+      siteId={siteId}
+    />
   )
 }
 
--- a/apps/studio/src/server/modules/collection/collection.router.ts
+++ b/apps/studio/src/server/modules/collection/collection.router.ts
@@ -405,7 +405,7 @@ export const collectionRouter = router({
           const blob = await updateBlobById(tx, {
             content: {
               ...content,
-              page: { description, ref, date, category, image, tags, tagged },
+              page: { description, ref, date, category, image, tagged },
             },
             pageId: linkId,
             siteId,
--- a/apps/studio/src/server/modules/collection/collection.router.ts
+++ b/apps/studio/src/server/modules/collection/collection.router.ts
@@ -358,7 +358,6 @@ export const collectionRouter = router({
           description,
           ref,
           image,
-          tags,
           tagged,
         },
         ctx,
--- a/apps/studio/src/server/modules/page/page.service.ts
+++ b/apps/studio/src/server/modules/page/page.service.ts
@@ -15,7 +15,12 @@ export const createDefaultPage = ({
     case "content": {
       const contentDefaultPage = {
         layout: ISOMER_USABLE_PAGE_LAYOUTS.Content,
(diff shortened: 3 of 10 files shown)


apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -405,7 +405,7 @@ export const collectionRouter = router({
           const blob = await updateBlobById(tx, {
             content: {
               ...content,
-              page: { description, ref, date, category, image, tags, tagged },
+              page: { description, ref, date, category, image, tagged },
```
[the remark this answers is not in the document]
seaerchin: this should no longer be required but will write a query and run against db to double check that all `tags` have a corresponding `tagged`

apps/studio/src/features/editing-experience/components/preview/Preview.tsx
```diff
@@ -5,19 +5,28 @@ import { trpc } from "~/utils/trpc"
 import type { PreviewProps } from "./PreviewWithCustomSitemap"
 import PreviewWithCustomSitemap from "./PreviewWithCustomSitemap"
 
-function SuspendablePreview({
-  siteId,
-  resourceId,
-  ...rest
-}: Omit<PreviewProps, "siteMap"> & { resourceId: number }) {
+// TypeScript's Omit collapses discriminated unions. This assertion function
```
dcshzj, line 8: suggestion(non-blocking): can consider [doing a distributive omit](https://github.com/sindresorhus/type-fest/blob/main/source/distributed-omit.d.ts) to solve this issue

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -41,10 +46,13 @@ export const createDefaultPage = ({
...
         layout: ISOMER_USABLE_PAGE_LAYOUTS.Database,
         page: {
-          title: "New database layout",
-          description:
-            "This is a layout where you can link your dataset from Data.gov.sg. Users can search through the table.",
+          contentPageHeader: {
+            summary:
+              "This is a layout where you can link your dataset from Data.gov.sg. Users can search through the table.",
+            showThumbnail: false,
+          },
```
dcshzj, line 53: nit: I think might need to run a query on our DB to see the extent of this issue in the default page causing schema invalid

dcshzj (approved):

apps/studio/src/server/modules/page/page.service.ts
seaerchin: Coming back to this — here's the query I'd run on prod to size the impact.

The new schema tightens three things relevant to the old database default page (and `ContentPageHeader` callers in general):

1. `ContentPageHeaderSchema.showThumbnail` is now required on `content`/`index`/`database` layouts.
2. Database pages now require `page.database.title` (nested) instead of the legacy `page.title`.
3. Database pages no longer carry top-level `page.title`/`page.description` (the old default fields).

**Headline query — counts by layout × role:**

```sql
WITH classified AS (
  SELECT
    b.id AS blob_id,
    b.content->>'layout' AS layout,
    CASE
      WHEN rd.id IS NOT NULL THEN 'draft'
      WHEN rp.id IS NOT NULL THEN 'currently_published'
      ELSE 'historical_version'
    END AS blob_role,
    (b.content #> '{page,contentPageHeader}') ? 'showThumbnail' AS has_show_thumbnail,
    (b.content #> '{page,contentPageHeader}') IS NOT NULL AS has_header,
    (b.content #> '{page,database,title}') IS NOT NULL AS has_db_nested_title,
    (b.content #>> '{page,title}') IS NOT NULL AS has_legacy_top_level_title,
    (b.content #>> '{page,description}') IS NOT NULL AS has_legacy_top_level_description
  FROM "Blob" b
  LEFT JOIN "Resource" rd ON rd."draftBlobId" = b.id
  LEFT JOIN "Version" v ON v."blobId" = b.id
  LEFT JOIN "Resource" rp ON rp."publishedVersionId" = v.id
)
SELECT
  layout,
  blob_role,
  COUNT(*) AS total,
  COUNT(*) FILTER (
    WHERE layout IN ('content','index','database')
      AND (NOT has_header OR NOT has_show_thumbnail)
  ) AS missing_show_thumbnail,
  COUNT(*) FILTER (
    WHERE layout = 'database' AND NOT has_db_nested_title
  ) AS db_missing_nested_title,
  COUNT(*) FILTER (
    WHERE layout = 'database' AND (has_legacy_top_level_title OR has_legacy_top_level_description)
  ) AS db_has_legacy_top_level_fields
FROM classified
WHERE layout IN ('content','index','database')
GROUP BY layout, blob_role
ORDER BY layout, blob_role;
```

**Per-site breakdown — to scope which sites need migration:**

```sql
WITH classified AS (
  SELECT
    b.id AS blob_id,
    b.content->>'layout' AS layout,
    COALESCE(rd."siteId", rp."siteId") AS site_id,
    CASE
      WHEN rd.id IS NOT NULL THEN 'draft'
      WHEN rp.id IS NOT NULL THEN 'published'
      ELSE 'historical'
    END AS blob_role,
    (b.content #> '{page,contentPageHeader}') ? 'showThumbnail' AS has_show_thumbnail,
    (b.content #> '{page,contentPageHeader}') IS NOT NULL AS has_header,
    (b.content #> '{page,database,title}') IS NOT NULL AS has_db_nested_title,
    (b.content #>> '{page,title}') IS NOT NULL AS has_legacy_title
  FROM "Blob" b
  LEFT JOIN "Resource" rd ON rd."draftBlobId" = b.id
  LEFT JOIN "Version" v ON v."blobId" = b.id
  LEFT JOIN "Resource" rp ON rp."publishedVersionId" = v.id
)
SELECT
  COALESCE(s.name, '(historical)') AS site,
  layout,
  blob_role,
  COUNT(*) AS invalid
FROM classified c
LEFT JOIN "Site" s ON s.id = c.site_id
WHERE layout IN ('content','index','database')
  AND (
    NOT has_header OR NOT has_show_thumbnail
    OR (layout = 'database' AND (NOT has_db_nested_title OR has_legacy_title))
  )
GROUP BY s.name, layout, blob_role
ORDER BY site, layout, blob_role;
```

Smoke-tested both against my local seeded DB — seed data already shows the issue (the one `content` and one `index` blob both lack `showThumbnail`), so I'd expect a non-trivial count on prod, particularly for `missing_show_thumbnail` (since that field never existed before). The `db_*` columns are the more interesting ones for sizing whether a backfill migration is needed vs. just relaxing `showThumbnail` to optional.

seaerchin: [rows requiring migration.csv](https://github.com/user-attachments/files/29187902/rows.requiring.migration.csv)

sadly there's quite abit

seaerchin:
marking as draft first since we can't merge off the bat

PR #2018, opened 2026-04-27
chore: prevent whitespace-only Hero title

dcshzj:
## Problem

The Hero component's `title` field accepts strings consisting solely of whitespace characters. A user can save a hero banner with a blank-looking title (e.g. `"   "`), which produces an empty-looking banner without any validation feedback.

## Solution

Apply the existing `NON_EMPTY_STRING_REGEX` pattern (already used by `InfoCards`) to the Hero `title` field so the schema rejects empty strings and whitespace-only input. A user-friendly `errorMessage` is added so the editor surfaces the message *"cannot be empty or contain only spaces"* instead of a raw regex failure.

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- Hero `title` field now validates against whitespace-only input across all hero variants (gradient, block, large image, floating, searchbar) since the rule is applied on the shared `HeroBaseSchema`.

**Bug Fixes**:

- Prevents users from saving a Hero block with a title containing only spaces.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] In Studio, open a page and add (or edit) a Hero block.
- [ ] In the Hero text field, enter only spaces (e.g. `"   "`) and confirm the field shows the validation error *"cannot be empty or contain only spaces"* and that saving is blocked.
- [ ] Clear the field entirely and confirm the same validation error appears.
- [ ] Enter a valid title with leading/trailing spaces (e.g. `"  Hello  "`) and confirm the field validates successfully and the hero block can be saved.
- [ ] Repeat the above for each Hero variant (gradient, block, large image, floating, searchbar) to confirm consistent validation behaviour.

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

### DIFF
--- a/packages/components/src/interfaces/complex/Hero.ts
+++ b/packages/components/src/interfaces/complex/Hero.ts
@@ -4,7 +4,7 @@ import type { IsomerSiteProps, LinkComponentType } from "~/types"
 import { Type } from "@sinclair/typebox"
 import { omit } from "lodash-es"
 import { IMAGE_ACCEPTED_MIME_TYPE_MAPPING } from "~/constants/image"
-import { LINK_HREF_PATTERN } from "~/utils/validation"
+import { LINK_HREF_PATTERN, NON_EMPTY_STRING_REGEX } from "~/utils/validation"
 
 import { ARRAY_RADIO_FORMAT } from "../format"
 import { generateImageSrcSchema } from "./Image"
@@ -23,6 +23,10 @@ const HeroBaseSchema = Type.Object({
     title: "Hero text",
     description: "The title of the hero banner",
     maxLength: 100,
+    pattern: NON_EMPTY_STRING_REGEX,
+    errorMessage: {
+      pattern: "cannot be empty or contain only spaces",
+    },
   }),
   subtitle: Type.Optional(
     Type.String({


seaerchin:

Single-file change in `packages/components/src/interfaces/complex/Hero.ts`:

- Adds `pattern: NON_EMPTY_STRING_REGEX` (`^(?=.*\\S)`) to `HeroBaseSchema.title`
- Adds an `errorMessage.pattern` of `"cannot be empty or contain only spaces"`

Because the rule is on `HeroBaseSchema`, all five Hero variants (gradient, block, large image, floating, searchbar) inherit the validation.

Reuses the existing regex already used by `InfoCards`. No new dependencies. **Stats:** +5 / -1, 1 file.

seaerchin (commented):

---

### **praise:** Right pattern, right place. **[Confidence: HIGH]**

The change is on `HeroBaseSchema`, so all five Hero variants (gradient, block, large image, floating, searchbar) inherit the rule with one edit. Reusing the existing `NON_EMPTY_STRING_REGEX` keeps validation message + behaviour consistent with `InfoCards`. This is the minimal correct change.

---

### **question:** Will already-published Hero blocks with whitespace-only titles fail validation on next save? **[Confidence: MEDIUM]**

`NON_EMPTY_STRING_REGEX` runs at validation time. Existing pages render fine, but the moment an editor opens and saves them they will be blocked from saving. Probably the intent (force a fix) but worth confirming. If a softer rollout is desired, consider a one-off backfill that nudges those titles to "Untitled".

---

### **suggestion:** Consider also trimming on save. **[Confidence: LOW]**

The regex only forbids fully whitespace strings. `"  Hello"` (leading/trailing spaces) still passes. If the design intent is "title is a non-empty string," `.trim()` (or a "remove whitespace" hint) would be a natural complement — not blocking.

---

### **note:** No tests added. **[Confidence: HIGH]**

A one-line `Value.Check`-based test confirming `"   "` fails while `"Hello"` passes would lock the behaviour in.

---

### **praise:** Clean reuse of `errorMessage`. **[Confidence: HIGH]**

Surfacing a human-readable `errorMessage.pattern` instead of a raw regex match failure is exactly the right UX move and matches `InfoCards`.

adriangohjw (approved):

PR #2019, opened 2026-04-27
feat(egazette): frontend components

seaerchin:
## Problem

This PR adds the E-Gazette management feature to Isomer Studio, enabling Toppan users to manage Government Gazettes. Toppan users need a dedicated interface to create, view, modify, and delete gazette entries with restricted access to only the gazette-related routes.

Closes ISOM-2254

## Solution

Implemented a complete E-Gazette feature with the following components:

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible

**Features**:

- **Gazette Table**: Data table displaying gazette entries with columns for notification number, title, category/subcategory, status, file ID, and publish time
- **Create Gazette Modal**: Form for adding new gazette entries with validation for title, category, subcategory, notification number, publish date/time, and file ID
- **Modify Gazette Modal**: Edit existing gazette entries with cancel publish functionality
- **View Gazette Modal**: View published gazette details with permanent deletion option (requires confirmation checkbox)
- **ToppanRouteGuard**: Route guard component that restricts Toppan users to `/sites/{siteId}/gazettes` path only

**Improvements**:

- Added `SiteMinimalLayout` template for pages without sidebar
- Made `CmsContainer` sidebar prop optional to support minimal layouts
- Added GrowthBook feature flag `is-toppan-user` for identifying Toppan users

**Bug Fixes**:

- Fixed `ToppanRouteGuard` to use `router.asPath` instead of `router.pathname` for correct URL comparison (pathname returns route pattern like `/sites/[siteId]/gazettes` instead of actual URL)
- Added GrowthBook `ready` check to prevent flash of content before feature flag resolves

## Before & After Screenshots

**AFTER**:

New Gazettes page at `/sites/{siteId}/gazettes` with:
- Dashboard layout with "Government Gazettes" header
- "Add a new Gazette" button
- Paginated data table with gazette entries
- Status indicators (Published, Scheduled, Scanning, various failure states)

## Tests

**Manual Verification Steps**:

- [ ] Navigate to `/sites/{siteId}/gazettes` and verify the gazette table renders with mock data
- [ ] Click "Add a new Gazette" and verify the create modal opens with form fields
- [ ] Verify form validation (title required, file ID must end in .pdf, etc.)
- [ ] Test route guard: If logged in as Toppan user (via feature flag), verify redirect to gazettes page from other routes
- [ ] Verify non-Toppan users can access all routes normally
[...]

PR #2021, opened 2026-04-28
feat: eGazette management - full backend integration

seaerchin:
## Problem

Toppan Next users need to manage Government Gazettes (create, modify, schedule, delete) through the Isomer Studio interface. The previous branch (`chore/egazette-frontend`) implemented the UI components with mock data. This PR connects those components to real backend APIs.

Closes 2256

## Solution

This PR implements the complete backend integration for the eGazette management feature, enabling Toppan Next users to:

- **Create gazettes** with file upload, scheduling, and metadata
- **Modify scheduled gazettes** including file replacement and rescheduling
- **Delete gazettes** before publication
- **View gazette list** with real-time data from the collection

**Breaking Changes**

**Features**:

- **File Upload Integration**: Gazettes can be uploaded as PDFs using presigned S3 URLs via `useUploadAssetMutation`
- **Create Gazette Flow**: Creates a collection link, updates metadata (category, subcategory, notification number), and schedules publication
- **Modify Gazette Flow**: Supports file replacement (upload new file or rename existing), metadata updates, and rescheduling
- **Delete Gazette**: Removes gazette resources before publication
- **Presigned GET URLs**: Generates on-demand download URLs for viewing gazette PDFs without exposing S3 paths
- **Asset Rename**: New `renameAsset` endpoint for renaming files in S3 (copy + soft delete)
- **Dynamic Subcategories**: Subcategories are fetched from backend collection tags and filtered by selected category

**Improvements**:

- **Refactored `createResourceWithBlob`**: Extracted common page/link creation logic into a shared service method in `resource.service.ts`
- **Enhanced `getCollectionTags`**: Now supports querying by `collectionId` directly (in addition to `resourceId`)
- **New `listWithBlob` endpoint**: Fetches collection items with blob content and file sizes, with gazette-specific ordering
- **Toppan Route Guard**: Sets email attribute in GrowthBook for feature flag evaluation and restricts non-Toppan users from gazette routes
- **Clickable Table Rows**: `Datatable` component now supports `onRowClick` and `isRowClickable` props
- **Subcategories Context**: `GazetteSubcategoriesProvider` manages subcategory options across create/modify modals

**Bug Fixes**:

- Fixed scheduling to allow past timestamps only for eGazette (for immediate publish)
- Prevented cancelling schedules for pages with past publish timestamps
- File upload mutation now accepts optional `fileName` parameter to override the file's original name

## Before & After Screenshots

**BEFORE**:

UI displays mock data with no backend connectivity.

**AFTER**:

Full CRUD operations with real data persistence, file uploads to S3, and scheduled publishing.

## Tests

- [ ] Log in as a Toppan Next user (email ending in `@toppannext.com`)
- [ ] Navigate to `/sites/{siteId}/gazettes` and verify the gazette list loads
- [ ] Click "Add a new Gazette" and create a gazette with PDF upload
- [ ] Verify the gazette appears in the list with "Scheduled" status
- [ ] Click a scheduled gazette row to open the modify modal
- [ ] Change the file and save - verify the new file is accessible
- [ ] Cancel a scheduled gazette and verify it is removed from the list
- [ ] Verify non-Toppan users are redirected away from gazette routes

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None


seaerchin:
* **#2021** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2021?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2027](https://github.com/opengovsg/isomer/pull/2027) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2027?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2039](https://github.com/opengovsg/isomer/pull/2039) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2039?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2021?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/studio/src/lib/s3.ts
```diff
@@ -42,6 +48,22 @@ export const generateSignedPutUrl = async ({
   )
 }
 
+export const generateSignedGetUrl = async ({
```
seaerchin, line 52: we dynamically generate a presigned url for the users so that they can see their gazettes even when it's being blocked by the AP

apps/studio/src/lib/s3.ts
```diff
@@ -73,3 +95,34 @@ export const deleteFile = async ({
...
+    const response = await storage.send(
+      new HeadObjectCommand({ Bucket, Key: Key?.slice(1) }),
+    )
+    return response.ContentLength ?? null
+  } catch {
+    return null
+  }
+}
+
+export const copyFile = async ({
```
seaerchin, line 113: `copyFile` still preserves the tags, which is required for blocking gazettes from being shown. this also works for copying an object in-place (same object -> same object)

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -311,6 +314,93 @@ export const collectionRouter = router({
           .execute()
       },
     ),
+  listWithBlob: protectedProcedure
```
seaerchin: considering shifting this method out into a dedicated `gazette.router.ts` - this is abit too specific :(

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -311,6 +314,93 @@ export const collectionRouter = router({
...
+
+      query = query
+        // 1. Status priority: Published last (8), Scheduled first (7)
+        .orderBy(
+          sql`CASE
+              WHEN "Resource"."state" = 'Published' THEN 8
+              WHEN "Resource"."scheduledAt" IS NOT NULL THEN 7
+              ELSE 9
+            END`,
+          "asc",
```
seaerchin: technically got more cases but our model isn't detailed enough to support

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -372,8 +374,21 @@ export const pageRouter = router({
...
+        .select("parentId")
+        .executeTakeFirstOrThrow()
+
+      // NOTE: check if the input.scheduledAt is after the current time
+      // we allow scheduling before the current time only for eGazette as
+      // this will trigger an immediate publish
+      if (
+        parent.parentId !== GAZETTE_COLLECTION_ID.toString() &&
+        isBefore(scheduledAt, new Date())
+      ) {
```
seaerchin: can consider removing this restriction but it's very egazette specific and idk when ppl will ever use this over publish now

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
```diff
@@ -2,33 +2,46 @@ import type { PropsWithChildren } from "react"
...
   const egazetteSiteId = useToppanSiteId()
   const router = useRouter()
 
-  const allowedBasePath = egazetteSiteId
+  // Set email attribute in GrowthBook - this triggers re-render when done
+  useEffect(() => {
+    if (gb && me.email) {
+      void gb.setAttributes({ email: me.email })
+    }
+  }, [gb, me.email])
```
dcshzj: question: we don't need this alr hor?

Anyway I think this is set also in `AvatarMenu`, actually maybe might make more sense to put here but may cause the whole app to re-render so up to you

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
```diff
@@ -2,33 +2,46 @@ import type { PropsWithChildren } from "react"
...
+
+  const isToppanUser = me.email.endsWith("@toppannext.com")
+  const isGazettePath = router.asPath.match(/^\/sites\/\d+\/gazettes/)
+  const isAllowedRoute = !isToppanUser || isGazettePath
```
dcshzj: nit: can change the const name? A bit confusing this allowed is referring to who

Actually why is `isToppanUser` involved here, I think that one is covered under the `useEffect` below?

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
```diff
@@ -2,33 +2,46 @@ import type { PropsWithChildren } from "react"
...
-  // Show spinner while GrowthBook is loading or redirecting unauthorized Toppan user
-  if (!gb?.ready || (egazetteSiteId && !isAllowedRoute)) {
+  // Show spinner while:
+  // - GrowthBook not ready
+  // - Toppan user but no siteId yet (waiting for attributes to propagate)
+  // - Redirecting unauthorized Toppan user
+  if (
+    !gb?.ready ||
+    (isToppanUser && !egazetteSiteId) ||
+    (isToppanUser && !isAllowedRoute)
```
dcshzj: suggestion: I think we can put the `isToppanUser` earlier/standalone to evaluate to false earlier for normal users?

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
```diff
@@ -38,21 +64,99 @@ const CreateGazetteModalContent = ({
...
+      })
+
+      // Step 4: Invalidate and close
+      await utils.collection.listWithBlob.invalidate()
+      toast({
+        status: "success",
+        title: "Gazette created successfully",
+        ...BRIEF_TOAST_SETTINGS,
+      })
+    } catch (error) {
```
dcshzj, line 129: issue: I think I get where you're coming from, to reuse existing logic as much as possible, but hmm I think this might be a problem if an error occurs halfway through this function (e.g. created collection page but failed to update collection link), then the site would be in an inconsistent state.

These 4 steps should be in a single procedure on the backend wrapped in a single transaction, so that anything happens we can rollback to a clean state. Wdyt?

apps/studio/src/features/gazettes/constants.ts
```diff
@@ -0,0 +1,56 @@
+export const GAZETTE_COLLECTION_ID = 30
```
dcshzj: issue: err wah I think this should be retrieved from Growthbook?

apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx
```diff
@@ -18,32 +24,52 @@ export const gazettePageSchema = z.object({
...
+
+  useEffect(() => {
+    if (!me.email.endsWith("@toppannext.com")) {
+      void router.replace("/")
+    }
+  }, [me.email, router])
+
+  if (!me.email.endsWith("@toppannext.com")) {
+    return null
+  }
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx:50]: suggestion: I think should allow core IsomerAdmins also?

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -428,26 +518,41 @@ export const collectionRouter = router({
...
-            .selectFrom("Resource")
-            .where("id", "=", String(resourceId))
-            .where("siteId", "=", siteId)
-            .select("parentId"),
-        )
-        .select("id")
-        .executeTakeFirst()
+      let indexPage
+
+      if (collectionId) {
```
dcshzj: nit: I think we can use the query builder then do an if-else just on the where `parentId` part?

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
```diff
@@ -38,21 +64,99 @@ const CreateGazetteModalContent = ({
...
+    trpc.page.schedulePage.useMutation()
+
+  const isPending = isCreating || isUpdating || isScheduling || isUploading
+
+  const onSubmit = handleSubmit(async (data) => {
+    try {
+      let fileId = data.fileId
+
+      // Step 0: Upload file if one was selected
+      if (file && fileId) {
```
dcshzj: question: I thought all gazettes would have a file? If fileId is missing, the gazette will still be created but the file is not uploaded

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
```diff
@@ -38,21 +64,99 @@ const CreateGazetteModalContent = ({
...
+      toast({
+        status: "error",
+        title: "Failed to create gazette",
+        description:
+          error instanceof Error ? error.message : "An error occurred",
+        ...BRIEF_TOAST_SETTINGS,
+      })
+    } finally {
+      onClose()
+    }
```
dcshzj: nit: should the onClose be only for the success case?

apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx
```diff
@@ -18,32 +24,52 @@ export const gazettePageSchema = z.object({
 const GazettesPage: NextPageWithLayout = () => {
   const { siteId } = useQueryParse(gazettePageSchema)
   const { isOpen, onOpen, onClose } = useDisclosure()
+  const { me } = useMe()
+  const router = useRouter()
+
+  useEffect(() => {
+    if (!me.email.endsWith("@toppannext.com")) {
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx]: nit: this email is duplicated uh I think should be a constant

apps/studio/src/lib/s3.ts
```diff
@@ -42,6 +48,22 @@ export const generateSignedPutUrl = async ({
...
+  Key,
+}: Pick<GetObjectCommandInput, "Bucket" | "Key">): Promise<string> => {
+  return getSignedUrl(
+    storage,
+    new GetObjectCommand({
+      Bucket,
+      Key,
+    }),
+    {
+      expiresIn: 60 * 60, // 1 hour
```
dcshzj: nit: maybe 5 min? Since we are refetching anyway

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
```diff
@@ -38,21 +64,99 @@ const CreateGazetteModalContent = ({
...
+        milliseconds: 0,
+      })
+      await schedulePage({
+        siteId,
+        pageId: Number(pageId),
+        scheduledAt,
+      })
+
+      // Step 4: Invalidate and close
+      await utils.collection.listWithBlob.invalidate()
```
dcshzj: nit: `void`

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
seaerchin: updated

seaerchin:

This PR wires the eGazette UI built in `chore/egazette-frontend` into real backend APIs.

**New / changed tRPC procedures**
- `collection.listWithBlob` — paginated listing of collection items with blob content + S3 file size, ordered by status (Scheduled → Published) → publish time → category → notification number
- `asset.getPresignedGetUrl` — on-demand 1-hour signed download URLs
- `asset.renameAsset` — copy-to-new-key + soft-delete-old-key in S3
- `collection.getCollectionTags` — XOR-refined to accept either `resourceId` or `collectionId`
- `page.schedulePage` — now permits past timestamps **only** if the resource's parent is `GAZETTE_COLLECTION_ID = 30`
- `page.cancelSchedulePage` — refuses to cancel a schedule whose `scheduledAt` is already in the past

**Frontend wiring**
- `CreateGazetteModal` / `ModifyGazetteModal` now perform the full create/update/cancel flows (upload → create link → update metadata → schedule → invalidate)
- `GazetteSubcategoriesProvider` uses `getCollectionTags` to drive the subcategory dropdown
- `GazetteTable` self-fetches via `listWithBlob`; row click opens modify modal (only for non-published rows)
- Page is gated client-side by `me.email.endsWith('@toppannext.com')`
- `Datatable` gained `onRowClick` / `isRowClickable`
- `useUploadAssetMutation` now accepts an optional `fileName` override

**Refactor**
- Common page+blob+permalink-conflict insert logic extracted into `createResourceWithBlob` in `resource.service.ts`; `createPage` migrated to use it
- Validates parent as `Folder | Collection` (was Folder-only); error message updated accordingly

**S3 helpers added** — `generateSignedGetUrl`, `getFileSize` (HeadObject), `copyFile` (CopyObject)

**Stats:** +1162 / -433 across 31 files

seaerchin (commented):

Reviewed against this branch's diff. Findings below use [Conventional Comments](https://conventionalcomments.org/), each with a confidence rating. Anchored line numbers omitted — comments are organised by topic.

---

### **issue (security):** Toppan-only gating is client-side only — no server-side enforcement. **[Confidence: HIGH]**

The page-level gate at `apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx` checks `me.email.endsWith("@toppannext.com")` on the client and redirects. None of the server-side procedures the page calls (`collection.listWithBlob`, `asset.getPresignedGetUrl`, `asset.renameAsset`, `collection.getCollectionTags`, `page.schedulePage`, `resource.delete`) verify the caller is a Toppan user — they only check standard site permissions.

A non-Toppan user with read/edit access to a site that happens to have collection ID 30 (or any user who passes `collectionId: 30` directly to `listWithBlob` / `getCollectionTags`) can bypass the email gate and use these endpoints. Add a server-side `assertToppanUser(ctx)` helper or a tRPC middleware and apply it to every gazette-only procedure, or feature-flag the surface server-side via GrowthBook with an email-attribute rule.

---

### **issue:** `GAZETTE_COLLECTION_ID = 30` hardcoded. **[Confidence: HIGH]**

`apps/studio/src/features/gazettes/constants.ts` exports `GAZETTE_COLLECTION_ID = 30`. Used to gate `schedulePage`'s past-timestamp branch in `page.router.ts`, to drive the gazette page query, and to build the form context. It will only resolve correctly in whichever environment has the gazette collection at row 30 — staging, dev, and per-tenant deployments will not. Move it to env / site config or look it up by a stable attribute on `Resource`.

---

### **issue:** `listWithBlob` issues N+1 S3 HEAD requests per page render. **[Confidence: HIGH]**

`collection.router.ts:listWithBlob` calls `getFileSize` (S3 HeadObject) for every result inside `Promise.all`. With `limit=25` per page that is 25 round-trips to S3 on every list query, plus the listing itself, on every pagination event and every `invalidate()` after create/modify/delete. Persist file size on the blob/resource row when the file is uploaded (or in `updateCollectionLink`) and read it from the DB.

---

### **issue:** `listWithBlob` uses `readCollectionSchema` instead of new `listWithBlobSchema`. **[Confidence: HIGH]**

You added `listWithBlobSchema = readCollectionSchema.omit({ orderBy: true })` (since the procedure ignores `orderBy` and uses its own ordering), but the procedure is still bound to `readCollectionSchema`. Either drop the unused schema or wire it in. Right now clients can pass `orderBy` and it will be silently ignored.

---

### **issue:** `renameAsset` is not transactional and the new key is never persisted server-side. **[Confidence: HIGH]**

`renameFile` in `asset.service.ts` does (1) `copyFile`, (2) `markFileAsDeleted` of source, (3) returns `newKey` to the client. Two problems: (a) the `Resource` / blob `ref` still points at the old key after this call — the client must then call `updateCollectionLink`, and a network drop in between leaves the resource referencing a soft-deleted file; (b) if `markFileAsDeleted` fails after `copyFile`, you have two live copies. Wrap the rename + DB update in a single tRPC procedure that takes the resource id, performs the copy, updates the blob `ref` in a transaction, and then schedules the soft-delete on success.

---

### **issue:** Modify flow has multiple non-atomic mutations. **[Confidence: MEDIUM]**

`ModifyGazetteModal.onSubmit` runs `uploadFile` OR `renameAsset` → `Promise.all([updateLink, updateSettings])` → `cancelSchedulePage` → `schedulePage`. Each is a separate tRPC call. Failure between any two leaves the gazette in an inconsistent state. Consider a server-side composite endpoint (`updateGazette`) that performs all of these inside a single DB transaction.

Related: cancel-then-reschedule has a visible window where the resource has `scheduledAt = null`. If a publishing worker picks up state in that window, or if the second call fails, the gazette is no longer scheduled. Reuse a server-side `reschedulePage` instead.

---

### **issue:** `cancelSchedulePage` now refuses past-time cancellations even for past-time eGazette schedules. **[Confidence: MEDIUM]**

The new check at `page.router.ts` line ~480:

```ts
if (isBefore(resource.scheduledAt, Date.now())) {
  throw new TRPCError({ ..."Unable to cancel schedule for a page that has already been scheduled to publish before the current time" })
}
```

But eGazette deliberately schedules in the past for "publish immediately." A user who schedules an eGazette and then immediately wants to cancel it has a race window where this check rejects them. Consider whether this guard should apply only to non-gazette resources.

---

### **issue:** `useEffect` redirect can flash protected content. **[Confidence: MEDIUM]**

The early `return null` mitigates the flash for the gating itself, but the underlying problem (no server enforcement) means data fetched by `GazetteSubcategoriesProvider` (rendered inside `Suspense`) and `GazetteTable` is still requested before the redirect, and anyone can bypass via DevTools. Fix the server-side guard and this becomes pure UX polish.

---

### **suggestion:** `Proxy` over `File` to inject size is fragile. **[Confidence: MEDIUM]**

`ModifyGazetteModal` builds a synthetic `File` then wraps it in a `Proxy` to expose `initialData.fileSize` as `.size`. Anything that calls a method that internally references `this.size` will hit the proxy trap with a different `prop` and may behave unexpectedly. Cleaner: thread `initialFileSize?: number` (and `initialFileName?`) explicitly into `GazetteFormFields` / `Attachment`.

---

### **suggestion:** Coupling `page.schedulePage` to `GAZETTE_COLLECTION_ID` is a code smell. **[Confidence: MEDIUM]**

Generic procedures should not import from feature-specific constants. Two cleaner options: (1) per-collection flag `allowImmediatePublish`, or (2) split into `schedulePage` (strict future-only) and `publishImmediately` for the gazette flow.

---

### **suggestion:** `enabled: false` + `refetch()` for an on-demand action is awkward. **[Confidence: MEDIUM]**

`FileIdCell` uses `useQuery` with `enabled: false` plus `refetch()` on click. Since this *generates* a new URL with side-effects (logs, expires), it is semantically a mutation. `useMutation` is a better fit.

---

### **suggestion:** `getCollectionTags` schema with XOR refinement is over-engineered. **[Confidence: LOW]**

Two separate procedures (`getCollectionTagsByCollectionId` / `…ByResourceId`) would be simpler than the XOR refinement plus branching body. Or accept just `collectionId` everywhere.

---

### **issue:** `getPresignedGetUrl` 1-hour expiry is long for a one-shot download. **[Confidence: LOW]**

The URL is opened in a new tab via `window.open`, so it lives in the user's tab history. A 5–10 minute expiry is more conventional for one-shot downloads.

---

### **nitpick:** `[...GAZETTE_CATEGORIES]` defeats the `as const` readonly. **[Confidence: LOW]**

If `SingleSelect` requires a mutable array, drop the `as const` or have the consumer accept readonly arrays.

---

### **nitpick:** Mixed semicolon style in `GazetteTable.tsx`. **[Confidence: HIGH]**

Lines 1–23 use semicolons; the rest of the file does not. Run the formatter.

---

### **nitpick:** Hardcoded magic strings — `@toppannext.com` appears in two places. **[Confidence: HIGH]**

Pull into a constant; it will be reused by the server-side guard.

---

### **chore:** Empty trailing-space test diffs in `page.router.test.ts`. **[Confidence: HIGH]**

Several `it.skip(... () => {})` got rewritten to `it.skip(... () => { })` (added space) — looks like an editor auto-format. Revert if possible.

---

### **chore:** Double-space typo in `renameFile` comment. **[Confidence: HIGH]**

`asset.service.ts`: `// NOTE: This will  also preserve existing tags for unlocked objects`. Also: the claim is correct only because the v3 SDK defaults to `TaggingDirective: COPY`; consider stating that explicitly.

---

### **note:** No automated tests added for the new endpoints. **[Confidence: HIGH]**

No tests for `collection.listWithBlob`, `asset.getPresignedGetUrl`, `asset.renameAsset`, or the new past-timestamp branch in `schedulePage`. Please add at least permission-failure and happy-path coverage, plus a test pinning the past-timestamp gazette branch.

---

### **praise:** Good extraction of `createResourceWithBlob`. **[Confidence: HIGH]**

Lifting the validate-folder/insert-blob/insert-resource/handle-permalink-conflict logic into a shared helper is a clean refactor that broadens the parent type from `Folder` to `Folder | Collection`, and the error message change accurately reflects that.

seaerchin (commented):
**THIS IS A AUTOMATED REVIEW BY CLAUDE — re-review of commits 44c4fdf21 and 1466db393**

The two new commits address most of the structural and security concerns from the prior review. Summary of what landed and what's still open below.

---

## What got resolved (praise)

| Prior issue | Fix |
|---|---|
| Toppan gating client-side only | New `assertGazetteAccess(userId)` in `gazette.service.ts` enforces Toppan email or `IsomerAdminRole.Core`, called at the top of every mutation/query in `gazette.router.ts`. |
| `GAZETTE_COLLECTION_ID = 30` hardcoded | Replaced with `useEgazetteInfo` reading `EGAZETTE_INFO_FEATURE_KEY` from GrowthBook (`siteId`, `gazettesCollectionId`). |
| N+1 S3 HEAD calls in `listWithBlob` | `fileSize` is persisted in the blob at create/update; list reads it from the blob. Comment clearly notes "Pre-fileSize gazettes return null and the UI degrades gracefully." |
| `listWithBlob` schema mismatch | `listWithBlob` removed entirely — replaced by `gazette.list` in dedicated `gazette.router.ts`. |
| `renameAsset` non-atomic | Replaced by a server-driven `gazette.update` flow that rebuilds the blob + updates the resource in a single DB transaction. |
| Modify flow non-atomic on the client | Client now calls a single `gazette.update` mutation that bundles ref/title/scheduledAt updates server-side. |
| Cancel-then-reschedule window | Eliminated — gazette flow no longer cancels-then-reschedules. |
| `page.schedulePage` coupled to `GAZETTE_COLLECTION_ID` | Past-timestamp branch removed from `page.router`; gazette uses its own router. |
| `getPresignedGetUrl` semantics | Now a `mutation` (was `query` with `enabled:false`). Comment justifies it. |
| `Proxy` over `File` | Replaced with `Object.defineProperty` in `buildPlaceholderFile` — keeps `this`-binding correct. |
| `@toppannext.com` magic string | Extracted to `~/constants/toppan.ts` `TOPPAN_EMAIL_DOMAIN`. |
| `renameFile` comment | Now explicit that v3 SDK defaults to `TaggingDirective: COPY`. |

Cleanly extracting the gazette feature into its own router with its own access guard is a substantial improvement.

---

## Remaining / new findings

### **issue:** `gazette.update` rename happens *before* the DB transaction — same orphaning problem in a new shape. **[Confidence: HIGH]**

`gazette.router.ts:update` (the rename-existing-file branch):

```ts
} else if (desiredFileName && existingRef) {
  const currentFileName = existingRef.split("/").pop()
  if (currentFileName !== desiredFileName) {
    const sourceKey = existingRef.replace(/^\//, "")
    const newKey = await renameFile({ sourceKey, newFileName: desiredFileName })
    finalRef = `/${newKey}`
    // renameFile already soft-deletes the old key on success, so we
    // don't need to clean it up again.
  }
}

// ... db.transaction().execute(...)
```

`renameFile` does (1) `copyFile` and (2) `markFileAsDeleted(sourceKey)` *before returning*. If the subsequent DB transaction fails, the resource still references the old ref but the old file is now soft-deleted (and will be reaped by the lifecycle policy). The `newRef` branch correctly defers cleanup via `oldRefToCleanUp` and only soft-deletes after the tx commits — please apply the same pattern here. Concretely: either (a) split `renameFile` into a copy-only helper that returns `newKey` and have the router soft-delete the old key in the same `try/catch` block as the `newRef` cleanup, or (b) move the rename inside the transaction (S3 then DB inside one logical unit, with an outer rollback that re-deletes the new key if the tx fails).

The accompanying comment ("the real consistency boundary is the tx below where we write the resolved ref") understates this — the soft-delete in `renameFile` is *outside* that boundary.

---

### **issue:** `gazette.update` silently ignores past `scheduledAt` (asymmetric with create). **[Confidence: MEDIUM]**

```ts
const newScheduledAt = isBefore(scheduledAt, new Date())
  ? existingResource.scheduledAt
  : scheduledAt
```

`gazette.create` allows past `scheduledAt` (treats it as immediate publish, skips the email). `gazette.update` *silently keeps the old scheduledAt* if the user picks a past time — no error, no warning, the user thinks they rescheduled but didn't. Pick one contract:

- Reject with a clear `BAD_REQUEST` ("scheduled time is in the past"),
- Or honour past = immediate publish on update too (mirror create), or
- Document explicitly that update can only set future times.

Whatever you choose, surface it to the client.

---

### **issue:** `fileSize` can be silently lost on update if the caller omits it. **[Confidence: MEDIUM]**

`gazetteMetadataSchema.fileSize` is `.optional()`. `buildGazetteBlobContent` rebuilds the entire `page` object from inputs. The frontend currently passes `fileSize = initialData.fileSize` as a defensive default, but any other caller (or a future refactor) that omits `fileSize` will overwrite the existing blob's `fileSize` with `undefined` — and the next `gazette.list` will return `null` for that row, regressing the UX.

Inside the `db.transaction(...)` in `update`, fetch the existing blob's `fileSize` and fall back to it when input is missing. The blob rebuild should preserve unchanged fields by default.

---

### **issue:** `useEgazetteInfo` returns `""` defaults — silent broken state when the flag isn't set. **[Confidence: MEDIUM]**

```ts
useFeatureValue<EgazetteInfo>(EGAZETTE_INFO_FEATURE_KEY, {
  siteId: "",
  gazettesCollectionId: "",
})
```

If the GrowthBook flag is missing or misconfigured for the current user, `Number("") = 0`. Two consequences:

1. `gazette.list` is called with `collectionId: 0`, which fails Zod (`min(1)`) and produces an opaque error in the table.
2. `ToppanRouteGuard` falls through (since `gazettesPath` is falsy), so Toppan users without the flag can browse the rest of the app — probably not what was intended.

Add an explicit "feature not configured" state to `useEgazetteInfo` (e.g. return `null` as the default and have callers branch on it), or surface a clean "Gazettes not enabled for your site — contact support" page.

---

### **issue:** `ToppanRouteGuard` traps Toppan users on the gazettes page. **[Confidence: HIGH — assuming this is intentional, please confirm]**

```ts
const isAllowedRoute = isToppanUser && router.asPath === gazettesPath
useEffect(() => {
  if (isToppanUser && !isAllowedRoute && gazettesPath) {
    void router.replace(gazettesPath)
  }
}, ...)
```

For a Toppan user with a configured `gazettesPath`, *any* navigation away from `/sites/{siteId}/gazettes` immediately redirects back. They cannot visit the dashboard, account, support, or any other route. The PR description says "restricts non-Toppan users from gazette routes" — the implementation actually restricts Toppan users *to* the gazettes route, which is the inverse phrasing. Two possible reads:

1. **Intentional:** Toppan is a tenant-only role and gazettes is the only valid surface for them. Then please make the description match, and consider: how does a Toppan user log out? Update their profile? Get redirected to a Toppan-friendly login on session expiry?
2. **Unintentional:** the route check is too strict and should also allow `/auth/*`, `/me/*`, etc.

Either way, this needs a test that pins the expected behaviour for both an allowed and a non-allowed Toppan navigation.

---

### **issue:** No tests for `gazette.router` or `assertGazetteAccess`. **[Confidence: HIGH]**

The new router is the access-control boundary for the entire feature. There's nothing covering: (a) Toppan email passes, (b) Core admin passes, (c) ordinary user with site write rejected, (d) past-date create allowed, (e) past-date update behaviour (whatever you settle on above), (f) rename-failure cleanup. A new feature with non-trivial security and ordering invariants and zero integration tests is the right place to draw a line.

---

### **suggestion:** The `asset.renameAsset` tRPC procedure is now dead code. **[Confidence: HIGH]**

`grep` shows no callers — the gazette router calls `renameFile` (the service) directly. Either remove the public procedure or document its intended future use. Keeping unused write endpoints around expands the attack surface for not much benefit.

---

### **suggestion:** `assertGazetteAccess` runs an extra `User.findOne` per call. **[Confidence: LOW]**

Every gazette mutation now does an extra `User` lookup before its main work. Fine for now, but consider attaching `email` (and `isomerAdminRole`) to `ctx.user` once at session-creation time. Avoids per-request DB round-trip for what is effectively session metadata.

---

### **suggestion:** `assertGazetteAccess` returning `UNAUTHORIZED` on missing user is unreachable. **[Confidence: LOW]**

`protectedProcedure` already requires a valid session. If the User row is somehow missing, that's a server-state inconsistency, not unauthorized. `INTERNAL_SERVER_ERROR` is more accurate.

---

### **nitpick:** `gazette.update` rewrites the blob even when nothing changed. **[Confidence: LOW]**

A no-op save re-writes the entire blob and emits an `AuditLogEvent.ResourceUpdate`. Not catastrophic — but if "save" is hit defensively, audit logs will fill with empty deltas. Consider short-circuiting when the rebuilt blob equals the existing one.

---

### **note:** Two log calls and two audit-log entries fire even when only metadata changed. **[Confidence: LOW]**

In `update`, you log both `ResourceUpdate` and conditionally `SchedulePublish`. The conditional is correct (only when `scheduledAtChanged`), but consider whether `ResourceUpdate` should also short-circuit on a no-op. Tied to the previous nit.

---

### **praise:** Clean separation of `gazette.router.ts` and `gazette.service.ts`. **[Confidence: HIGH]**

The router is thin and reads top-to-bottom; the service holds `assertGazetteAccess` and could grow more guards over time. Comments explaining "the real consistency boundary is the tx below" and "Skip the email when scheduledAt is in the past" are exactly the kind of breadcrumbs reviewers/future-you will want.

---

### **praise:** `buildPlaceholderFile` via `Object.defineProperty` is the right replacement for the Proxy. **[Confidence: HIGH]**

Mutating the real File rather than wrapping it preserves `this`-binding for built-in methods. The comment explaining *why* this beats a Proxy is great.

seaerchin (commented):
**THIS IS A AUTOMATED REVIEW BY CLAUDE — review of commit cc158e79**

Targeted follow-up commit on top of the prior re-review. Scope is small, focused, and addresses the open issues directly.

---

## What landed (praise)

| Prior issue (HIGH/MEDIUM) | Resolution in this commit |
|---|---|
| `gazette.update` rename happens before DB tx | Split `renameFile` into `copyFileWithNewName` (copy only) and routed the soft-delete through the existing `oldRefToCleanUp` post-tx path. The two cleanup branches (`newRef` and `desiredFileName`) now share the same compensating-action shape. Comment in the router calls out "S3 copy now, soft-delete old AFTER the DB commit so a tx rollback never strands the resource pointing at a tombstoned key." |
| `gazette.update` silently ignores past `scheduledAt` | Now uses the input directly and lets create's "skip email when past" rule apply on update too — comment explicitly says "We don't silently rewrite the value the caller sent." |
| `fileSize` overwritten with `undefined` on partial updates | `resolvedFileSize = fileSize ?? existingPage?.fileSize` reads the existing blob's value when omitted. Tested. |
| `useEgazetteInfo` empty-string defaults | Replaced with a discriminated union (`ConfiguredEgazetteInfo | UnconfiguredEgazetteInfo`); the page renders an explicit "Gazettes are not enabled for this account" state instead of firing a `collectionId: 0` query. |
| `ToppanRouteGuard` traps Toppan users | Falls through when `!isConfigured`, and `startsWith` lets child paths under `/sites/{id}/gazettes/*` work. |
| `asset.renameAsset` dead code | Procedure + Zod schema removed. |
| `assertGazetteAccess` returns `UNAUTHORIZED` on missing user | Now `INTERNAL_SERVER_ERROR`, with comment referencing `protectedProcedure`. |
| No tests for the gazette router | New `gazette.router.test.ts` with 8 cases covering Toppan/Core/regular access, past-date create, and fileSize preservation on partial update. |

Nice clean execution — every resolution carries a commit-message bullet, every behaviour change has an explanatory comment, every behaviour change has a test.

---

## New / residual findings

### **issue:** `ToppanRouteGuard.startsWith(gazettesPath)` matches sibling-prefix paths. **[Confidence: MEDIUM]**

```ts
const isAllowedRoute =
  !shouldPin || (gazettesPath !== null && router.asPath.startsWith(gazettesPath))
```

With `gazettesPath = "/sites/1/gazettes"`, `startsWith` is `true` for `/sites/1/gazettes`, `/sites/1/gazettes/foo`, **and** `/sites/1/gazettes-foo` / `/sites/1/gazettesfoo`. There's no route at the latter today so this is harmless right now, but the moment someone adds e.g. `/sites/{id}/gazettesArchive` the guard will silently let Toppan users land there. Tighten to:

```ts
const isAllowedRoute =
  !shouldPin || router.asPath === gazettesPath ||
  router.asPath.startsWith(`${gazettesPath}/`) ||
  router.asPath.startsWith(`${gazettesPath}?`)
```

(Or use `URL` parsing if you'd rather not hand-roll the boundary.)

---

### **issue:** No test for the `copyFileWithNewName` + post-tx cleanup ordering. **[Confidence: MEDIUM]**

The whole point of splitting `renameFile` is to ensure that when the DB transaction fails, the old key is *not* soft-deleted. The current test suite covers fileSize preservation and past-date acceptance but not the rename-failure invariant. Worth one test that mocks `copyFileWithNewName` to succeed and `updateBlobById` (or `updatePageById`) to throw, then asserts `markFileAsDeleted` was *not* called for the source key. Without this, the regression you just fixed has nothing pinning it.

---

### **suggestion:** No coverage for the `desiredFileName` happy path either. **[Confidence: LOW]**

Tests for `update` cover (a) fileSize preservation when fileSize is omitted, and (b) past-date update. They don't exercise the branch that sends `desiredFileName` and verifies `copyFileWithNewName` was called and the old key got soft-deleted *after* the tx. One more case would close the gap.

---

### **suggestion:** Test for `IsomerAdmin` with non-Core role would close the access matrix. **[Confidence: LOW]**

The 4 access tests are: Toppan → allow, Core admin → allow, ordinary → reject, unauth → reject. A `Migrator` (or any non-Core admin role) caller — to confirm the role list is enforced rather than treated as "any admin role" — would round it out.

---

### **nitpick:** Update mutation still rebuilds and writes the blob on no-op saves. **[Confidence: LOW]**

Still flagged from the previous round; not addressed by this commit, and not a regression. Mentioned again only because the rebuild now includes a fileSize fallback fetch, so a no-op save costs slightly more than before.

---

### **nitpick:** Comment in `update` says "Any S3 failure here aborts before the DB transaction starts" — now slightly imprecise. **[Confidence: LOW]**

That sentence predates the split; it's still true (the `copyFileWithNewName` call is still pre-tx), but the *meaning* has shifted because the soft-delete is no longer in there. Worth a one-line edit so future-you doesn't hunt for a delete that isn't there.

---

### **praise:** The discriminated-union shape on `useEgazetteInfo` is the right move.

`isConfigured: true | false` lets every consumer narrow once and stop sprinkling `Number("") || 0` defensive coercions. The "not enabled" UI in the page is also nicely paired — clean degradation with a clear support hook.

---

### **praise:** Lifting `gazettesCollectionId` to a prop on `CreateGazetteModal` / `ModifyGazetteModal` / `GazetteTable` / `GazetteSubcategoriesProvider`.

Calling `useEgazetteInfo` once at the page level and threading the value through is much easier to reason about than each component reading the same flag independently. Bonus: the components are now testable without mocking GrowthBook.

seaerchin (commented):
**THIS IS A AUTOMATED REVIEW BY CLAUDE — re-review of commit(s) since cc158e79**

## Scope

Single commit `4a89ab47b` — a deliberate, narrow revert of one decision from cc158e79: `fileSize` is no longer persisted on the gazette blob. Specifically:

- `gazetteMetadataSchema.fileSize` field removed from `apps/studio/src/schemas/gazette.ts`.
- `GazetteBlobInputs.fileSize` and the `fileSize` write-through removed from `buildGazetteBlobContent`, `create`, and `update` in `apps/studio/src/server/modules/gazette/gazette.router.ts`. The `existingPage` lookup that backfilled `fileSize` on partial update is also gone.
- `gazette.list` now `Promise.all`s an S3 `HeadObject` (`getFileSize` from `~/lib/s3`) per result row, mapping `{ ...result, fileSize }` (or `null` when no `ref`).
- Client callers (`CreateGazetteModal`, `ModifyGazetteModal`) no longer thread `file.size` / `initialData.fileSize` into the mutation input.
- Test updates in `gazette.router.test.ts`: the create test no longer sends/asserts `fileSize`; the "preserves fileSize on partial update" test is replaced with a "rewrites blob metadata + resource title" test exercising `newRef`, `category`, `description`, `tagged`.

The intent (from the commit message + new comment) is to keep the `BlobJsonContent` contract free of feature-specific fields. Trade-off accepted: ~25 S3 HEADs per list page in exchange for a clean shared contract.

## Findings

### **issue:** Re-introduces the N+1 S3 HEAD problem flagged in the very first review. (confidence: high)

The first Claude review explicitly called out: *"`listWithBlob` issues N+1 S3 HEAD requests per page render … 25 round-trips to S3 on every list query, plus the listing itself, on every pagination event and every `invalidate()` after create/modify/delete."* That issue was fixed in `1466db393` by persisting `fileSize` on the blob. This commit reverts the fix.

The new comment argues the trade-off is fine because S3 HEAD scales to thousands of QPS and the page is bounded to ~25. Latency is the bigger concern than QPS though: list latency is now `tableQuery + max(headLatency_i)` (Promise.all is bounded by the slowest HEAD, typically 30–100ms p99 from EC2/ECS to S3 in-region, much worse from cold lambdas / cross-region). Every `utils.gazette.list.invalidate()` after create/update/delete pays this cost. With the prior persistence approach, list was a single SQL round-trip.

If keeping `fileSize` off the blob is non-negotiable on contract grounds, two cleaner options:
1. Add a dedicated column on `Resource` or a sibling table for "file metadata" — outside `BlobJsonContent`, still avoids S3 on list.
2. Cache the HEAD response (per `ref`, immutable on a given key) at the request or process level — `ref` is content-addressed within the rename flow.

If the contract argument *does* win, please at least time the new list latency against the old one before merging.

### **issue:** Test coverage for `fileSize` persistence is now gone, with no equivalent for the S3 HEAD path. (confidence: medium)

The test "preserves the existing fileSize when the caller omits it" is replaced (not added-to) by an unrelated "rewrites blob metadata" test. There is now zero coverage for:
1. `gazette.list` returning the correct `fileSize` for rows whose object exists in S3.
2. `gazette.list` returning `fileSize: null` when (a) `ref` is missing on the blob, or (b) `getFileSize` returns null (HEAD failed / object missing).
3. The Promise.all is well-behaved when one of N HEADs returns null — `getFileSize` swallows errors via `catch { return null }`, but a test pinning that contract would prevent a future change to that swallow from silently breaking list rendering.

Given the previous reviewer-flagged sensitivity around list behaviour, please add at least case (2a) (cheap — no S3 mock needed).

### **suggestion:** `getFileSize` swallowing all errors makes ops invisible to the list endpoint. (confidence: low)

`s3.ts:99` does `catch { return null }`. With this revert, the only signal that S3 access is broken is "every gazette shows blank file size" in the UI — no logs, no metrics. Consider `console.warn` or a counter at minimum, or distinguishing 404 (legitimately missing — return null) from 5xx / network errors (worth surfacing).

### **suggestion:** The new comment in `buildGazetteBlobContent` slightly overstates the case. (confidence: low)

> S3 HEAD scales to thousands of QPS, so the N HEADs are fine

QPS is rarely the bottleneck for an internal CMS — latency, $0.0004/1000 HEAD billing on hot pages, and tail-latency amplification on Promise.all are. Worth softening the comment so a future reader doesn't read it as "this is unconditionally fine."

### **praise:** The revert itself is mechanically clean. (confidence: high)

The diff is consistent: schema field, blob builder input, router writes (create + update), the partial-update preservation logic, client mutation inputs, and the test all line up. No dangling `fileSize` references. The `getFileSize({ Key: ref })` call passes `/1/abc/notice.pdf`-style refs into a function whose first action is `Key?.slice(1)` to strip the leading slash — that contract is preserved by the existing client-side ref convention. The Promise.all correctly preserves row order (since `.map` is index-stable).

## Verdict

The revert is mechanically clean but reintroduces a high-confidence performance regression flagged in the original review and fixed in `1466db393`. If the `BlobJsonContent` contract concern is the real driver, please consider a non-blob persisted column instead of going back to per-row HEAD. Otherwise, document the latency trade-off explicitly and add the missing list-path test.

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
```diff
@@ -1,37 +1,39 @@
...
+  // so child paths and query strings under /sites/{siteId}/gazettes/* don't
+  // bounce. Non-Toppan users and Toppan users with no configured egazette
+  // flag are not pinned.
+  const shouldPin = isToppanUser && egazette.isConfigured
+  const gazettesPath = egazette.isConfigured
+    ? `/sites/${egazette.siteId}/gazettes`
+    : null
+  const isAllowedRoute =
+    !shouldPin ||
+    (gazettesPath !== null && router.asPath.startsWith(gazettesPath))
```
dcshzj: issue(non-blocking): I think I might have commented this earlier but the naming of `shouldPin` and `isAllowedRoute` is a bit confusing and misleading. Perhaps something like:
- `isLockToEgazetteRequired` instead of `shouldPin`
- `isEgazetteRoute` instead of `isAllowedRoute`

Then the useEffect conditional below can be understood easier as `isLockToEgazetteRequired && !isEgazetteRoute`

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
```diff
@@ -1,37 +1,39 @@
...
-    // as they should only have access there
-    if (isToppanUser && !isAllowedRoute && gazettesPath) {
+    if (shouldPin && !isAllowedRoute && gazettesPath) {
       void router.replace(gazettesPath)
     }
-  }, [isAllowedRoute, router, isToppanUser, gazettesPath, siteId])
+  }, [shouldPin, isAllowedRoute, router, gazettesPath])
 
-  if (!isToppanUser || (isToppanUser && isAllowedRoute)) {
+  if (isAllowedRoute) {
```
dcshzj: (continued) then for this, it would be `!isLockToEgazetteRequired || isEgazetteRoute`, something like that

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
```diff
@@ -7,52 +7,140 @@ import {
...
+    if (!file) {
+      toast({
+        status: "error",
+        title: "Please attach a PDF before submitting",
+        ...BRIEF_TOAST_SETTINGS,
+      })
+      return
+    }
+
+    const [hours, minutes] = data.publishTime.split(":").map(Number)
```
dcshzj: nit: should we rely on `date-fns` to perform the parsing instead?

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
```diff
@@ -7,52 +7,140 @@ import {
...
+    const scheduledAt = set(data.publishDate, {
+      hours,
+      minutes,
+      seconds: 0,
+      milliseconds: 0,
+    })
+
+    try {
+      const { path: ref } = await uploadFile({ file, fileName: data.fileId })
```
dcshzj, line 107: question(non-blocking): confirming here that the upload is idempotent?

Would it also be possible if Toppan for some reason uses the same `fileId` from an existing gazette, then that gets overwritten?

apps/studio/src/features/gazettes/components/ModifyGazetteModal/ModifyGazetteModal.tsx
```diff
@@ -88,21 +123,91 @@ const ModifyGazetteModalContent = ({
...
-  const onCancelPublish = () => {
-    // TODO: Implement API call to cancel gazette publication
-    console.log("Cancelling gazette publication:", gazetteId)
-    onClose()
+  const onCancelPublish = async () => {
+    try {
+      await deleteResource({
+        resourceId: String(gazetteId),
+        siteId,
+      })
```
dcshzj, line 184: nit: can add a comment on why we choose to delete for the publish cancel? Might not be clear on initial glance cos I would think the resource should remain present but there is no schedule at date.

apps/studio/src/features/gazettes/constants.ts
```diff
@@ -0,0 +1,54 @@
+export const GAZETTE_CATEGORIES = [
```
dcshzj: nit: should the type be `{ label: GazettesCategory; value: string }[]` or something?

apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx
```diff
@@ -18,32 +27,86 @@ export const gazettePageSchema = z.object({
 const GazettesPage: NextPageWithLayout = () => {
   const { siteId } = useQueryParse(gazettePageSchema)
   const { isOpen, onOpen, onClose } = useDisclosure()
+  const { me } = useMe()
+  const router = useRouter()
+  const { isAdmin: isCoreAdmin, isLoading: isAdminCheckLoading } =
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx]: nit

```suggestion
  const { isAdmin: isIsomerAdmin, isLoading: isAdminCheckLoading } =
```

apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx
```diff
@@ -18,32 +27,86 @@ export const gazettePageSchema = z.object({
...
+  // rejects with an opaque error).
+  if (!egazette.isConfigured) {
+    return (
+      <Center minH="50vh">
+        <VStack spacing="0.5rem">
+          <Text textStyle="subhead-1" color="base.content.strong">
+            Gazettes are not enabled for this account
+          </Text>
+          <Text textStyle="body-2" color="base.content.medium">
+            Please contact support if you believe this is incorrect.
```
Comment (dcshzj) [inline apps/studio/src/pages/sites/[siteId]/gazettes/index.tsx]: nit: can link to support@isomer.gov.sg, think got some other place that does this

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -0,0 +1,462 @@
...
+      // bounded to ~25 rows per request and S3 HEAD scales to thousands of
+      // QPS, so the N HEADs are fine — keeping the size out of the blob
+      // preserves the PrismaJson.BlobJsonContent contract with the
+      // components package.
+      return Promise.all(
+        results.map(async (result) => {
+          const ref = (result.content as { page?: { ref?: string } } | null)
+            ?.page?.ref
+          if (!ref) {
+            return { ...result, fileSize: null }
```
dcshzj: issue: should this contain `scheduledAt: result.scheduledAt ?? result.publishedAt` also?

apps/studio/src/lib/s3.ts
```diff
@@ -75,3 +97,34 @@ export const deleteFile = async ({
...
   )
 }
+
+export const getFileSize = async ({
+  Key,
+  Bucket,
+}: Pick<HeadObjectCommandInput, "Key" | "Bucket">): Promise<number | null> => {
+  try {
+    const response = await storage.send(
+      new HeadObjectCommand({ Bucket, Key: Key?.slice(1) }),
```
dcshzj: nit: should we let the caller do the `slice`? Also to make it more explicit that we are stripping the leading `/` if it exists and not just any first character

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -0,0 +1,462 @@
...
+          "asc",
+        )
+        // 2. Publish date descending
+        .orderBy("Resource.scheduledAt", (ob) => ob.desc().nullsLast())
+        // 3. Category priority from blob content
+        .orderBy(
+          sql`CASE
+              WHEN COALESCE("DraftBlob"."content", "PublishedBlob"."content")->'page'->>'category' = 'Government Gazette' THEN 1
+              WHEN COALESCE("DraftBlob"."content", "PublishedBlob"."content")->'page'->>'category' = 'Legislation Supplements' THEN 2
+              WHEN COALESCE("DraftBlob"."content", "PublishedBlob"."content")->'page'->>'category' = 'Other Supplements' THEN 3
```
dcshzj: issue(non-blocking): this part is a bit hard-coded, should we import it from the constants that are already present for this?

apps/studio/src/features/gazettes/constants.ts
```diff
@@ -0,0 +1,54 @@
...
+  BANKRUPTCY_ACT_NOTICE: "Bankruptcy Act Notice",
+  COMPANIES_ACT_NOTICE: "Companies Act Notice",
+  OTHERS: "Others",
+  ADVERTISEMENTS: "Advertisements",
+  TENDERS: "Tenders",
+  REVOCATION: "Revocation",
+  AUDITED_REPORTS: "Audited Reports",
+} as const
+
+export const governmentGazetteSubcategoriesKeys = Object.values(
```
dcshzj, line 29: minor nit: these are Values right not Keys

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
```diff
@@ -8,26 +13,58 @@ import {
...
+// Build a synthetic File with the right name + size for display purposes.
+// `defineProperty` is used instead of a `Proxy` because it mutates the real
+// File rather than wrapping it, so any built-in method (e.g. `.slice()`) that
+// reads `this.size` still resolves correctly without proxy method-rebinding.
+const buildPlaceholderFile = (name: string, size?: number): File => {
+  const file = new File([], name, { type: "application/pdf" })
+  if (size !== undefined) {
+    Object.defineProperty(file, "size", { value: size, writable: false })
+  }
+  return file
```
dcshzj, line 49: question(non-blocking): why not do `new File([new ArrayBuffer(size)], name, { type: "application/pdf" })`

(this one AI suggest one so double check if it works)

apps/studio/src/components/ToppanRouteGuard/ToppanRouteGuard.tsx
seaerchin: suggestion: honestly the naming suggested also quite confusing, i'm gonna go with `shouldRestrictToGazettesRoute` and `isGazettesRoute` - let me know if any issues

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
seaerchin: it's the same upload mutation hook - in the backend, we generate a random UUID even for the same fileId.

apps/studio/src/features/gazettes/components/ModifyGazetteModal/ModifyGazetteModal.tsx
seaerchin: sounds good, to be clear - i did this for parity with the exsting egazettes behaviour where it deletes the row entry. will also put this down in code

apps/studio/src/features/gazettes/components/CreateGazetteModal/CreateGazetteModal.tsx
seaerchin: done, edited for modify gaztte modal too

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: this is technically impossible since it would mean that they bypassed the modal - but still going to add regardless

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: updated to use kysely case statement, credits to claude

apps/studio/src/features/gazettes/constants.ts
seaerchin: they are keys in the single select

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
seaerchin: this would actually allocate memory isn't it? on our dashboard, since we limit to 25, this would allocate, at an estimate, 25 * 4 = 100 mb of memory just to show the sizes :monkas:

dcshzj: Makes sense!

dcshzj (approved):

seaerchin:
### Merge activity

* **May 12, 3:35 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2021).
* **May 12, 3:37 AM UTC**: [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2021) rebased this pull request as part of a merge.
* **May 12, 3:45 AM UTC**: @seaerchin merged this pull request with [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2021).

PR #2024, opened 2026-04-30
chore(studio): replace BlobJsonContent double cast with satisfies in MSW handler

seaerchin:
The readCollectionLink MSW handlers used `as unknown as PrismaJson.BlobJsonContent`
with a TODO noting confusion about the typing issue. Production code (e.g.
createCollectionLinkJson in collection.service.ts) constructs the same shape using
`satisfies UnwrapTagged<PrismaJson.BlobJsonContent>`, which is the established
pattern. Switch the handlers to match — same runtime behaviour, but the type is
now actually verified instead of silently bypassed.

https://claude.ai/code/session_01XZ1sf9frZjrLYCxxT5EsCU

### DIFF
--- a/apps/studio/tests/msw/handlers/collection.ts
+++ b/apps/studio/tests/msw/handlers/collection.ts
@@ -1,3 +1,4 @@
+import type { UnwrapTagged } from "type-fest"
 import { ResourceState, ResourceType } from "~prisma/generated/generatedEnums"
 
 import { MOCK_STORY_DATE } from "../constants"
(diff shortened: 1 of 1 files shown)


seaerchin (commented):

## Diff summary

Tiny, surgical type-level cleanup in MSW handlers.

**Code flow**
- [`apps/studio/tests/msw/handlers/collection.ts`](https://github.com/opengovsg/isomer/blob/claude/collect-todo-comments-jXVqu/apps/studio/tests/msw/handlers/collection.ts): two MSW handler factories that build mock `BlobJsonContent` payloads.
- Before: `} as unknown as PrismaJson.BlobJsonContent,` with a "not too sure why this error happens" TODO above it. The double cast silently bypassed the opaque-type check.
- After: `} satisfies UnwrapTagged<PrismaJson.BlobJsonContent>,` matching the established pattern in [`apps/studio/src/server/modules/collection/collection.service.ts`](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/server/modules/collection/collection.service.ts) (lines 24 and 41 — `createCollectionLinkJson`).

**Net effect**: no runtime behaviour change, two TODOs removed, type now actually verified.

---

## Review

**praise:** Aligns the test mocks with the production pattern and reclaims compile-time verification that the prior `as unknown as` cast threw away. Net reduction in tech-debt comments. (confidence: high)

LGTM — nothing to flag.

PR #2027, opened 2026-05-04
feat: add gazette management system with SearchSG integration and S3 compliance lock

seaerchin:
## Problem

Government gazettes need to be managed, scheduled for publication, and indexed in SearchSG for public search. Currently there is no system to handle gazette-specific workflows including PDF uploads, scheduled publishing, and automatic ingestion into the SearchSG index. Additionally, published gazettes must be protected from deletion to comply with regulatory requirements.

Closes ISOM-2248

## Solution

Implemented a complete gazette management system with the following capabilities:

1. **New Gazette UI** - Full CRUD interface for managing gazette documents including create, view, modify, and cancel modals with category/subcategory tagging support

2. **Scheduled Publishing** - Gazettes can be scheduled for future publication with automatic processing via a cron job

3. **SearchSG Integration** - A new `PushDocumentJob` system that:
   - Queues documents for ingestion into SearchSG at scheduled times
   - Extracts full text from PDFs for search indexing
   - Derives categories and subcategories from collection tag mappings
   - Removes S3 access restrictions when documents are published

4. **S3 Compliance Lock** - Published gazette PDFs are protected with S3 Object Lock in COMPLIANCE mode:
   - Sets retention for 10,000 days (~27 years) from publication date
   - Ensures documents cannot be deleted even by administrators
   - Matches existing eGazette system behavior

5. **S3 Enhancements** - New utilities for file operations including:
   - `getBlob` - Fetch file contents from S3
   - `getFileSize` - Get file size via HEAD request
   - `copyFile` - Copy files within S3
   - `setAssetAsPublished` - Set compliance lock and remove scheduled access tags
   - `markScheduledAssetAsCancelled` - Tag cancelled assets for cleanup

**Breaking Changes**

**Features**:

- New `PushDocumentJob` database table for scheduling document ingestion
- Cron job (`schedulePushDocumentJob`) that runs every minute to process scheduled documents
- Gazette router with full CRUD operations (list, create, update, cancelScheduledPublish)
- `cancelScheduledPublish` endpoint that atomically deletes resource, job, and marks S3 asset as cancelled
- PDF text extraction for SearchSG content indexing
- Integration with SearchSG API for document ingestion
- Gazette UI components (table, create/modify/view modals)
- Toppan route guard for gazette access control
- GrowthBook feature flag integration for gazette feature
- S3 compliance lock (10,000 days retention) for published gazettes

**Improvements**:

- Extended S3 utilities with signed GET URLs, file size retrieval, file copy, and compliance lock operations
- Added tagging support to asset uploads for scheduled publishing
- Extracted page creation logic into shared service method

## Before & After Screenshots

N/A - New feature

## Tests

- Unit tests for `schedulePushDocumentJob` covering PDF extraction, SearchSG ingestion, and job cleanup
- Unit tests for `gazette.router` covering list, create, update, and cancelScheduledPublish operations
- Tests verify atomic transaction behavior and audit logging

**New dependencies**:

- None

**New dev dependencies**:

- None


seaerchin:
* **#2074** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2074?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2027** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2027?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2027?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2021** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2021?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2039](https://github.com/opengovsg/isomer/pull/2039) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2039?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

apps/studio/src/env.mjs
```diff
@@ -58,6 +58,9 @@ const server = z
     SESSION_SECRET: z.string().min(32),
     GROWTHBOOK_CLIENT_KEY: z.string().optional(),
     STUDIO_SSM_WEBHOOK_API_KEY: z.string().optional(),
+    S3_GAZETTE_BUCKET_NAME: z.string(),
+    S3_GAZETTE_DOMAIN_NAME: z.string().optional(),
+    EGAZETTE_DOCUMENT_INDEX: z.string().optional(),
```
adriangohjw: why optional ah?

apps/studio/next.config.mjs
```diff
@@ -94,6 +94,11 @@ const ContentSecurityPolicy = `
         ? `https://${env.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME}`
         : "https://*.by.gov.sg"
     }
+    ${
+      !!env.S3_GAZETTE_DOMAIN_NAME
+        ? `https://${env.S3_GAZETTE_DOMAIN_NAME}`
+        : ""
+    }
```
adriangohjw: re: why optional ah?

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -431,6 +451,22 @@ export const gazetteRouter = router({
...
 
             if (scheduledAtChanged) {
+              // NOTE: Need to update the associated PushDocumentJob
+              await tx
+                .updateTable("PushDocumentJob")
+                .set({
+                  scheduledAt,
+                  scheduledBy: user.id,
+                })
+                .where("resourceId", "=", String(gazetteId))
```
adriangohjw, line 521: defence-in-depth: we should also check for siteId

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -471,4 +507,185 @@ export const gazetteRouter = router({
...
+
+        // 2. Log the cancellation audit event
+        await logResourceEvent(tx, {
+          siteId,
+          eventType: AuditLogEvent.CancelSchedulePublish,
+          by: user,
+          delta: {
+            before: { resource: existingResource, blob: existingBlob },
+            after: { resource: existingResource, blob: existingBlob },
+          },
```
adriangohjw: question: actually is the typing for this correct? seems to me that all `CancelSchedulePublish` will have delta where the `before` and `after` is the same? the delta seems redundant

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -471,4 +507,185 @@ export const gazetteRouter = router({
...
+            .deleteFrom("Blob")
+            .where("id", "=", existingResource.draftBlobId)
+            .execute()
+        }
+
+        // 5. Delete the Resource
+        return await tx
+          .deleteFrom("Resource")
+          .where("id", "=", String(gazetteId))
+          .returningAll()
```
adriangohjw, line 679: defence-in-dept: add siteId check

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -471,4 +507,185 @@ export const gazetteRouter = router({
         return { gazetteId: updatedResource.id }
       },
     ),
+
+  cancelScheduledPublish: protectedProcedure
```
adriangohjw, line 576: non-blocking question: is there a X time before cancellation is no longer possible? e.g. like 1 min. before its scheduled to publish. in case there's a super edge case race condition between it being cancelled and it being processed

apps/studio/src/hooks/useUploadGazetteMutation.ts
```diff
@@ -0,0 +1,104 @@
+import type { z } from "zod"
+import type { getPresignedPutUrlSchema } from "~/schemas/asset"
```
adriangohjw: should this be from `schemas/gazette`?

apps/studio/src/schemas/gazette.ts
```diff
@@ -65,3 +65,63 @@ export const updateGazetteServerSchema = gazetteMetadataSchema.extend({
...
+      z.object({
+        key: z.string(),
+        value: z.string(),
+      }),
+    )
+    .optional(),
+  resourceId: z.string().optional(),
+  year: z.number().min(1000).max(9999),
+  category: z.string().min(1),
+  subcategory: z.string().min(1),
```
adriangohjw: suggestion(non-blocking): should we prevent whitespaces string?

apps/studio/src/schemas/gazette.ts
```diff
@@ -65,3 +65,63 @@ export const updateGazetteServerSchema = gazetteMetadataSchema.extend({
...
+        const extension = fileName
+          .toLowerCase()
+          .substring(fileName.lastIndexOf("."))
+        // NOTE: Gazettes should only be PDF files
+        return extension === ".pdf"
+      },
+      {
+        message: "File type not allowed. Please upload a supported file type.",
+      },
+    ),
```
adriangohjw: is this more direct?
```ts
.refine(
  (fileName) => fileName.trim().toLowerCase().endsWith(".pdf"),
  {
    message: "Only PDF files are allowed.",
  },
)
```

apps/studio/src/schemas/gazette.ts
```diff
@@ -65,3 +65,63 @@ export const updateGazetteServerSchema = gazetteMetadataSchema.extend({
...
+
+export const cancelScheduledPublishSchema = z.object({
+  siteId: z.number().min(1),
+  gazetteId: z.number().min(1),
+})
+
+export const getPresignedGetUrlSchema = z.object({
+  siteId: z.number().min(1),
+  fileKey: z.string().min(1),
+})
```
adriangohjw, line 85: security: should we ensure `fileKey` fit a certain constraint and do not start contain `/` to prevent IDOR? user might be able to run `fileKey: "/SITE_ID/some_stuff.pdf"` 

though since this is to get an asset (which should NOT be confidential), the risk is much lesser

though just thought to point out!

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -37,3 +53,108 @@ export const assertGazetteAccess = async (userId: string): Promise<void> => {
...
+
+  const presignedPutUrl = await generateSignedPutUrl({
+    Bucket: S3_GAZETTE_BUCKET_NAME,
+    Key: key,
+    ContentType: contentType,
+    ContentDisposition: contentDisposition,
+    Tagging: tags && stringifiedTags,
+  })
+  return { presignedPutUrl, contentType, contentDisposition }
+}
```
adriangohjw, line 128: non-blocker: should we add test? since there's an additional `tags` compared to asset's version

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -37,3 +53,108 @@ export const assertGazetteAccess = async (userId: string): Promise<void> => {
...
+ */
+export const copyFileWithNewName = async ({
+  sourceKey,
+  newFileName,
+}: {
+  sourceKey: string
+  newFileName: string
+}): Promise<string> => {
+  const parts = sourceKey.split("/")
+  // NOTE: This is in `/year/category/subcategory/filename` format
```
adriangohjw, line 156: non-blocking: actually i wonder if we want to do some OOP here to create a class/object for this so that we can encapsulate the logic inside it. Otherwise we have to add comments here and there haha

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -37,3 +53,108 @@ export const assertGazetteAccess = async (userId: string): Promise<void> => {
...
+  if (parts.length < 4) {
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: "Invalid source key format",
+    })
+  }
+  const prefix = parts.slice(0, 3).join("/")
+
```
adriangohjw, line 164: should this be 5 and 4 respectively? (instead of 4 and 3) 

/year/category/subcategory/filename -> ["", "year", "category", "subcategory", "filename"]

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -0,0 +1,239 @@
...
+
+export const schedulePushDocumentJob = async () => {
+  return await registerPgbossJob(
+    logger,
+    JOB_NAME,
+    CRON_SCHEDULE,
+    schedulePushDocumentJobHandler,
+    { retryLimit: 3, singletonKey: JOB_NAME },
+    env.SCHEDULED_PUBLISHING_HEARTBEAT_URL
+      ? { heartbeatURL: env.SCHEDULED_PUBLISHING_HEARTBEAT_URL }
```
adriangohjw: reminder: rmb to add to infra repo

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -0,0 +1,239 @@
...
+    throw new Error(`Failed to get SearchSG auth token: ${response.statusText}`)
+  }
+
+  const { accessToken, tokenType } = (await response.json()) as {
+    accessToken: string
+    tokenType: string
+  }
+
+  return { accessToken, tokenType }
+}
```
adriangohjw, line 202: import from `apps/studio/src/server/modules/searchsg/searchsg.service.ts` instead? or centralize it there

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -0,0 +1,239 @@
...
+    {
+      method: "POST",
+      headers: {
+        Authorization: `${tokenType} ${accessToken}`,
+        "Content-Type": "application/json",
+        "User-Agent": ISOMER_UA,
+      },
+      body: JSON.stringify({ documentsToAdd: documents }),
+    },
+  )
```
adriangohjw, line 222: should we also move this into `searchsg.service.ts`?

apps/studio/src/lib/s3.ts
```diff
@@ -98,6 +103,71 @@ export const deleteFile = async ({
...
+  // the gazette cannot be deleted
+  await storage.send(
+    new PutObjectRetentionCommand({
+      Bucket,
+      Key,
+      Retention: {
+        Mode:
+          env.NEXT_PUBLIC_APP_ENV === "production"
+            ? "COMPLIANCE"
+            : "GOVERNANCE",
```
adriangohjw, line 152: why this ah? for testing purposes?

apps/studio/src/lib/s3.ts
```diff
@@ -98,6 +103,71 @@ export const deleteFile = async ({
...
+        RetainUntilDate: retainUntilDate,
```
adriangohjw: super nit: can inline `addDays(new Date(), EGAZETTE_COMPLIANCE_HOLD_IN_DAYS)` directly instead of higher up in the code

apps/studio/src/lib/s3.ts
```diff
@@ -98,6 +103,71 @@ export const deleteFile = async ({
...
+    new PutObjectTaggingCommand({
+      Bucket,
+      Key,
+      // NOTE: We perform a soft delete here so the file can be kept available
+      // until the page is published
+      Tagging: {
+        TagSet: [...originalTagSet.filter(({ Key }) => Key !== "scheduledAt")],
+      },
+    }),
+  )
```
adriangohjw, line 169: question: i assume we can still change the tags after locking the gazette? and that locking only prevent deletion?

apps/studio/src/lib/s3.ts
```diff
@@ -98,6 +103,71 @@ export const deleteFile = async ({
...
 
+// NOTE: In order to set the asset as published, we have to do 2 things:
+// 1. we have to set the object lock retention
+// 2. we have to remove the scheduledAt tag
+// this is required to guarantee that the gazettes
+// can be seen and cannot be deleted
+export const setAssetAsPublished = async ({
+  Key,
+  Bucket,
+}: Pick<PutObjectTaggingCommandInput, "Key" | "Bucket">) => {
```
adriangohjw, line 114: given that this is for egazettes only, should this be in s3? or perhaps can be in something like "s3.gazette.ts" and remove Bucket props so we can reduce chance of accidentally using for other assets?

apps/studio/src/lib/s3.ts
```diff
@@ -126,3 +196,56 @@ export const copyFile = async ({
...
+      // NOTE: Remove scheduledAt tag and set deletedAt tag to mark the asset as cancelled
+      Tagging: {
+        TagSet: [
+          ...originalTagSet.filter(
+            ({ Key }) => Key !== "scheduledAt" && Key !== "deletedAt",
+          ),
+          {
+            Key: "deletedAt",
+            Value: Date.now().toString(),
+          },
```
adriangohjw, line 235: question: if there's an existing `deletedAt`, we will override it - is that intentional? feels like we shouldn't touch it if it aldy have a `deletedAt`?

apps/studio/src/lib/s3.ts
```diff
@@ -126,3 +196,56 @@ export const copyFile = async ({
...
+        ],
+      },
+    }),
+  )
+}
+
+export const getBlob = async (bucketName: string, key: string) => {
```
adriangohjw, line 242: should we standardize to use named arguments instead of positional one?

apps/studio/src/hooks/useUploadGazetteMutation.ts
```diff
@@ -0,0 +1,104 @@
...
+    },
+    method: "PUT",
+    body: file,
+  })
+
+  if (!response.ok) {
+    const data = (await response.json()) as unknown as { error: string }
+    throw new Error(data.error)
+  }
+}
```
adriangohjw: suggestion: `useUploadGazetteMutation` can be standalone but maybe this can be shared? since it touches security stuff

adriangohjw (commented):
- can add unit tests for the gazette router and service (the utils)?
- disclaimer: i skipped reviewing the tests

apps/studio/src/env.mjs
seaerchin: ~~got fixed downstream hahaha~~

we don't need domain name - it's jsut used to generate the document url that we send to searchsg hence i left it as optional

apps/studio/next.config.mjs
seaerchin: see above sir

apps/studio/src/hooks/useUploadGazetteMutation.ts
seaerchin: good catch, lemme update

apps/studio/src/lib/s3.ts
seaerchin: yea, don't wanna block deletion unless it's production

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: updated in downstream

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: updated downstream already

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Done — scoped the `PushDocumentJob` update via a `resourceId in (select Resource.id where siteId = ...)` subquery (PushDocumentJob has no direct siteId column). Fixed in #2200.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Done — same subquery pattern on the `PushDocumentJob` delete, plus an explicit `siteId` predicate on the `Resource` delete itself. Fixed in #2200.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Good catch. The `after` now carries a resource snapshot with `scheduledAt` / `scheduledBy` cleared, so the delta is the inverse of `SchedulePublish`. The subsequent `ResourceDelete` event still handles the actual deletion. Fixed in #2200.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Leaving as-is — the worst-case race window is bounded:

- SearchSG ingestion is keyed on `documentId` (deterministic from `url + resourceId`), so a duplicate ingest after cancellation just produces an upsert we can clear out-of-band.
- The S3 tag flip is best-effort and idempotent.
- A 1-min buffer would surprise users who hit cancel right before publish.

If this race materialises in prod we can revisit.

apps/studio/src/hooks/useUploadGazetteMutation.ts
seaerchin: Fixed in #2200 — the hook now imports from `~/schemas/gazette`. Also extracted `handleAssetUpload` into a shared module so the asset + gazette upload hooks share the same fetch logic.

apps/studio/src/schemas/gazette.ts
seaerchin: Done — `category` and `subcategory` now use `z.string().trim().min(1)`, which rejects whitespace-only input and also trims valid input before it flows into the S3 key. Fixed in #2200.

apps/studio/src/schemas/gazette.ts
seaerchin: Done, used your version verbatim. Behaviour is equivalent for the strings we care about, and now also tolerates trailing whitespace. Fixed in #2200.

apps/studio/src/schemas/gazette.ts
seaerchin: Done — `fileKey` now rejects values that start with `/` or contain a `..` segment. Two new tests cover both cases. Risk is low as you noted (it just bound an asset URL signing), but the guard makes the intent explicit. Fixed in #2200.

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: Added — two tests now assert `Tagging` is passed through to `generateSignedPutUrl` when tags are supplied, and omitted when not. Covered at both the router level (`getPresignedPutUrl > passes the supplied tags through to the underlying signer`) and the service level (`gazette.service > getPresignedPutUrl > includes Tagging in the signer call when tags are supplied`). Fixed in #2200.

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: Skipping — YAGNI for now. Single call site, the surrounding comments already make the intent clear, and an OOP wrapper would add an abstraction without removing the comments. Happy to revisit if a second use case shows up.

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: Pushing back — the current `4` and `3` are correct. Trace:

- `gazette.router.ts` (around the `update` mutation): `sourceKey = existingRef.replace(/^\//, "")` — the leading `/` is stripped before the call.
- So at the point of `sourceKey.split("/")`, the string looks like `year/category/subcategory/filename`.
- `split("/")` on that yields **4** parts: `["year", "category", "subcategory", "filename"]`.
- `parts.slice(0, 3)` keeps `year/category/subcategory`, which is the prefix we want.

Your reasoning would hold if the leading `/` were still present at split time, but the caller strips it. No change needed.

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: Noted — tracking separately. Need `SCHEDULE_PUSH_DOCUMENT_JOB_HEARTBEAT_URL` (Datadog) and `EGAZETTE_DOCUMENT_INDEX` + `SEARCHSG_API_KEY` provisioned in the infra repo before this can ship to staging/prod.

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: Done — `getSearchSGAuthToken` and `pushDocumentsForIngestion` now live in `searchsg.service.ts` alongside the existing config helpers. The cron worker just imports `pushDocumentsForIngestion` and the `PushDocument` type. Fixed in #2200.

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: Done — `pushDocumentsForIngestion` is now in `searchsg.service.ts` (same move as the auth-token helper above). Fixed in #2200.

apps/studio/src/lib/s3.ts
seaerchin: Done — inlined the `addDays(...)` call directly into the `RetainUntilDate` field. Fixed in #2200.

apps/studio/src/lib/s3.ts
seaerchin: Correct — `PutObjectRetention` (Object Lock) only blocks DELETE / overwrite of the object body. `PutObjectTagging` is unaffected, which is exactly why we lock first and then strip the `scheduledAt` tag in the second call. Both succeed.

apps/studio/src/lib/s3.ts
seaerchin: Agreed in principle — gazette-only functions in a generic `s3.ts` is sketchy. Skipping the split in this PR though to keep the diff focused on the review-feedback items; happy to do it as a follow-up. (`setAssetAsPublished`, `markScheduledAssetAsCancelled`, `getBlob`, `getFileSize`, and `copyFile` all have only gazette call sites today, so the move would be mechanical — drop the `Bucket` prop, default to `S3_GAZETTE_BUCKET_NAME`.)

apps/studio/src/lib/s3.ts
seaerchin: Good catch — fixed. `markScheduledAssetAsCancelled` now preserves an existing `deletedAt` instead of overwriting it (the original soft-delete timestamp is more meaningful than the cancellation moment). Fixed in #2200.

apps/studio/src/lib/s3.ts
seaerchin: Done — `getBlob` now takes `{ Bucket, Key }` to match the rest of `s3.ts`. Updated the one call site in `schedulePushDocumentJob.ts`. Fixed in #2200.

apps/studio/src/hooks/useUploadGazetteMutation.ts
seaerchin: Done — extracted `handleAssetUpload` into `src/hooks/handleAssetUpload.ts`. Both `useUploadAssetMutation` and `useUploadGazetteMutation` import from the same place now, so the signed Content-Type / Content-Disposition logic stays in lockstep. Fixed in #2200.

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: Update: #2074 also restructures this code (moves `pushDocumentsForIngestion` into `gazette.service.ts`, exports the related constants from `searchsg.service.ts`). To avoid duplicate work in two open PRs, I reverted the move from #2200 — the location decision belongs in #2074, where the wider restructure lives. Will let you and @seaerchin settle the final landing spot there.

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: Update: same as the auth-token thread above — reverted from #2200; #2074 carries the equivalent restructure.

apps/studio/src/lib/s3.ts
seaerchin: Update: I reverted the named-args refactor in #2200 to align with #2074, which keeps `getBlob` positional. The named-args cleanup can be done as a follow-up after both PRs land — happy to pick it up then.

apps/studio/src/env.mjs
adriangohjw: non-blocking: should we make it compulsory and put a placeholder domain for local/test? since its something that's required for production

apps/studio/src/server/modules/gazette/gazette.router.ts
adriangohjw: issue: it's changed to the following, but it means
- we are manually editing updating it before logging. Practically okay but not ideologically correct IMO. What about only we move it to after Step 5 where we actually delete the resource?
- in egazette case, we are actually deleting the resource but we are logging updated scheduledAt and scheduledBy in `CancelSchedulePublish` which isn't the whole truth? feels iffy

```ts
            after: {
              resource: {
                ...existingResource,
                scheduledAt: null,
                scheduledBy: null,
              },
              blob: existingBlob,
            },
```

apps/studio/src/lib/s3.ts
adriangohjw: ask: can add a comment to give future context on why this is being done?

adriangohjw:
@seaerchin reviewed. main blocker for me is the audit log stuff which seems iffy since the logs don't quite reflect the reality

adriangohjw (approved):
LGTM

PR #2032, opened 2026-05-08
feat: pass-through showThumbnail fallback for CollectionBlock

dcshzj:
## Problem

[CollectionBlock](packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx) ignored the referenced Collection's `showThumbnail` setting and always rendered the item's own image (or, if absent, the site logo). After [#1973](https://github.com/opengovsg/isomer-next/pull/1973) introduced a `showThumbnail.fallback` setting on Collection pages (`"logo"` vs `"first-image"`), the linked Collection block on the homepage no longer matched the Collection page's own thumbnail behaviour — e.g. a Collection configured to fall back to the page's first image still showed the site logo when surfaced via CollectionBlock.

## Solution

Thread the referenced Collection's `showThumbnail` setting through the sitemap into CollectionBlock, and reuse the existing `getItemImage` resolver from the Collection layout. When the referenced Collection does not have `showThumbnail` set (i.e. the Collection page itself hides thumbnails), CollectionBlock substitutes `{ fallback: "logo" }` so it always renders something — preserving the existing block-level "always show a thumbnail" expectation.

The block-level `displayThumbnail` toggle remains the master switch: when off, no thumbnail is rendered regardless of the Collection's setting.

**Breaking Changes**

**Features**:

- CollectionBlock now respects the referenced Collection's `showThumbnail.fallback` setting (`"logo"` or `"first-image"`), matching the Collection page's own thumbnail rendering.

**Improvements**:

- `showThumbnail` is now serialised onto `IsomerCollectionPageSitemap.collectionPagePageProps` during publishing, alongside the existing `sortOrder`, `defaultSortBy`, etc.
- `SingleCard` in CollectionBlock no longer duplicates the logo-fallback logic — it consumes the already-resolved `image` and `isContainNeeded` from `getCollectionItems`, the same path the Collection page uses.

**Bug Fixes**:

- A linked Collection on the homepage that points at a Collection configured with `showThumbnail.fallback = "first-image"` will now correctly use the item's first image as the fallback thumbnail, rather than the site logo.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

Unit tests added in [`getCollectionPages.test.ts`](packages/components/src/templates/next/components/complex/CollectionBlock/utils/__tests__/getCollectionPages.test.ts) cover:

- Item's own image is used regardless of `showThumbnail`.
- `showThumbnail` undefined → site logo (with `isContainNeeded`).
- `showThumbnail.fallback = "logo"` → site logo.
- `showThumbnail.fallback = "first-image"` → uses `firstImage`, with logo as a backstop when `firstImage` is absent.

`npm run typecheck` passes across the monorepo. `oxlint` passes (the 2 new warnings on the publishing line match the existing `no-unsafe-*` pattern on adjacent fields like `tagCategories` and `sortOrder`).

- [ ] On a site with at least one Collection containing items where some items have their own thumbnail and some do not, set the Collection page's "Display thumbnail on all items" setting to fall back to **first image on page**.
- [ ] On the same site, add a CollectionBlock on the homepage that references that Collection, with the block's "Display thumbnail of all pages" toggle **on**.
- [ ] Publish the site and load the homepage. Confirm that for items without their own thumbnail, the CollectionBlock now shows the item's first inline image (matching the Collection page's behaviour), rather than the site logo.
- [ ] Update the same Collection's "Display thumbnail on all items" setting to fall back to **site logo**, republish, and confirm CollectionBlock now shows the site logo as the fallback.
- [ ] Toggle the Collection page's "Display thumbnail on all items" setting **off** entirely, republish, and confirm the Collection page itself hides thumbnails while the CollectionBlock still renders the site logo as a fallback.
- [ ] Toggle the CollectionBlock's "Display thumbnail of all pages" **off**, republish, and confirm no thumbnails are rendered in the block (block-level toggle still takes precedence).

**New scripts**:

- None.

**New dependencies**:

- None.

**New dev dependencies**:

- None.

### DIFF
--- a/packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
+++ b/packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
@@ -77,6 +77,7 @@ const compoundStyles = createInfoCardsStyles()
 const SingleCard = ({
   title,
   image,
+  isContainNeeded,
   category,
   referenceLinkHref,
   displayThumbnail,
@@ -90,20 +91,18 @@ const SingleCard = ({
   const isExternalLink = !!referenceLinkHref && isExternalUrl(referenceLinkHref)
 
   const renderImage = () => {
-    const hasImage = image?.src !== undefined
-
-    // Fallback to site logo if user want to display thumbnail but no image is provided
-    const imageSrc = hasImage ? image.src : site.logoUrl
-    const imageAlt = hasImage ? image.alt : `Site logo for ${site.siteName}`
+    if (!image?.src) {
+      return null
+    }
 
     return (
       <div className={compoundStyles.cardImageContainer({ numberOfCards })}>
         <ImageClient
-          src={imageSrc}
-          alt={imageAlt}
+          src={image.src}
+          alt={image.alt}
           width="100%"
           className={compoundStyles.cardImage({
-            imageFit: hasImage ? "cover" : "contain",
+            imageFit: isContainNeeded ? "contain" : "cover",
           })}
           lazyLoading={shouldLazyLoad}
           assetsBaseUrl={site.assetsBaseUrl}
--- a/packages/components/src/templates/next/components/complex/CollectionBlock/utils/__tests__/getCollectionPages.test.ts
+++ b/packages/components/src/templates/next/components/complex/CollectionBlock/utils/__tests__/getCollectionPages.test.ts
@@ -49,10 +49,14 @@ describe("getCollectionPages", () => {
     id,
     permalink,
     date = "2021-01-01",
+    image,
+    firstImage,
   }: {
     id: string
     permalink: string
     date?: string
+    image?: { src: string; alt: string }
+    firstImage?: { src: string; alt: string }
   }): IsomerSitemap => ({
     id,
     title: `${id} title`,
@@ -61,6 +65,8 @@ describe("getCollectionPages", () => {
     lastModified: date,
     layout: "article",
     date,
+    image,
+    firstImage,
   })
 
   it("should return an empty array when the collection exists but has no items", () => {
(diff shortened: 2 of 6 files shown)


packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx
```diff
@@ -90,20 +91,18 @@ const SingleCard = ({
...
 
   const renderImage = () => {
-    const hasImage = image?.src !== undefined
-
-    // Fallback to site logo if user want to display thumbnail but no image is provided
-    const imageSrc = hasImage ? image.src : site.logoUrl
-    const imageAlt = hasImage ? image.alt : `Site logo for ${site.siteName}`
+    if (!image?.src) {
+      return null
+    }
```
seaerchin, line 96: comment: the flow changed - previously, if there was no image, `hasImage` would be false then it uses the `site.logoUrl` 

now, we aim to respect the config the user set - but that isn't present here, it seems like we're changing the defaults rather than following the config (ie, show/hide site logo, use first image blah blah)

dcshzj: This part of the code is quite convoluted, but not doing a refactor now due to blast radius vs urgency. To clarify:
1. The `getCollectionPages` function sets the `showThumbnail` to be `{ fallback: 'logo' }` when the original collection has it undefined. This is to make the site logo be the fallback image if the original collection is not showing any thumbnails
    1. To clarify, in this case we will still show the item's thumbnail if it exists, even though the collection explicitly says no --> this is something that we might need to discuss further on the behaviour, cos it is quite uncertain right now how users are using this
2. The `getCollectionItems` then uses the `showThumbnail` setting, then passes it to the `getItemImage` function which then determines the correct image to show depending on the circumstances and the fallback image selected

So by the time it reaches here, each item would have either the item thumbnail, or the fallback image set, which makes this `null` very unlikely to occur but is here more for defensive purposes.

The config in the referenced collection would still be respected in this sense, except for the actual show/hide thumbnail on the CollectionBlock itself which we would also need to discuss whether we want to remove this option from users.

adriangohjw (commented):
## Summary

- **What it does:** Threads the referenced Collection's `showThumbnail` setting through the sitemap into `CollectionBlock` so the homepage block matches the Collection page's fallback strategy (`logo` vs `first-image`), instead of always defaulting to the site logo.
- **Codeflow:**
  - `tooling/build/scripts/publishing/index.ts:150` serialises `showThumbnail` onto the sitemap entry's `collectionPagePageProps`.
  - `packages/components/src/types/sitemap.ts:37` extends the sitemap type accordingly.
  - `packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionPages.ts:25` reads `collectionParent.collectionPagePageProps?.showThumbnail` and substitutes `{ fallback: "logo" }` when unset, then delegates to the shared `getCollectionItems` resolver.
  - `packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx:80-112` drops the in-card duplicated logo-fallback logic, consuming the already-resolved `image` + `isContainNeeded`.
- **Overall:** Clean, well-scoped refactor that removes duplication and properly aligns block rendering with Collection page semantics. Tests cover the four interesting branches.

---

### Findings

**thought (non-blocking):** `packages/components/src/templates/next/components/complex/CollectionBlock/utils/getCollectionPages.ts:25-26` — substituting `{ fallback: "logo" }` when the Collection's own `showThumbnail` is `undefined` is a deliberate divergence from how the Collection page itself behaves (which hides thumbnails entirely in that case). The PR description calls this out, but it is worth confirming with design/product: an author who explicitly disables thumbnails on their Collection page may still see logo thumbnails on the homepage CollectionBlock. The current behavior is defensible (the block has its own `displayThumbnail` master toggle) but it does mean the two surfaces never fully agree when `showThumbnail` is off. Confidence: medium

**suggestion (non-blocking):** `packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx:93-96` — `renderImage()` now returns `null` whenever `image?.src` is missing, while the old code always rendered the aspect-ratio container with the logo. In practice the upstream default `{ fallback: "logo" }` should always produce an `image`, so this branch is unreachable today — but it creates a fragile coupling: if a future change ever lets `getItemImage` return `undefined` (or yields an image with empty `src`), only some cards in the grid would lose their thumbnail container, producing uneven card heights instead of a visible fallback. Consider either (a) asserting/narrowing the type so callers can't pass an image-less card while `displayThumbnail` is true, or (b) keeping a logo fallback inside `renderImage` as a defensive backstop. Confidence: medium

**nitpick (a11y, if-minor):** `packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts:46` (existing) vs. removed `CollectionBlock.tsx` line — the alt text for the logo fallback shifts from `"Site logo for {siteName}"` to `"{siteName} site logo"` as a side effect of switching to the shared resolver. Functionally equivalent and arguably more natural, but it is a user-visible (screen-reader-visible) change that the PR description doesn't mention. No action needed; flagging for awareness. Confidence: high

**question:** `packages/components/src/templates/next/components/complex/CollectionBlock/utils/__tests__/getCollectionPages.test.ts:243-250` — the assertion checks `item.image.isContainNeeded === true` AND `item.isContainNeeded === true`, i.e. the flag is duplicated both inside the `image` object and as a sibling field on the processed card. That mirrors what `getCollectionItems` + `processCollectionItems` produce today, but it's worth asking whether the duplication is intentional going forward, or whether `image.isContainNeeded` should be stripped in `processCollectionItems` to leave a single source of truth. Not a blocker for this PR — pre-existing shape — but a good cleanup target. Confidence: high

**praise:** `packages/components/src/templates/next/components/complex/CollectionBlock/CollectionBlock.tsx:91-112` — nice removal of the parallel fallback logic. Centralising thumbnail resolution in `getItemImage` is exactly the right call and makes the two surfaces (Collection page + CollectionBlock) provably consistent. Confidence: high

**note:** `tooling/build/scripts/publishing/index.ts:150` — this adds a new field to every published sitemap entry under `collectionPagePageProps`. Already-published sites won't have it until they republish, and the consumer code defaults to `{ fallback: "logo" }` (preserving prior behavior), so no migration is required. Just flagging that the visible behavior change only kicks in after a republish. Confidence: high

packages/components/src/templates/next/components/complex/CollectionBlock/utils/__tests__/getCollectionPages.test.ts
adriangohjw, line 1: ultra nit: add the comment Arrange/Assert/Act for consistency

adriangohjw (approved):
no blocker for me

but FYI should only merge this during release, otherwise the branch will be auto deleted on merge and the sites referencing this infra will have build failure

PR #2035, opened 2026-05-11
feat: wire Redirects table into publishing pipeline

dcshzj:
## Problem

The `Redirect` table introduced in the base branch stores per-site redirect paths, but nothing in the publishing pipeline reads from it yet. As a result, redirects configured in Studio do not actually take effect on published sites.

Closes ISOM-2260.

## Solution

Extend the publishing pipeline so each site build fetches its redirect rows from Postgres and materialises them as S3 objects that the existing CloudFront Function can act on.

- Add `GET_REDIRECTS` query and `fetchAndWriteRedirects` step in [publishing/index.ts](tooling/build/scripts/publishing/index.ts) that writes `redirects.json` for the current site. On query error the step writes an empty array so the build does not fail.
- Move `redirects.json` into the template output and capture its absolute path in [publisher.sh](tooling/build/scripts/publisher.sh) before changing CWD.
- Add [uploadRedirects.ts](tooling/build/scripts/publishing/uploadRedirects.ts), a standalone script invoked via `npm run upload-redirects`. For each `{source, destination}` row it uploads an empty `index.html` to `s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest/<source>/index.html` with `x-amz-meta-redirect-destination` set to the destination. The CloudFront Function already reads this header to issue the HTTP redirect.
- The upload runs **after** the `aws s3 sync --delete` of `_next`, so redirect objects are not swept away.
- Source paths are normalised and validated before upload: leading/trailing/duplicate slashes are collapsed, control characters and backslashes are rejected, `..` traversal segments are rejected, and duplicate normalised sources are skipped. Destinations that are not absolute (`http(s)://` or `/`) log a warning but still upload.
- Uploads run through a fixed-size worker pool (default concurrency 20, configurable via `UPLOAD_CONCURRENCY`). If any upload fails the script exits non-zero, but the calling shell step swallows the error so a few bad redirects do not fail the whole publish.

**Breaking Changes**

**Features**:

- Redirects configured for a site are now applied at publish time: each row in the `Redirect` table becomes an S3 object that triggers a CloudFront 301-style redirect to the configured destination.

**Improvements**:

- Input hardening for redirect sources (path normalisation, traversal rejection, deduplication) so malformed rows cannot produce unexpected S3 keys.
- Bounded concurrency on `aws s3 cp` calls so sites with many redirects do not fork hundreds of processes.

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] In Studio (or directly in the DB), add a `Redirect` row for a test site with a source like `old-page` and a destination like `https://example.com/new-page`.
- [ ] Trigger a publish for that site and confirm the CodeBuild logs include `Uploading redirect files to S3...` followed by `Uploaded N/N redirects.`
- [ ] In the S3 bucket, confirm an object exists at `<SITE_NAME>/<BUILD_NUMBER>/latest/old-page/index.html` with metadata header `x-amz-meta-redirect-destination: https://example.com/new-page`.
- [ ] Visit `https://<site-domain>/old-page` and confirm the browser is redirected to `https://example.com/new-page`.
- [ ] Add a redirect with a malformed source (e.g. `../etc/passwd` or one containing control characters) and confirm the publish logs `Skipping invalid redirect source: ...` and that no corresponding S3 object is created.
- [ ] Add two redirects whose sources normalise to the same path (e.g. `/foo/` and `foo`) and confirm only one is uploaded with a `Skipping duplicate normalised source` warning.
- [ ] Publish a site with zero redirect rows and confirm the build succeeds with `No valid redirects to upload.` in the logs.
- [ ] Confirm pre-existing `_next` assets are still present after publish (i.e. the redirect upload step running after the `--delete` sync has not regressed normal site assets).

**New scripts**:

- `upload-redirects` (in `tooling/build/scripts/publishing`): runs `uploadRedirects.ts` to upload S3 redirect objects from `redirects.json`.

**New dependencies**:

- None

**New dev dependencies**:

- None

### DIFF
--- a/tooling/build/scripts/publisher.sh
+++ b/tooling/build/scripts/publisher.sh
@@ -141,6 +141,9 @@ mv schema/ ../../../template/
 mv data/ ../../../template/
 cp sitemap.json ../../../template/public/
 mv sitemap.json ../../../template/
+mv redirects.json ../../../template/
+# Capture absolute path now; the upload step runs from a different CWD later.
+REDIRECTS_JSON="$(realpath ../../../template/redirects.json)"
 cd ../../../template
 # Create not-found.json by copying _index.json if it doesn't exist
 # Refer to tooling/template/app/not-found.tsx for more context
@@ -189,6 +192,19 @@ aws s3 sync --only-show-errors . s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUIL
 # Set all files in the _next folder to be cached indefinitely (1 year) on users' browsers
 # Next.js uses unique content hashes in filenames, allowing updated content to have different filenames and invalidate the cache on new builds.
 aws s3 sync --only-show-errors _next s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUILD_NUMBER/latest/_next --delete --no-progress --cache-control "max-age=31536000, public"
+
+# Upload redirect objects AFTER the --delete sync so they are not swept away.
+# Each redirect becomes an empty index.html with x-amz-meta-redirect-destination metadata
+# that the CloudFront Function reads to issue the HTTP redirect response.
+echo "Uploading redirect files to S3..."
+(
+  cd ../../build/scripts/publishing
+  REDIRECTS_JSON="$REDIRECTS_JSON" \
+    S3_BUCKET_NAME="$S3_BUCKET_NAME" \
+    SITE_NAME="$SITE_NAME" \
+    CODEBUILD_BUILD_NUMBER="$CODEBUILD_BUILD_NUMBER" \
+    npm run upload-redirects
+) || echo "Warning: some redirects failed to upload, continuing..."
 
 calculate_duration $start_time
 
--- a/tooling/build/scripts/publishing/queries.ts
+++ b/tooling/build/scripts/publishing/queries.ts
@@ -60,3 +60,7 @@ SELECT content FROM public."Footer" WHERE "siteId" = $1;
 export const GET_CONFIG = `
 SELECT name, config, theme FROM public."Site" WHERE "id" = $1;
 `
+
+export const GET_REDIRECTS = `
+SELECT source, destination FROM public."Redirect" WHERE "siteId" = $1;
+`
--- a/tooling/build/scripts/publishing/queries.ts
+++ b/tooling/build/scripts/publishing/queries.ts
@@ -62,5 +62,5 @@ SELECT name, config, theme FROM public."Site" WHERE "id" = $1;
 `
 
 export const GET_REDIRECTS = `
-SELECT source, destination FROM public."Redirect" WHERE "siteId" = $1;
+SELECT source, destination FROM public."Redirect" WHERE "siteId" = $1 AND "deletedAt" IS NULL;
 `
--- a/tooling/build/scripts/publishing/uploadRedirects.ts
+++ b/tooling/build/scripts/publishing/uploadRedirects.ts
@@ -1,5 +1,7 @@
 import { spawn } from "child_process"
 import * as fs from "fs"
+import * as os from "os"
+import * as path from "path"
 
 const REDIRECTS_JSON = process.env.REDIRECTS_JSON
 const S3_BUCKET = process.env.S3_BUCKET_NAME
@@ -27,13 +29,15 @@ function normalizeSource(source: string): string | null {
   return trimmed
 }
 
+const EMPTY_FILE = path.join(os.tmpdir(), "isomer-redirect-empty")
+
 function uploadOne(source: string, destination: string): Promise<void> {
   return new Promise((resolve, reject) => {
     const proc = spawn("aws", [
       "s3",
       "cp",
       "--only-show-errors",
-      "/dev/null",
+      EMPTY_FILE,
       `s3://${S3_BUCKET}/${SITE_NAME}/${BUILD_NUMBER}/latest/${source}/index.html`,
       "--content-type",
       "text/html",
@@ -120,6 +124,8 @@ async function main(): Promise<void> {
     console.log("No valid redirects to upload.")
     return
   }
+
+  fs.writeFileSync(EMPTY_FILE, "")
 
   console.log(
     `Uploading ${valid.length} redirect(s) with concurrency ${CONCURRENCY}...`,
--- a/tooling/build/scripts/publishing/uploadRedirects.ts
+++ b/tooling/build/scripts/publishing/uploadRedirects.ts
@@ -1,0 +1,132 @@
+import { spawn } from "child_process"
+import * as fs from "fs"
+
+const REDIRECTS_JSON = process.env.REDIRECTS_JSON
+const S3_BUCKET = process.env.S3_BUCKET_NAME
+const SITE_NAME = process.env.SITE_NAME
+const BUILD_NUMBER = process.env.CODEBUILD_BUILD_NUMBER
+const CONCURRENCY = Number(process.env.UPLOAD_CONCURRENCY) || 20
+
+interface Redirect { source: string; destination: string }
+
+// Returns the normalised S3 key segment, or null if the source is unsafe/empty.
+function normalizeSource(source: string): string | null {
+  if (typeof source !== "string") return null
+  // Reject control chars and backslashes
+  if (/[\x00-\x1f\\]/.test(source)) return null
+  const trimmed = source
+    .replace(/^\/+/, "")
+    .replace(/\/+$/, "")
+    .replace(/\/+/g, "/")
+  if (!trimmed) return null
+  // Reject any path-traversal segment
+  if (trimmed.split("/").some((seg) => seg === "..")) return null
+  return trimmed
+}
+
+function uploadOne(source: string, destination: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const proc = spawn("aws", [
+      "s3",
+      "cp",
+      "--only-show-errors",
+      "/dev/null",
+      `s3://${S3_BUCKET}/${SITE_NAME}/${BUILD_NUMBER}/latest/${source}/index.html`,
+      "--content-type",
+      "text/html",
+      "--cache-control",
+      "max-age=600",
+      // JSON form avoids shorthand parsing of commas/equals in destination URLs.
+      "--metadata",
+      JSON.stringify({ "redirect-destination": destination }),
+    ])
+    let stderr = ""
+    proc.stderr.on("data", (d) => {
+      stderr += d.toString()
+    })
+    proc.on("close", (code) => {
+      if (code === 0) resolve()
+      else
+        reject(
+          new Error(`aws s3 cp exited ${code} for ${source}: ${stderr.trim()}`),
+        )
+    })
+    proc.on("error", reject)
+  })
+}
+
+// Worker-pool: each worker pulls from the shared queue until it is empty.
+async function runWithConcurrency(
+  items: Redirect[],
+  limit: number,
+): Promise<{ failed: number }> {
+  const queue = items.slice()
+  let failed = 0
+  const worker = async () => {
+    while (queue.length > 0) {
+      const r = queue.shift()
+      if (!r) break
+      try {
+        await uploadOne(r.source, r.destination)
+      } catch (err) {
+        console.error("Redirect upload failed:", err)
+        failed++
+      }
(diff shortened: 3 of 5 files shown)


adriangohjw (commented):
## Summary

This PR wires the new `Redirect` table into the publish pipeline:

- `tooling/build/scripts/publishing/index.ts:545-560` adds `fetchAndWriteRedirects` which writes `redirects.json` next to the other publish artefacts; on DB error it writes `[]`.
- `tooling/build/scripts/publisher.sh:144-146` moves `redirects.json` into `template/` and captures the absolute path before the build changes CWD.
- After the `aws s3 sync --delete` block (`tooling/build/scripts/publisher.sh:190-194`), `publisher.sh:199-207` invokes the new `upload-redirects` npm script, which streams the file through a fixed-size worker pool (`uploadRedirects.ts:62-84`) and uploads one empty `index.html` per redirect with `x-amz-meta-redirect-destination` set (`uploadRedirects.ts:30-59`).

Overall: structure is sound, ordering vs. `--delete` is correct, and `spawn(argv)` avoids shell injection on the upload path. Notes below cover correctness gaps (the soft-delete column is ignored, DB failure silently empties live redirects, destination validation has a hole that allows protocol-relative open redirects) and a couple of robustness items worth fixing before this is exercised in prod.

## Findings

**issue:** `tooling/build/scripts/publishing/queries.ts:64-66` — the `Redirect` table has a `deletedAt` soft-delete column (`apps/studio/prisma/schema.prisma:184`), but `GET_REDIRECTS` selects unconditionally. Once Studio (or any tooling) starts soft-deleting redirects, deleted rows will keep being published and the only way to actually retract a redirect will be a hard `DELETE`. Add `AND "deletedAt" IS NULL` to the WHERE clause.

Confidence: high

**issue:** `tooling/build/scripts/publishing/index.ts:552-558` — on any DB error, this silently writes `[]` to `redirects.json`. Combined with `publisher.sh:200-207` swallowing the upload step’s exit code, a transient DB blip during publish will result in a successful build that publishes **zero** redirects — every previously-working redirect goes 404 with no signal. For a piece of infrastructure that exists specifically so old URLs keep working, "fail open to empty" is the wrong default. Prefer to let the error propagate so the publish fails loudly, or at minimum log at `error` level *and* exit non-zero so the CodeBuild job surfaces it.

Confidence: high

**issue (security):** `tooling/build/scripts/publishing/uploadRedirects.ts:110` — the regex `/^(https?:\/\/|\/)/` treats `//evil.com/path` as valid because it starts with `/`. Protocol-relative URLs are valid `Location:` header values and resolve to `https://evil.com/path` in the browser — i.e. an open redirect off the government domain. Since Studio has no write-side validator yet (`apps/studio/src/pages/sites/[siteId]/settings/redirects.tsx:30` is still "coming soon"), this script is the only line of defense. Tighten to e.g. `^(https?:\/\/|\/(?!\/))/`, or parse with `new URL(...)` and explicitly require `http:`, `https:`, or a same-origin pathname.

Confidence: high

**issue (security):** `tooling/build/scripts/publishing/uploadRedirects.ts:30-45` — `destination` is passed straight into S3 metadata with no validation of control characters or length. `normalizeSource` rejects `\x00-\x1f` and backslashes for the *source* (line 19) but the destination has the same trust profile (admin-controlled DB column, no write-side validation today) and ends up in an HTTP response header the browser consumes. The AWS SDK will likely reject CR/LF in metadata, but relying on that is fragile; apply the same control-char rejection to `destination`, and reject empty / non-string values.

Confidence: high

**issue:** `tooling/build/scripts/publishing/uploadRedirects.ts:98-117` — the loop assumes `r.destination` is a string but never checks. If the DB returns `null`/`undefined`/non-string (or if the JSON is malformed), `r.destination` falls through: the regex on line 110 stringifies it (so `null` matches nothing, warning logged), then line 116 pushes it into `valid`, and the upload sets metadata to the literal string `"null"` / `"undefined"` — or fails opaquely. Add `if (typeof r.destination !== "string" || !r.destination) continue` alongside the `normalizeSource` check.

Confidence: high

**issue:** `tooling/build/scripts/publisher.sh:199-207` — `( cd …; npm run upload-redirects ) || echo "Warning…"` masks a non-zero exit from the upload script. The script already distinguishes "no valid redirects" (success) from "some uploads failed" (`uploadRedirects.ts:127-129`, `process.exit(1)`). Swallowing that here means partial failures (e.g. throttling, IAM hiccup on a subset) never page anyone — over time a site can lose redirects with no signal. Consider letting the script’s exit code propagate, or at minimum push a metric / explicit error log instead of an `echo`.

Confidence: medium

**issue:** `tooling/build/scripts/publishing/uploadRedirects.ts:30-45` — uploading `<source>/index.html` will silently overwrite a real site page that lives at the same path. The main `aws s3 sync … --delete` runs first (`publisher.sh:190`), then redirects upload on top, so a redirect with `source = "about"` would replace the actual `/about/` page with an empty redirect stub. There’s no detection or warning. Worth either rejecting at write time in Studio (when that lands) or, defensively here, checking against the sitemap that the build also writes (`publisher.sh:142-143`) and skipping/erroring on collisions.

Confidence: medium

**suggestion:** `tooling/build/scripts/publishing/uploadRedirects.ts:8` — `Number(process.env.UPLOAD_CONCURRENCY) || 20` means `UPLOAD_CONCURRENCY=0` gets coerced back to 20 and a non-numeric value silently becomes 20. Probably fine in practice, but the parse-then-validate pattern (`const n = Number(...); CONCURRENCY = Number.isFinite(n) && n > 0 ? n : 20`) makes misconfigurations more obvious.

Confidence: low

**suggestion:** `tooling/build/scripts/publishing/uploadRedirects.ts:30-59` — spawning a fresh `aws` process per redirect is fine for tens, marginal for hundreds, and slow for thousands (process-fork overhead dominates the network call). If this is ever expected to handle large redirect sets, the AWS SDK's `PutObject` (already a transitive dep in this repo) would be faster and easier to surface partial-failure metadata from. Non-blocking.

Confidence: medium

**nitpick:** `tooling/build/scripts/publisher.sh:144` — `mv redirects.json ../../../template/` puts the file inside `template/` but not `template/public/`, so unlike `sitemap.json` (`publisher.sh:142-143`) it never ends up in `out/`. The only consumer is the upload step that reads it via `$REDIRECTS_JSON`. Leaving it in `tooling/build/scripts/publishing/` and capturing the path there would avoid the misleading `mv` and the need to traverse back later. Non-blocking.

Confidence: medium

**praise:** `tooling/build/scripts/publishing/uploadRedirects.ts:42-44` — good call using `JSON.stringify` for `--metadata` to dodge shorthand parsing on commas/equals in destination URLs, and good call on `spawn(argv)` rather than a shell string. Both are easy things to get wrong.

Confidence: high

tooling/build/scripts/publisher.sh
```diff
@@ -190,6 +193,19 @@ aws s3 sync --only-show-errors . s3://$S3_BUCKET_NAME/$SITE_NAME/$CODEBUILD_BUIL
...
+# Each redirect becomes an empty index.html with x-amz-meta-redirect-destination metadata
+# that the CloudFront Function reads to issue the HTTP redirect response.
+echo "Uploading redirect files to S3..."
+(
+  cd ../../build/scripts/publishing
+  REDIRECTS_JSON="$REDIRECTS_JSON" \
+    S3_BUCKET_NAME="$S3_BUCKET_NAME" \
+    SITE_NAME="$SITE_NAME" \
+    CODEBUILD_BUILD_NUMBER="$CODEBUILD_BUILD_NUMBER" \
+    npm run upload-redirects
```
adriangohjw, line 206: question: since this is being used in the publisher script, does it make sense to be a bash script here instead? then we can use similar commands like `aws s3` and concurrency is preset based on the above

dcshzj: There are some processing being done by the uploadRedirects script that would not be fully captured inside a bash command, like the normalisation of the redirects, hence opted to put it as a TS script for clarity

adriangohjw: does it make sense then to do the normalization in TS and leave the S3 operations as bash script?
+ will be great if we can add unit test for the normalization utils to prevent regressions

dcshzj: Hmm for the s3 cp part not really, because the copy is done on a per-file basis but the concurrency part that you mentioned is when we run an s3 sync, so we won't be able to benefit from the presets.

Right now this separate script is the one that does the concurrency part.

tooling/build/scripts/publishing/uploadRedirects.ts
```diff
@@ -0,0 +1,141 @@
+import { spawn } from "child_process"
+import * as fs from "fs"
+import * as os from "os"
+import * as path from "path"
+
+const REDIRECTS_JSON = process.env.REDIRECTS_JSON
+const S3_BUCKET = process.env.S3_BUCKET_NAME
+const SITE_NAME = process.env.SITE_NAME
+const BUILD_NUMBER = process.env.CODEBUILD_BUILD_NUMBER
+const CONCURRENCY = Number(process.env.UPLOAD_CONCURRENCY) || 20
```
adriangohjw: question: where is this `UPLOAD_CONCURRENCY` from?

tooling/build/scripts/publishing/uploadRedirects.ts
```diff
@@ -0,0 +1,141 @@
...
+interface Redirect {
+  source: string
+  destination: string
+}
+
+// Returns the normalised S3 key segment, or null if the source is unsafe/empty.
+function normalizeSource(source: string): string | null {
+  if (typeof source !== "string") return null
+  // Reject control chars and backslashes
+  if (/[\x00-\x1f\\]/.test(source)) return null
```
adriangohjw, line 21: question: should we put this in a shared utils that studio can use for single source of truth when validating inputs?

tooling/build/scripts/publishing/uploadRedirects.ts
```diff
@@ -0,0 +1,141 @@
...
+        reject(
+          new Error(`aws s3 cp exited ${code} for ${source}: ${stderr.trim()}`),
+        )
+    })
+    proc.on("error", reject)
+  })
+}
+
+// Worker-pool: each worker pulls from the shared queue until it is empty.
+async function runWithConcurrency(
```
adriangohjw, line 66: non-blocking: thoughts on using `p-limit` (or something equivalent) instead of handrolling it ourselves?

tooling/build/scripts/publishing/queries.ts
```diff
@@ -60,3 +60,7 @@ SELECT content FROM public."Footer" WHERE "siteId" = $1;
 export const GET_CONFIG = `
 SELECT name, config, theme FROM public."Site" WHERE "id" = $1;
 `
+
+export const GET_REDIRECTS = `
+SELECT source, destination FROM public."Redirect" WHERE "siteId" = $1;
```
adriangohjw: **issue**: need to filter away deleted redirects? (there's `deletedAt` in `Redirect` table)

adriangohjw (changes requested):

tooling/build/scripts/publishing/uploadRedirects.ts
dcshzj: Right now we don't share anything between studio and the publishing scripts because the context is not available at the point this script runs (similar to the actual publishing script), hence there is duplication involved here.

But point taken, I think let's follow up on this with moving the shared utils and imports as a separate package consumed by both Studio and the publishing scripts.

tooling/build/scripts/publishing/uploadRedirects.ts
dcshzj: Prefer not adding dependencies to the publishing scripts right now but not a strong opinion

tooling/build/scripts/publishing/queries.ts
dcshzj: Good point, updated in 35166d843b1019975f222170e6567cbfab1b61ad

tooling/build/scripts/publishing/uploadRedirects.ts
dcshzj: Hmm removed in 83245853d94c2ba473913292aa0c3bb87f1a4338.

adriangohjw (approved):

PR #2037, opened 2026-05-11
chore: remove character limit from tag labels

seaerchin:
## Problem

Tag labels were limited to 70 characters, which may be too restrictive for some use cases where more descriptive tag names are needed.

## Solution

Removed the `maxLength: 70` constraint from the `DropdownItemSchema` in `packages/components/src/types/page.ts`. This affects both tag category labels and tag option labels.

**Breaking Changes**

**Improvements**:

- Tag category labels and tag option labels can now be any length (previously limited to 70 characters)

## Before & After Screenshots

N/A - Schema change only, no visual changes.

## Tests

- Verify tag categories and tag options can be created with labels longer than 70 characters
- Verify existing tags continue to work as expected

**New scripts**: None

**New dependencies**: None

### DIFF
--- a/packages/components/src/types/page.ts
+++ b/packages/components/src/types/page.ts
@@ -29,7 +29,7 @@ export const TagCategoryUuidSchema = generateUuidSchema({
 // NOTE: single value for now but we might extend this in the future with additional metadata,
 // so we will leave it as is
 const DropdownItemSchema = Type.Object({
-  label: Type.String({ maxLength: 70 }),
+  label: Type.String(),
   id: TagOptionUuidSchema,
 })
 const TagOptionSchema = DropdownItemSchema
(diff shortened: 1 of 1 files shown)


packages/components/src/types/page.ts
```diff
@@ -29,7 +29,7 @@ export const TagCategoryUuidSchema = generateUuidSchema({
 // NOTE: single value for now but we might extend this in the future with additional metadata,
 // so we will leave it as is
 const DropdownItemSchema = Type.Object({
-  label: Type.String({ maxLength: 70 }),
+  label: Type.String(),
   id: TagOptionUuidSchema,
 })
```
adriangohjw, line 34: non-blocker/suggestion: but should we add `NON_EMPTY_STRING_REGEX`?

adriangohjw (approved):

packages/components/src/types/page.ts
seaerchin: oh yea yes you right

PR #2043, opened 2026-05-12
feat: implement redirects management UI

dcshzj:
## Problem

The redirects settings page at `/sites/[siteId]/settings/redirects` was a placeholder showing only "Redirects — coming soon". Site admins had no way to manage URL redirects from the Studio UI.

Closes ISOM-2281.

## Solution

Build the redirects management UI behind the existing `useIsRedirectionsEnabled` feature flag, matching the latest Figma design. Redirects go live as soon as they are added — there is no separate publish step, so the page is a single table of live redirects:

- A page header with a one-line description and a "how to use redirects" support link.
- An add-redirect card that validates source and destination via Zod (rejects control characters, backslashes, and `..` segments; enforces that destinations start with `/` or `https://`) and normalises sources to a single leading slash. Adding shows a success toast, or an error toast if a redirect already exists for that path.
- A paginated, sortable table (25 entries per page) built on `@tanstack/react-table` + the existing `Datatable`/`DatatablePagination` components, with columns for source (wildcard `*` badge), destination, and an "Added" column with friendly timestamps ("just now", "today", "yesterday", then a date). Deleted redirects are never shown.
- Deleting a row opens a confirmation modal summarising the redirect before it is removed.

API calls are stubbed via jotai-backed mock state that simulates the server contract — including limit/offset pagination and server-side sorting — and are marked with `TODO: Replace with trpc.redirect.*` comments, so the backend PR swaps in real endpoints without UI churn. Bulk upload of redirects is excluded from this PR and will be implemented separately.

**Breaking Changes**

**Features**:

- Redirects settings page renders a functional management UI (header, add form, paginated sortable table, delete confirmation modal) behind the redirections feature flag.
- Immediate-publish flow: no publish button or draft states — adds and confirmed deletes take effect right away (mocked in this PR).
- Pagination at 25 entries per page, with sorting handled "server-side" by the mock so it orders the full dataset, not just the visible page. Changing the sort returns to the first page, and deleting the last row of a page steps back to the new last page.

**Improvements**:

- Centralises redirect-related UI under `apps/studio/src/features/settings/Redirects/`, with the mock API mirroring the planned tRPC contract (`list` with limit/offset/sort, `count`, `create`, `delete`) for a clean swap in the backend PR.

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**: Placeholder page showing "Redirects — coming soon"

**AFTER**:

> Note: the screenshot below is from an earlier iteration (it still shows the Publish button and status column, which have since been removed) — to be refreshed.

image

## Tests

- [ ] With the redirections GrowthBook feature flag enabled for your user, navigate to `/sites/<siteId>/settings/redirects` and confirm the page renders the header, add form, and table.
- [ ] With the flag disabled, confirm the page renders nothing (no crash) and other settings pages still work.
- [ ] In the add form, submit with an empty source or destination and confirm validation blocks it.
- [ ] Enter an invalid destination (e.g. `foo` without a leading slash, or `http://example.com`) and confirm a validation error appears.
- [ ] Add a valid redirect (e.g. source `test`, destination `/about`) — confirm the success toast and that the row appears at the top with "just now" under Added.
- [ ] Add the same source again and confirm an error toast appears and nothing is added.
- [ ] Click the trash icon on a row, confirm the modal shows the redirect's source and destination, confirm — the row disappears with a success toast.
- [ ] The mock data seeds 60 rows: confirm the pagination footer shows 3 pages and page navigation works.
- [ ] Click each sortable column header and confirm rows reorder across the whole dataset (not just the visible page) and the table returns to page 1.

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A


apps/studio/src/features/settings/Redirects/types.ts
```diff
@@ -0,0 +1,52 @@
...
+    { message: 'Source path cannot contain ".." segments' },
+  )
+
+const destinationSchema = z
+  .string()
+  .min(1, "Destination is required")
+  .max(2000, "Destination is too long")
+  .refine((val) => /^(https?:\/\/|\/)/.test(val), {
+    message: "Destination must start with / or https://",
+  })
```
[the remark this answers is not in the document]
dcshzj: Not doing yet since this is just the frontend

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const prevIsPending = useRef(false)
+  useEffect(() => {
+    if (prevIsPending.current && !isPending) {
+      toast({
+        ...SETTINGS_TOAST_MESSAGES.success,
+        status: "success",
+      })
+    }
+    prevIsPending.current = isPending
+  }, [isPending, toast])
```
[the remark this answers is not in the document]
dcshzj: Not doing yet since this is just the frontend

adriangohjw (commented):
## Summary

- Replaces the placeholder `/sites/[siteId]/settings/redirects` page with a functional redirects management UI (header + add-redirect card + sortable Datatable + per-row delete + Publish action), all gated behind the existing `useIsRedirectionsEnabled` GrowthBook flag (`apps/studio/src/pages/sites/[siteId]/settings/redirects.tsx:27`).
- Data layer is a deliberate stub: a module-level Jotai atom seeded with 5 rows lives in `apps/studio/src/features/settings/Redirects/api.ts:49`, exposing `useListRedirects`, `useCreateRedirect`, `useDeleteRedirect`, `usePublishRedirects` with `TODO: Replace with trpc.redirect.*` markers. The contract (`{ data, isLoading }` / `{ mutate, isPending }`) mirrors tRPC's shape so the swap-in should be mechanical.
- Validation lives in `apps/studio/src/features/settings/Redirects/types.ts` with a Zod schema rejecting control chars, backslashes, empty paths, and `..` segments on the source side, and requiring `/` or `https://` prefix on the destination side.
- Overall: nicely scoped, designed for backend swap-in. Several issues below are worth fixing before the tRPC routes ship so the validation contract doesn't quietly carry over.

## Findings

**issue (security):** `apps/studio/src/features/settings/Redirects/types.ts:L43` — the destination regex `/^(https:\/\/|\/)/` accepts protocol-relative URLs like `//evil.com/path`. `"//evil.com/path"` starts with `/`, so it passes the check, but browsers resolve it as `https://evil.com/path` — i.e. an open-redirect off-site. Verified: `/^(https:\/\/|\/)/.test("//attacker.com/foo")` → `true`. Even though backend validation will be authoritative, this UI schema is exactly what tends to get lifted and reused server-side. Reject any value starting with `//` (or require the second char to be non-`/`), e.g. `/^(https:\/\/[^/]|\/[^/])/` or an explicit `!val.startsWith("//")` refine.

Confidence: high

**issue (security):** `apps/studio/src/features/settings/Redirects/types.ts:L39-L45` — destination validation only checks the prefix; nothing rejects control chars, `\r\n` injection, whitespace, or `..` traversal in destinations. Source paths get the full `hasInvalidChars` + `..` treatment but destinations don't, so `/foo\r\nSet-Cookie:%20x=1` or `/../../etc/passwd` passes. If destination is rendered back to the user or eventually used in `Location:` headers downstream, this matters. Apply the same `hasInvalidChars` and `..`-segment refinements to `destinationSchema` (skipping the segment check when it starts with `https://`).

Confidence: medium

**issue (bug):** `apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx:L49-L60` — `isAddDisabled` only checks whitespace-trimmed non-emptiness, but `handleAdd` runs the full Zod validation and silently returns when it fails. Net effect: with both fields non-empty but invalid (e.g. destination `foo` with no leading slash and no blur), the Add button is enabled, the user clicks it, and nothing happens — no error appears either (validate sets errors, but the user already saw the button was clickable). Either run `validate()` in a `useEffect`/derived state and use its result for `isAddDisabled`, or show errors immediately on submit failure with a visible affordance (focus the first invalid field). The current state is "click does nothing".

Confidence: high

**issue (bug):** `apps/studio/src/features/settings/Redirects/api.ts:L49,L52` — the `redirectsAtom` is module-scoped and ignores `siteId` (it's `_siteId` everywhere). Two sites in the same Studio session share one list: navigate to site A, add a redirect, switch to site B → site B shows site A's redirects. This is mock-only state, but reviewers manually testing across sites will hit this and file it as a bug. Cheapest fix: key the atom by `siteId` (`atomFamily` or a `Map<number, RedirectRow[]>`). Or call this out in a code comment so the next person knows it's intentional.

Confidence: high

**issue (bug):** `apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx:L264-L266` — `handleDelete` is a fresh arrow on every render, so `useMemo(() => getColumns(handleDelete), [handleDelete])` re-runs every render and never memoizes. Re-creating the columns array breaks `useReactTable`'s internal memoization and can churn the header cell components. Wrap `handleDelete` in `useCallback([deleteRedirect, siteId])`, or just inline the call into the `cell` renderer.

Confidence: high

**suggestion:** `apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx:L62` — `React.FormEvent<HTMLDivElement>` is the wrong element type. `HStack as="form"` produces a `<form>` at runtime, but Chakra's `as` prop doesn't narrow the event type. Use `React.FormEventHandler<HTMLFormElement>` or `(e: React.FormEvent<HTMLFormElement>)` so the type matches reality.

Confidence: high

**suggestion:** `apps/studio/src/features/settings/Redirects/api.ts:L19,L27,L35,L43` — `SEED_ROWS` evaluates `new Date()` and `Date.now() - 2 * 24 * 60 * 60 * 1000` at module load. In dev (HMR aside) and in long-lived sessions, "Active" rows will keep reading `"a few seconds ago"` for the lifetime of the tab. Either move the seed into a function evaluated on first hook call, or hardcode an ISO string so the staleness is obvious during review.

Confidence: medium

**suggestion:** `apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx:L54` — source is normalised to a single leading `/` before send, but destination isn't touched at all. A user typing `"/about "` (trailing space) submits `/about ` verbatim; `"/about//foo"` likewise. If the backend will normalise, fine — leave a comment saying so. If not, add a `.trim()` and collapse repeated slashes to keep frontend display consistent.

Confidence: medium

**question:** `apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx:L25-L34` — the success toast fires off a `prevIsPending` ref tracking, so it'll also fire if `usePublishRedirects` ever ends up returning `isPending: true → false` for a reason other than a user click (e.g. a future cache-driven refetch). Once the real tRPC mutation lands, prefer the mutation's `onSuccess` callback to drive the toast — it's the standard pattern in this codebase. Worth flagging as a comment so it isn't missed during the swap.

Confidence: medium

**nitpick:** `apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx:L57` — `https://support.isomer.gov.sg` is a generic homepage link, not a deep link to "how to use redirects". The Publish action handler also uses `BRIEF_TOAST_SETTINGS` + generic `SETTINGS_TOAST_MESSAGES.success` ("Changes published. They'll appear on your site in 5-10 minutes.") — that message wording assumes a publish pipeline that may not match how redirects actually deploy. Worth a second look once the real backend is wired.

Confidence: low

**praise:** good call on clearly marking the mock layer with `TODO: Replace with trpc.redirect.*` and matching the tRPC shape (`{ data, isLoading }` / `{ mutate, isPending }`). Makes the eventual swap a near-mechanical change rather than a refactor.

Confidence: high

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
adriangohjw: can we have `usePublishRedirects` accept a `onSuccess` such that we can trigger the toast on success instead of having a `useRef` and `useEffect`? since useEffect is only checking for `isPending` switching from true to false

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+      </Stack>
+      <Button
+        isDisabled={!hasDirtyRows}
+        isLoading={isPending}
+        onClick={() => publishRedirects(siteId)}
+        flexShrink={0}
+        size="xs"
+      >
+        Publish
+      </Button>
```
adriangohjw: question: we not supporting bulk upload as per figma?

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const handleAdd = () => {
+    if (isAddDisabled) return
+    if (!validate()) return
+    const normalisedSource = `/${source.replace(/^\/+/, "")}`
+    createRedirect({ siteId, source: normalisedSource, destination })
+    setSource("")
+    setDestination("")
+    setSourceError("")
+    setDestinationError("")
```
adriangohjw: issue(?): we clear the content on createRedirect regardless of the outcome, which means its cleared even if its error (for whatever reason) - is this intended?

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,141 @@
...
+    return false
+  }
+
+  const isAddDisabled = !source.trim() || !destination.trim()
+
+  const handleAdd = () => {
+    if (isAddDisabled) return
+    if (!validate()) return
+    const normalisedSource = `/${source.replace(/^\/+/, "")}`
+    createRedirect({ siteId, source: normalisedSource, destination })
```
adriangohjw: small issue: `validate` is done on `source` but we create redirection with a normalisedSource. i think they should be the same?

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,141 @@
...
+}
+
+export const AddRedirectCard = ({
+  siteId,
+}: AddRedirectCardProps): JSX.Element => {
+  const [source, setSource] = useState("")
+  const [destination, setDestination] = useState("")
+  const [sourceError, setSourceError] = useState("")
+  const [destinationError, setDestinationError] = useState("")
+  const { mutate: createRedirect, isPending } = useCreateRedirect()
```
adriangohjw, line 46: suggestion: should trigger error toast on error?

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,141 @@
...
+  siteId: number
+}
+
+export const AddRedirectCard = ({
+  siteId,
+}: AddRedirectCardProps): JSX.Element => {
+  const [source, setSource] = useState("")
+  const [destination, setDestination] = useState("")
+  const [sourceError, setSourceError] = useState("")
+  const [destinationError, setDestinationError] = useState("")
```
adriangohjw: question: should we use `useZodForm` here?

apps/studio/src/features/settings/Redirects/api.ts
adriangohjw, line 1: _(i skipped since mostly TODO)_

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
dcshzj: Not yet, that's a separate task

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Updated

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
dcshzj: It's like this for now as the API is being mocked. It will be switched to the tanstack query ones which will have these out of the box in the downstream backend PR

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Will do in downstream backend PR since this is just scaffolding

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Hmm I'm excluding this part from the validation so that the user can input a leading slash and it still passes the validation (e.g. if they are copying and pasting entries which would have the leading slash).

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Yeah not intended, but this one is scaffolding still so will do in the downstream backend PR

apps/studio/src/features/settings/Redirects/components/DeleteRedirectModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <ModalCloseButton size="lg" />
+
+        <ModalBody>
+          <Stack spacing="1.5rem">
+            <Text textStyle="body-1">
+              Are you sure you want to delete this redirect?
+            </Text>
+            <HStack
+              spacing="1.5rem"
+              align="top"
```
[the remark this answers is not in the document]
dcshzj: Fixed, changed to `flex-start`

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+    sortBy: (sorting[0]?.id ?? "publishedAt") as RedirectSortField,
+    sortDirection: (sorting[0]?.desc ?? true) ? "desc" : "asc",
+  })
+
+  // Changing the sort order reshuffles rows across pages, so jump back to
+  // the first page to avoid showing a stale slice
+  const handleSortingChange: OnChangeFn<SortingState> = (updater) => {
+    setSorting(updater)
+    onPaginationChange((prev) => ({ ...prev, pageIndex: 0 }))
+  }
```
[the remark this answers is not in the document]
dcshzj: Fixed by setting `enableMultiSort: false` on the table, so the misleading state can't arise at all

apps/studio/src/features/settings/Redirects/api.ts
```diff
@@ -0,0 +1,165 @@
...
+// misleading), and must exclude soft-deleted redirects — only live rows are
+// ever shown.
+export function useListRedirects(
+  _siteId: number,
+  { limit, offset, sortBy, sortDirection }: ListRedirectsParams,
+): {
+  data: RedirectRow[]
+  isLoading: boolean
+} {
+  const rows = useAtomValue(redirectsAtom)
```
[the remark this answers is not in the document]
dcshzj: Fixed, the mock state is now keyed by siteId so each site stays isolated (every site starts from the same seed rows)

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+          toast({ ...SETTINGS_TOAST_MESSAGES.success, status: "success" })
+        },
+      },
+    )
+
+  const tableInstance = useReactTable<RedirectRow>({
+    columns,
+    data: redirects,
+    state: { sorting, pagination },
+    onSortingChange: handleSortingChange,
```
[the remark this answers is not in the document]
dcshzj: The defaulting deliberately lives in the hook rather than the table: the mock always returns an array, and in the downstream backend PR `useListRedirects` returns `data ?? []` with a non-optional return type, so the table can't receive `undefined` by construction. A `?? []` here would be dead code.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+      borderRadius="0.5rem"
+      p="1.25rem"
+      bgColor="base.canvas.default"
+    >
+      <Text textStyle="subhead-1" mb="0.25rem">
+        Add new redirects
+      </Text>
+      <Text textStyle="body-2" color="base.content.medium" mb="1.25rem">
+        Redirects go live as soon as you add them.
+      </Text>
```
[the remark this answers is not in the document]
dcshzj: The PR description has been updated to describe the immediate-publish behaviour, including the revised manual test plan

apps/studio/src/features/settings/Redirects/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { differenceInMinutes, format, isToday, isYesterday } from "date-fns"
+
+// Coarse relative time for the "Added" column, matching the design:
+// "just now" → "today" → "yesterday" → "12 Sep 2024"
+export const formatAddedAt = (date: Date): string => {
+  if (differenceInMinutes(new Date(), date) < 5) return "just now"
+  if (isToday(date)) return "today"
+  if (isYesterday(date)) return "yesterday"
+  return format(date, "d MMM yyyy")
```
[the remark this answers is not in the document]
dcshzj: Added unit tests for `formatAddedAt` covering the boundaries (the 5-minute "just now" cutoff, today/yesterday transitions, and the midnight crossover). The validation schema boundary cases (control characters, `..` segments, normalisation) are already covered by unit tests in the downstream backend PR (#2093) against the shared `~/schemas/redirect`, which replaces the feature-local copy here.

seaerchin:
## Summary

- Replaces the placeholder redirects settings page with a functional UI: header → add-redirect card with `useZodForm` validation → paginated/sortable `Datatable` → confirm-delete modal, gated by `useIsRedirectionsEnabled` (`apps/studio/src/pages/sites/[siteId]/settings/redirects.tsx:13-28`).
- API is intentionally stubbed via a Jotai atom keyed by `siteId` (`apps/studio/src/features/settings/Redirects/api.ts:47-179`), with `TODO: Replace with trpc.redirect.*` markers so the next PR can swap it out cleanly. Source paths are zod-validated then normalised at submit time (`AddRedirectCard.tsx:52`).
- Overall the scaffolding is clean and the deferrals to the backend PR are reasonable. Most findings below are about pre-empting issues at the moment that swap happens, so the boundary doesn't drift.

This is a convention-focused pass on top of the existing Copilot / @adriangohjw threads — not a re-litigation of points already deferred.

## Findings

**suggestion (maintainability):** `apps/studio/src/features/settings/Redirects/types.ts:47` — `addRedirectSchema` lives in `types.ts`. Per `apps/studio/src/features/CLAUDE.md`, feature-local schemas belong in `schema.ts`. And per the `react-forms-usezodform-over-usestate` convention, *"reuse the schema from `~/schemas/` — ideally the same schema the tRPC procedure validates with"*. Since the backend PR is imminent and will validate the same shape, the schema should land in `~/schemas/redirect.ts` so client + server validate against one source of truth. Renaming now (or as the first move in the backend PR) avoids silent client/server drift.

Confidence: high

**issue (correctness):** `apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx:52` — `normalisedSource = \`/${source.replace(/^\/+/, "")}\`` only normalises **leading** slashes; the validator (`types.ts:26,33`) trims leading **and** trailing slashes to check emptiness/`..` segments. Net effect: a user can submit `foo` and `foo/` and end up with two distinct rows (`/foo` and `/foo/`) that point at logically the same source. `useCreateRedirect` dedupes by exact string equality and the backend will likely do the same, so this is a real footgun. Folding normalisation into the schema via `.transform()` makes "validated value" and "stored value" the same string and eliminates the drift:

```ts
const sourceSchema = z
  .string()
  .min(1, ...)
  .transform((val) => `/${val.replace(/^\/+/, "").replace(/\/+$/, "")}`)
  .refine((val) => val.length > 1, { message: "Source path cannot be empty" })
  // ...further refinements run on the already-normalised value
```

This also addresses @adriangohjw's "validate vs normalisedSource should be the same" thread.

Confidence: high

**suggestion (readability):** `apps/studio/src/features/settings/Redirects/types.ts:26,33,43` and `AddRedirectCard.tsx:52` — per the `document-regex-with-verified-comments` convention, every non-trivial regex needs a "what + why" comment. None of the four regexes here are documented. The destination one (`/^(https:\/\/|\/)/`) is the most consequential: a future reader "simplifying" it could accidentally re-allow `http://` (downgrade vector). Worth a one-liner explaining why `http://` is deliberately excluded.

Confidence: high

**issue (tests):** `apps/studio/src/features/settings/Redirects/__tests__/` — only `formatAddedAt` is covered. `features/CLAUDE.md` requires *"at least one happy-path component test"* for new features, and the bulk of the risk surface in this PR is the validation rules (`addRedirectSchema`) and the table's pagination/sort/delete-page-recovery logic. Suggest:
- Schema tests for the boundary cases the PR body explicitly calls out (control chars, backslashes, `..` segments, prefix rules) — easy to write, easy to regress.
- A component test that adds a row → asserts toast + row presence → deletes → confirms it's gone.

Confidence: high

**question:** `apps/studio/src/features/settings/Redirects/types.ts:20-21` — `source.max(1999)` vs `destination.max(2000)`. Is the off-by-one intentional (e.g. reserving a char for the leading `/` after normalisation)? If so, worth a comment; if not, align them.

Confidence: medium

**question:** `apps/studio/src/features/settings/Redirects/types.ts:47` — the schema permits `*` anywhere in the source, but the table only renders the wildcard badge when the source ends with `*` (`RedirectsTable.tsx:97`). If wildcards are conventionally trailing-only, the schema should enforce that (`/^[^*]*\*?$/` or similar) so `/foo/*/bar` isn't silently mis-classified. If non-suffix wildcards are intended later, the badge logic needs to broaden. Either way, schema + UI should agree on what "wildcard" means.

Confidence: medium

**suggestion (non-blocking):** `apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx:48-49` — `watch(["source", "destination"])` triggers a full re-render of the card on every keystroke just to compute `isAddDisabled`. `formState.isValid` (with `mode: "onChange"`) gives the same disable behaviour without the watch and is what the rest of the codebase reaches for. Minor.

Confidence: medium

**note (non-blocking):** `apps/studio/src/pages/sites/[siteId]/settings/redirects.tsx:13-28` — this page re-implements feature-flag gating (returns `null` while gb loads, `router.replace` when disabled) that no other settings page under `apps/studio/src/pages/sites/[siteId]/settings/` does. Once the flag is on for everyone the gate goes away, so it's fine to land — but consider whether `useIsRedirectionsEnabled` should just gate visibility in `SettingsSidenav` (it likely already does at `SettingsSidenav.tsx:51`) and let the page render unconditionally, so this isn't a special-snowflake page.

Confidence: low — depends on whether direct-URL access by flag-disabled users is part of the threat model.

**praise:** Tests follow AAA (`__tests__/utils.test.ts`) and the form is wired through `useZodForm` per the forms convention. Splitting `DeleteRedirectModal` out as a separate component rather than overloading a generic modal with role-specific props is consistent with the `react-new-component-over-prop-overload` smell — nice.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { useZodForm } from "~/lib/form"
+
+import { useCreateRedirect } from "../api"
+import { addRedirectSchema, type AddRedirectInput } from "../types"
+
+interface AddRedirectCardProps {
+  siteId: number
+}
+
+export const AddRedirectCard = ({
```
seaerchin, line 32: comment: haven't read the remaining stuff yet, but just putting some stuff down here that i feel are important

1. circular redirects
2. non utf-8 redirects (eg: chinese/tamil characters)
3. collision of redirects (+ reconciliation - should it be user driven or we auto-decide first)
4. escaped characters - we need to sanitise, especially for reserved chars like /

backend downstream probably needs to do some form of normalisation and enforce it. other runtime errors that we should be wary about are, perhaps, linking to an offline domain but idt that is part of our scope for redirections.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+        onSubmit={handleSubmit(onSubmit)}
+      >
+        <FormControl flex={1} isInvalid={!!errors.source} isRequired>
+          <FormLabel>When someone visits</FormLabel>
+          <InputGroup size="xs">
+            <InputLeftAddon
+              borderColor="base.divider.strong"
+              bgColor="interaction.support.disabled"
+            >
+              /
```
seaerchin, line 99: comment: i think we need a info label to prevent users from directing from privileged internal routes (/_next/asdf for example) to other routes

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+interface RedirectsTableProps {
+  siteId: number
+}
+
+export const RedirectsTable = ({
+  siteId,
+}: RedirectsTableProps): JSX.Element => {
+  const toast = useToast(BRIEF_TOAST_SETTINGS)
+  const { mutate: deleteRedirect, isPending } = useDeleteRedirect()
```
seaerchin, line 209: comment: should we do this eagerly? seems low risk and users might want to delete many redirects at once which might be better ux if we do it eagerly

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+            <Icon as={BiWrench} boxSize="1rem" />
+          </Center>
+          <Text as="h1" textStyle="h3">
+            Redirects
+          </Text>
+        </Flex>
+        <Text textStyle="body-2" color="base.content.medium">
+          When someone visits a link that is no longer in use, redirects send
+          them elsewhere so they don&apos;t get lost. Learn{" "}
+          <Link variant="inline" href="https://support.isomer.gov.sg">
```
seaerchin, line 25: nit: should link to the support page for it directly if available; don't want to make them do extra work when we are making a ask for the user

apps/studio/src/features/settings/Redirects/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  id: string
+  source: string
+  destination: string
+  publishedAt: Date
+}
+
+export type RedirectSortField = "source" | "destination" | "publishedAt"
+
+const hasInvalidChars = (val: string) =>
+  Array.from(val).some((c) => c.charCodeAt(0) < 32 || c === "\\")
```
seaerchin, line 15: question: wouldn't regex be more understandable here ._.

apps/studio/src/features/settings/Redirects/types.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const sourceSchema = z
+  .string()
+  .min(1, "Source path is required")
+  .max(1999, "Source path is too long")
```
seaerchin, line 20: issue: accoding to aws docs, https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-choose, s3 object keys have max length of 1024.

suggest reducing to 100, so that we can expand if required and also to account for expansion of non latin letters into their respective encodings (eg: a chinese word turns into 4 latin chars in hanyu pinyin)

seaerchin (approved):
no major issues; ok w/ approving but length limits is a simple fix (albeit very unlikely)

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: These are handled in #2466 now that the real schema + service exist:

- **Circular**: `refineSourceDestinationDiffer` rejects self-redirects; the `validate` preflight returns a `REDIRECT_LOOP` error for longer chains.
- **Collision + reconciliation**: creating an existing source is rejected with `ALREADY_EXISTS`; a destination that is itself another redirect's source surfaces a non-blocking `DESTINATION_IS_REDIRECT_SOURCE` warning — so it's user-driven, not auto-decided.
- **Escaped/reserved chars**: `INVALID_PATH_CHARS_REGEX` rejects control chars + backslashes, and `normalizeRedirectPath` collapses slashes to a canonical form before persisting.
- **Non-UTF-8**: only control chars/backslashes are rejected. CJK/Tamil are valid UTF-8 and intentionally allowed — flag if you want punycode/percent-encoding normalisation, but I'd keep that out of scope unless the publish layer needs it.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: ~Leaving this out of scope for now: redirect sources are matched against site content paths, which can't collide with framework-internal routes like `/_next/*`, so a redirect from one wouldn't take effect anyway. Happy to add a denylist later if we see a real need.~

Fking claude hallucinating, let me address this inside the validation PR #2466.

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
dcshzj: We're intentionally not doing optimistic delete here. Deletes need to be confirmed as persisted before we let the user add new redirects — applying the change eagerly in the UI ahead of the server risks write conflicts (e.g. re-adding a source whose delete hasn't actually landed yet). So the flow stays: confirm → server delete → then the table reflects it. Bulk/multi-select delete is separate net-new UX that I'll raise as a follow-up rather than expand this stack.

apps/studio/src/features/settings/Redirects/components/RedirectsHeader.tsx
dcshzj: The support article isn't ready yet, so this stays on the generic support landing page for now and will be updated with the direct link once the article is published.

apps/studio/src/features/settings/Redirects/types.ts
dcshzj: Agreed — the char-by-char `charCodeAt` loop is now a single `INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/` in #2466.

apps/studio/src/features/settings/Redirects/types.ts
dcshzj: Done in #2466, with a split: source is capped at 100 (it becomes the S3 object key, and 100 leaves headroom for percent-encoding expansion of non-latin chars under the 1024 limit). Destination keeps a 2000 cap — an external `https://` URL with a query string is legitimately long, so 100 there would reject valid inputs. `MAX_REDIRECT_SOURCE_LENGTH` / `MAX_REDIRECT_DESTINATION_LENGTH` in `schemas/redirect.ts`, with boundary tests.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Added in https://github.com/opengovsg/isomer/commit/2e1f8e5d2ff0a6a0d4f285a4e3484f624e443b20

PR #2051, opened 2026-05-12
Potential fix for 1 code quality finding

adriangohjw:
_This PR applies 1/2 suggestions from code quality [AI findings](https://github.com/opengovsg/isomer/security/quality/ai-findings). 1 suggestion was skipped to avoid creating conflicts._


dcshzj (approved):
I think this is okay but we should take a stand if we want to rely on the globals or the named imports, then either remove the globals or suppress the finding

PR #2073, opened 2026-05-13
chore(studio): disable "Move to..." for default Search page

adriangohjw:
## Problem

The Search resource (`permalink === "search"` at root) already has its **Delete** action disabled in the resource dropdown menus, but **Move to...** was still active. Moving the default Search page would put it in an unexpected location and break the site's search experience.

## Solution

Mirror the existing disabled-Delete treatment for **Move to...** in both dropdown menus that render this action:

- `apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx` — when `isSearchPage`, render a disabled `Move to...` MenuItem with the tooltip `"This is a default page that cannot be moved."`; otherwise keep the existing `Can do="move"`-gated active item.
- `apps/studio/src/features/dashboard/components/CollectionTable/CollectionTableMenu.tsx` — same ternary on `isSearchPage`.

The tooltip wording follows the parallel pattern already used for Delete (`"This is a default page that cannot be removed."`).

**Breaking Changes**

**Features**:

- N/A

**Improvements**:

- Default Search page can no longer be moved via the dashboard dropdown menus, with a tooltip explaining why.

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] In Studio, open a site dashboard that contains the default Search page (permalink `search` at root level).
- [ ] Click the `...` (options) dropdown on the Search row in the resource table.
- [ ] Verify `Move to...` appears greyed out / disabled with the tooltip "This is a default page that cannot be moved." on hover.
- [ ] Verify `Delete` is also still disabled with its existing tooltip.
- [ ] Open the dropdown on a non-Search page/folder and confirm `Move to...` is still active and functional (clicking opens the move modal).
- [ ] Repeat the same checks inside a collection view (CollectionTable dropdown).

### DIFF
--- a/apps/studio/src/features/dashboard/components/CollectionTable/CollectionTableMenu.tsx
+++ b/apps/studio/src/features/dashboard/components/CollectionTable/CollectionTableMenu.tsx
@@ -68,13 +68,23 @@ export const CollectionTableMenu = ({
               Edit settings
             </MenuItem>
           )}
-          <MenuItem
-            as="button"
-            icon={<BiFolderOpen fontSize="1rem" />}
-            onClick={handleMoveResourceClick}
-          >
-            Move to...
-          </MenuItem>
+          {isSearchPage ? (
+            <MenuItem
+              isDisabled
+              icon={<BiFolderOpen fontSize="1rem" />}
+              tooltip="This is a default page that cannot be moved."
+            >
+              Move to...
+            </MenuItem>
+          ) : (
+            <MenuItem
+              as="button"
+              icon={<BiFolderOpen fontSize="1rem" />}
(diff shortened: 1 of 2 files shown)


seaerchin (approved):
comment: we should also disallow this on the router level at the backend. right now this is frontend only, meaning that calling the API directly w/ the session token still work

adriangohjw:
> comment: we should also disallow this on the router level at the backend. right now this is frontend only, meaning that calling the API directly w/ the session token still work

Yeah agree, just a hotfix for now to prevent agency users weird behaviours hahaha

PR #2039, opened 2026-05-11
chore: update to imda

seaerchin:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [ ] No - this PR is backwards compatible

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**Manual Verification Steps**:

- [ ] Step 1
- [ ] Step 2
- [ ] Step 3

**New scripts**:

- `script` : script details

**New dependencies**:

- `dependency` : dependency details

**New dev dependencies**:

- `dependency` : dependency details

PR #2074, opened 2026-05-13
feat: gazette deletion and push document job functionality

seaerchin:
## Problem

The gazette system needed the ability to delete gazettes and schedule documents for indexing in SearchSG.

## Solution

**Breaking Changes**

**Features**:

- Added gazette deletion functionality with email notifications to admins when a gazette is deleted
- Added a new `PushDocumentJob` table and cron job to schedule gazette documents for ingestion into SearchSG
- Added PDF text extraction using `pdfreader` library to support full-text search indexing
- Created dedicated gazette S3 bucket configuration (`S3_GAZETTE_BUCKET_NAME`, `S3_GAZETTE_DOMAIN_NAME`)

**Improvements**:

- Refactored SearchSG integration into a dedicated service (`searchsg.service.ts`)
- Moved gazette-specific S3 operations to `gazette.service.ts` for better isolation
- Added `useUploadGazetteMutation` hook for gazette-specific uploads
- Updated ViewGazetteModal to support deletion workflow with confirmation

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- Added unit tests for `schedulePushDocumentJob` handler
- Added integration tests for gazette router deletion endpoints

**New scripts**:

- N/A

**New dependencies**:

- `pdfreader` (^3.0.8): Used for parsing PDF content for full-text search indexing

**New dev dependencies**:

- N/A


seaerchin:
* **#2074** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2074?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2027** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2027?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2021** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2021?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2039](https://github.com/opengovsg/isomer/pull/2039) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2039?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

seaerchin (commented):
## Summary

This PR adds gazette deletion functionality with a 15-minute grace period and email notifications, plus refactors the SearchSG integration into a shared service. The codeflow:

1. `ViewGazetteModal` checks if deletion is allowed based on `publishedAt` timestamp (`apps/studio/src/features/gazettes/components/ViewGazetteModal/ViewGazetteModal.tsx:214-216`)
2. `gazette.delete` procedure validates Toppan/admin access, checks grace period, removes from SearchSG, deletes S3 asset, then deletes DB resource (`apps/studio/src/server/modules/gazette/gazette.router.ts:700-810`)
3. Email notifications are sent to site admins after deletion (`apps/studio/src/server/modules/gazette/gazette.router.ts:783-808`)

**Overall assessment**: Solid implementation with good test coverage. A few issues worth addressing before merge.

---

## Findings

**issue (security):** `apps/studio/src/server/modules/gazette/gazette.service.ts:4` — Importing `logger` from `storybook/internal/client-logger` in server-side code. This is a Storybook internal module not intended for production use. It may not be available in production builds and could leak client-side implementation details. Should use the existing `createBaseLogger` from `~/lib/logger` instead.

Confidence: high

---

**issue:** `apps/studio/src/server/modules/gazette/gazette.service.ts:196` — Debug `console.log` left in production code. The JSON response is logged after a failed SearchSG removal. This should be removed or converted to proper structured logging via the logger.

Confidence: high

---

**issue:** `apps/studio/src/server/modules/gazette/gazette.router.ts:763-764` — External service calls (`removeGazetteFromSearchIndex`, `deleteGazetteAsset`) happen BEFORE the database transaction. If the DB delete fails (e.g., FK constraint, connection issue), the gazette is already removed from SearchSG and S3 but still exists in the database — leaving an inconsistent state where users can see the gazette in the UI but the file is gone.

Consider either:
1. Moving external calls inside the transaction and using a saga/compensation pattern on rollback
2. Doing external calls AFTER the DB transaction commits (current `cancelScheduledPublish` does this pattern at line 611-625)

Confidence: high

---

**suggestion:** `apps/studio/src/server/modules/gazette/gazette.router.ts:785-790` — The feature flag lookup uses `gazetteSiteId` to find admins, but if the feature flag isn't configured or returns empty strings, `Number("")` returns `0`, and the query will find no admins (or potentially wrong admins if site id 0 exists). Consider adding a guard:

```typescript
if (!gazetteSiteId) {
  ctx.logger.warn("EGAZETTE_INFO_FEATURE_KEY not configured, skipping admin notification")
  return
}
```

Confidence: medium

---

**suggestion:** `apps/studio/src/server/modules/gazette/gazette.service.ts:145-160` — `parseFullTextFromPDF` creates a new `pdfReader` instance at module load (`line 30`) but the `parseBuffer` callback pattern suggests the reader may accumulate state between calls. For a cron job processing multiple PDFs, consider creating a fresh `PdfReader` instance per call to avoid potential state leakage.

Confidence: low

---

**note:** `apps/studio/src/components/Datatable/Datatable.tsx:59-60` — All rows are now clickable (`cursor: "pointer"`) and always trigger `onRowClick`. Previously, `isRowClickable` allowed conditional behavior. If any consumers relied on non-clickable rows, this could be a subtle behavior change.

Confidence: medium

---

**praise:** Good test coverage for the deletion flow in `apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts:328-911` — tests the grace period boundary conditions (10 min, exactly 15 min, 16 min), access control for non-Toppan users, and edge cases like missing published version.

adriangohjw (commented):
## Summary

This PR adds gazette deletion with a 15-minute grace period, refactors SearchSG integration into a dedicated service, and adds a `PushDocumentJob` cron pipeline for ingesting gazette PDFs into SearchSG. The flow: `GazetteTable` opens `ViewGazetteModal` for published rows → `canDelete` check based on `publishedAt` → `gazette.delete` tRPC procedure validates Toppan/admin access, checks grace period, removes from SearchSG + S3, deletes DB record, then emails site admins.

The prior review comments (Storybook logger, console.log) appear fixed in the CI-fix commit. Good test coverage on the deletion boundary conditions. A few real issues below.

---

## Findings

**issue (bug):** `apps/studio/src/server/modules/gazette/gazette.service.ts:L252-L260` — `deleteGazetteAsset` swallows S3 deletion failures entirely (just a `logger.warn`). At that point in the router, `removeGazetteFromSearchIndex` has already succeeded and the DB delete is about to happen. If S3 throws, the user receives a success response, the gazette disappears from SearchSG and the internal UI, but the PDF file remains at its public S3 URL indefinitely — anyone with the direct link can still access it. Given the deletion feature exists precisely to remove a gazette from public view, a silent S3 failure defeats the purpose. Either surface the error to the caller or, at minimum, upgrade the log to `error` and explicitly note the file is still publicly accessible.

Confidence: high

---

**issue:** `apps/studio/src/server/modules/gazette/gazette.router.ts:L861-L864` — The delete transaction is missing the `siteId` scope present everywhere else. Compare `cancelScheduledPublish` at line 685–690 which includes `.where("siteId", "=", siteId).returningAll()`. The auth checks upstream should prevent cross-site deletion in practice, but skipping the defense-in-depth constraint is inconsistent and leaves a latent footgun if the authorization layer ever changes. Should match the pattern.

Confidence: high

---

**issue:** `apps/studio/src/server/modules/gazette/gazette.router.ts:L886` / `apps/studio/src/features/mail/templates/templates.ts:L31` — `fileId: ref` passes the full S3 key (e.g. `/2024/Government Gazette/gazette/GN_1234.pdf`) to `sendGazetteDeletionEmail`. The email subject renders as "The gazette with file id: /2024/Government Gazette/gazette/GN_1234.pdf and title: ..." — recipients see an internal S3 path rather than the human-readable file ID displayed in the UI. The intent is probably `ref.split("/").pop()` (just the filename).

Confidence: high

---

**issue:** `apps/studio/src/server/modules/gazette/gazette.router.ts:L868-L882` — When `EGAZETTE_INFO_FEATURE_KEY` is not configured or returns its default (`siteId: ""`), `Number("")` evaluates to `0`. The Site query finds no admins and the `Promise.all` succeeds silently — no emails sent, no warning logged. A guard would make this visible:

```ts
if (!gazetteSiteId) {
  ctx.logger.warn("EGAZETTE_INFO_FEATURE_KEY siteId unset — skipping admin notification")
  return
}
```

Confidence: medium

---

**note:** `apps/studio/src/server/modules/gazette/gazette.router.ts:L842-L865` — The comment justifies running `removeGazetteFromSearchIndex` and `deleteGazetteAsset` before the DB transaction. That reasoning holds for SearchSG, but combined with `deleteGazetteAsset` silently swallowing errors (finding #1), there is a path where SearchSG removal succeeds, S3 deletion silently fails, then the DB delete succeeds — leaving the file publicly accessible with no indication it was not fully cleaned up.

Confidence: high

---

**praise:** `apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts:L759-L963` — Solid test coverage: boundary conditions at 10 min, exact 15 min, 16 min; access control for non-Toppan/non-admin; gazette without a published version; IsomerAdmin Core happy path. The external service mocking via `vi.spyOn` is clean.

apps/studio/src/features/gazettes/components/ViewGazetteModal/ViewGazetteModal.tsx
```diff
@@ -44,29 +51,58 @@ interface ViewGazetteModalProps extends Pick<
...
   onClose,
+  siteId,
   gazetteId,
   data,
   initialView = "view",
 }: ViewGazetteModalProps): JSX.Element => {
+  const { subcategoryMap } = useGazetteSubcategoriesContext()
+
+  const canDelete = data.publishedAt
+    ? differenceInMinutes(new Date(), data.publishedAt) <= 15
```
adriangohjw: ask: can add comment or use constant instead of magic number 15? thanks!

i think you did that in gazette.router.ts as ALLOWED_GAZETTE_DELETION_TIMEFRAME_IN_MINUTES

apps/studio/src/features/gazettes/components/ViewGazetteModal/ViewGazetteModal.tsx
```diff
@@ -44,29 +51,58 @@ interface ViewGazetteModalProps extends Pick<
...
+    : false
+
+  const subcategoryLabel = subcategoryMap[data.subcategory] ?? data.subcategory
+
   const [view, setView] = useState<ModalView>(initialView)
   const [isConfirmed, setIsConfirmed] = useState(false)
-  const [isDeleting, setIsDeleting] = useState(false)
+
+  const toast = useToast()
```
adriangohjw, line 72: nit (non-blocking): can pass BRIEF_TOAST_SETTINGS here directly once

apps/studio/src/features/mail/templates/templates.ts
```diff
@@ -24,6 +25,21 @@ import type {
...
+  return {
+    subject: `[Isomer Studio] The gazette with file id: ${fileId} and title: ${gazetteTitle} has been deleted`,
+    body: `<p>Hi ${recipientEmail},</p>
+<p>The gazette ${gazetteTitle} has been deleted from your site and removed from the search results</p>
+<p>If you believe this was a mistake or need assistance, please contact <a href="${ISOMER_SUPPORT_LINK}">${ISOMER_SUPPORT_EMAIL}</a>.</p>
+<p>Best,</p>
+<p>Isomer team</p>`,
+  }
+}
+
```
adriangohjw, line 48: issue: can sanitize as per https://github.com/opengovsg/isomer/pull/2223? thanks HAHA!

apps/studio/src/env.mjs
```diff
@@ -111,6 +110,7 @@ const processEnv = {
     process.env.NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY,
   NEXT_PUBLIC_INTERCOM_APP_ID: process.env.NEXT_PUBLIC_INTERCOM_APP_ID,
   SEARCHSG_API_KEY: process.env.SEARCHSG_API_KEY,
+  EGAZETTE_DOCUMENT_INDEX: process.env.EGAZETTE_DOCUMENT_INDEX,
```
adriangohjw: curious/nit: why move it down ah? i thought it being with `S3_GAZETTE_BUCKET_NAME` and `S3_GAZETTE_DOMAIN_NAME` will be more intuitive

apps/studio/src/server/modules/searchsg/searchsg.service.ts
```diff
@@ -142,6 +143,10 @@ export const updateSearchSGConfig = async (
   return res
 }
 
+export const generateDocumentId = (url: string, resourceId: string) => {
+  return `${url}-${resourceId}`
+}
```
adriangohjw: question: this seems quite specific to egazette? not sure if it makes sense to be in searchsg.service.ts?

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -771,4 +780,115 @@ export const gazetteRouter = router({
...
+        .selectAll()
+        .executeTakeFirstOrThrow(() => new TRPCError({ code: "BAD_REQUEST" }))
+
+      // Next, fetch the gazettes that they are trying to delete and
+      // make sure that the time is within the allowed timeframe.
+      // This has to be on the version already cos otherwise they should be able to cancel the publish
+      const gazette = await db
+        .selectFrom("Resource")
+        .innerJoin("Version", "Version.id", "Resource.publishedVersionId")
+        .where("Resource.id", "=", String(gazetteId))
```
adriangohjw, line 808: nit: defence-in-depth - to add filter for siteId

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -771,4 +780,115 @@ export const gazetteRouter = router({
...
+      })
+      // NOTE: Send email out to IMDA so that they get visibility on what gazettes are deleted
+      const gb = ctx.gb
+      const { siteId: gazetteSiteId } = gb.getFeatureValue(
+        EGAZETTE_INFO_FEATURE_KEY,
+        {
+          siteId: "",
+          gazettesCollectionId: "",
+        },
+      )
```
adriangohjw: question: why need growthbook? we are able to know gazetteSiteId from ctx.siteId?

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -771,4 +780,115 @@ export const gazetteRouter = router({
...
+      const admins = await db
+        .selectFrom("Site")
+        .where("Site.id", "=", Number(gazetteSiteId))
+        .innerJoin("ResourcePermission", "Site.id", "ResourcePermission.siteId")
+        .innerJoin("User", "ResourcePermission.userId", "User.id")
+        .where("ResourcePermission.role", "=", "Admin")
+        .select("User.email")
+        .execute()
```
adriangohjw, line 901: issue: need to check that both User and ResourcePermission is not deleted (`deletedAt` is null)

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -771,4 +780,115 @@ export const gazetteRouter = router({
...
+            before: { resource: gazette, blob },
+            after: null,
+          },
+        })
+
+        await tx
+          .deleteFrom("Resource")
+          .where("id", "=", String(gazetteId))
+          .execute()
+      })
```
adriangohjw: nit(non-blocking): i think should switch the ordering of these 2 operations - we should only log after it is deleted (though shouldnt matter since its in a tx)

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -771,4 +780,115 @@ export const gazetteRouter = router({
...
+          by: user,
+          delta: {
+            before: { resource: gazette, blob },
+            after: null,
+          },
+        })
+
+        await tx
+          .deleteFrom("Resource")
+          .where("id", "=", String(gazetteId))
```
adriangohjw: defence-in-depth: should add siteId filter here

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -771,4 +780,115 @@ export const gazetteRouter = router({
...
+          message: "Gazette does not have a valid S3 reference",
+        })
+      }
+
+      // Always try to delete the asset and search record first
+      // before proceeding to remove the database resource.
+      // This is because the public uses those to access the gazette
+      // but the database resource is purely for internal view.
+      await removeGazetteFromSearchIndex(ref, gazette.id)
+      await deleteGazetteAsset(ref)
```
adriangohjw, line 848: question: what happens when this happened but the subsequent queries fail?

adriangohjw (changes requested):

apps/studio/src/features/gazettes/components/ViewGazetteModal/ViewGazetteModal.tsx
seaerchin: Done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — lifted `ALLOWED_GAZETTE_DELETION_TIMEFRAME_IN_MINUTES` into `~/constants/gazette` so the modal and the router share the same source.

apps/studio/src/features/mail/templates/templates.ts
seaerchin: Done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — `escapeHtml` applied to `fileId`, `gazetteTitle`, and `recipientEmail` (matches the #2223 pattern).

apps/studio/src/env.mjs
seaerchin: Done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — moved it back next to `S3_GAZETTE_BUCKET_NAME` / `S3_GAZETTE_DOMAIN_NAME`.

apps/studio/src/server/modules/searchsg/searchsg.service.ts
seaerchin: Good call — done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c). `generateDocumentId` moved into `gazette.service.ts` (the only consumers are gazette delete + the gazette push job). Updated `schedulePushDocumentJob` to import from the new location too.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — added `.where("Resource.siteId", "=", siteId)` on the fetch, matching the `cancelScheduledPublish` pattern.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Good catch — you're right, the input `siteId` and the growthbook `siteId` are the same value for this procedure, and the growthbook default (`siteId: ""` → `Number("") === 0`) would have silently sent zero emails. Dropped the growthbook lookup entirely in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c); now uses the input `siteId` directly.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — added `.where("User.deletedAt", "is", null)` and the same on `ResourcePermission.deletedAt`.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Agreed — fixed in [00dc9b64](https://github.com/opengovsg/isomer/pull/2074/commits/00dc9b644). Swapped the order in both this `delete` procedure and `cancelScheduledPublish` so the audit log entry describes a deletion that has actually happened. AuditLog has no FK on `resourceId` so it's safe inside the same transaction.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Done in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — added `.where("siteId", "=", siteId)` on the `deleteFrom("Resource")` too.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Addressed in [aa9c073c](https://github.com/opengovsg/isomer/pull/2074/commits/aa9c073c) — `deleteGazetteAsset` now throws on S3 failure instead of swallowing. So the new order `removeGazetteFromSearchIndex` → `deleteGazetteAsset` → DB tx means any failure aborts before the DB delete; we never end up with a deleted DB row and a still-public PDF.

There's still a small window where SearchSG removal succeeds but S3 fails (gone from search, still at the direct URL) — fully closing that gap needs a saga/compensation pattern, which felt over-engineered for a 15-min Toppan-only grace window. Happy to revisit if you'd rather have it.

adriangohjw (commented):
## Summary

This PR adds gazette deletion (15-minute grace period, SearchSG + S3 cleanup, admin email notification) and refactors the SearchSG integration into a shared service. The prior two reviews caught and fixed the main issues (S3 silent failure, missing siteId scope, full S3 key in email subject, stale Storybook logger import). Good test coverage on the deletion boundary conditions and access control.

**One new critical bug** that the tests missed — details below.

---

## Findings

**issue (bug, critical):** `apps/studio/src/server/modules/gazette/gazette.router.ts:L853-L869` — The `delete` transaction deletes the `Resource` directly without first deleting the associated `PushDocumentJob`. The schema defines `PushDocumentJob.resource` with `onDelete: Restrict` (see `prisma/schema.prisma`), so the `deleteFrom("Resource")` will fail with a FK constraint error whenever a `PushDocumentJob` row still exists for that gazette.

In practice this hits every time: the cron runs every minute, so for up to ~60 seconds after publication the job is still present. Since the grace period is 15 minutes and eGazette uses immediate-publish (`scheduledAt ≈ now`), the very first deletion attempt in production will 500.

The tests miss this because `seedPublishedGazette` creates a Resource + Version but **no** `PushDocumentJob` — so the FK constraint is never exercised.

Fix mirrors `cancelScheduledPublish` (lines 633–643): delete the `PushDocumentJob` first inside the same transaction, before deleting the `Resource`:

```ts
await db.transaction().execute(async (tx) => {
  // 1. Delete PushDocumentJob (FK onDelete: Restrict — must go before Resource)
  await tx
    .deleteFrom("PushDocumentJob")
    .where("resourceId", "=", String(gazetteId))
    .execute()

  // 2. Delete the Resource
  await tx
    .deleteFrom("Resource")
    .where("siteId", "=", siteId)
    .where("id", "=", String(gazetteId))
    .execute()

  await logResourceEvent(...)
})
```

A matching test case would be: create via `gazette.create` (which inserts the PushDocumentJob), do NOT let the cron fire, then call `gazette.delete` — it should succeed.

Confidence: high

---

**issue (bug):** `apps/studio/src/components/Datatable/Datatable.tsx:L119` — `cursor={"pointer"}` is now hardcoded for every row. `UserTable`, `ResourceTable`, and `CollectionTable` all call `<Datatable>` without passing `onRowClick` — clicking their rows does nothing, but the pointer cursor now misleadingly suggests they are clickable. This is a regression on three non-gazette tables.

Fix: make cursor conditional on the prop:

```tsx
cursor={onRowClick ? "pointer" : undefined}
onClick={() => onRowClick?.(row)}
```

Confidence: high

---

**suggestion:** `apps/studio/src/server/modules/gazette/gazette.service.ts:L34` — `pdfReader` is a module-level singleton. `schedulePushDocumentJobHandler` processes multiple PDFs concurrently via `Promise.all` (line 150 in `schedulePushDocumentJob.ts`), which means multiple `parseBuffer` calls run on the same `PdfReader` instance simultaneously. `pdfreader` is built on top of `pdf2json` whose C++ bindings maintain internal state; concurrent calls on one instance can produce interleaved results or corrupt internal state.

Consider creating a fresh `PdfReader` instance per call:
```ts
export const parseFullTextFromPDF = async (pdfBuffer: Uint8Array) => {
  const reader = new PdfReader({})
  const data: string[] = await new Promise((resolve, reject) => {
    ...
    reader.parseBuffer(...)
  })
  return data.join(" ")
}
```

Confidence: low (depends on pdfreader internals not exposed in source)

---

**note:** `apps/studio/src/server/modules/gazette/gazette.router.ts:L847-L848` — The ordering `removeGazetteFromSearchIndex` → `deleteGazetteAsset` is intentional (the comment explains it), but there is a narrow failure window: if SearchSG removal succeeds and the S3 soft-delete fails, the gazette is deindexed from SearchSG but still publicly accessible. On retry within the grace period this is fine (both calls are idempotent). If the transient S3 failure eats enough of the 15-minute window that the retry lands past the deadline, the gazette ends up permanently deindexed but still reachable at its S3 URL with no UI recovery path.

Not blocking, but worth documenting (or adding a compensating `logger.error` that names the S3 URL to aid manual recovery).

Confidence: medium

---

**praise:** `apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts:L769-L963` — Solid boundary tests (10 min, exact 15 min, 16 min), access control coverage, and clean `vi.spyOn` mocking for the external service calls. The `seedPublishedGazette` helper is easy to read and reuse.

seaerchin:
Thanks for catching the FK bug — that one would have 500'd in prod on the first delete. Addressed in [d911edbd](https://github.com/opengovsg/isomer/pull/2074/commits/d911edbd9):

**issue (critical) — PushDocumentJob FK constraint:** Fixed. The `delete` transaction now deletes the `PushDocumentJob` first (scoped via `Resource.siteId` subquery, defence-in-depth) before the `Resource`. Added a regression test (`deletes a gazette that still has an unprocessed PushDocumentJob`) that seeds an active job and verifies both the resource and the job are gone after delete. 26/26 router tests pass.

**issue — Datatable cursor regression:** Fixed. `cursor` is now `onRowClick ? "pointer" : undefined`, so `UserTable` / `ResourceTable` / `CollectionTable` (which don't pass `onRowClick`) no longer mislead.

**suggestion — PdfReader singleton:** Fixed. `parseFullTextFromPDF` now instantiates a fresh `PdfReader` per call. You're right that the cron uses `Promise.all` over multiple documents — sharing the instance was a latent corruption risk under concurrency. Took the conservative path.

**note — SearchSG → S3 partial-failure window:** Added a `logger.error` inside `deleteGazetteAsset` that surfaces the public S3 URL when the soft-delete throws. Ops now has a recovery target without having to derive the URL from logs piecewise. Full saga/compensation still feels over-engineered for a 15-min Toppan-only window, but happy to revisit if you want it tracked.

**praise — test coverage:** ty 🙏 also added the FK test mentioned above to keep this case from regressing.

Re-requesting review.

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -252,11 +256,23 @@ export const removeGazetteFromSearchIndex = async (
...
-  })
+  const key = ref.slice(1) // Remove leading slash
+  try {
+    await markFileAsDeleted({ key })
+  } catch (err) {
+    const publicUrl = `https://${env.S3_GAZETTE_DOMAIN_NAME}${ref}`
+    logger.error(
+      { err, key, publicUrl },
+      "Failed to soft-delete gazette in S3; the file may still be publicly reachable at publicUrl until manually cleaned up",
+    )
```
adriangohjw, line 273: non-blocking: shall we set a DD monitor?

adriangohjw (approved):
LGTM

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: sounds good, planning to repro existing ones and add abit more for failure modes

PR #2075, opened 2026-05-13
chore(studio): gazette form UX improvements and duplicate file ID prevention

seaerchin:
## Problem

When creating gazette entries, users had to manually type the File ID field even though they just uploaded a file with a name. Additionally, long file IDs would overflow in the gazette table, making them difficult to read. Furthermore, there was no validation to prevent duplicate file IDs within the same gazette collection, which could lead to confusion and data integrity issues.

## Solution

**Breaking Changes**

**Improvements**:

- Auto-populate the File ID field with the uploaded PDF filename when a file is selected, reducing manual data entry
- File ID can still be edited manually after auto-population if needed
- Long file IDs now wrap properly in the gazette table instead of overflowing

**Bug Fixes**:

- Added duplicate file ID validation in both `create` and `update` gazette mutations
- The duplicate check uses `COALESCE(PublishedBlob, DraftBlob)` to prefer the published blob when checking for existing file IDs, ensuring accurate detection even when drafts differ from published versions
- Returns a `CONFLICT` error with a clear message when a duplicate file ID is detected

## Before & After Screenshots

**BEFORE**:
- File ID field required manual entry after uploading a file
- Long file IDs could overflow in table cells
- Could create gazettes with duplicate file IDs

**AFTER**:
- File ID field is automatically filled with the uploaded filename
- Long file IDs wrap within their table cells using `wordBreak="break-all"`
- Attempting to create/update a gazette with a duplicate file ID returns an error

## Tests

Run gazette router unit tests:
```bash
npm run test:unit -- src/server/modules/gazette/__tests__/gazette.router.test.ts
```

Manual testing:
1. Open gazette creation modal
2. Upload a PDF file
3. Verify the File ID field is auto-populated with the filename
4. Verify the File ID field can still be edited
5. Create a gazette with a long filename
6. Verify the file ID wraps properly in the table view
7. Attempt to create a gazette with the same file ID as an existing one
8. Verify an error is shown preventing the duplicate

**New scripts**: None

**New dependencies**: None

### DIFF
--- a/apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
+++ b/apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
@@ -186,6 +186,9 @@ export const GazetteFormFields = ({
           onChange={(newFile) => {
             setFile(newFile)
             onFileChange?.(newFile)
+            if (newFile) {
+              setValue("fileId", newFile.name)
+            }
           }}
         />
       </FormControl>
--- a/apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
+++ b/apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
@@ -21,6 +21,7 @@ import { TimeSelect } from "~/components/Select/TimeSelect"
 import type { GazettesCategory } from "../../types"
 import { GAZETTE_CATEGORIES } from "../../constants"
 import { useGazetteSubcategoriesContext } from "../../contexts/GazetteSubcategoriesContext"
+import { toFileId } from "../../utils/toFileId"
 
 type GazetteFormData = CreateGazetteInput
 
@@ -187,15 +188,7 @@ export const GazetteFormFields = ({
             setFile(newFile)
             onFileChange?.(newFile)
             if (newFile) {
-              // Sanitize filename to match fileId regex: ^[_\-a-zA-Z0-9]+\.pdf$
-              // 1. Remove .pdf extension (we'll add it back)
-              // 2. Replace spaces with hyphens
-              // 3. Strip any characters not in [_\-a-zA-Z0-9]
-              const baseName = newFile.name.replace(/\.pdf$/i, "")
-              const sanitized =
-                baseName.replace(/\s+/g, "-").replace(/[^_\-a-zA-Z0-9]/g, "") ||
-                "file"
-              setValue("fileId", `${sanitized}.pdf`, {
+              setValue("fileId", toFileId(newFile.name), {
                 shouldValidate: true,
                 shouldDirty: true,
               })
--- a/apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
+++ b/apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
@@ -187,7 +187,18 @@ export const GazetteFormFields = ({
             setFile(newFile)
             onFileChange?.(newFile)
             if (newFile) {
-              setValue("fileId", newFile.name)
+              // Sanitize filename to match fileId regex: ^[_\-a-zA-Z0-9]+\.pdf$
+              // 1. Remove .pdf extension (we'll add it back)
+              // 2. Replace spaces with hyphens
+              // 3. Strip any characters not in [_\-a-zA-Z0-9]
+              const baseName = newFile.name.replace(/\.pdf$/i, "")
+              const sanitized =
+                baseName.replace(/\s+/g, "-").replace(/[^_\-a-zA-Z0-9]/g, "") ||
+                "file"
+              setValue("fileId", `${sanitized}.pdf`, {
+                shouldValidate: true,
+                shouldDirty: true,
+              })
             }
           }}
         />
--- a/apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts
+++ b/apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts
@@ -227,6 +227,43 @@ describe("gazette.router", async () => {
       // Only the collection itself exists — no link was created.
       expect(resources.map((r) => r.type)).toEqual([ResourceType.Collection])
     })
+
+    it("rejects creation when a gazette with the same file ID already exists", async () => {
+      const { site, collection } = await seedToppanWithCollection()
+
+      // Create first gazette with a specific filename
+      await caller.create({
+        siteId: site.id,
+        collectionId: Number(collection.id),
+        title: "First Notice",
+        permalink: crypto.randomUUID(),
+        ref: "/sites/1/gazettes/uuid1/duplicate-file.pdf",
+        category: "Government Gazette",
+        date: "30/04/2026",
+        tagged: ["sub-1"],
+        scheduledAt: PAST_DATE,
+      })
+
+      // Attempt to create second gazette with the same filename (different path prefix)
+      await expect(
+        caller.create({
+          siteId: site.id,
+          collectionId: Number(collection.id),
+          title: "Second Notice",
+          permalink: crypto.randomUUID(),
+          ref: "/sites/1/gazettes/uuid2/duplicate-file.pdf", // Same filename
+          category: "Government Gazette",
+          date: "30/04/2026",
+          tagged: ["sub-1"],
+          scheduledAt: PAST_DATE,
+        }),
+      ).rejects.toThrowError(
+        new TRPCError({
+          code: "CONFLICT",
+          message: "A gazette with the same file ID already exists",
+        }),
+      )
+    })
   })
 
   describe("update", () => {
@@ -319,6 +356,55 @@ describe("gazette.router", async () => {
         .selectAll()
         .executeTakeFirstOrThrow()
       expect(after.scheduledAt).toEqual(PAST_DATE)
+    })
+
+    it("rejects update when changing to a file ID that already exists", async () => {
+      const { site, collection } = await seedToppanWithCollection()
+
+      // Create first gazette
+      await caller.create({
+        siteId: site.id,
+        collectionId: Number(collection.id),
+        title: "First Notice",
+        permalink: crypto.randomUUID(),
+        ref: "/sites/1/gazettes/uuid1/existing-file.pdf",
+        category: "Government Gazette",
+        date: "30/04/2026",
+        tagged: ["sub-1"],
+        scheduledAt: PAST_DATE,
+      })
+
+      // Create second gazette with a different filename
+      const { gazetteId } = await caller.create({
+        siteId: site.id,
+        collectionId: Number(collection.id),
+        title: "Second Notice",
+        permalink: crypto.randomUUID(),
+        ref: "/sites/1/gazettes/uuid2/different-file.pdf",
+        category: "Government Gazette",
+        date: "30/04/2026",
+        tagged: ["sub-1"],
+        scheduledAt: PAST_DATE,
+      })
+
+      // Attempt to update second gazette to use the same filename as first
+      await expect(
+        caller.update({
+          siteId: site.id,
+          gazetteId: Number(gazetteId),
+          title: "Second Notice",
+          newRef: "/sites/1/gazettes/uuid3/existing-file.pdf", // Same filename as first
+          category: "Government Gazette",
+          date: "30/04/2026",
+          tagged: ["sub-1"],
+          scheduledAt: PAST_DATE,
+        }),
+      ).rejects.toThrowError(
+        new TRPCError({
+          code: "CONFLICT",
+          message: "A gazette with the same file ID already exists",
+        }),
+      )
     })
   })
 })
--- a/apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts
+++ b/apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts
@@ -71,6 +71,7 @@ describe("gazette.router", async () => {
 
   describe("assertGazetteAccess (via gazette.list)", () => {
     it("rejects an ordinary site member with no Toppan email and no admin role", async () => {
+      // Arrange
       const user = await setupUser({
         userId: session.userId ?? undefined,
         email: "user@example.com",
@@ -82,6 +83,7 @@ describe("gazette.router", async () => {
         siteId: site.id,
       })
 
+      // Act & Assert
       await expect(
         caller.list({
           siteId: site.id,
@@ -98,6 +100,7 @@ describe("gazette.router", async () => {
     })
 
     it("allows an IsomerAdmin Core user even without a Toppan email", async () => {
+      // Arrange
       const user = await setupUser({
         userId: session.userId ?? undefined,
         email: "admin@example.com",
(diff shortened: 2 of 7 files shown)


adriangohjw (commented):
## Summary

- Three changes bundled: auto-populate `fileId` on file upload (`GazetteFormFields.tsx:189-191`), wrap long file IDs in the table (`FileIdCell.tsx:49,58`), and reject duplicate file IDs within a collection on both `create` and `update` (`gazette.router.ts:198-232`, `gazette.router.ts:416-451`).
- Duplicate-check codepath: filename is extracted with `ref.split("/").pop()`, then a kysely query joins `Resource` → `DraftBlob` / `PublishedBlob` and compares `split_part(COALESCE(PublishedBlob.content, DraftBlob.content)->'page'->>'ref', '/', -1)` against the candidate filename. Throws `TRPCError({ code: "CONFLICT" })` on a hit. Update path also excludes self (`Resource.id != gazetteId`).
- Overall direction is good and tests cover the happy paths, but a couple of correctness issues are worth fixing before merge — most importantly the `COALESCE` order disagreeing with the list query, and the duplicate check happening *after* the S3 copy in the update flow.

---

## Findings

**issue (bug):** `apps/studio/src/server/modules/gazette/gazette.router.ts:219` and `:438` — the duplicate check uses `COALESCE("PublishedBlob"."content", "DraftBlob"."content")` (prefers published), but the `list` query at `:102,116,121` uses `COALESCE("DraftBlob"."content", "PublishedBlob"."content")` (prefers draft). The user sees the draft filename in the UI, but the check compares against the published filename. Concrete failure mode: gazette A is published as `old.pdf`, then its draft is edited to `new.pdf`. A user creating gazette B with filename `new.pdf` will pass the check (because the check sees A's published `old.pdf`), and the collection ends up with two `new.pdf` drafts. The PR description states the published-preference is intentional, but it conflicts with what "file ID already exists" means to the user. Recommend either (a) prefer draft to match the list view, or (b) check against *both* blobs (`OR`) so neither a draft nor a published filename can collide.

Confidence: high

**issue:** `apps/studio/src/server/modules/gazette/gazette.router.ts:407-414` — in the `desiredFileName` rename flow, `copyFileWithNewName` runs *before* the duplicate-file-ID check at `:416-451`. If the duplicate check trips, the freshly copied S3 object is orphaned and relies on the lifecycle policy to reap it. Move the duplicate check above the S3 copy (you already have `existingRef` and can compute the target filename from `desiredFileName` without writing anything to S3).

Confidence: high

**issue (race):** `apps/studio/src/server/modules/gazette/gazette.router.ts:202-231` and `:420-450` — the duplicate check is a read outside the transaction with no DB-level uniqueness, so two concurrent `create`s (or two `update`s targeting the same filename) can both pass and both commit. Probably low-likelihood for a CMS, but the PR copy ("Added duplicate file ID validation") suggests a stronger guarantee than the code provides. Options: (a) move the check inside the transaction and accept it's still TOCTOU under `READ COMMITTED`, (b) add a partial unique index on `(parentId, lower(split_part(...->'page'->>'ref', '/', -1)))` for CollectionLinks and let PG enforce it (would need to be a computed/generated column or expression index), or (c) document this as best-effort.

Confidence: medium

**suggestion:** `apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx:190` — `setValue("fileId", newFile.name)` doesn't trigger validation, and the fileId regex at `schemas/gazette.ts:21` is strict (`^[_\-a-zA-Z0-9]+\.pdf$`). Real-world filenames frequently contain spaces, parens, or dots (e.g. `Gazette Notice (2026).pdf`), so the autofill will often produce a value that only fails at submit. Two improvements worth doing together:
1. Pass `{ shouldValidate: true, shouldDirty: true }` so the error renders immediately.
2. Sanitize before setting — e.g. replace spaces with `-`, strip disallowed characters — so the common case auto-populates a valid value the user doesn't have to re-edit.

Confidence: high

**suggestion (maintainability):** `apps/studio/src/server/modules/gazette/gazette.router.ts:202-232` and `:420-451` — the duplicate-check query is duplicated near-verbatim between `create` and `update`. Extract into a helper in `gazette.service.ts` like `findCollectionLinkWithFilename({ db, siteId, parentId, filename, excludeId? })` so the two call sites can't drift (this would also have made the COALESCE-order bug above harder to introduce).

Confidence: high

**nitpick:** `apps/studio/src/server/modules/gazette/gazette.router.ts:417-418` — `finalRef.split("/").pop()` returns `string | undefined`, but `existingRef` is `?? ""`, so `oldFilename` ends up as `""` when there's no existing ref. The `newFilename && newFilename !== oldFilename` guard handles it, but consider being explicit (`newFilename && oldFilename !== newFilename`) for readability — and note that an existing gazette with no ref at all (shouldn't happen, but the schema doesn't forbid it) will then *always* run the duplicate check, which is the right behavior but worth a comment.

Confidence: low

**thought (test coverage):** `apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts:230-266` — tests only exercise the all-draft case (no `publishedVersionId`), so the `COALESCE` issue raised above isn't caught. Adding a test where one gazette has a *published* version with filename X and a draft with filename Y, then attempting to create another with filename Y, would have surfaced it.

Confidence: high

**praise:** `apps/studio/src/features/gazettes/components/GazetteTable/FileIdCell.tsx:49,58` — good call replacing `noOfLines={1}` with `wordBreak="break-all"` rather than truncating; for file IDs (no whitespace, often long) `break-all` is the right wrap mode.

Confidence: high

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -195,6 +195,42 @@ export const gazetteRouter = router({
...
+            .leftJoin(
+              "Blob as PublishedBlob",
+              "Version.blobId",
+              "PublishedBlob.id",
+            )
+            .where("Resource.siteId", "=", siteId)
+            .where("Resource.parentId", "=", String(collectionId))
+            .where("Resource.type", "=", ResourceType.CollectionLink)
+            .where(
+              sql`split_part(COALESCE("PublishedBlob"."content", "DraftBlob"."content")->'page'->>'ref', '/', -1)`,
```
dcshzj: question: hmm a bit confused by this, we prefer using the published version then the draft one, does this mean:
1. We allow Toppan to edit an existing published gazette?
2. If yes, then wouldn't it be possible to edit a published gazette's file ID to `new.pdf` without publishing, then creating a new gazette with the same `new.pdf`, causing conflict?

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -195,6 +195,42 @@ export const gazetteRouter = router({
           resourceIds: [String(collectionId)],
         })
 
+        // Check for duplicate file ID (filename portion of ref) in the same collection
+        // ref format: /sites/{siteId}/gazettes/{uuid}/filename.pdf
+        const filename = ref.split("/").pop()
+        if (filename) {
+          const duplicateRef = await db
```
dcshzj: issue: I think this should be inside the `created` transaction below so we can maintain the atomicity of checking-then-insert

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
```diff
@@ -186,6 +186,9 @@ export const GazetteFormFields = ({
           onChange={(newFile) => {
             setFile(newFile)
             onFileChange?.(newFile)
+            if (newFile) {
+              setValue("fileId", newFile.name)
```
dcshzj: issue: the filename might be invalid, so we might need to force a revalidation?

Also maybe we don't update the file name if the `fileId` field is already dirty?

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -377,6 +413,43 @@ export const gazetteRouter = router({
           }
         }
 
+        // Check for duplicate file ID if ref is changing
+        const newFilename = finalRef.split("/").pop()
+        const oldFilename = existingRef.split("/").pop()
+        if (newFilename && newFilename !== oldFilename) {
+          const duplicateRef = await db
```
dcshzj: issue: this query is duplicated, might want to abstract into a common function

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -195,6 +195,42 @@ export const gazetteRouter = router({
           resourceIds: [String(collectionId)],
         })
 
+        // Check for duplicate file ID (filename portion of ref) in the same collection
+        // ref format: /sites/{siteId}/gazettes/{uuid}/filename.pdf
```
dcshzj: question(non-blocking): sorry a bit confused, I thought the ref format would be something different like `/{year}/{month}/{filename}.pdf`?

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: 1. yeap, within the 30 minutes window
2. this has been fixed by checkign on both duplicate and published ref

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: done for both create/update

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
seaerchin: > issue: the filename might be invalid, so we might need to force a revalidation?

followed claude's suggestion to autofix for them

> Also maybe we don't update the file name if the fileId field is already dirty?

i think we should, because most of the time, the `fieldId` field will not be dirty as it will be autofilled. per Toppan, their file is always numbers/letters only. hence, this dirty case will be rare

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: done

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: down in downstream, post infra changes

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
```diff
@@ -186,6 +186,20 @@ export const GazetteFormFields = ({
...
+            if (newFile) {
+              // Sanitize filename to match fileId regex: ^[_\-a-zA-Z0-9]+\.pdf$
+              // 1. Remove .pdf extension (we'll add it back)
+              // 2. Replace spaces with hyphens
+              // 3. Strip any characters not in [_\-a-zA-Z0-9]
+              const baseName = newFile.name.replace(/\.pdf$/i, "")
+              const sanitized =
+                baseName.replace(/\s+/g, "-").replace(/[^_\-a-zA-Z0-9]/g, "") ||
+                "file"
+              setValue("fileId", `${sanitized}.pdf`, {
```
adriangohjw: ask: put into an utils and write unit test for it? since it contains regex. 

Also, thoughts on using filenamify?

apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts
adriangohjw, line 1: nit: add Arrange/Act/Assert comments to be consistent

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -37,3 +37,43 @@ export const assertGazetteAccess = async (userId: string): Promise<void> => {
...
+    .leftJoin("Version", "Resource.publishedVersionId", "Version.id")
+    .leftJoin("Blob as PublishedBlob", "Version.blobId", "PublishedBlob.id")
+    .where("Resource.siteId", "=", siteId)
+    .where("Resource.parentId", "=", parentId)
+    .where("Resource.type", "=", ResourceType.CollectionLink)
+    .where(
+      sql<boolean>`(
+        split_part("DraftBlob"."content"->'page'->>'ref', '/', -1) = ${filename}
+        OR split_part("PublishedBlob"."content"->'page'->>'ref', '/', -1) = ${filename}
+      )`,
```
adriangohjw, line 74: question(perf): with gazettes having a lot of resources, how long does this take? will it be an issue?

_note: im okay with merging first optimizing after launch if its not an ridiculously long time_

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -354,27 +373,44 @@ export const gazetteRouter = router({
...
+          desiredFileName &&
+          existingRef &&
+          desiredFileName !== oldFilename
+        ) {
+          const sourceKey = existingRef.replace(/^\//, "")
+          const newKey = await copyFileWithNewName({
+            sourceKey,
+            newFileName: desiredFileName,
+          })
+          finalRef = `/${newKey}`
```
adriangohjw, line 418: question(non-blocking): should we combine to avoid repeating the conditional statements?

adriangohjw (commented):
generally non-blocking, but just wanna clarify some stuff before approving! 💃

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -210,6 +213,22 @@ export const gazetteRouter = router({
...
         const created = await db.transaction().execute(async (tx) => {
+          // Check for duplicate file ID (filename portion of ref) in the same collection
+          // ref format: /sites/{siteId}/gazettes/{uuid}/filename.pdf
+          const filename = ref.split("/").pop()
+          if (filename) {
+            const duplicate = await findCollectionLinkWithFilename({
+              siteId,
+              parentId: String(collectionId),
+              filename,
+            })
```
dcshzj, line 225: issue: should pass `tx` in uh

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: for now not likely cos we're starting from clean slate. at an average rate of 50 gazettes/day, this is ~18k gazettes a year which seems ok to me

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: good catch, will fix

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
seaerchin: sounds good on filenamify!

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Right — `findCollectionLinkWithFilename` was using the module-level `db`, so the check ran outside the transaction. Added a `trx` param and now pass `tx` from both create and update (86c36747f). The update path also pre-checks before the S3 copy so a conflict doesn't orphan the copied object.

apps/studio/src/features/gazettes/components/GazetteModal/GazetteFormFields.tsx
seaerchin: Extracted into `toFileId` with unit tests (bc8589390). Went with filenamify as suggested — but it only strips filesystem-illegal characters (`/ \ : * ?` …) and keeps spaces/parens/unicode, so it can't satisfy the `^[_\-a-zA-Z0-9]+\.pdf$` schema regex on its own. So filenamify runs first (path separators become hyphens instead of merging tokens), then a charset strip enforces the schema.

apps/studio/src/server/modules/gazette/__tests__/gazette.router.test.ts
seaerchin: Added Arrange/Act/Assert comments across the gazette router tests (8c5d9e827).

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: The query is scoped to a single collection's CollectionLinks (filtered by `parentId` + `siteId` + `type`) with a `split_part` per row, so it's O(N) over just that collection's links — fine for the volumes we expect. No index on the computed filename yet; happy to add an expression index post-launch if it ever becomes hot, per your note.

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: Combined the two branches into one (86c36747f). The duplicate check now also runs as a pre-check before the S3 copy so a conflict never orphans a copied object, with the authoritative check kept inside the transaction.

adriangohjw (approved):

PR #2077, opened 2026-05-13
chore(studio): scope collection editing flag to tags management

adriangohjw:
## Problem

The GrowthBook flag `is-new-collection-editing-experience-enabled` was being used to gate three unrelated changes: the collection variant layout radio (from PR #1825), and the refreshed collection root drawer + dashboard subtitle copy (from PR #2002). With those refreshes ready to roll out by default, we want to retire the broad flag and keep only the still-unfinished Filters block behind a more narrowly named flag.

## Solution

Renames `is-new-collection-editing-experience-enabled` to `is-new-collection-tags-management-enabled` (constant `IS_NEW_COLLECTION_TAGS_MANAGEMENT_ENABLED_FEATURE_KEY`, hook `useNewCollectionTagsManagement`) and rewires usages. The flag is now only checked for the Filters BaseBlock in the collection root drawer; everything else that previously consulted the flag now renders unconditionally.

**Breaking Changes**

**Features**:

- The Filters block in the collection root drawer is gated by both `isUserIsomerAdmin` and the new `is-new-collection-tags-management-enabled` flag.

**Improvements**:

- The refreshed collection root drawer layout (Manage Collection section, Collection display block) renders by default for `collection` pages.
- The dashboard collection row subtitle now always reads "Manage the Collection's layout, filters, and sorting."
- The 1-column/2-column collection variant radio in `JsonFormsCollectionVariantControl` always renders.
- Removed the old `useNewCollectionEditingExperience` hook and renamed the variable/parameter `isNewCollectionEditingExperienceEnabled` to `isNewCollectionTagsManagementEnabled`.

## Tests

- [ ] On the dashboard, open a site with a collection resource and confirm the subtitle reads "Manage the Collection's layout, filters, and sorting." regardless of any flag state.
- [ ] Open a collection index page in Studio and confirm the drawer shows the "Manage Collection" section with the "Collection display" block by default (no flag needed).
- [ ] As a non Isomer Admin user with `is-new-collection-tags-management-enabled` off, confirm the Filters block does NOT appear.
- [ ] As an Isomer Admin (Core or Migrator) with `is-new-collection-tags-management-enabled` on, confirm the Filters block appears and clicking it does not crash.
- [ ] As an Isomer Admin with `is-new-collection-tags-management-enabled` off, confirm the Filters block does NOT appear.
- [ ] Open a page that renders the collection variant control and confirm the 1-column / 2-column radio is visible (no flag needed).

### DIFF
--- a/apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx
+++ b/apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx
@@ -508,17 +508,12 @@ export default function RootStateDrawer() {
                 <VStack gap="1.5rem" w="100%">
                   <VStack w="100%" h="100%" gap="1rem">
                     <Flex flexDirection="row" w="100%">
-                      {pageLayout !== ISOMER_USABLE_PAGE_LAYOUTS.Collection && (
-                        <VStack gap="0.25rem" align="start" flex={1}>
-                          <Text textStyle="subhead-1">Custom blocks</Text>
-                          <Text
-                            textStyle="caption-2"
-                            color="base.content.medium"
-                          >
-                            Use blocks to display your content.
-                          </Text>
-                        </VStack>
-                      )}
+                      <VStack gap="0.25rem" align="start" flex={1}>
+                        <Text textStyle="subhead-1">Custom blocks</Text>
+                        <Text textStyle="caption-2" color="base.content.medium">
+                          Use blocks to display your content.
+                        </Text>
+                      </VStack>
                       {/* TODO: we should swap over to using the `resource.type` */}
                       {/* rather than the `page.layout` but we are unable to do so due */}
                       {/* to the existence of custom index page that are `layout: */}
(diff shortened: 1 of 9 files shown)


apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx
```diff
@@ -456,7 +471,8 @@ export default function RootStateDrawer() {
           )}
 
           {pageLayout === ISOMER_USABLE_PAGE_LAYOUTS.Collection &&
-          isNewEditingExperienceEnabled ? (
+          isNewCollectionTagsManagementEnabled ? (
+            // New collection editing UI introduced in https://github.com/opengovsg/isomer/pull/2002
```
dcshzj, line 475: issue: hmm this would mean they no longer have access to the new collection editing experience view anymore, which is not really the intent. The `or manage filters` part of the copy can be the one being gated behind the feature flag, and the filters editing itself inside the drawer also

adriangohjw: @dcshzj 

oh this section is actually not part of `isNewEditingExperienceEnabled` and is only introduced in the collection tags management epic. 

so it's not something they already have access to, at the moment in production

Image

reference: https://github.com/opengovsg/isomer/pull/2002/changes#diff-dafba518ed56c8209efd53b0288fc4f6615100d952b1a315014fe396b09ce96bR458-R472

dcshzj (approved):

PR #2093, opened 2026-05-14
feat(redirect): add backend tRPC router with paginated listing and immediate publish

dcshzj:
## Problem

The redirects management UI (from `feat/redirects-management-frontend`) uses hardcoded seed data and local-only Jotai atoms. There is no backend API, so nothing persists, no permissions are enforced, and no changes reach the published site.

Fixes ISOM-2262.

## Solution

Adds a tRPC backend for redirects where every mutation publishes immediately (no separate publish step), and rewires the frontend hooks to use it instead of mock data. Listing is paginated and sorted server-side, matching the contract the frontend's mock simulated.

**Deploy Notes**

⚠️ **This PR requires a DB migration**: `20260610000001_add_redirect_audit_log_events`

- Adds `RedirectCreate` and `RedirectDelete` values to the `AuditLogEvent` enum via `ALTER TYPE ... ADD VALUE`
- Purely additive: no table rewrite, no locks, and old app code is unaffected while the migration rolls out, so it is safe to apply via the standard `migrate:<env>` flow ahead of the deploy
- The migration **must** be applied before this app code serves traffic — the new code writes the new enum values on every redirect create/delete
- Per `apps/studio/prisma/CLAUDE.md`, `ALTER TYPE` requires an explicit approving reviewer: please confirm approval of this enum addition in review

**Breaking Changes**

**Features**:

- **Backend tRPC router** (`redirect.list`, `redirect.count`, `redirect.create`, `redirect.delete`) registered on the app router — `read` permission for listing/counting, site-level `update` permission (site admins) for mutations
- **Server-side pagination and sorting**: `list` takes limit/offset (the UI requests 25 per page) plus a sort field (`source`, `destination`, or `publishedAt`) and direction, with an `id` tie-breaker so rows with equal sort values don't shuffle between pages; `count` returns the number of live redirects for deriving the page count
- **Immediate publish**: `create` and `delete` each run in a single transaction that writes the row, writes the audit entries, and triggers a CodeBuild site publish — redirects go live without a separate publish action
- **Duplicate handling on create**: creating a source that already has a live redirect is rejected with `CONFLICT` (surfaced as an error toast); a soft-deleted row for the same source is revived instead, with its publish timestamp refreshed
- **Soft deletes**: `delete` sets `deletedAt` (scoped to the requesting site, `NOT_FOUND` otherwise); `list` and `count` only ever consider live redirects
- **New audit log event types**: `RedirectCreate` and `RedirectDelete` (DB enum + migration), logged via a new typed `logRedirectEvent` with real DB rows as the before/after delta; the triggered site publish is still logged as a `Publish` event, and the audit-log report script renders the new events
- **Shared Zod schemas** (`schemas/redirect.ts`) used by both the router and the form — sources reject control characters, backslashes and `..` segments, and are normalised to a single leading slash; destinations must start with `/` or `https://`; the list schema composes the shared `offsetPaginationSchema`

**Improvements**:

- Replaced the mock `api.ts` with thin `trpc.redirect.*` hooks: the table's page/sort state passes straight through to `list` (with `keepPreviousData` so the table doesn't flash while changing pages), and mutations invalidate the whole redirect router so both list and count refetch
- `RedirectRow` is now derived from `RouterOutput` so the table type cannot drift from the API response
- The add form reuses the server schema (`createRedirectSchema.omit({ siteId })`), keeping client and server validation identical
- Added MSW handlers for `redirect.list` (honouring limit/offset) and `redirect.count`, wired into the Default and Empty Storybook stories

## Before & After Screenshots

**BEFORE**: UI renders hardcoded seed rows; nothing persists across page loads and nothing reaches the live site

**AFTER**: UI reads from and writes to the database via tRPC; adds and confirmed deletes publish to the site immediately, and the table pages through live redirects 25 at a time

## Tests

Integration tests (`redirect.router.test.ts`, 27 tests) cover auth (401), permissions (403), listing (live-only, cross-site isolation, default newest-first ordering, limit/offset pagination without overlap, sorting by field and direction), counting (live rows only, scoped to the site), create (normalisation, conflict, revive), delete (soft-delete, not-found, already-deleted), and the audit entries (`RedirectCreate`/`RedirectDelete` deltas with real rows, plus the `Publish` entry). Schema unit tests (11 tests) cover the validation boundary cases.

- [ ] Apply the DB migration and confirm `SELECT unnest(enum_range(NULL::"AuditLogEvent"))` includes `RedirectCreate` and `RedirectDelete`
- [ ] Navigate to Site Settings → Redirects for a site with the redirections feature flag enabled
- [ ] Add a redirect (e.g. source: `old-page`, destination: `/new-page`) — verify the success toast and that the row appears with "just now" under Added
- [ ] Refresh the page and verify the redirect persists
- [ ] Add a redirect with the same source again — verify the error toast ("A redirect already exists for this path") and that nothing is added
- [ ] With more than 25 redirects, verify the pagination footer appears, page navigation fetches the next page, and sorting a column reorders across the whole dataset and returns to page 1
- [ ] Click the trash icon on a row, verify the confirmation modal shows the redirect's source and destination, confirm — verify the row disappears with a success toast
- [ ] Re-add the deleted source — verify it succeeds (revive) with the new destination
- [ ] Verify the `AuditLog` table has `RedirectCreate`, `RedirectDelete` and `Publish` entries for the actions above
- [ ] For a launched site (with a CodeBuild project), verify a site build is triggered after an add and after a delete
- [ ] Open redirects for a different site and verify no data leaks across sites

**New scripts**: None

**New dependencies**: None


apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,95 @@
...
+      destination: row.destination,
+      createdAt: row.createdAt,
+      status: row.deletedAt ? ("deleted" as const) : ("active" as const),
+      hasUnpublishedChanges: false,
+    })),
+    totalCount: countResult.count,
+  }
+}
+
+export const publishRedirects = async ({
```
dcshzj: issue: I forgot audit logs

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -91,7 +92,15 @@ export const AddRedirectCard = ({
...
               placeholder="redirect-from"
               value={source}
-              onChange={(e) => setSource(e.target.value)}
+              onChange={(e) =>
+                setSource(
+                  transliterate(e.target.value)
+                    .toLowerCase()
+                    .replace(/[^a-z0-9/-]/g, "-")
+                    .replace(/^\/+/, "")
+                    .replace(/\/{2,}/g, "/"),
```
adriangohjw: ask: ps, can trouble you to add comments so its easier to understand at first glance? 😬

apps/studio/src/server/modules/redirect/__tests__/redirect.router.test.ts
adriangohjw, line 1: ask: can add the Arrange/Act/Assert comments to be consistent? thanks!

apps/studio/src/schemas/redirect.ts
adriangohjw, line 1: sorry a bit confused - why do we have two different schema in here and in `apps/studio/src/features/settings/Redirects/types.ts`?

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  ])
+
+  return {
+    items: items.map((row) => ({
+      id: String(row.id),
+      source: row.source,
+      destination: row.destination,
+      createdAt: row.createdAt,
+      status: row.deletedAt ? ("deleted" as const) : ("active" as const),
+      hasUnpublishedChanges: false,
```
adriangohjw: question: why always false? and if always false, should we even return it?

apps/studio/src/features/settings/Redirects/api.ts
```diff
@@ -1,63 +1,104 @@
...
   data: RedirectRow[]
+  totalCount: number
+  draftCount: number
   isLoading: boolean
+  isFetching: boolean
 } {
-  const data = useAtomValue(redirectsAtom)
-  return { data, isLoading: false }
+  const pageIndex = options?.pageIndex ?? 0
+  const pageSize = options?.pageSize ?? 25
```
adriangohjw: non-blocking: since we hardcode to 25 throughout the codebase, should we just hardcode to 25 here and save the trouble of passing in a prop? i don't think we gonna support the option to change "items per page"

then we also can have this hook return `pageCount` such that the consuming component (RedirectTablet) can just use it

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -253,34 +255,84 @@ function DeleteCell({
...
+  })
 
-  const handleDelete = (id: string) => deleteRedirect({ siteId, id })
+  const prevDraftCount = useRef(draftCount)
+  useEffect(() => {
+    if (draftCount > prevDraftCount.current) {
+      setPagination((prev) => ({ ...prev, pageIndex: 0 }))
+    }
+    prevDraftCount.current = draftCount
+  }, [draftCount])
```
adriangohjw: question: a bit confused why we need `draftCount > prevDraftCount.current`? this means if the list shrinks (e.g. user deletes the records, the pageindex will be stale and not updated?

apps/studio/src/features/settings/Redirects/api.ts
```diff
@@ -76,60 +118,120 @@ export function useCreateRedirect(): {
...
     }
-    setRows((prev) => [newRow, ...prev])
+    setDraftsMap((prev) => {
+      const next = new Map(prev)
+      const drafts = next.get(siteId) ?? []
+      next.set(siteId, [newRow, ...drafts.filter((d) => d.source !== source)])
+      return next
+    })
   }
   return { mutate, isPending: false }
```
adriangohjw: is there hardcoded isPending a placeholder?

apps/studio/src/features/settings/Redirects/api.ts
```diff
@@ -76,60 +118,120 @@ export function useCreateRedirect(): {
...
-        ]
-      }),
-    )
-  }
+  const [, setDraftsMap] = useAtom(localDraftsAtom)
+  const [, setPendingDeletesMap] = useAtom(pendingDeletesAtom)
+
+  const mutate = useCallback(
+    ({ siteId, id }: { siteId: number; id: string }) => {
+      if (id.startsWith("draft-")) {
```
adriangohjw: ask: can we extract `draft-` out as a const like `DRAFT_PREFIX` since it's used in line 121? for single source of truth + easier to draw the connection on what it is, instead of magic string

apps/studio/src/features/settings/Redirects/api.ts
```diff
@@ -1,63 +1,104 @@
...
+  const draftCount = localDrafts.length
+  const draftSources = useMemo(
+    () => localDrafts.map((d) => d.source),
+    [localDrafts],
+  )
+  const virtualStart = pageIndex * pageSize
+  const virtualEnd = virtualStart + pageSize
+
+  const draftsForPage = localDrafts.slice(
+    Math.max(0, virtualStart),
```
adriangohjw: nit: `virtualStart` can't be negative, so don't need `Math.max(...)`

apps/studio/src/features/settings/Redirects/api.ts
```diff
@@ -76,60 +118,120 @@ export function useCreateRedirect(): {
...
+                (id) => !submittedDeleteIds.has(id),
+              ),
+            )
+            if (remaining.size > 0) {
+              next.set(siteId, remaining)
+            } else {
+              next.delete(siteId)
+            }
+            return next
+          })
```
adriangohjw: question: on `serverPublish` success, technically there's no more "drafts" or "pending delete" records. So instead of manually reconcilling here, we can just wipe clean and start from empty state?

adriangohjw (commented):
ask: can add storybook in follow-up PR? since we have the API, so we can now mock it for storybook. thanks!

apps/studio/src/server/modules/redirect/__tests__/redirect.router.test.ts
```diff
@@ -0,0 +1,308 @@
...
+        source: "/old",
+        destination: "/new",
+        status: "active",
+        hasUnpublishedChanges: false,
+      })
+      expect(result.items[0]!.createdAt).toBeInstanceOf(Date)
+      expect(typeof result.items[0]!.id).toBe("string")
+    })
+
+    it("should include soft-deleted redirects with deleted status", async () => {
```
seaerchin: comment: should we add a test for adding soft-deleted redirects

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,188 @@
...
+          destination,
+          deletedAt: null,
+          // A revived redirect is republished now, so its createdAt (shown
+          // to users as the publish time) is refreshed
+          createdAt: new Date(),
+        }),
+      )
+      .returningAll()
+      .executeTakeFirstOrThrow()
+
```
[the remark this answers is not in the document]
dcshzj: Fixed — the upsert's conflict update is now gated with `WHERE "Redirect"."deletedAt" IS NOT NULL`, so only soft-deleted rows can be revived. If the statement returns no row (i.e. a concurrent create committed a live redirect after the pre-check), we throw `CONFLICT` instead of overwriting it. The pre-check stays for the common serial case since it also feeds the audit delta.

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,76 @@
...
+  siteId: z.number().min(1),
+  source: sourceSchema,
+  destination: destinationSchema,
+})
+export type CreateRedirectInput = z.infer<typeof createRedirectSchema>
+
+export const deleteRedirectSchema = z.object({
+  siteId: z.number().min(1),
+  id: z.string().min(1),
+})
```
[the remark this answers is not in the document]
dcshzj: Fixed — the id is now validated as a numeric bigint-like string (`/^[1-9][0-9]*$/`, same shape as the bigint id schema in `resource.ts`), with a test asserting that a non-numeric id is rejected by validation instead of becoming a DB cast error.

seaerchin:
## Summary

Adds a tRPC backend for redirects (`list`, `count`, `create`, `delete`) and rewires the frontend hooks at `apps/studio/src/features/settings/Redirects/api.ts` away from the Jotai mock. Mutations publish immediately via `publishSite` inside the same transaction, audit entries are written with real DB rows per the `audit-log-real-db-rows` convention, and a new `AuditLogEvent` enum migration adds `RedirectCreate`/`RedirectDelete`.

Codeflow:
- `redirect.router.ts:19-76` — thin router; permission check (`read` for list/count, site-level `update` for mutations) before delegating to the service.
- `redirect.service.ts:31-50` (`listRedirects`) — server-side sort + offset/limit + `id` tie-breaker so equal sort values don't shuffle between pages; soft-deletes excluded.
- `redirect.service.ts:62-130` (`createRedirect`) — pre-check captures audit `before` state; `INSERT ... ON CONFLICT DO UPDATE WHERE deletedAt IS NOT NULL` guards revivals from silently overwriting live rows under concurrent creates.
- `redirect.service.ts:132-167` (`deleteRedirect`) — soft-delete scoped to `(siteId, id, deletedAt IS NULL)`, with `before`/`after` real-row audit entries.
- `schemas/redirect.ts:23-46` — single source of truth shared between server and the `AddRedirectCard` form (via `.omit({ siteId })`).
- `RedirectsTable.tsx:223-256` — server-driven sort/page state; resets to page 0 on sort change; steps back when the last page empties.

Overall: well-scoped, conventions followed, concurrency carefully thought through. A few non-blocking issues below — mostly defence in depth and small test-rigor nits.

## Findings

**praise:** `redirect.service.ts:90-103` — the `onConflict … doUpdateSet … where("Redirect.deletedAt", "is not", null)` pattern is exactly the right shape for the revival race. The pre-check captures the audit `before` state, and the conditional update prevents a concurrent live create from being silently overwritten. The matching test (`redirect.router.test.ts:1027-1055`) provably exercises that the existing live row is not clobbered on CONFLICT. Worth keeping as a pattern reference for other "soft-delete + revive" surfaces.

Confidence: high

---

**praise (testing):** `redirect.router.test.ts` — solid coverage across auth (401), site-level permission (403), cross-site isolation, soft-delete invisibility, sort+direction, pagination without overlap, audit-entry deltas with real rows, and the `Publish` side-effect. Follows the AAA convention from `conventions/tests-arrange-act-assert.md`.

Confidence: high

---

**issue (security):** `schemas/redirect.ts:71-75` — `listRedirectsSchema` composes `offsetPaginationSchema`, which has no upper bound on `limit` (`pagination.ts:5`: `z.number().int().positive().default(10)`). A logged-in user with read access can request `limit: 10_000_000` and force the server to return an unbounded Postgres result set + JSON-serialise it across the wire. The frontend always passes 25, but the server schema is the trust boundary.

This is a codebase-wide gap (gazette/resource/folder/user paginators all inherit the same shape), so I'd suggest either (a) adding a `.max(MAX_PAGE_SIZE)` to `offsetPaginationSchema` and updating call sites, or (b) extending the schema per-router with a redirect-local cap (e.g. `limit: z.number().int().positive().max(100).default(25)`). Option (a) is the proper fix; option (b) is a local hedge that wouldn't require touching unrelated routers.

Confidence: high

---

**issue (correctness, non-blocking):** `redirect.router.test.ts:1200-1208` — the test claims to verify a 400 for a non-numeric id but asserts only the error message string. Other tests in the same file (e.g. `:702-705`, `:716-722`) use `new TRPCError({ code: "..." })` so the *code* is verified. Suggest tightening to:

```ts
await expect(result).rejects.toThrowError(
  new TRPCError({ code: "BAD_REQUEST", message: /* … */ }),
)
```

so a future refactor that swaps the error code for a different reason doesn't silently slip through.

Confidence: medium

---

**suggestion:** `redirect.service.ts:34-43` — the mapping from `sortBy: "publishedAt"` to the underlying `createdAt` column is implicit:

```ts
.orderBy(sortBy === "publishedAt" ? "createdAt" : sortBy, sortDirection)
```

If someone adds a fourth value to `redirectSortFieldSchema` later and forgets to update this branch, Kysely will either error at runtime or silently sort on a wrong field. A small explicit map closes the gap:

```ts
const SORT_FIELD_TO_COLUMN = {
  source: "source",
  destination: "destination",
  publishedAt: "createdAt",
} as const satisfies Record<RedirectSortField, keyof Redirect>
…
.orderBy(SORT_FIELD_TO_COLUMN[sortBy], sortDirection)
```

The `satisfies` plus exhaustive map gives a compile error the moment the enum drifts.

Confidence: medium

---

**suggestion (non-blocking):** `schemas/redirect.ts:48-52` — the redirect-id regex `^[1-9][0-9]*$` duplicates `bigIntSchema` in `schemas/resource.ts:~/* the private one */` (which uses `^[0-9]+$` + `!v.startsWith("0")` to express the same thing). Worth lifting to `schemas/common.ts` so both modules consume a single named primitive — see `schemas/CLAUDE.md` § "Reuse `common.ts` primitives where possible".

Confidence: medium

---

**suggestion (non-blocking):** `prisma/scripts/getAuditLogs.ts:156` — `RedirectCreate` renders as `Redirect from "<after.source>" to "<after.destination>" created` for *both* fresh creates and revivals. On revival, the audit `before` carries the soft-deleted row with the previous destination, but the report drops it. A `CASE` on `al.delta -> 'before'` would let the report distinguish "created" from "revived (was: /old-dest)". Not a security or correctness issue — just a fidelity loss in the auditor's view.

Confidence: medium

---

**question:** `redirect.service.ts:114-122` and `:155-161` — both mutations write `metadata: { redirects: { created: [created] } }` / `{ deleted: [deleted] }` on the `Publish` audit entry. That shape isn't typed anywhere central (it's accepted by `RepublishEventLogProps`'s `Record<string, unknown>` escape hatch). Is the intent to evolve this into a typed metadata variant for redirect-driven publishes, or is it informational only for ad-hoc DB queries? If informational, fine; if you expect other consumers to read it, a typed `RedirectPublishMetadata` next to the other audit types in `audit.service.ts` would prevent shape drift.

Confidence: medium

---

**nitpick:** `_app.ts:24-29` — `redirectRouter` is registered between `page` and `folder`, which doesn't match the alphabetical-ish ordering of the surrounding routers (`asset, page, folder, collection, gazette, …`). Trivial — just noting if you want the file to stay scannable.

Confidence: low

---

**thought (non-blocking):** `redirect.service.ts:144-150` and `:165-167` — `publishSite` is invoked inside the DB transaction, which holds the connection open while the AWS `StartBuild` call goes out. This mirrors `resource.service.ts:808` and `:841`, so it's consistent with the codebase, not a regression introduced here. But the consistent pattern is also a consistent risk: under bursty redirect edits on a high-traffic site, the connection pool can stay tied up on AWS round-trips. If you ever observe pool exhaustion in prod, this whole family of services is a candidate for "commit DB, then enqueue publish via pg-boss" rather than calling AWS inline.

Confidence: low

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,79 @@
...
+export const MAX_REDIRECT_PATH_LENGTH = 2000
+
+// Matches ASCII control characters (0x00-0x1f, 0x7f) and backslashes.
+// These are never valid in a URL path and can corrupt the generated
+// redirect rules on the published site.
+const INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/
+
+// Strips slashes from both ends of a path so "/foo/", "foo" and "foo//"
+// all normalise to the same inner segments before validation.
+const trimSlashes = (value: string) => value.replace(/^\/+|\/+$/g, "")
```
seaerchin, line 19: comment: will we allow users to submit a fully formed redirect? eg: /original-page -> /nested/page. if this is hte case, then we need to trim the slashes in the middle also

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,79 @@
...
+  .max(MAX_REDIRECT_PATH_LENGTH, { message: "Source path is too long" })
+  .refine((value) => !INVALID_PATH_CHARS_REGEX.test(value), {
+    message: "Source must not contain control characters or backslashes",
+  })
+  .refine((value) => trimSlashes(value).length > 0, {
+    message: "Source path cannot consist only of slashes",
+  })
+  .refine((value) => !trimSlashes(value).split("/").includes(".."), {
+    message: "Source must not contain '..' path segments",
+  })
```
seaerchin, line 34: comment: why not just ban '.'

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// Strips slashes from both ends of a path so "/foo/", "foo" and "foo//"
+// all normalise to the same inner segments before validation.
+const trimSlashes = (value: string) => value.replace(/^\/+|\/+$/g, "")
+
+const sourceSchema = z
+  .string()
+  .min(1, { message: "Source path is required" })
+  .max(MAX_REDIRECT_PATH_LENGTH, { message: "Source path is too long" })
+  .refine((value) => !INVALID_PATH_CHARS_REGEX.test(value), {
```
seaerchin: comment: i feel like we should go with a whitelisted characters approach rather than a black list approach as we've done here. this bounds the possible cases and is negligible to most users.

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,79 @@
...
+  .transform((value) => `/${trimSlashes(value).replace(/\/{2,}/g, "/")}`)
+
+const destinationSchema = z
+  .string()
+  .min(1, { message: "Destination is required" })
+  .max(MAX_REDIRECT_PATH_LENGTH, { message: "Destination is too long" })
+  // Destinations must be a path on the same site ("/...") or an external
+  // https URL — anything else (http://, javascript:, ...) is rejected.
+  .refine((value) => value.startsWith("/") || value.startsWith("https://"), {
+    message: "Destination must start with '/' or 'https://'",
```
seaerchin, line 46: comment: we've seen people use mailto:// as a contact-us - i think that this is a valid use case but happy to omit first

apps/studio/src/schemas/__tests__/redirect.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      const result = createRedirectSchema.safeParse({
+        ...VALID_REDIRECT,
+        source: "/bad\\path",
+      })
+
+      // Assert
+      expect(result.success).toBe(false)
+    })
+
+    it("should reject sources containing '..' path segments", () => {
```
seaerchin, line 54: test: since we're going with a blacklist approach, is there value in checking that paths containing a single . are preserved?

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,79 @@
+import { z } from "zod"
+
+import { offsetPaginationSchema } from "./pagination"
+
+export const MAX_REDIRECT_PATH_LENGTH = 2000
+
+// Matches ASCII control characters (0x00-0x1f, 0x7f) and backslashes.
+// These are never valid in a URL path and can corrupt the generated
+// redirect rules on the published site.
+const INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/
```
seaerchin: comment: asked claude what is the valid list of characters for redirects, answer below:

                                                                                
  Mapping back to your schema                                                   
                                                                                
  Comparing what RFC 3986 says to INVALID_PATH_CHARS_REGEX =                    
  /[\x00-\x1f\x7f\\]/:                                                          
                                                                                
  ┌─────────────────────────────┬─────────────────────┬────────────────────┐    
  │            Class            │      RFC 3986       │   Current schema   │    
  ├─────────────────────────────┼─────────────────────┼────────────────────┤    
  │ Control chars (0x00–0x1F,   │ must be %-encoded   │ rejected ✓         │    
  │ 0x7F)                       │                     │                    │    
  ├─────────────────────────────┼─────────────────────┼────────────────────┤    
  │ Backslash \                 │ must be %-encoded   │ rejected ✓         │    
  ├─────────────────────────────┼─────────────────────┼────────────────────┤    
  │ Space, ", <, >, ^, `, {, |, │ must be %-encoded   │ accepted as        │    
  │  }                          │                     │ literal ✗          │    
  ├─────────────────────────────┼─────────────────────┼────────────────────┤    
  │ ?, #, [, ]                  │ reserved as         │ accepted as        │    
  │                             │ gen-delims          │ literal ✗          │    
  ├─────────────────────────────┼─────────────────────┼────────────────────┤    
  │ Non-ASCII (0x80+)           │ must be %-encoded   │ accepted as        │    
  │                             │                     │ literal ✗          │    
  └─────────────────────────────┴─────────────────────┴────────────────────┘    
                                                                                
  So the current rule is necessary but not sufficient against the spec. If you  
  want the schema to be a baseline-correct gate, the strict form is "every byte 
  must be in pchar ∪ / or be %XX", which expressed as a regex on each segment is
   roughly:                                                                     
                                                                                
  ^([A-Za-z0-9\-._~!$&'()*+,;=:@]|%[0-9A-Fa-f]{2})*$                            
                                                                                
  A looser, practical baseline that matches what browsers/servers will actually 
  tolerate without surprise is "no whitespace, no control chars, no             
  fragment/query/gen-delim chars" — that catches the common footguns without    
  requiring users to learn percent-encoding.

apps/studio/src/server/modules/redirect/redirect.router.ts
```diff
@@ -0,0 +1,77 @@
...
+  // separate publish step for redirects. Site-wide abilities only define CRUD
+  // actions ("publish" exists only for resources), so both gate on "update" —
+  // granted solely to site admins, matching other site-wide settings.
+  create: protectedProcedure
+    .input(createRedirectSchema)
+    .mutation(async ({ ctx, input }) => {
+      await validateUserPermissionsForSite({
+        siteId: input.siteId,
+        userId: ctx.user.id,
+        action: "update",
```
seaerchin: nit: this should be "create"

apps/studio/src/server/modules/redirect/redirect.router.ts
```diff
@@ -0,0 +1,77 @@
...
+      })
+    }),
+
+  delete: protectedProcedure
+    .input(deleteRedirectSchema)
+    .mutation(async ({ ctx, input }) => {
+      await validateUserPermissionsForSite({
+        siteId: input.siteId,
+        userId: ctx.user.id,
+        action: "update",
```
seaerchin: nit: this should be `delete`

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,201 @@
...
+  const { count } = await db
+    .selectFrom("Redirect")
+    .select((eb) => eb.fn.countAll().as("count"))
+    .where("siteId", "=", siteId)
+    .where("deletedAt", "is", null)
+    .executeTakeFirstOrThrow()
+  return Number(count)
+}
+
+export const createRedirect = async ({
```
seaerchin, line 78: issue: i thought part of our success criteria as outlined [here](https://app.notion.com/p/opengov/Redirection-Scoping-32f77dbba78880c8bc21f97f8e12dfc4?source=copy_link#33b77dbba788808b9d71f73c1765eb45) was that the redirect would update automatically whenever the page it's referencing updates their internal permalink;

in our current implementation, this is passed and stored as string rather than a reference to the page entity, which means that updating the page permalink will break our redirectse.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,201 @@
...
+  return (
+    db
+      .selectFrom("Redirect")
+      .select(["id", "source", "destination", "createdAt as publishedAt"])
+      .where("siteId", "=", siteId)
+      .where("deletedAt", "is", null)
+      .orderBy(sortBy === "publishedAt" ? "createdAt" : sortBy, sortDirection)
+      // Tie-break on id so rows with equal sort values don't shuffle between
+      // pages across requests
+      .orderBy("id", sortDirection)
```
seaerchin: nit: `id` is an implementation detail and wouldn't make much sense to users; prefer to have it sorted by source/destination, so that it's something users can make sense of

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,201 @@
...
+            deletedAt: null,
+            // A revived redirect is republished now, so its createdAt (shown
+            // to users as the publish time) is refreshed
+            createdAt: new Date(),
+          })
+          // Only soft-deleted rows may be revived: a concurrent create for
+          // the same source can commit after the pre-check above, and must
+          // surface as a conflict instead of silently overwriting the live
+          // redirect
+          .where("Redirect.deletedAt", "is not", null),
```
seaerchin, line 122: idk enough to know if this is correct for pg but ok.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,201 @@
...
+      )
+      .returningAll()
+      .executeTakeFirst()
+    if (!created) {
+      throw new TRPCError({
+        code: "CONFLICT",
+        message: `A redirect already exists for ${source}`,
+      })
+    }
```
seaerchin: style: prefer to inline and use `executeTakeFirstOrThrow`

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    // NOTE: the delta holds real DB rows — `existing` as fetched before the
+    // write (the soft-deleted row when reviving) and the row returned by the
+    // insert — so the audit entry reflects exactly what was committed
+    await logRedirectEvent(tx, {
```
seaerchin, line 135: question: should we log outside of the tx? not a big deal for me, just wondering if it could cause any issues with transaction isolation or consistency.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,201 @@
...
+      .where("siteId", "=", siteId)
+      .where("id", "=", id)
+      .where("deletedAt", "is", null)
+      .executeTakeFirst()
+    if (!before) {
+      throw new TRPCError({
+        code: "NOT_FOUND",
+        message: "Redirect not found",
+      })
+    }
```
seaerchin: style: prefer to inline as per above

seaerchin (commented):
i don't think storing a redirect as a string is robust enough for our use-case; we should ideally support a redirect that's stored by a pointer so that we get teh update permalink part for free. i think we can reuse (with some refactoring done) the existing `LinkSelectionModal` so that we can account for these on the frontend

apps/studio/src/server/modules/redirect/__tests__/redirect.router.test.ts
dcshzj: Covered — there's `should revive a soft-deleted redirect for the same source` (and the matching audit-delta test) which create over a soft-deleted row and assert it's revived, plus `should not return soft-deleted redirects` on the list side.

apps/studio/src/schemas/redirect.ts
dcshzj: Single internal slashes are path separators and are preserved — only runs (`//`) are collapsed and leading/trailing slashes trimmed. So a fully-formed `/original-page` → `/nested/page` round-trips unchanged; added a nested-path test to confirm.

apps/studio/src/schemas/redirect.ts
dcshzj: Kept `.` allowed — it's legitimate in real paths (`/files/report.v2.pdf`, versioned URLs). Only `..` path segments are rejected. The new whitelist permits `.` while excluding the dangerous cases.

apps/studio/src/schemas/redirect.ts
dcshzj: Done — switched to a whitelist of RFC 3986 `pchar` characters (`^[A-Za-z0-9\-._~!$&'()*+,;=:@%/]+$`) instead of the blacklist, so only known-safe characters reach the generated rules.

apps/studio/src/schemas/redirect.ts
dcshzj: ~Holding off on `mailto:` for now — https + internal paths only. Easy to add later as another destination variant if it's needed.~

Hallucinate again. I don't think we will support a redirect to mailto: (at least at the moment) cos the behaviour of having a 301 redirect to the mail client might be unexpected to the user.

apps/studio/src/schemas/__tests__/redirect.test.ts
dcshzj: Added a test that `/files/report.v2.pdf` is accepted and preserved, plus one asserting out-of-whitelist characters (spaces, `<`, `{`) are rejected.

apps/studio/src/schemas/redirect.ts
dcshzj: Adopted the `pchar` whitelist baseline from this analysis: `^[A-Za-z0-9\-._~!$&'()*+,;=:@%/]+$` for the source path.

apps/studio/src/server/modules/redirect/redirect.router.ts
dcshzj: Fixed — gated on `"create"`.

apps/studio/src/server/modules/redirect/redirect.router.ts
dcshzj: Fixed — gated on `"delete"`.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Addressed in a stacked follow-up PR: #2490. Internal destinations are now stored as `[resource:siteId:resourceId]` references and resolved to the live permalink at publish time (so they follow the page when its permalink changes) and for display, reusing `ResourceSelector` in the add form. Split out so these backend review fixes can merge independently.

---

Adding on, agreed this is the intent although fundamentally my approach is changing. Previously I was intending to just store the actual permalinks and have any moves/deletes adjust the Redirects table, but will proceed with just storing the reference links so it can be resolved at build time.

Using the ResourceSelector was also intended to be in a future PR (now #2490).

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Now tie-breaks on `source` first (user-meaningful), falling back to `id` only as the final unique key so pagination stays stable across requests.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Confirmed correct for Postgres — the partial `WHERE "deletedAt" IS NOT NULL` on the `ON CONFLICT DO UPDATE` only revives soft-deleted rows; a concurrent live row makes the upsert a no-op (no row returned) which we surface as a 409. Exercised by the revival + conflict tests.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Done — inlined with `executeTakeFirstOrThrow`.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Keeping it inside the transaction on purpose — the audit row must commit atomically with the write (and it reads the real committed rows), per our audit-log-real-db-rows convention. Same tx, so no isolation/consistency concern.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Done — inlined as above.

dcshzj:
Thanks for the thorough review @seaerchin — inline threads answered individually. Summarising the review-level points:

- **Pagination upper bound (security):** added a redirect-local `.max(100)` on `limit` (the list input is the trust boundary). Left the shared `offsetPaginationSchema` untouched to avoid changing unrelated paginators — happy to file the codebase-wide follow-up you flagged.
- **Test asserts message not code:** tightened the bad-id test to assert the `BAD_REQUEST` code.
- **Sort-field map:** replaced the implicit branch with an exhaustive `SORT_FIELD_TO_COLUMN ... satisfies Record<RedirectSortField, ...>`.
- **bigint-id schema:** lifted to `common.ts` as `generateBigIntSchema`, now shared with `resource.ts`.
- **Audit revival fidelity:** `RedirectCreate` now distinguishes `revived (was: ...)` from `created`.
- **Router ordering:** moved the `redirectRouter` registration.
- **Typed `RedirectPublishMetadata`:** left as the `Record<string, unknown>` hatch for now (informational); will type it if another consumer reads it.
- **`publishSite` inside the tx:** agreed it's a latent risk shared across this service family; out of scope here since it mirrors `resource.service` — worth a separate "commit DB, then enqueue publish" ticket if we see pool pressure.

The string-vs-reference success criterion (auto-following permalink changes) is handled in the stacked follow-up #2490.

apps/studio/src/schemas/redirect.ts
seaerchin: this already exists in production haha

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,95 @@
...
+  .min(1, { message: "Destination is required" })
+  .max(MAX_REDIRECT_PATH_LENGTH, { message: "Destination is too long" })
+  // Same-site path ("/...") or external https URL; anything else (http://,
+  // javascript:, ...) is rejected.
+  .refine((value) => value.startsWith("/") || value.startsWith("https://"), {
+    message: "Destination must start with '/' or 'https://'",
+  })
+  // Collapse a leading run of slashes on an internal path so a protocol-relative
+  // "//evil.com" can't pass as an open redirect ("//evil.com" -> "/evil.com").
+  .transform((value) =>
```
seaerchin, line 50: question: will https://https://www.isomer.gov.sg be accepted as a valid redirection destination?

apps/studio/src/schemas/redirect.ts
```diff
@@ -0,0 +1,95 @@
...
+  // equivalent inputs map to one source (keeps the unique constraint meaningful).
+  .transform((value) => `/${trimSlashes(value).replace(/\/{2,}/g, "/")}`)
+
+const destinationSchema = z
+  .string()
+  .min(1, { message: "Destination is required" })
+  .max(MAX_REDIRECT_PATH_LENGTH, { message: "Destination is too long" })
+  // Same-site path ("/...") or external https URL; anything else (http://,
+  // javascript:, ...) is rejected.
+  .refine((value) => value.startsWith("/") || value.startsWith("https://"), {
```
seaerchin, line 45: comment: we probably want to define our own `internalLinkSchema` then at the refine stage, we validate that the destination is either:

1. valid https:// URL or 
2. valid internal URL

this would also allow us to reuse `internalLinkSchema` for the `sourceSchema`

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,215 @@
...
+  siteId,
+  sortBy,
+  sortDirection,
+  limit,
+  offset,
+}: ListRedirectsInput) => {
+  // Live rows only; createdAt is the publish time (creates publish immediately).
+  let query = db
+    .selectFrom("Redirect")
+    .select(["id", "source", "destination", "createdAt as publishedAt"])
```
seaerchin, line 50: comment: we need to remember to update the `createdAt` if it's deleted and re-created

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    })
+
+    // Every mutation republishes the site; audited as a separate Publish event.
+    await logPublishEvent(tx, {
+      siteId,
+      by: byUser,
+      delta: { before: null, after: null },
+      eventType: AuditLogEvent.Publish,
+      metadata: { redirects: { created: [created] } },
+    })
```
seaerchin, line 149: question: we don't actually publish until we exit this tx, up to you whether we want to eagerly log, i don't see an issue personally

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -49,22 +49,31 @@ export const AddRedirectCard = ({
...
+          toast(
+            error.data?.code === "CONFLICT"
+              ? {
+                  title: "A redirect already exists for this path",
+                  description: `Delete the redirect for ${source} first if you want to change where it points to.`,
+                  status: "error",
+                }
+              : {
+                  title: "Failed to add redirect",
+                  description: "Something went wrong. Please try again.",
```
seaerchin: question: should we use the `error.data.message` here?

seaerchin (approved):
i think the validation is abit too lenient here - i didn't see the router do additional validation (which is ok, i mean the schema should have caught it anw) which seems to imply that malformed external redirects are accepted

i think we should try to prevent this; additionally, we also don't prevent self-redirects - for a page that has a full permalink of `/a`, a user can trivially add a redirect of `https://www.isomer.gov.sg/a` which would lead to infinite redirects at run-time haha

i'm still going to approve this PR because these exploits do require a deliberate sabotage on the part of the actor + some tech know-how, which doesn't fit our user profile. that being said, we should fix this before it gets too long in the tooth

apps/studio/src/schemas/redirect.ts
dcshzj: Yes — on this branch it would. The check is just `startsWith("https://")`, so a doubled scheme slips through; `new URL("https://https://www.isomer.gov.sg")` even parses, with hostname `"https"`. Fixed on #2466: the destination now has to parse as an `https:` URL with a dotted host, which rejects the doubled scheme and bare single-label hosts. Tests added there.

apps/studio/src/schemas/redirect.ts
dcshzj: Done on #2466, slightly differently from the literal suggestion: destination is refined to internal path | parseable `https:` URL | `[resource:...]` reference, and source/destination share the `normalizeRedirectPath`/`normalizeRedirectSource` primitives rather than one `internalLinkSchema` — source is lowercased to match page permalinks and destination isn't, so they can't be the identical schema. The strict https parse closes the `https://https://` gap from the thread above.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Already handled — the revival path (`onConflict … doUpdateSet`) sets `createdAt: new Date()`, so a revived redirect shows the new publish time. The comment anchored to line 50 but the logic is in the create upsert.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Keeping the Publish audit row inside the tx on purpose — it should commit atomically with the redirect write so we can't end up with a published change and no audit trail (or vice-versa). `publishSite()` already runs after commit, so the external CodeBuild call isn't holding the row locks.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Switched the fallback toast to surface the server message — it's `error.message` on the tRPC client, not `error.data.message` (`data` only carries `code`/`zodError`). Matches the other modals (PageSettingsModal, AddUserModal, …). Client-side zod validation catches malformed input before submit, so what reaches this toast is a clean `TRPCError` message, not the zod JSON blob.

seaerchin (commented):
## Summary

This is the foundational backend PR for per-site URL redirects: a Prisma migration adds `RedirectCreate`/`RedirectDelete` to the `AuditLogEvent` enum (`packages/db/prisma/migrations/20260610000001_add_redirect_audit_log_events/migration.sql`), shared Zod schemas live in `apps/studio/src/schemas/redirect.ts`, the service in `apps/studio/src/server/modules/redirect/redirect.service.ts`, the tRPC router in `apps/studio/src/server/modules/redirect/redirect.router.ts`, plus a `logRedirectEvent` audit logger (`apps/studio/src/server/modules/audit/audit.service.ts:182`) and the frontend rewire.

Codeflow: `redirect.router.ts` runs `validateUserPermissionsForSite` first — `read` for `list`/`count` (`:18`,`:31`), site-level `create`/`delete` CRUD for the mutations (`:44`,`:60`) — which map to `Site` abilities granted only to site Admins/Isomer admins (`permissions.service.ts:79-83`). `createRedirect` (`redirect.service.ts:78`) runs a transaction: pre-check for a live row → `CONFLICT`, else upsert gated on `deletedAt IS NOT NULL` to revive soft-deleted rows, then `logRedirectEvent` + a `Publish` audit entry, then `publishSite` **after** the txn commits (`:159`). `deleteRedirect` (`:152`) soft-deletes scoped to `siteId` (`NOT_FOUND` otherwise) with the same audit+publish shape. `listRedirects` (`:39`) is live-only, site-scoped, with a `source`→`id` tie-break for stable pagination.

Overall assessment: strong, security-conscious PR. Authorization (the highest-risk axis for this stack) is correct and site-scoped on every endpoint; input validation moved to a sound RFC-3986 whitelist; open-redirect vectors (`//`, `javascript:`, `http://`) are closed; cross-site isolation, conflict/revive races, and publish-failure persistence are all tested (27 integration + 11 schema tests). Findings below are mostly non-blocking correctness/consistency notes.

## Findings

**praise:** Authorization is done right and is the load-bearing part of this PR. Every procedure calls `validateUserPermissionsForSite` with `input.siteId` *before* touching data (`redirect.router.ts:18-66`), mutations require site-level CRUD which is Admin-only (`permissions.service.ts:79-83`), and `listRedirects`/`countRedirects`/`deleteRedirect` are all `where("siteId", "=", siteId)`-scoped — so there's no IDOR across sites. The delete-other-site and list-other-site tests confirm it. Confidence: high

**praise:** The open-redirect surface is handled thoughtfully — `destinationSchema` rejects `http://`/`javascript:` via an allowlist of `/` or `https://`, and collapses protocol-relative `//evil.com` → `/evil.com` (`redirect.ts:42-52`), with tests pinning both. Source uses an RFC-3986 pchar whitelist plus `..`-segment and slash-only rejection (`redirect.ts:14-39`). Confidence: high

**issue: (non-blocking)** `destinationSchema` (`redirect.ts:38-53`) validates only the prefix — unlike `sourceSchema`, it does **not** reject control characters, whitespace, or backslashes. A `destination` like `/foo\r\nbar`, `https://x\\y`, or one containing raw control bytes is accepted, persisted, and later written into the published redirect rules / a `Location` header. Depending on how the static site consumes these rows this is a header/rule-injection or rendering footgun, and it's an inconsistent trust boundary vs. `source`. Suggest applying the same control-char/backslash/whitespace rejection to the path portion of `destination` (the `startsWith("/")` branch at minimum). Confidence: medium

**issue: (non-blocking)** The `Publish` audit entry is written *inside* the transaction (`redirect.service.ts:135`, `:188`) but `publishSite` runs *after* commit (`:159`, `:197`) and can throw — the persistence test even asserts `create` rejects with "CodeBuild unavailable" while the row stays committed (`redirect.router.test.ts:467`). Result: on a publish failure you get a committed `Publish` audit row for a build that never started, and the user sees an error. The data state is self-healing (the next successful mutation republishes), but the audit log can claim a publish that didn't happen. Consider logging the `Publish` event only after `publishSite` succeeds, or noting the decoupling explicitly. Confidence: medium

**question:** `publishSite` no-ops for sites without a CodeBuild project (`codebuild.service.ts:28-37`) — i.e. unlaunched sites. For those, `create`/`delete` still write a `Publish` audit entry every time (`:135`/`:188`) even though nothing publishes. Is a `Publish` audit row the right record for a site that has no build pipeline yet, or should the publish-event logging be conditional on an actual build being triggered? Not blocking, just confirming intent. Confidence: low

**note:** The pre-check `existing` query in `createRedirect` (`:90`) selects without `deletedAt` filter and feeds the audit `before` delta, while the upsert's `WHERE deletedAt IS NOT NULL` guard handles the concurrent-live-create race (returns no row → `CONFLICT` via `executeTakeFirstOrThrow`, `:128`). This is correct and the prior Copilot/seaerchin thread on this is resolved. One residual: under a concurrent revive of the *same* soft-deleted row, both txns pass the pre-check, both match `deletedAt IS NOT NULL`, and the second to commit overwrites the first's `destination`/`createdAt` (last-writer-wins) rather than conflicting. Benign for this feature, but worth being aware of. Confidence: low

**note:** `getByUser` throws `BAD_REQUEST` "Please ensure that you are logged in!" (`redirect.service.ts:31`) — but the router is already `protectedProcedure` + a permission check, so reaching this means a valid session whose `User` row vanished mid-request. `BAD_REQUEST`/"logged in" is a slightly misleading code+message for that case (`INTERNAL_SERVER_ERROR` or `UNAUTHORIZED` would read truer). Cosmetic; matches an existing pattern elsewhere if copied. Confidence: low

**praise:** Audit logging is wired into the same transaction as every state change with real before/after rows (`redirect.service.ts:131`,`:181`), the migration is purely additive `ADD VALUE` with deploy notes calling out the apply-before-deploy ordering, and the `getAuditLogs.ts` report renders the new events including the revived-vs-created distinction (`getAuditLogs.ts:156-157`). This closes the earlier "I forgot audit logs" thread cleanly. Confidence: high

seaerchin (commented):
## Summary

This is the foundational backend PR for per-site URL redirects: a Prisma migration adds `RedirectCreate`/`RedirectDelete` to the `AuditLogEvent` enum (`packages/db/prisma/migrations/20260610000001_add_redirect_audit_log_events/migration.sql`), shared Zod schemas live in `apps/studio/src/schemas/redirect.ts`, the service in `apps/studio/src/server/modules/redirect/redirect.service.ts`, the tRPC router in `apps/studio/src/server/modules/redirect/redirect.router.ts`, plus a `logRedirectEvent` audit logger (`apps/studio/src/server/modules/audit/audit.service.ts:182`) and the frontend rewire.

Codeflow: `redirect.router.ts` runs `validateUserPermissionsForSite` first — `read` for `list`/`count` (`:18`,`:31`), site-level `create`/`delete` CRUD for the mutations (`:44`,`:60`) — which map to `Site` abilities granted only to site Admins/Isomer admins (`permissions.service.ts:79-83`). `createRedirect` (`redirect.service.ts:78`) runs a transaction: pre-check for a live row → `CONFLICT`, else upsert gated on `deletedAt IS NOT NULL` to revive soft-deleted rows, then `logRedirectEvent` + a `Publish` audit entry, then `publishSite` **after** the txn commits (`:159`). `deleteRedirect` (`:152`) soft-deletes scoped to `siteId` (`NOT_FOUND` otherwise) with the same audit+publish shape. `listRedirects` (`:39`) is live-only, site-scoped, with a `source`→`id` tie-break for stable pagination.

Overall assessment: strong, security-conscious PR. Authorization (the highest-risk axis for this stack) is correct and site-scoped on every endpoint; input validation moved to a sound RFC-3986 whitelist; open-redirect vectors (`//`, `javascript:`, `http://`) are closed; cross-site isolation, conflict/revive races, and publish-failure persistence are all tested (27 integration + 11 schema tests). Findings below are mostly non-blocking correctness/consistency notes.

## Findings

**praise:** Authorization is done right and is the load-bearing part of this PR. Every procedure calls `validateUserPermissionsForSite` with `input.siteId` *before* touching data (`redirect.router.ts:18-66`), mutations require site-level CRUD which is Admin-only (`permissions.service.ts:79-83`), and `listRedirects`/`countRedirects`/`deleteRedirect` are all `where("siteId", "=", siteId)`-scoped — so there's no IDOR across sites. The delete-other-site and list-other-site tests confirm it. Confidence: high

**praise:** The open-redirect surface is handled thoughtfully — `destinationSchema` rejects `http://`/`javascript:` via an allowlist of `/` or `https://`, and collapses protocol-relative `//evil.com` → `/evil.com` (`redirect.ts:42-52`), with tests pinning both. Source uses an RFC-3986 pchar whitelist plus `..`-segment and slash-only rejection (`redirect.ts:14-39`). Confidence: high

**issue: (non-blocking)** `destinationSchema` (`redirect.ts:38-53`) validates only the prefix — unlike `sourceSchema`, it does **not** reject control characters, whitespace, or backslashes. A `destination` like `/foo\r\nbar`, `https://x\\y`, or one containing raw control bytes is accepted, persisted, and later written into the published redirect rules / a `Location` header. Depending on how the static site consumes these rows this is a header/rule-injection or rendering footgun, and it's an inconsistent trust boundary vs. `source`. Suggest applying the same control-char/backslash/whitespace rejection to the path portion of `destination` (the `startsWith("/")` branch at minimum). Confidence: medium

**issue: (non-blocking)** The `Publish` audit entry is written *inside* the transaction (`redirect.service.ts:135`, `:188`) but `publishSite` runs *after* commit (`:159`, `:197`) and can throw — the persistence test even asserts `create` rejects with "CodeBuild unavailable" while the row stays committed (`redirect.router.test.ts:467`). Result: on a publish failure you get a committed `Publish` audit row for a build that never started, and the user sees an error. The data state is self-healing (the next successful mutation republishes), but the audit log can claim a publish that didn't happen. Consider logging the `Publish` event only after `publishSite` succeeds, or noting the decoupling explicitly. Confidence: medium

**question:** `publishSite` no-ops for sites without a CodeBuild project (`codebuild.service.ts:28-37`) — i.e. unlaunched sites. For those, `create`/`delete` still write a `Publish` audit entry every time (`:135`/`:188`) even though nothing publishes. Is a `Publish` audit row the right record for a site that has no build pipeline yet, or should the publish-event logging be conditional on an actual build being triggered? Not blocking, just confirming intent. Confidence: low

**note:** The pre-check `existing` query in `createRedirect` (`:90`) selects without `deletedAt` filter and feeds the audit `before` delta, while the upsert's `WHERE deletedAt IS NOT NULL` guard handles the concurrent-live-create race (returns no row → `CONFLICT` via `executeTakeFirstOrThrow`, `:128`). This is correct and the prior Copilot/seaerchin thread on this is resolved. One residual: under a concurrent revive of the *same* soft-deleted row, both txns pass the pre-check, both match `deletedAt IS NOT NULL`, and the second to commit overwrites the first's `destination`/`createdAt` (last-writer-wins) rather than conflicting. Benign for this feature, but worth being aware of. Confidence: low

**note:** `getByUser` throws `BAD_REQUEST` "Please ensure that you are logged in!" (`redirect.service.ts:31`) — but the router is already `protectedProcedure` + a permission check, so reaching this means a valid session whose `User` row vanished mid-request. `BAD_REQUEST`/"logged in" is a slightly misleading code+message for that case (`INTERNAL_SERVER_ERROR` or `UNAUTHORIZED` would read truer). Cosmetic; matches an existing pattern elsewhere if copied. Confidence: low

**praise:** Audit logging is wired into the same transaction as every state change with real before/after rows (`redirect.service.ts:131`,`:181`), the migration is purely additive `ADD VALUE` with deploy notes calling out the apply-before-deploy ordering, and the `getAuditLogs.ts` report renders the new events including the revived-vs-created distinction (`getAuditLogs.ts:156-157`). This closes the earlier "I forgot audit logs" thread cleanly. Confidence: high

PR #2099, opened 2026-05-14
chore(deps): upgrade storybook from 9.1.20 to 10.3.6

adriangohjw:
## Problem

Storybook 9.1.20 is one major version behind. Storybook 10 (released 2025-10-28) adds support for Next 16 / Vitest 4, ships smaller installs, and is the supported track for further updates.

## Solution

Bumped the `storybook` catalog from `9.1.20` to `10.3.6` across both Storybook instances (`apps/studio`, `packages/components`), then resolved every v10 compatibility gap so the upgrade is a no-op behaviourally.

Pinned to `10.3.6` rather than the very latest `10.4.0` because the repo's `minimumReleaseAge: 1440` rule blocks packages younger than 24h, and `10.4.0` shipped < 24h ago.

**Breaking Changes**

**Improvements**:

- Bumped `storybook`, `@storybook/nextjs`, `@storybook/react-vite`, `@storybook/addon-{a11y,docs,links,themes}`, `storybook-addon-pseudo-states`, and `eslint-plugin-storybook` to `10.3.6`.
- Made both `.storybook/main.ts` files ESM-clean (Storybook 10 is ESM-only): dropped the CJS `require.resolve`-based `getAbsolutePath` helper and use bare package names — Storybook resolves them itself.
- Replaced the unmaintained `storybook-mock-date-decorator` (only supports `storybook ^9`, no v10 release) in `apps/studio/.storybook/preview.tsx` with an inline `mockdate`-based decorator that mirrors the existing pattern in `packages/components/.storybook/preview.tsx` and preserves `parameters.date` / `parameters.disableMockDate` semantics.
- Switched `packages/components/tsconfig.json` to `moduleResolution: "bundler"` + `module: "esnext"` so TypeScript honours Storybook 10's package `exports`; kept the published CJS bundle intact by pinning `moduleResolution: "node"` + `module: "commonjs"` in `tsconfig.build.json` (the ESM build inherits the new bundler base, which is what it should have been doing anyway).
- Removed dead `viteFinal` block in `packages/components/.storybook/main.ts` that pre-optimised `storybook-dark-mode` — that package isn't installed and was never used.

## Tests

CI: `pnpm typecheck`, `pnpm lint`, `pnpm build-storybook` all pass. `pnpm exec storybook doctor` is clean for both Storybooks. Both dev servers boot.

- [ ] Run `pnpm storybook` and confirm both Storybooks boot (Components on `:6006`, Studio on `:6007`) without console errors.
- [ ] In Studio (`:6007`), open a story that depends on a mocked date (e.g. `Page/EditPage` stories that don't set `disableMockDate`) and confirm the displayed time matches the preview `parameters.date` default.
- [ ] In Studio, open a story listed with `parameters: { disableMockDate: true }` (e.g. `EditUserModal`, `EditIndexPage`) and confirm it renders without errors.
- [ ] In Studio, open any story that uses MSW-mocked tRPC and confirm data loads.
- [ ] In Studio, toggle the theme and a11y panels from the toolbar and confirm both work.
- [ ] In Components (`:6006`), switch the theme toggle between "Isomer Classic" and "Isomer Next" and confirm both render.
- [ ] In Components, open the pseudo-states addon panel on a button/link story and confirm hover/focus/active variants render.
- [ ] Confirm the Chromatic CI run on this PR shows no unintended visual diffs.

**New dependencies**:

None.

**New dev dependencies**:

None. `storybook-mock-date-decorator` was removed from `apps/studio`; `mockdate` was already a dependency.


adriangohjw (commented):
## Summary

Bumps `storybook` and its addons from `9.1.20` → `10.3.6` (caret-pinned) and resolves the v10 compatibility gaps: drops the CJS `require.resolve`-based `getAbsolutePath` helper in both `.storybook/main.ts` files (Storybook 10 is ESM-only), replaces the unmaintained `storybook-mock-date-decorator` with an inline `mockdate` decorator in `apps/studio/.storybook/preview.tsx:139-152`, and switches `packages/components/tsconfig.json` to `moduleResolution: "bundler"` + `module: "esnext"` while pinning the CJS emit back to `node` / `commonjs` in `packages/components/tsconfig.build.json`.

Codeflow:
- Catalog bump in `pnpm-workspace.yaml:207-216` drives all consumers.
- Studio (`apps/studio/.storybook/main.ts`) and Components (`packages/components/.storybook/main.ts`) configs trimmed to bare-string addon names.
- The studio `webpackFinal` React-unalias workaround (`apps/studio/.storybook/main.ts:43-58`) is preserved.
- Components TS resolution split: ESM build inherits `bundler` from base; CJS build overrides to `node`.

Overall: low-risk, well-scoped upgrade. Build + typecheck + lint pass; both Storybooks boot. A few small behavioural/maintenance items worth a second look before merge.

## Findings

**issue (correctness):** `apps/studio/.storybook/preview.tsx:139-152` — the inline decorator's behaviour drifts from `storybook-mock-date-decorator`'s. The old library only called `MockDate.set(parameters.date)` *if* a date was provided, leaving the global alone otherwise (resetting on unmount). The new code calls `mockdate.reset()` **unconditionally** on every render before the date check. Effects:
1. Stories without `parameters.date` set (only possible if a story explicitly overrides the preview-level default to `undefined`) will now wipe any mock established by another decorator earlier in the chain.
2. The old library's unmount-cleanup is gone — fine in Storybook because the next mount immediately re-mocks, but it's worth being deliberate about.

The preview-level default at `apps/studio/.storybook/preview.tsx:175` (`date: new Date(2024, 10, 28)`) means the practical impact today is near-zero. Still, since `disableMockDate: true` is used by several stories (e.g. `EditUserModal.stories.tsx:86`, `EditHomePage.stories.tsx:128`), please smoke-test those + at least one default-date story in the actual Studio Storybook before merging — types passing isn't enough here.

Confidence: medium

**question (correctness):** `apps/studio/.storybook/main.ts:43-58` — the `webpackFinal` React-unalias block was added against Storybook v9's `@storybook/nextjs` aliasing (issue #30646). Did you confirm v10 still aliases the same keys (`react`, `react-dom/test-utils`, `react-dom$`, `react-dom/client`, `react-dom/server`) in `config.resolve.alias`? If v10 already uses the host's React (via the Next 15 framework integration), this block is now dead code; if v10 aliases additional keys, the block is incomplete and you'll silently get React 19 inside Storybook. Worth logging `config.resolve.alias` once during a dev run to confirm.

Confidence: medium

**question (correctness):** `msw-storybook-addon` stayed at `^2.0.7` (catalog) but its peer dependency only constrains `msw`, not `storybook` — so a v9-era addon binding to internal Storybook APIs that moved in v10 (e.g. the `storybook/*` consolidations of `manager-api` / `preview-api`) could silently no-op at runtime. `build-storybook` succeeding doesn't guarantee MSW interception actually fires. Please verify one tRPC-mocked story in `apps/studio` (`:6007`) shows the mocked payload rather than a network error against the bare-string MSW handler before merge.

Confidence: medium

**suggestion (maintainability):** `packages/components/tsconfig.build.json:2-6` — the split (CJS build uses `moduleResolution: "node"` + `module: "commonjs"`, ESM build inherits `bundler` + `esnext` from the base) is intentional but undocumented. The PR description has the reasoning; please move a one-liner into the build config so a future maintainer doesn't "unify" the two and silently break CJS consumers. e.g. `// CJS emit pinned to node resolution; base uses bundler for typecheck against Storybook 10's package exports.`

Confidence: high

**note:** `pnpm-workspace.yaml:207-216` — caret range `^10.3.6` combined with `minimumReleaseAge: 1440` pins to 10.3.6 today, but in 24h+ any fresh `pnpm install` will silently float to 10.4.x / future 10.x. That's the intended caret behaviour, but worth noting since the PR title commits to a specific version — if `pnpm-lock.yaml` were ever regenerated mid-flight (e.g. by another dependabot bump) the actual installed version would drift. CI on each PR will catch regressions, so this is informational.

Confidence: high

**praise:** `packages/components/.storybook/main.ts` — nice catch removing the `viteFinal` `optimizeDeps.include` for `storybook-dark-mode`. That package isn't installed anywhere in the repo and was effectively a dead include for a long time.

Confidence: high

seaerchin (approved):

PR #2101, opened 2026-05-15
🧪 experiment: migrate typecheck from tsc to tsgo

adriangohjw:
## Problem

`tsc` is the slowest step in our `pnpm typecheck` pipeline. Microsoft now ships a Go-based TypeScript compiler (`tsgo`, distributed as `@typescript/native-preview`) that is a near-drop-in for `tsc --noEmit` and runs roughly 2× faster end-to-end.

## Solution

Swap `tsc --noEmit` → `tsgo --noEmit` in every workspace's `typecheck` script. Build / emit pipelines (`build:main`, `build:module`, `dev`) keep stock `tsc` — tsgo's emit is still preview-quality and we depend on `tsc-alias` post-processing.

`packages/components`: tsgo dropped the legacy `moduleResolution: "node"` algorithm, so its base tsconfig moves to `moduleResolution: bundler` + `module: esnext` (matching the direction of #2099). `tsconfig.build.json` is pinned back to `node` + `commonjs` so the published CJS artifact is byte-identical to before. Adds tiny ambient `*.css` / `*.scss` / `@fontsource/ibm-plex-mono` shims that bundler resolution requires (legacy `node` silently ignored these side-effect imports).

`apps/studio/tests/integration/helpers/seed/index.ts`: replaces a single misplaced `@ts-expect-error` (which `tsc` happened to honor at the outer `.values()` call while tsgo binds it to the inner `config:` key) with two `@ts-ignore` directives — one at each line where the kysely JSONB brand-tag error attaches. `@ts-ignore` is silent when unused, so both compilers stay clean. No business-logic change.

**Breaking Changes**

**Improvements**:

rough non-scientific benchmarking

| | before | after | diff |
| - | - | - | - |
| CI (turbo) | 61s | 45s | 1.35x faster |
| local (turbo) | 25s | 7s | 3.57x faster |
| local (packages/components) | 2.4 | 0.6s | 4x faster |
| local (apps/studio) | 5.3s | 2.2s | 2.4x faster |

_for reference, 1.7x faster on CI for starter-kit https://github.com/opengovsg/starter-kit/pull/585_

**Caveat**:

- `tsgo` is still a preview build (`@typescript/native-preview`). It is currently slightly softer than `tsc` on a few checks. Keeping an occasional `tsc` run as a backstop is reasonable until parity is reached.

## Tests

- [ ] Run `pnpm install` at the repo root and confirm it succeeds.
- [ ] Run `pnpm typecheck` at the repo root and confirm all 7 tasks pass.
- [ ] From `packages/components`, run `pnpm build` and confirm `dist/cjs/index.js` + `dist/esm/index.js` are produced and the package's existing consumers (e.g. `apps/studio`) still import/build correctly.
- [ ] Run the existing E2E / integration test suites (`pnpm test:e2e`, `pnpm --filter isomer-studio test:unit`) to confirm the `seed/index.ts` `@ts-ignore` swap has no runtime effect.

**New dev dependencies**:

- `@typescript/native-preview` (pinned to `7.0.0-dev.20260513.1`, added via the pnpm catalog): the Go-based TypeScript compiler providing the `tsgo` binary used by the new typecheck scripts.


adriangohjw (commented):
## Summary

- Swaps `tsc --noEmit` → `tsgo --noEmit` in the four workspaces that have a `typecheck` script: `apps/studio/package.json:53`, `packages/components/package.json:57`, `tooling/storybook/package.json:15`, `tooling/template/package.json:15`. Build/emit pipelines remain on stock `tsc`.
- `packages/components`: base `tsconfig.json:7-8` flips to `moduleResolution: bundler` + `module: esnext`; `tsconfig.build.json:3-6` is then pinned back to `commonjs`/`node` so the CJS artifact is unchanged. Ambient shims added at `apps/studio/src/types/shims.d.ts`, `packages/components/src/types/css.d.ts`, `tooling/template/styles/css.d.ts` to satisfy bundler-mode side-effect imports.
- `apps/studio/tests/integration/helpers/seed/index.ts:93,96`: one `@ts-expect-error` replaced with two `@ts-ignore` directives to satisfy both tsc and tsgo simultaneously.
- Overall: the change is mechanically sound and CI's `typecheck` task (`.github/workflows/ci.yml:64`) now runs end-to-end on tsgo. The notable risks are around the deliberately-softer suppressions in seed and the broader CI gate change.

## Findings

**issue (maintainability):** `apps/studio/tests/integration/helpers/seed/index.ts:93,96` — replacing one `@ts-expect-error` with two `@ts-ignore` trades a tripwire for a silent suppression at two sites. `@ts-expect-error` will fail when the underlying types change such that the suppression is no longer needed (a useful signal that the test fixture can be cleaned up). `@ts-ignore` won't. Two of them widen the surface where a real type regression could be hidden inside this insertion. Options that avoid both: (a) cast just the offending field — `config: { ... } as unknown as SiteJsonConfig` — so no suppression directive is needed at all; (b) introduce a small typed `seedSite()` helper that returns the correct kysely value expression; or (c) drop tsc once tsgo is the only gate and revert to a single `@ts-expect-error` placed where tsgo binds the error.

Confidence: high

**issue (security/quality, CI gate softened):** `.github/workflows/ci.yml:64-73` — the `typecheck` job now invokes `pnpm run typecheck` which exclusively runs tsgo. The PR body acknowledges tsgo currently catches *fewer* errors than tsc (the seed regression in this PR was itself one such case where the two diverged). For the duration that tsgo is a preview, real type bugs that tsc would catch can land on `main`. Consider adding a parallel `tsc --noEmit` step (or a separate job) in this workflow until tsgo reaches feature parity — costs ~20s of CI time and preserves the previous safety net.

Confidence: high

**suggestion:** `apps/studio/src/types/shims.d.ts:3` — `declare module "@fontsource/ibm-plex-mono"` makes *any* import from that package `any`. The actual usage is side-effect-only (`.storybook/preview.tsx:1`, `src/pages/_app.tsx:1`), so the shim can be tightened to the side-effect entrypoint: `declare module "@fontsource/ibm-plex-mono/index.css"` (or a `@fontsource/*` glob in case more font packages are added). Same applies generally to the `*.css` / `*.scss` shims — those are conventionally fine but you could narrow them with `declare module "*.css" { const css: string; export default css }` if you ever import the URL.

Confidence: medium

**question:** `packages/components/tsconfig.module.json:1-7` — this file extends the base but doesn't override `moduleResolution`, so the ESM build (`tsconfig.module.build.json`) now sees `moduleResolution: bundler` at typecheck-time where it previously saw `node`. Emit bytes are governed by `module`, so the `.js` output should be identical; however, the `.d.ts` output can shift if a dependency's `exports` map distributes different types under the `import` condition than under the legacy `node` resolution. Worth a `pnpm --filter @opengovsg/isomer-components build` + diff of `dist/esm/**/*.d.ts` against the pre-PR build before merging.

Confidence: medium

**issue (dev watcher behavior change):** `packages/components/package.json:45` — `"dev": "tsc -w"` uses the base `tsconfig.json` directly, which now has `module: esnext`, `moduleResolution: bundler`, and still `outDir: dist/cjs`, `declaration: true`. Previously the watch emitted CJS into `dist/cjs/`; now it will emit ESM into `dist/cjs/`. Anyone running `pnpm dev` for live-reload against a downstream CJS consumer (e.g. parts of `apps/studio` import via `require`) may get broken intermediate output. Either point `dev` at `tsconfig.build.json` (which is now pinned to CJS) or add `--noEmit` to make this a pure typecheck watcher.

Confidence: medium

**note:** `pnpm-workspace.yaml:82` — `@typescript/native-preview` is pinned to a specific daily build (`7.0.0-dev.20260513.1`). That's the right call for now, but daily preview tarballs can be unpublished/retitled, and there's no stable channel yet. Worth documenting an owner (or a recurring chore) for bumping this so the team doesn't end up stuck on a vanished version when someone next regenerates the lockfile.

Confidence: low

**praise:** `packages/components/tsconfig.build.json:3-6` — pinning the CJS build back to `module: commonjs` + `moduleResolution: node` via the build-specific tsconfig (instead of leaving the base on the legacy settings) is the right shape — keeps the published artifact identical while unblocking the modern resolution mode for typecheck. Matches what #2099 is doing.

Confidence: high

seaerchin (approved):

PR #2119, opened 2026-05-17
chore: migrate Next.js from 15.5.18 to 16.2.6

adriangohjw:
## Problem

Next.js 15 reaches end-of-life on **2026-10-21** (~5 months away). Without a proactive upgrade, we risk being forced into a rushed migration to stay on a supported version receiving security patches. Migrating now also unlocks Turbopack as the default dev/build engine for Studio — a meaningful DX improvement (faster HMR, faster cold starts).

Closes [insert issue #]

## Solution

Bump the `next` catalog entry from `^15.5.18` to `^16.2.6` and adopt the new defaults for Studio, while keeping the static-export template pinned to webpack to protect runtime bundle size.

### Why Studio uses Turbopack but `tooling/template` stays on `--webpack`

- **Studio** (`apps/studio`): Now uses Turbopack (the Next 16 default). Studio is an authenticated CMS app — bundle size has minimal impact on Lighthouse scores since it isn't a public-facing site, but dev-iteration speed matters a lot.
- **`tooling/template`**: Kept on webpack via the per-invocation `--webpack` flag on `next dev` and `next build`. Community reports for Next 16 Turbopack consistently flag larger client bundles vs. webpack. Template output ships to every public Isomer site, where bundle size directly hits **Lighthouse TBT (Total Blocking Time)** and other Core Web Vitals. We'd rather preserve current static-export parity until Turbopack's production bundle matches or beats webpack's.

### Breaking changes from Next 15 → 16 and how each affects us

| Breaking change                                            | Affects us? | Notes                                                                                                |
| ---------------------------------------------------------- | ----------- | ---------------------------------------------------------------------------------------------------- |
| Node.js 20.9+ required (Node 18 dropped)                   | No          | CI/runtime already on Node 20+.                                                                      |
| Turbopack is now the default for `next dev` and `next build` | Yes (intentional) | Adopted for Studio; opted out for `tooling/template` via `--webpack`.                          |
| `experimental.turbo` → top-level `turbopack` config        | Yes         | Removed `turbo.moduleIdStrategy` from `tooling/template/next.config.mjs` (now irrelevant — webpack). |
| `images.domains` removed → `images.remotePatterns` only    | Yes         | Migrated `apps/studio/next.config.mjs` to `remotePatterns`.                                          |
| `publicRuntimeConfig` / `serverRuntimeConfig` discouraged  | Yes         | Removed unused `publicRuntimeConfig.NODE_ENV` from `apps/studio/next.config.mjs` (read from `env` directly elsewhere). |
| `next lint` / built-in ESLint integration removed          | Yes         | We use Oxlint via a separate CI task. Dropped `eslint: { ignoreDuringBuilds: true }` — no longer a valid key. |
| `indexExtensions` config removed                           | Yes         | Removed from `tooling/template/next.config.mjs`.                                                     |
| AMP support removed                                        | No          | Not used.                                                                                            |
| `next/legacy/image` removed                                | No          | Not used.                                                                                            |
| `swcMinify` removed (always on)                            | No          | Not set in our config.                                                                               |
| App Router async `params` / `searchParams`                 | No          | Studio uses the **Pages Router**.                                                                    |
| App Router caching changes (`'use cache'`, `cacheLife`)    | No          | Studio uses the **Pages Router**.                                                                    |
| React 19 requirement                                       | No          | Next 16 still supports React 18; we remain on React 18.3.1.                                          |
| Browser console logs forwarded to terminal by default      | Yes         | New `[browser] ...` lines in dev terminal — added `logging.browserToTerminal: false` to suppress, since third-party noise (`findDOMNode` from `react-input-mask` in the OGP design system's `DatePicker`) is unfixable from this repo. Warnings still surface in browser DevTools. |
| `next build` runs type checking even with separate CI step | Yes         | Added `typescript.ignoreBuildErrors: true` to `apps/studio/next.config.mjs` — typecheck is already a CI task. |

### Studio config changes (`apps/studio/next.config.mjs`)

- Removed the custom `webpack()` function: `dd-trace` is server-only (Turbopack handles this automatically), and `hidden-source-map` overrides are no longer needed.
- Replaced `images.domains` with `images.remotePatterns`.
- Removed `publicRuntimeConfig` (unused; `env.NODE_ENV` is read directly in the codebase).
- Removed `eslint.ignoreDuringBuilds` (config key removed in Next 16).
- Added `typescript.ignoreBuildErrors: true` since typechecking runs separately in CI.
- Added `logging.browserToTerminal: false` to suppress noisy third-party browser warnings being forwarded to the terminal.

### Template config changes (`tooling/template/next.config.mjs` + `package.json`)

- `dev` and `build:template` scripts now invoke `next dev --webpack` / `next build --webpack` to keep the existing webpack-based static export.
- Removed `indexExtensions`, `turbo.moduleIdStrategy`, and `eslint.ignoreDuringBuilds` (all removed/no-op in Next 16).

### `apps/studio/tsconfig.json`

- `jsx: "preserve"` → `jsx: "react-jsx"` (Next 16 codemod default; aligns with the new JSX transform).
- Cosmetic reformatting of array values (Oxfmt).

**Breaking Changes**

  - This is a framework upgrade with no user-facing or API surface changes. All config changes track Next.js 16's renamed/removed keys and our existing behavior is preserved.

**Features**:

- Turbopack adopted as the default dev/build engine for `apps/studio` — significantly faster HMR and cold starts.

**Improvements**:

- Next.js bumped to `16.2.6`, putting us comfortably ahead of Next 15 EOL (2026-10-21).
- Browser console warnings no longer spam the dev terminal (`logging.browserToTerminal: false`); they remain available in browser DevTools.
- Static-export template (`tooling/template`) explicitly pinned to webpack to protect public-site bundle size and Lighthouse TBT.

**Bug Fixes**:

- None — this is a framework upgrade.

## Before & After Screenshots

On CI build, 4 mins 11 seconds -> 1 min 36 seconds (~2.65x faster)

| before | after |
| - | - |
| Screenshot 2026-05-18 at 6 03 12 PM | Screenshot 2026-05-18 at 6 03 36 PM |

## Tests

- [ ] `pnpm install` succeeds with the updated lockfile.
- [ ] From `apps/studio`, run `pnpm dev` and confirm the studio boots on `localhost:3000` with Turbopack (look for "Turbopack" in the boot banner).
- [ ] In the running studio dev server, log in and open a page in the editor — confirm the editing drawer (`EditLink`, `FormBuilder`, `JsonForms*` renderers) loads without crashing.
- [ ] In the same dev session, confirm the terminal no longer shows `[browser] ...` warning lines, while the browser DevTools console still does.
- [ ] From `apps/studio`, run `pnpm build` and confirm a successful production build with Turbopack.
- [ ] From `tooling/template`, run `pnpm dev` and confirm the template boots on `localhost:3001` using **webpack** (look for the webpack banner, not Turbopack).
- [ ] From `tooling/template`, run `pnpm build:template` and confirm the static export completes successfully.
- [ ] Smoke-test image rendering on a Studio page that uses an asset from `NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME` — confirms `images.remotePatterns` migration works.
- [ ] Run `pnpm typecheck` and `pnpm lint` from the repo root and confirm both pass.
- [ ] Run `pnpm test:e2e` and confirm the Playwright suite passes.

**New scripts**:

- None.

**New dependencies**:

- `next@^16.2.6` — bumped from `^15.5.18` (catalog entry).

**New dev dependencies**:

- None.


adriangohjw:
## Blocked on #2099

This PR cannot be merged until #2099 is merged first.

### Why

The Storybook CI failure (`TypeError: swc.isWasm is not a function`) is caused by `@storybook/nextjs@9.1.20` being incompatible with Next.js 16. Next.js 16 removed `isWasm` as a top-level SWC function — bindings must now be explicitly loaded via `loadBindings()` first.

The Storybook maintainers have declined to backport the fix to 9.x ([storybookjs/storybook#33088](https://github.com/storybookjs/storybook/issues/33088)). The proper fix is to upgrade to **Storybook 10**, which has native Next.js 16 support.

### Merge order

1. #2099 (Storybook 10 upgrade) → merge first, remove the patch from this PR
2. This PR → merge after

adriangohjw (commented):
## Summary

- Bumps `next` from `^15.5.18` → `^16.2.6` across the catalog, replaces removed/renamed config keys, and adopts Turbopack as the default in `apps/studio` while pinning `tooling/template` to webpack via `--webpack`.
- Codeflow: catalog bump in `pnpm-workspace.yaml:131` propagates to all importers in `pnpm-lock.yaml`; `apps/studio/next.config.mjs:145-208` is rewritten (custom `webpack()` deleted, `publicRuntimeConfig` dropped, `images.domains` → `remotePatterns`, new `logging.browserToTerminal: false` and `typescript.ignoreBuildErrors: true`); `tooling/template/next.config.mjs:1-12` drops `indexExtensions`, `turbo.moduleIdStrategy`, and `eslint.ignoreDuringBuilds`; `tooling/template/package.json:7,9` flips to `--webpack`; `apps/studio/tsconfig.json:16` switches `jsx` to `react-jsx`. The PR is also self-blocked on #2099 (Storybook 10) per the existing PR comment.
- Overall assessment: The change is well-scoped, the PR body documents every breaking-change axis, and the audited keys all match Next 16's renames. Two judgment calls warrant a second look (source-map exposure on Turbopack, and silently disabling build-time TS checks); the rest reads as a clean, deliberate upgrade.

## Findings

**issue (security):** `apps/studio/next.config.mjs:155` — `productionBrowserSourceMaps: true` is kept, but the deleted `webpack()` override was the thing that downgraded the client devtool to `hidden-source-map`, i.e. emit source maps without a `//# sourceMappingURL=` comment in the bundle. With that gone, Turbopack's default behavior for production source maps will likely re-attach the `sourceMappingURL` reference (or expose `*.map` files at predictable URLs), so anyone visiting the Studio with DevTools open can fetch the original TS sources — internal route shapes, tRPC contracts, auth-flow logic, env-derived constants, etc. That was an intentional defense before; removing it without a replacement weakens the production posture of an authenticated CMS. Please verify by building (`pnpm build` in `apps/studio`) and inspecting whether the client chunks under `.next/static/chunks/*.js` contain `sourceMappingURL` and whether `.map` files are emitted+served. If they are, either set `productionBrowserSourceMaps: false` for the public bundle and continue uploading source maps to Datadog out-of-band, or check whether Next 16 exposes a `hiddenSourceMaps`-style option to restore parity.

Confidence: medium

**issue:** `apps/studio/next.config.mjs:157` — `typescript: { ignoreBuildErrors: true }` is being added on the rationale that "typecheck is already a CI task," but that's a behavioral regression vs. Next 15, which would have failed the build on type errors. The PR body frames this as a side-effect of Next 16; it isn't — this key exists in 15 too, and not setting it was the safety net. If the CI typecheck job is ever skipped (path filters, required-status drift, a forced merge, a Docker-only build path), a type-broken bundle can now ship to prod. At minimum, please confirm that `pnpm typecheck` is a required status check on `main` for this repo and call that out explicitly so the guardrail is documented. Better: drop this key and let `next build` fail fast on the same machine that produces the artifact.

Confidence: high

**question:** `apps/studio/next.config.mjs:148-150` — is `logging.browserToTerminal` the actual Next 16 key name? Next 16's `logging` config historically exposes `logging.fetches.*`; the new browser-log-forwarding feature may be configured differently (e.g. nested under a different namespace, or controlled via a CLI flag / `experimental` block). I wasn't able to verify against the Next 16 docs in this session. Worth double-checking against the official release notes/upgrade guide so this doesn't silently no-op and re-introduce the noisy `[browser] ...` lines later.

Confidence: low

**thought:** `apps/studio/next.config.mjs` (removed `webpack()` block) — the explicit `dd-trace` external was a belt-and-suspenders guard against any future client import accidentally pulling the server-only tracer into a browser chunk. Today `tracer.ts` is only loaded from `instrumentation.ts` behind a `NEXT_RUNTIME === "nodejs"` check, and `database/database.ts` is server-only, so this is currently safe. Worth a mental note that with Turbopack we now rely entirely on import-graph hygiene; consider adding a lint rule or an explicit `serverExternalPackages: [..., "dd-trace"]` entry to make the boundary load-bearing rather than implicit.

Confidence: medium

**praise:** PR body is unusually good — every Next 15→16 breaking change is enumerated with an "affects us?" column and a justification for the Studio-vs-template Turbopack split tied to a concrete metric (Lighthouse TBT on public sites). Easy to review against.

Confidence: high

**note:** Existing PR comment already calls out the Storybook 10 dependency (#2099). No action here — just flagging so the merge order isn't forgotten.

Confidence: high

seaerchin (approved):

PR #2159, opened 2026-05-19
fix: await getCollectionItemByPermalink to prevent unhandled rejection

adriangohjw:
## Problem

An intermittent unhandled rejection error was occurring in CI during the Vitest test run. The error manifested as:

```
Error: no result
 ❯ SelectQueryBuilderImpl.executeTakeFirstOrThrow
```

The query was looking for a \`Resource\` with \`parentId IS NULL\` and \`permalink = 'test-page'\`, but finding no result. This caused false positive test failures that were timing-dependent and hard to reproduce locally.

## Solution

The root cause was a missing \`await\` on a call to \`getCollectionItemByPermalink()\` in the test "should create a new \`draftBlob\` if it is currently \`null\`". Without \`await\`, the Kysely \`executeTakeFirstOrThrow()\` promise was left pending when the test completed. The next test's \`beforeEach\` hook then called \`resetTables()\`, clearing the database while the query was still in-flight — so when it resolved, the row was gone and Kysely threw an unhandled rejection.

**Breaking Changes**

**Bug Fixes**:

- Added missing `await` on `getCollectionItemByPermalink()` call in `collection.router.test.ts` (line 1356), preventing a dangling promise that caused intermittent unhandled rejections in CI.

## Before & After Screenshots

N/A — test-only change.

## Tests

No manual verification steps required — this is a test-only fix.

### DIFF
--- a/apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
+++ b/apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
@@ -1357,7 +1357,8 @@ describe("collection.router", async () => {
         page.permalink,
         page.parentId,
       )
-      expect(expected).toMatchObject(actual)
+      expect(actual.draftBlobId).toEqual(expected.id)
+      expect(expected.content.content).toEqual([])
     })
 
     it("should update the collection link successfully", async () => {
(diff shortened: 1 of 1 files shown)


adriangohjw:
damn, claude fixes the flakey test that has been happening for so long

seaerchin (approved):

PR #2191, opened 2026-05-21
fix(storybook): resolve node: URI scheme error in webpack 5

seaerchin:
## Problem

Storybook (\`apps/studio\`) fails to start with the following error:

\`\`\`
Module build failed: UnhandledSchemeError: Reading from "node:path" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
\`\`\`

Webpack 5 does not natively handle the \`node:\` URI prefix (e.g. \`node:path\`) used by modern packages. The error is thrown during scheme detection — which runs **before** \`resolve.alias\` is consulted — so alias-based fixes don't work.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Added a \`webpack.NormalModuleReplacementPlugin\` in \`webpackFinal\` inside \`apps/studio/.storybook/main.ts\`. It intercepts every module request matching \`/^node:/\` **before** the resolver runs and strips the prefix, so webpack sees a plain built-in module name (e.g. \`path\`) that it already knows how to handle.

## Before & After Screenshots

**BEFORE**: Storybook fails with \`UnhandledSchemeError\` on \`node:path\`.

**AFTER**: Storybook starts successfully.

## Tests

**New scripts**: none

**New dependencies**: none

### DIFF
--- a/apps/studio/.storybook/main.ts
+++ b/apps/studio/.storybook/main.ts
@@ -54,6 +54,29 @@ const config: StorybookConfig = {
         delete config.resolve.alias[unalias]
       }
     }
+
+    // Webpack 5 doesn't handle the node: URI scheme used by some packages.
+    // Map node:-prefixed builtins to their unprefixed equivalents so webpack
+    // can resolve them via its existing node polyfill / externals handling.
+    config.resolve = config.resolve ?? {}
+    config.resolve.alias = {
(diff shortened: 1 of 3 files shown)


adriangohjw:
haha fixed in https://github.com/opengovsg/isomer/pull/2190

PR #2199, opened 2026-05-22
chore(studio): add Folder→Collection conversion scripts

adriangohjw:
## Problem

Some sites have legacy Folders that should be Collections (e.g. newsrooms, resources). Today the only way to convert one is to manually edit the database — risky, slow, and easy to get wrong, especially when source pages contain blocks that aren't allowed in the `article` layout.

## Solution

Adds a developer workflow under `apps/studio/prisma/scripts/convert-folder-to-collection/` that is explicitly split into two steps so a human can review the planned changes before any DB write happens.

**Breaking Changes**

**Features**:

- `exportConversionPlan.ts` (read-only) — prompts for site ID + folder ID, verifies both exist in the DB, then writes:
  - `.out/convert-folder-<id>-<ts>.json` (machine-readable plan)
  - `.out/convert-folder-<id>-<ts>.report.md` (human-readable report listing per-page block counts and any disallowed-in-article blocks)
- `applyConversionPlan.ts` (writes) — prompts for site ID + folder ID, verifies they match the loaded plan, then writes draft blobs and flips `Resource.type` (Folder→Collection, Page→CollectionPage) in a single transaction. Optionally mass-publishes all converted resources with one site rebuild at the end.

**Improvements**:

- Pure conversion logic (`buildCollectionIndexBlob`, `buildArticleBlob`, `findDisallowedBlocks`, etc.) lives in `helpers.ts` with 30 unit tests in `helpers.test.ts` (no DB needed to run).
- Both scripts validate IDs against `Site` and `Resource` tables before doing anything, and `applyConversionPlan.ts` cross-checks the loaded plan against the prompted IDs to prevent applying the wrong file.

**Bug Fixes**:

- N/A

## Tests

`pnpm test:unit -- prisma/scripts/convert-folder-to-collection/helpers.test.ts` → 30 passing tests covering the pure conversion helpers.

- [ ] On a non-production environment, identify a test Folder with one IndexPage and a few Page children. Run `source .env && pnpm exec tsx prisma/scripts/convert-folder-to-collection/exportConversionPlan.ts` and supply that site/folder ID.
- [ ] Verify the script refuses to proceed if either ID does not exist or the resource is not type=Folder.
- [ ] Open the generated `.out/*.report.md` and confirm the page list and any disallowed-in-article block warnings look correct. Confirm no DB rows were modified at this point.
- [ ] Run `source .env && pnpm exec tsx prisma/scripts/convert-folder-to-collection/applyConversionPlan.ts`, supply the same IDs, accept the suggested plan path, and confirm the apply prompt.
- [ ] In Studio, verify the Folder now appears as a Collection, the IndexPage renders as a collection landing page, and each former Page renders as an article with the chosen default category.
- [ ] Re-run `applyConversionPlan.ts` against the now-converted resource and verify it errors out (type is no longer Folder) — confirms accidental re-apply is blocked.
- [ ] (Optional) Re-run the apply flow and choose "mass-publish": verify a site rebuild is triggered and the published version reflects the converted layout.

**New scripts**:

- `prisma/scripts/convert-folder-to-collection/exportConversionPlan.ts` — workflow 1, exports a conversion plan + report to local files.
- `prisma/scripts/convert-folder-to-collection/applyConversionPlan.ts` — workflow 2, applies a reviewed plan to the DB and optionally mass-publishes.

**New dev dependencies**:

- `@inquirer/prompts` — interactive CLI prompts used by both scripts.


adriangohjw (commented):
## Summary

This PR introduces a two-step developer workflow (`exportConversionPlan.ts` → `applyConversionPlan.ts`) for converting legacy Folder resources to Collections in the DB. The human-review checkpoint before any write is the key safety feature. Pure conversion helpers in `helpers.ts` are well-isolated and have 30 unit tests covering them without a DB.

**Codeflow:** `exportConversionPlan.ts` reads the DB, builds `ConversionPlan` in memory, and writes `.out/*.json` + a report. `applyConversionPlan.ts` loads those files, cross-checks plan IDs against the prompted IDs (`shared.ts:L160-168`), then writes draft blobs + flips `Resource.type` in a single transaction (`applyConversionPlan.ts:L67-97`). Optional mass-publish calls `incrementVersion` inside its own transaction (`applyConversionPlan.ts:L112-128`).

**Overall:** Solid design. A few correctness and usability concerns worth addressing before this is run on any production-like data.

---

### Findings

**issue:** `shared.ts:L52-61` — `getBlobOfResource` throws a cryptic Kysely "no rows" error when `publishedVersionId` is also null. The subquery `SELECT blobId FROM Version WHERE id = NULL` matches nothing; `executeTakeFirstOrThrow()` fires with no context. Any page that was created but never saved (common in test/staging environments) will produce a confusing failure mid-conversion rather than a clear "this page has no blob" message.

```ts
// After the draftBlobId guard, add:
if (!publishedVersionId) {
  throw new Error(`Resource ${resourceId} has no draft blob and no published version`)
}
```

Confidence: high

---

**suggestion:** `shared.ts:L89-108` — `updateBlobById` does a redundant write when no draft exists. When `draftBlobId` is null, the new Blob is INSERTed with `content` already set (`L92`), then immediately UPDATEd with the same `content` (`L103-108`). The final `updateTable` could be moved inside an `else` branch (or the insert path could just `returningAll()` directly), saving an extra round-trip per page inside what may already be a long transaction.

Confidence: high

---

**note:** `shared.ts:L273-275` and `L253` — plan JSON files are loaded with a bare `JSON.parse(...) as FolderPlan` / `as PagePlan`. There is no schema validation. A manually edited or partially-written plan file (e.g., from a crashed export) would be silently accepted and could write structurally invalid blobs to the DB. Given the `as unknown as IsomerSchema` casts that flow all the way into `updateBlobById`, there is nothing downstream to catch the corruption before it lands in the `Blob` table.

At a minimum, checking for the presence of required top-level keys (`id`, `siteId`, `indexPageId`, `pageIds`) before trusting the file would make failures much easier to diagnose.

Confidence: high

---

**note:** `exportConversionPlan.ts:L269-273` — `FolderMeta` children are explicitly excluded from the `unexpected` guard and silently skipped. After conversion the `FolderMeta` resource remains as a child of the now-Collection. If Studio or the renderer queries `WHERE parentId = <collectionId>` and doesn't expect a `FolderMeta` sibling, it could surface as a rendering bug or an orphaned entry in the CMS UI. Worth confirming this is safe or adding a step to handle/delete it.

Confidence: medium

---

**note:** `shared.ts:L234-239` — `writeJson` silently overwrites existing plan files. If a developer re-runs `exportConversionPlan` for the same folder with a different `defaultCategory`, the `.out/*.json` files are updated in place with no warning. The prior-reviewed plan is gone. Since `applyConversionPlan` defaults to the existing plan path and shows the plan before confirming, this is recoverable, but adding a "Plan files already exist, overwrite? (y/N)" prompt would prevent the subtle foot-gun of reviewing stale output.

Confidence: medium

---

**note:** `applyConversionPlan.ts:L67-72` vs `L79-83` and `L92-95` — the Folder resource's `state` is not changed when its `type` is flipped to `Collection`, but the IndexPage and every child page are explicitly set to `ResourceState.Draft`. If the original Folder was `Published`, the Collection container becomes `Published` while its content pages are `Draft`. Whether this is intentional (the collection shell is live, pages pending review) should be documented; it's non-obvious and the inconsistency could confuse Studio's state display.

Confidence: medium

---

**praise:** The plan cross-check at `applyConversionPlan.ts:L160-168` (folder ID + site ID must match the loaded plan) is exactly the right guard against accidentally applying the wrong plan file. Nicely done.

**praise:** `helpers.ts` keeping all pure logic isolated from DB imports means the 30 unit tests run without any infrastructure, which is the right call for logic this critical to data integrity.

seaerchin (approved):

PR #2207, opened 2026-05-24
fix: reject whitespace-only site names

jiaxinnns:
## Problem

Whitespace-only siteName values pass UI (AJV) and backend (Zod) validation and are persisted; expected is to reject whitespace-only site names (must contain at least one non-whitespace character).
Closes https://github.com/opengovsg/isomer/issues/1953

## Solution

Applied non-whitespace validation consistently in both frontend and backend to reject whitespace-only site names.

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- Whitespace-only site names are no longer accepted.

## Tests

`cd apps/studio && pnpm test:unit src/schemas/__tests__/site.test.ts`
(Related unit tests were added for the Zod schema as it was stated as a recommendation in the linked issue. They can be omitted if they add little value.)

- [ ] Create a site with a whitespace-only name and verify that an error message is shown.


packages/components/src/types/site.ts
```diff
@@ -10,6 +10,7 @@ import {
   VicaSchema,
 } from "~/interfaces"
 import { NotificationSettingsSchema } from "~/interfaces/internal/Notification"
+import { NON_EMPTY_STRING_REGEX } from "~/utils/validation"
```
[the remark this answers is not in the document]
jiaxinnns: NON_EMPTY_STRING_REGEX is already a string

packages/components/src/types/site.ts
```diff
@@ -18,6 +19,10 @@ export const AgencySettingsSchema = Type.Object({
     title: "Site name",
     description:
       "This is displayed on browser tabs, the footer, and the Search Results page. It’s also the default meta title of your homepage.",
+    pattern: NON_EMPTY_STRING_REGEX,
```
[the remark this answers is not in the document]
jiaxinnns: NON_EMPTY_STRING_REGEX is already a string

packages/components/src/types/site.ts
```diff
@@ -18,6 +19,10 @@ export const AgencySettingsSchema = Type.Object({
...
+    errorMessage: {
+      pattern: "Site name cannot be empty or whitespace-only",
+    },
```
adriangohjw, line 25: ask: can we ignore copilot and stick to `"cannot be empty or contain only spaces"` for consistency?

apps/studio/src/schemas/site.ts
```diff
@@ -85,7 +85,7 @@ export const updateSiteConfigSchema = z
   .and(
     z.object({
       siteId: z.number(),
-      siteName: z.string(),
+      siteName: z.string().trim().min(1, { message: "Site name is required" }),
     }),
   )
```
adriangohjw, line 90: can we add unit test for the router to ensure invalid inputs are rejected? `apps/studio/src/server/modules/site/__tests__/site.router.test.ts`

apps/studio/src/schemas/__tests__/site.test.ts
adriangohjw, line 1: thanks for the initiative for this but we can skip this test. Nothing wrong with it, but we generally trust the schema/type generated by jsonform so there's not as much value in having this test. Having unit test on the router is sufficient IMO

adriangohjw (changes requested):
thanks for the PR! left some comments - nothing technically wrong, just a preference on our end. thanks!

adriangohjw (commented):
## Summary

This PR fixes a bug where whitespace-only site names (e.g. `" "`, `"\t"`) bypassed both UI (AJV) and backend (Zod) validation and were persisted. The fix applies validation consistently at both layers:

- **`packages/components/src/types/site.ts`**: Adds `pattern: NON_EMPTY_STRING_REGEX` + `errorMessage` to `AgencySettingsSchema.siteName`, which gates the AJV validator compiled in `site.router.ts:80` and used in `updateSiteConfigSchema`.
- **`apps/studio/src/schemas/site.ts`**: Adds `.trim().min(1)` to `createSiteSchema.siteName` (L73) and `updateSiteConfigSchema.siteName` (L88).
- **`apps/studio/src/schemas/__tests__/site.test.ts`**: New test file covering both schemas for whitespace-only rejection and valid name acceptance.

The two validation layers are independent and will both reject whitespace-only names; `ajv-errors` is correctly wired up in `apps/studio/src/utils/ajv.ts` so the `errorMessage` keyword will fire.

---

**praise:** The `NON_EMPTY_STRING_REGEX` constant from `utils/validation.ts:206` was reused rather than inlining a raw regex string. Consistent with the existing pattern in that file.

**issue (non-blocking):** `apps/studio/src/schemas/site.ts:L73,L88` — The Zod error message `"Site name is required"` will fire for both an empty string and a whitespace-only string. For the whitespace case a user who submits `"   "` will see `"Site name is required"` on the backend, but the frontend AJV message correctly says `"Site name cannot be empty or whitespace-only"`. The mismatch is confusing for API consumers and for any future path where the Zod error reaches the UI. Consider aligning the Zod message to something like `"Site name cannot be empty or whitespace-only"`.

Confidence: high

**note:** `apps/studio/src/schemas/site.ts:L73,L88` — `.trim()` in Zod is a transform, not just a validator. A caller who submits `" My Site "` will have their value silently trimmed to `"My Site"` before it reaches the DB (`site.router.ts:136` reads the Zod-parsed `siteName` directly). This is almost certainly the desired UX, but the test suite only checks rejection; there's no assertion that the trimmed value is what gets returned after a successful parse. Adding one test case like `expect(createSiteSchema.parse({ siteName: " My Site " }).siteName).toBe("My Site")` would document this intent explicitly.

Confidence: medium (trim-on-save is idiomatic and may be considered obvious)

**note:** `apps/studio/src/schemas/site.ts:L83-L90` — `updateSiteConfigSchema` validates `siteName` twice: once via the AJV `pattern` in `SiteConfigSchema` (which operates on the raw input) and once via Zod's `.trim().min(1)` in the `.and()` clause. Both reject whitespace-only names, but with different error messages and at different stages. Worth being aware of if the error-handling path is ever surfaced to users.

Confidence: high

**question:** `apps/studio/src/schemas/__tests__/site.test.ts:L40-L53` — For `updateSiteConfigSchema`, the valid-name test passes `{ siteId, logoUrl, url, theme, siteName: "My Site" }` through the full AJV + Zod pipeline. Does `SiteConfigSchema` actually consider this a valid config (no missing required fields)? Worth a quick sanity check that the test is passing because of the schema logic, not because AJV is lenient about extra/missing optional keys in `IntegrationsSettingsSchema` or `NotificationSettingsSchema`.

Confidence: low (likely fine — all required fields appear to be covered)

apps/studio/src/server/modules/site/__tests__/site.router.test.ts
```diff
@@ -467,6 +467,29 @@ describe("site.router", async () => {
...
+        const result = caller.updateSiteConfig({
+          siteName,
+          logoUrl: MOCK_LOGO_URL,
+          url: "https://www.isomer.gov.sg",
+          theme: "isomer-next",
+          siteId: site.id,
+        })
+
+        // Assert
+        await expect(result).rejects.toThrowError()
```
adriangohjw: ask:
- to specify the specified error (code + message) - can refer to the test above
- nit: add line break before the start of this test and the previous + after the end of this and the next test

apps/studio/src/server/modules/site/__tests__/site.router.test.ts
adriangohjw, line 1: ask: there's only test for `updateSiteConfig` but given that we also updated `createSiteSchema`, we should also add unit test for the `create` route

adriangohjw (approved):

PR #2210, opened 2026-05-25
feat(navbar): add isPinnedOnMobile option to pin CTA next to logo on mobile

adriangohjw:
## Problem

Closes ISO-202

## Solution

**Breaking Changes**

**Features**:

- Adds an `isPinnedOnMobile` option to the navbar CTA (`callToAction`). When enabled, the CTA button appears next to the site logo on mobile, and the search icon moves into the hamburger menu.

**Improvements**:

- Introduces a `BoxedGroupControl` JSON Forms renderer for optional object fields (those with `format: "boxedGroup"` in the schema). It renders a labelled toggle switch; when enabled, the object's fields expand inside a bordered box. Used for the CTA and utility links sections in the navbar Customise tab.
- Reduces the CTA button text max length from 50 to 25 characters to better suit the pinned mobile layout.
- Applies `boxedGroup` format to both `callToAction` and `utility` in `NavbarAddonsSchema`, so both sections in the Customise tab use the new togglable control.
- When CTA is pinned on mobile, the search input renders inside the mobile nav menu instead of the top bar, keeping the header uncluttered.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Navigate to a site's navbar settings (`/sites/:siteId/settings/navbar`) and open the **Customise** tab.
- [ ] Confirm the **Primary Call-to-Action** section renders as a togglable boxed group (switch off by default when no CTA is configured).
- [ ] Toggle the CTA on, fill in a button label (≤ 25 chars) and URL, then enable **Pin Call-to-Action on mobile**. Publish and preview the site on a mobile viewport.
- [ ] Confirm the CTA button appears next to the site logo on mobile, and the search icon is no longer visible in the top bar (it moves into the hamburger menu).
- [ ] Confirm that on desktop the CTA still renders in the normal position in the nav bar.
- [ ] Toggle **Pin Call-to-Action on mobile** off. Confirm the CTA reverts to appearing at the top of the mobile menu (not next to the logo), and the search icon returns to the top bar.
- [ ] Confirm the **Utility links** section also renders as a togglable boxed group in the Customise tab.

**New scripts**:

None

**New dependencies**:

None

**New dev dependencies**:

None


sehyunidaaa:
@adriangohjw pushed styling changes for small viewports directly

packages/components/src/templates/next/components/internal/Navbar/MobileNavMenu/MobileNavbarMenu.tsx
```diff
@@ -53,7 +58,20 @@ export const MobileNavMenu = forwardRef<HTMLDivElement, MobileNavMenuProps>(
...
-            {!!callToAction && (
+            {isPinned && !!search && (
+              <div className="border-y border-b-base-divider-subtle bg-base-canvas-alt px-6 py-3">
+                {search.type === "localSearch" && (
+                  <LocalSearchInputBox searchUrl={search.searchUrl} />
+                )}
+                {search.type === "searchSG" && (
+                  <NavbarSearchSGInputBox
+                    clientId={search.clientId}
+                    isOpen={true}
```
dcshzj:
```suggestion
                    isOpen
```

packages/components/src/templates/next/components/internal/Navbar/NavbarClient.tsx
```diff
@@ -203,17 +221,23 @@ export const NavbarClient = ({
...
                   size="sm"
-                  className={navbarStyles.callToAction()}
+                  className={navbarStyles.callToAction({ isPinned })}
                   isWithFocusVisibleHighlight
                 >
-                  {callToAction.label}
+                  {isPinned ? (
+                    <span className="max-w-[10rem] truncate max-xs:line-clamp-2 max-xs:whitespace-normal">
+                      {callToAction.label}
+                    </span>
```
dcshzj, line 230: issue: this span seems to cause the CTA button to overflow to 2 lines, which then causes the navbar to expand. I think it should always only have a single line due to the limited height of the button?

Image

Image

seaerchin:
## Review

### Summary

- Adds an `isPinnedOnMobile` flag to the navbar CTA so it renders next to the logo on mobile, with the search icon relocating into the hamburger menu (`packages/components/src/templates/next/components/internal/Navbar/NavbarClient.tsx`, `MobileNavbarMenu.tsx`), plus a new `BoxedGroupControl` JSON Forms renderer for optional object groups toggled via `format: "boxedGroup"` (`apps/studio/.../JsonFormsBoxedGroupControl.tsx`).
- Codeflow: `NavbarAddonsSchema` gains the new boolean + `boxedGroup` format → studio renders CTA/utility as togglable boxed groups via the new renderer (registered in `FormBuilder.tsx:209`) → template reads `callToAction.isPinnedOnMobile` in `NavbarClient.tsx:88` to drive `tv` variants and to pass `search` into `MobileNavMenu`.
- Overall: clean implementation that follows existing renderer and `tv` variant patterns, with good story coverage. Two real concerns below: a likely-broken SearchSG search box in the pinned mobile menu (duplicate DOM id), and the `maxLength` 50→25 reduction which can lock existing sites out of publishing navbar changes.

### Findings

**issue:** `MobileNavbarMenu.tsx:61-73` — when pinned and the hamburger is open, a second `NavbarSearchSGInputBox` is rendered. `BaseSearchSGInputBox` renders a div with a *hardcoded* `id="searchsg-searchbar"` (`SearchSGInputBox.tsx:26`), and the top search bar section in `NavbarClient.tsx:288-301` is still always rendered (just visually hidden). The SearchSG script attaches to the element by id, and `getElementById` resolves to the first match in DOM order — the hidden top-bar instance — so the search box inside the pinned mobile menu will most likely stay empty for SearchSG sites. Additionally, `useSearchSGScript`'s cleanup removes the shared `searchsg-config` script tag whenever the menu unmounts, which can clobber a script loaded by the top-bar instance. Consider not rendering the top-bar SearchSG instance when pinned-on-mobile (e.g. gate it with the same condition used for the menu), so only one `#searchsg-searchbar` exists at a time. `localSearch` is unaffected since it's a self-contained form.

Confidence: high (duplicate id is certain; exact SearchSG script behavior inferred)

**issue:** `packages/components/src/interfaces/internal/Navbar.ts:72` — reducing `maxLength` from 50 → 25 invalidates existing data. Any site whose saved CTA label is 26–50 chars will fail `validateAddonsFn` in `NavbarEditor.tsx:35` the moment an editor opens navbar settings, and since the single Publish button is disabled on any schema error (`NavbarEditor.tsx:209`), they're blocked from publishing *any* navbar change (including unrelated menu-link edits) until they shorten the label. The PR is marked "backwards compatible", but this is effectively a breaking change for affected sites. Consider a data migration to truncate/flag existing long labels, or only enforcing the tighter limit when `isPinnedOnMobile` is set.

Confidence: high

**suggestion:** `JsonFormsBoxedGroupControl.tsx:46-54` — toggling the switch *on* for a previously-empty group calls `handleChange(path, dataSnapshot)` where `dataSnapshot` is `undefined`, so no data is created at the path. Because the object is `Type.Optional`, ajv reports no errors, Publish stays enabled, and the nested required fields (label/url) show no validation nudge — a user can flip the CTA on, see the form, and publish thinking it's configured while nothing is saved. `handleChange(path, dataSnapshot ?? {})` would surface the required-field errors and block publish until the group is filled in or toggled back off.

Confidence: medium

**suggestion (non-blocking):** `NavbarClient.tsx:64-74, 227-233` — despite the setting being named "Pin Call-to-Action **on mobile**", enabling it also changes desktop rendering: the pinned variant drops `mx-5` (spacing now comes from `gap-3` on `primaryNavigationSection`, 20px → 12px), and the `max-w-[10rem] truncate` span applies at all viewports, so a pinned CTA truncates at ~160px on desktop while an unpinned one doesn't. If the desktop layout is meant to be unchanged, scope these to `max-lg:` (e.g. `max-lg:max-w-[10rem] max-lg:truncate`).

Confidence: medium (may be intended by design)

**nitpick:** `JsonFormsBoxedGroupControl.tsx:80` — `alignItems="space-between"` is not a valid `align-items` value (you likely meant `justifyContent="space-between"`). The layout currently works only because Chakra's `FormControl` defaults to `width: 100%`, pushing the Switch right as a side effect.

Confidence: high

**nitpick (non-blocking):** `NavbarClient.tsx:311-312` / `MobileNavbarMenu.tsx:61` — the pinned state is triple-encoded: the parent passes `isPinned`, pre-gates `search` with `isPinned && layout !== "search"`, and the child re-checks `isPinned && !!search`. Since `callToAction` is already a prop of `MobileNavMenu`, the child could derive `isPinned` from `callToAction.isPinnedOnMobile` and the parent could just pass `search` through, removing one prop and the redundant double-gating.

Confidence: medium

**praise:** Good story coverage — the Chromatic stories for pinned CTA (default, expanded menu, 25-char truncation at `mobileSmall`) and the studio Customise-tab stories with MSW handlers for both empty and pre-populated CTA/utility states cover the interesting states of this change.

packages/components/src/templates/next/components/internal/Navbar/NavbarClient.tsx
sehyunidaaa: Otherwise the button will say Repor... on a 360px screen... I thought safer to wrap although it's ugly. We alr wrap the cta text for hero too

dcshzj: Hmm okay, I think it looks a little bit weird to expand the height of the navbar but ok noted on the text would be more important to show

dcshzj (approved):

PR #2211, opened 2026-05-25
fix: allow collections to be moved into folders

adriangohjw:
## Problem

Collections had no \"Move to...\" option in the dashboard row menu and were silently blocked at the DB level, making them impossible to reorganise after creation.

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- N/A

**Bug Fixes**:

- Added `ResourceType.Collection` to the `"Move to..."` menu condition in `ResourceTableMenu.tsx` so the option now appears for Collections.
- Added `ResourceType.Collection` to the `WHERE Resource.type IN (...)` guard in the `move` mutation's `UPDATE` statement in `resource.router.ts` so the DB write executes for Collections.
- Updated the `MoveResourceModal` infobox copy from "Moving a page or folder…" to "Moving a page, folder, or collection changes its URL, effective immediately".
- Added two unit tests to `resource.router.test.ts` covering Collection → Folder (success) and Collection → Collection (rejected with 400).

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] On the site dashboard, open the `⋯` menu on a Collection row — confirm "Move to..." now appears.
- [ ] Click "Move to..." — confirm `MoveResourceModal` opens with infobox text "Moving a page, folder, or collection changes its URL, effective immediately".
- [ ] In the modal, confirm only Folders are shown as destinations (not Collections or Pages).
- [ ] Select a Folder and click "Move here" — confirm toast "Resource moved!" appears and the Collection now lives under the chosen Folder.
- [ ] Navigate into the Folder and confirm the Collection is listed in its ResourceTable.
- [ ] Confirm Pages and Folders still move correctly (no regression).

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A


adriangohjw (commented):
## Summary

This PR unblocks moving Collections in the dashboard by threading  through three places: the UI row menu condition ([](https://github.com/opengovsg/isomer/pull/2211/files#diff-...)), the backend UPDATE type guard ([](https://github.com/opengovsg/isomer/pull/2211/files#diff-...)), and the modal copy. Two integration tests cover the happy path (Collection → Folder) and the negative case (Collection → Collection rejected).

The change is surgical and correct. The existing destination validation at  already blocks Collection→Collection moves at the backend, and the recursive descendant cycle check at lines 443-480 already includes Collection — so no new logic was needed, just unlocking.

---

**note (non-blocking):** `apps/studio/src/server/modules/resource/resource.router.ts:501-516` — The router re-fetches the resource after the UPDATE with `executeTakeFirst()` but never verifies the UPDATE actually affected a row (no `numAffectedRows` check). Before this fix, a Collection move would have silently returned `200` with the original `parentId` because the type guard excluded Collection — the mutation looked successful but nothing changed. The PR fixes the Collection case correctly. Worth keeping in mind: any future resource type not added to the IN list at L487-492 will exhibit the same silent no-op behaviour, returning a seemingly-valid response with the pre-move `parentId`.

Confidence: high

---

**thought:** `ResourceTableMenu.tsx:121-123` vs `resource.router.ts:487-492` — These two type lists encode related but intentionally different concepts: the UI list controls which row-menu entries render a "Move to..." item (Page, Folder, Collection), while the DB list controls which rows the UPDATE can affect (Page, CollectionPage, Folder, Collection, CollectionLink). The asymmetry is correct today, but as new resource types are added they'll need to be threaded through both places independently. A brief co-located comment explaining the intended invariant on each list would help the next engineer avoid either over- or under-including a new type.

Confidence: medium

adriangohjw (commented):
## Summary

This PR unblocks moving Collections in the dashboard by threading `ResourceType.Collection` through three places: the UI row menu condition (`ResourceTableMenu.tsx:121-123`), the backend UPDATE type guard (`resource.router.ts:487-492`), and the modal copy. Two integration tests cover the happy path (Collection → Folder) and the negative case (Collection → Collection rejected).

The change is surgical and correct. The existing destination validation at `resource.router.ts:421-429` already blocks Collection→Collection moves at the backend, and the recursive descendant cycle check at lines 443-480 already includes Collection — so no new logic was needed, just unlocking.

---

**note (non-blocking):** `apps/studio/src/server/modules/resource/resource.router.ts:501-516` — The router re-fetches the resource after the UPDATE with `executeTakeFirst()` but never verifies the UPDATE actually affected a row (no `numAffectedRows` check). Before this fix, a Collection move would have silently returned `200` with the original `parentId` because the type guard excluded Collection — the mutation looked successful but nothing changed. The PR fixes the Collection case correctly. Worth keeping in mind: any future resource type not added to the IN list at L487-492 will exhibit the same silent no-op behaviour, returning a seemingly-valid response with the pre-move `parentId`.

Confidence: high

---

**thought:** `ResourceTableMenu.tsx:121-123` vs `resource.router.ts:487-492` — These two type lists encode related but intentionally different concepts: the UI list controls which row-menu entries render a "Move to..." item (Page, Folder, Collection), while the DB list controls which rows the UPDATE can affect (Page, CollectionPage, Folder, Collection, CollectionLink). The asymmetry is correct today, but as new resource types are added they'll need to be threaded through both places independently. A brief co-located comment explaining the intended invariant on each list would help the next engineer avoid either over- or under-including a new type.

Confidence: medium

seaerchin:
@dcshzj i think we made a product decision to disallow collections inside folders - i don't quite recall the exact reason why, pinging incase you do

dcshzj:
I don't recall also but got quite a bit of sites that have collections inside folders now anyway so I think can just relax this requirement?

We can create collections inside folders now anyway.

dcshzj (approved):

PR #2212, opened 2026-05-25
refactor: migrate to t3-oss/env-* for env handling

### LINKED ISSUE #2205: Migrate to t3-oss/env-* for env handling
Can refer to how starter kit does it in

https://github.com/opengovsg/starter-kit/blob/c6f1b84a19324b71f7a1001c46ae6b3030cfdc0f/pnpm-workspace.yaml#L52-L54

and the subsequent imports

adriangohjw:
## Problem

Closes #2205

## Solution

Replaces the hand-rolled Zod + Proxy env validation in `apps/studio/src/env.mjs` and the manual `schema.parse` patterns in the shared packages with [`@t3-oss/env-nextjs`](https://env.t3.gg) and [`@t3-oss/env-core`](https://env.t3.gg), following the same approach used in the starter-kit.

**Breaking Changes**

  - Details ...

**Improvements**:

- `apps/studio/src/env.mjs` — replaced ~160 lines of custom validation/Proxy logic with `createEnv` from `@t3-oss/env-nextjs`; `SCHEDULED_PUBLISHING_HEARTBEAT_URL` and `DEACTIVATE_INACTIVE_USERS_HEARTBEAT_URL` correctly moved to `server`-only (they were never client-accessible); Storybook skip handled via `!!process.env.STORYBOOK` in `skipValidation`
- `packages/logging/src/env.ts` — migrated to `createEnv` from `@t3-oss/env-core`
- `packages/pgboss/src/env.ts` — migrated to `createEnv` from `@t3-oss/env-core`
- `tooling/site-launch/env.ts` — migrated to `createEnv` from `@t3-oss/env-core`
- `pnpm-workspace.yaml` — added a dedicated `env` named catalog for `@t3-oss/env-core` and `@t3-oss/env-nextjs`

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Run `pnpm dev` from the repo root and confirm the app starts without env validation errors
- [ ] Run `SKIP_ENV_VALIDATION=true pnpm build` from `apps/studio` and confirm it builds successfully
- [ ] Confirm that accessing a server-side env var (e.g. `SESSION_SECRET`) from client-side code throws the expected error in development

**New dependencies**:

- `@t3-oss/env-nextjs` (`catalog:env`): type-safe env validation for the Next.js studio app
- `@t3-oss/env-core` (`catalog:env`): type-safe env validation for `logging`, `pgboss`, and `site-launch` packages

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -83,6 +83,7 @@
     "@paralleldrive/cuid2": "catalog:",
     "@prisma/client": "catalog:prisma",
     "@sinclair/typebox": "catalog:",
+    "@t3-oss/env-nextjs": "catalog:env",
     "@tanstack/query-devtools": "catalog:tanstack-query",
     "@tanstack/react-query": "catalog:tanstack-query",
     "@tanstack/react-query-devtools": "catalog:tanstack-query",
--- a/apps/studio/src/env.mjs
+++ b/apps/studio/src/env.mjs
@@ -1,22 +1,39 @@
+import { createEnv } from "@t3-oss/env-nextjs"
 import { z } from "zod"
 
-const s3Schema = z.object({
-  NEXT_PUBLIC_S3_REGION: z.string().default("us-east-1"),
-  NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME: z.string(),
-  NEXT_PUBLIC_S3_ASSETS_BUCKET_NAME: z.string(),
-})
-
-const cronHeartbeatSchema = z.object({
-  SCHEDULED_PUBLISHING_HEARTBEAT_URL: z.string().url().optional(),
-  DEACTIVATE_INACTIVE_USERS_HEARTBEAT_URL: z.string().url().optional(),
-})
-
-/**
- * Specify your client-side environment variables schema here. This way you can ensure the app isn't
- * built with invalid env vars. To expose them to the client, prefix them with `NEXT_PUBLIC_`.
- */
-const client = z
-  .object({
+export const env = createEnv({
+  shared: {
+    NODE_ENV: z.enum(["development", "test", "production"]),
+  },
+  /**
+   * Specify your server-side environment variables schema here. This way you can ensure the app
+   * isn't built with invalid env vars.
+   */
+  server: {
+    DATABASE_URL: z.string().url(),
+    CI: z.coerce.boolean().default(false),
+    OTP_EXPIRY: z.coerce.number().positive().optional().default(600),
+    POSTMAN_API_KEY: z.string().optional(),
+    SESSION_SECRET: z.string().min(32),
+    GROWTHBOOK_CLIENT_KEY: z.string().optional(),
+    STUDIO_SSM_WEBHOOK_API_KEY: z.string().optional(),
+    SEARCHSG_API_KEY: z.string(),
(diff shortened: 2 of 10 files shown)


adriangohjw (commented):
## Summary

This PR replaces a hand-rolled Zod + Proxy env validation system with `@t3-oss/env-nextjs` / `@t3-oss/env-core` across the monorepo, following the starter-kit pattern. The diff is clean and the migration is mechanically correct — `createEnv` handles the server/client split, the Proxy guard, and `experimental__runtimeEnv` destructuring that the old code did manually.

One structural change worth noting: `SCHEDULED_PUBLISHING_HEARTBEAT_URL` and `DEACTIVATE_INACTIVE_USERS_HEARTBEAT_URL` were previously merged into both `client` and `server` schemas (via `cronHeartbeatSchema`). They are now `server`-only, which is correct — both are only accessed in `apps/studio/src/server/cron/jobs/` and were never reachable from client code anyway.

Overall this is a solid refactor — fewer moving parts, better type inference, alignment with the starter-kit.

---

**note:** `packages/logging/src/env.ts:L18-L20` — `NODE_ENV` is validated in the schema and added to `runtimeEnv`, but `env.NODE_ENV` is never actually read by any consumer. `tracer.ts` only reads `env.NEXT_PUBLIC_APP_ENV` and `env.NEXT_PUBLIC_APP_VERSION`; `logger.ts` receives `nodeEnv` as a call-site argument rather than reading it off `env`. The field is dead weight in this schema (and the whole file is slated for deletion anyway per the TODO). Not a bug, but removing it would keep the schema honest.

Confidence: high

---

**suggestion:** `tooling/site-launch/env.ts:L20-L22` — The original script called `envSchema.safeParse(processEnv)` unconditionally — it always validated, no escape hatch. Adding `skipValidation: !!process.env.SKIP_ENV_VALIDATION || process.env.npm_lifecycle_event === "lint"` is a new behaviour: if `SKIP_ENV_VALIDATION` happens to be set in the shell (common in CI/CD environments that run everything from a single env), the script will skip validating `GITHUB_TOKEN`, `DATABASE_URL`, `S3_BUCKET_URI` etc. and produce a cryptic downstream error instead of a clear "missing required env var" message. For an interactive CLI that deliberately requires all vars to be present, consider omitting `SKIP_ENV_VALIDATION` from the `skipValidation` condition and keeping only the lint guard, or omitting `skipValidation` entirely since the defaults are meaningless for a script with no `default()` values.

Confidence: medium

---

**note (non-blocking):** `apps/studio/src/env.mjs:L79` — The old code had a dedicated Storybook branch: it parsed `STORYBOOK_ENVIRONMENT` (a JSON-stringified snapshot of `process.env` injected by `main.ts`) and ran `client.partial().safeParse()` on it, giving Zod-defaulted values. The new code adds `!!process.env.STORYBOOK` to `skipValidation`, so `createEnv` returns raw `process.env` instead. In practice this is equivalent (both read the same underlying env vars, and `preview.tsx` merges the result with explicit mock overrides anyway), but the `STORYBOOK_ENVIRONMENT` JSON injection in `main.ts:L29` is now completely unused — it was load-bearing under the old design and is now dead config. Worth a follow-up cleanup in `main.ts` to remove `STORYBOOK_ENVIRONMENT: JSON.stringify(process.env)`.

Confidence: high

adriangohjw:
@cursor pull from main and resolve conflict

PR #2217, opened 2026-05-27
fix: add default header summary to database pages

### LINKED ISSUE #1955: [DETAIL] Studio: Database pages can be published with an empty header summary
> [!NOTE]
> From: https://app.detail.dev/oss/opengovsg/isomer/bug_8f67d1ea-fa23-4a9d-a1e8-5f0d25f6b6e3

# Summary
- **Context**: The `createDefaultPage` function in `page.service.ts` creates default page content when users create new pages in Isomer Studio.
- **Bug**: The database layout case returns a page object missing the required `contentPageHeader` field.
- **Actual vs. expected**: The `page` object should contain `contentPageHeader: { summary: "..." }` but instead includes unused `title` and `description` fields.
- **Impact**: Database pages created through Studio have an empty summary in the header. The page renders with `<p class="prose-title-lg-regular mt-5"></p>` instead of meaningful text.

# Code with Bug
```typescript
case "database": {
  const databaseDefaultPage = {
    layout: ISOMER_USABLE_PAGE_LAYOUTS.Database,
    page: {
      title: "New database layout",           // <-- Overwritten at publish time
      description:                             // <-- Only sitemap fallback, NOT used in page header
        "This is a layout where you can link your dataset from Data.gov.sg. Users can search through the table.",
      database: {
        dataSource: {
          type: "dgs",
          resourceId: "d_3c55210de27fcccda2ed0c63fdd2b352",
        },
      },
      // BUG: Missing contentPageHeader: { summary: "This is the page summary" }
    },
    content: [],
    version: "0.1.0",
  } satisfies UnwrapTagged<PrismaJson.BlobJsonContent>
  return databaseDefaultPage
}
```

# Explanation
- `Database` layout rendering expects `page.contentPageHeader` and passes it into `ContentPageHeader`; when missing, the component spreads `undefined`, resulting in an empty summary paragraph.
- This invalid shape is not caught by validation:
  - The TypeBox/AJV schema for the `page` union includes `HomePagePageSchema = Type.Object({})`, which compiles to an `anyOf` branch that matches any object. As a result, AJV reports `Valid: true` even when `contentPageHeader` is missing.
  - Studio page creation inserts the draft blob returned by `createDefaultPage` directly into the DB without validation, and publishing increments a version using the existing draft blob ID without validating content.
  - Users can publish immediately after creation without saving, so even “validate on save” would not prevent publishing malformed content.

## Codebase Inconsistency
- The schema defines `contentPageHeader` as required for database pages:
  ```typescript
  export const DatabasePagePageSchema = Type.Object({
    contentPageHeader: ContentPageHeaderSchema,  // REQUIRED
    database: SearchableTableSchema,
  })
  ```
- Other layouts’ default templates set the appropriate required header fields (e.g., `contentPageHeader` for content pages, `articlePageHeader` for article pages), but the database layout does not.

# Recommended Fix
Add `contentPageHeader` to the database default page object (and remove `title`/`description` if they are not used for the page header):
```typescript
case "database": {
  const databaseDefaultPage = {
    layout: ISOMER_USABLE_PAGE_LAYOUTS.Database,
    page: {
      contentPageHeader: { summary: "This is the page summary" },  // ADD
      database: {
        dataSource: {
          type: "dgs",
          resourceId: "d_3c55210de27fcccda2ed0c63fdd2b352",
        },
      },
    },
    content: [],
    version: "0.1.0",
  } satisfies UnwrapTagged<PrismaJson.BlobJsonContent>
  return databaseDefaultPage
}
```

# History
This bug was introduced in commit 1ebe6d6a. The commit added the "database" layout case to `createDefaultPage`, but omitted the required `contentPageHeader` field. The schema had required this field since commit e0fd6bb3 (Oct 2024), but the developer creating the new database page defaults forgot to include it, likely because TypeScript's TypeBox union with `Type.Object({})` matches any object, silently accepting the malformed data.

Benjam11n:
## Problem

Database pages created in Studio were initialized with an invalid default page shape.

The database page renderer expects the page summary to come from `page.contentPageHeader.summary`, but the default database page data used `title` and `description` fields instead. As a result, newly created database pages could render an empty header summary.

Closes #1955

## Solution

Updated the default database page payload to include `page.contentPageHeader.summary` and added an unit test. 

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**BEFORE**:

Screenshot 2026-05-27 at 5 06 04 PM

**AFTER**:

Screenshot 2026-05-27 at 5 05 21 PM

## Tests

- `pnpm --filter isomer-studio test:unit -- src/server/modules/page/__tests__/page.router.test.ts`

- [ ] Create a new page with the database page layout
- [ ] Add a title and start editing
- [ ] Verify that the "This is the page summary" text appears below the title

### DIFF
--- a/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
+++ b/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
@@ -1753,6 +1753,12 @@ describe("page.router", async () => {
         ...expectedPageArgs,
         content: createDefaultPage({ layout: "database" }),
       })
+      expect(actual.content).toMatchObject({
+        layout: "database",
+        page: {
+          contentPageHeader: { summary: "This is the page summary" },
+        },
+      })
       await assertAuditLogRows(1)
       const auditLog = await db.selectFrom("AuditLog").selectAll().execute()
       expect(auditLog).toHaveLength(1)
--- a/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
+++ b/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
@@ -1753,12 +1753,6 @@ describe("page.router", async () => {
         ...expectedPageArgs,
         content: createDefaultPage({ layout: "database" }),
       })
-      expect(actual.content).toMatchObject({
-        layout: "database",
-        page: {
-          contentPageHeader: { summary: "This is the page summary" },
-        },
-      })
       await assertAuditLogRows(1)
       const auditLog = await db.selectFrom("AuditLog").selectAll().execute()
       expect(auditLog).toHaveLength(1)
--- a/apps/studio/src/server/modules/page/page.service.ts
+++ b/apps/studio/src/server/modules/page/page.service.ts
@@ -41,9 +41,7 @@ export const createDefaultPage = ({
       const databaseDefaultPage = {
         layout: ISOMER_USABLE_PAGE_LAYOUTS.Database,
         page: {
-          title: "New database layout",
-          description:
-            "This is a layout where you can link your dataset from Data.gov.sg. Users can search through the table.",
+          contentPageHeader: { summary: "This is the page summary" },
           database: {
             dataSource: {
               type: "dgs", // we only support DGS creation on studio for now


apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -1753,6 +1753,12 @@ describe("page.router", async () => {
         ...expectedPageArgs,
         content: createDefaultPage({ layout: "database" }),
       })
+      expect(actual.content).toMatchObject({
+        layout: "database",
+        page: {
+          contentPageHeader: { summary: "This is the page summary" },
+        },
+      })
```
adriangohjw: These assertion looks overlapping?

adriangohjw (commented):
Thanks for working on this! No blockers for me, just wanna clarify something regarding the test

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
Benjam11n: Removed the overlapping assertion. Thanks

adriangohjw (approved):

PR #2221, opened 2026-05-28
Fix webhook API key auth bypass

adriangohjw:
## Problem

Fixes a webhook authorization bypass where authenticated users could invoke webhook procedures without providing the shared webhook API key.

## Solution

**Breaking Changes**

  - Webhook procedure callers must now provide the `x-api-key` header even if they have an authenticated Studio session.

**Features**:

- N/A

**Improvements**:

- N/A

**Bug Fixes**:

- Always validate `x-api-key` in `webhookMiddleware` before allowing webhook procedures to proceed.
- Updated webhook tests to include the API key on successful webhook calls.
- Added a regression test that rejects an authenticated caller without the webhook API key.
- Corrected the missing API key test helper so it actually omits the header.

## Before & After Screenshots

**BEFORE**:

N/A

**AFTER**:

N/A

## Tests

- [x] `pnpm exec oxlint --type-aware src/server/trpc.ts src/server/modules/webhook/webhook.router.ts src/server/modules/webhook/__test__/webhook.router.test.ts src/schemas/__tests__/webhook.test.ts`
- [x] `pnpm exec oxfmt --check src/server/trpc.ts src/server/modules/webhook/webhook.router.ts src/server/modules/webhook/__test__/webhook.router.test.ts src/schemas/__tests__/webhook.test.ts`
- [x] `pnpm exec oxlint --type-aware src/server/modules/webhook/webhook.router.ts`
- [x] `pnpm exec oxfmt --check src/server/modules/webhook/webhook.router.ts`
- [ ] `pnpm test:unit -- src/server/modules/webhook/__test__/webhook.router.test.ts src/schemas/__tests__/webhook.test.ts` (blocked in this cloud environment: Docker is not installed, and Vitest global setup requires testcontainers.)

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

<div><a href="https://cursor.com/agents/bc-ac7fb604-6c12-4a46-87f3-3fc8a753562e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ac7fb604-6c12-4a46-87f3-3fc8a753562e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/studio/src/server/trpc.ts
+++ b/apps/studio/src/server/trpc.ts
@@ -173,29 +173,26 @@ const authMiddleware = t.middleware(async ({ next, ctx }) => {
 
 /**
  * Webhook middleware to protect endpoints that do not need a user context
- * but still need to be protected via an API key. We still check the session
- * in case the request is made by a logged in user (via the FE), and to allow for easier testing.
+ * but still need to be protected via an API key.
  * */
 
 export const WEBHOOK_X_API_KEY_HEADER = "x-api-key"
 
 const webhookMiddleware = t.middleware(async ({ next, ctx }) => {
-  if (!ctx.session?.userId) {
-    const apiKey = ctx.req.headers[WEBHOOK_X_API_KEY_HEADER]
-    // Ensure that the API key is set in the env
-    if (!env.STUDIO_SSM_WEBHOOK_API_KEY) {
-      throw new TRPCError({
-        code: "INTERNAL_SERVER_ERROR",
-        message: "Webhook API key is not configured",
-      })
-    }
-    // Ensure that the API key is valid and matches
-    if (!apiKey || apiKey !== env.STUDIO_SSM_WEBHOOK_API_KEY) {
-      throw new TRPCError({
-        code: "UNAUTHORIZED",
-        message: "Invalid Webhook API key provided",
-      })
-    }
+  const apiKey = ctx.req.headers[WEBHOOK_X_API_KEY_HEADER]
+  // Ensure that the API key is set in the env
+  if (!env.STUDIO_SSM_WEBHOOK_API_KEY) {
+    throw new TRPCError({
+      code: "INTERNAL_SERVER_ERROR",
+      message: "Webhook API key is not configured",
+    })
+  }
+  // Ensure that the API key is valid and matches
+  if (!apiKey || apiKey !== env.STUDIO_SSM_WEBHOOK_API_KEY) {
+    throw new TRPCError({
+      code: "UNAUTHORIZED",
+      message: "Invalid Webhook API key provided",
+    })
   }
   return next()
 })
--- a/apps/studio/src/schemas/__tests__/webhook.test.ts
+++ b/apps/studio/src/schemas/__tests__/webhook.test.ts
@@ -1,5 +1,6 @@
 import type { NextApiRequest, NextApiResponse } from "next"
 import type { z } from "zod"
+import type { env } from "~/env.mjs"
 import { createMocks } from "node-mocks-http"
 import { resetTables } from "tests/integration/helpers/db"
 import { createTestUser } from "tests/integration/helpers/iron-session"
@@ -11,10 +12,10 @@ import type { codeBuildWebhookSchema } from "../webhook"
 
 vi.mock("~/env.mjs", async () => {
   // Import the real module first to get all default values
-  const actual = await vi.importActual("~/env.mjs")
+  const actual = await vi.importActual<{ env: typeof env }>("~/env.mjs")
   return {
     env: {
-      ...actual,
+      ...actual.env,
       // override some env variables for testing
       STUDIO_SSM_WEBHOOK_API_KEY: "test-webhook-api-key",
       GROWTHBOOK_CLIENT_KEY: "test-growthbook-client-key",
@@ -33,7 +34,7 @@ const createMockRequest = ({
   apiKey = "test-webhook-api-key",
 }: {
   arn: string
-  apiKey?: string
+  apiKey?: string | null
 }) => {
   const body: z.input<typeof codeBuildWebhookSchema> = {
     projectName: "test-project",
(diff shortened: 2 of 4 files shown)


apps/studio/src/server/trpc.ts
```diff
@@ -173,29 +173,26 @@ const authMiddleware = t.middleware(async ({ next, ctx }) => {
...
  * Webhook middleware to protect endpoints that do not need a user context
- * but still need to be protected via an API key. We still check the session
- * in case the request is made by a logged in user (via the FE), and to allow for easier testing.
+ * but still need to be protected via an API key.
  * */
 
 export const WEBHOOK_X_API_KEY_HEADER = "x-api-key"
 
 const webhookMiddleware = t.middleware(async ({ next, ctx }) => {
-  if (!ctx.session?.userId) {
```
KenLSM, line 183: Just to clarify, if the user session is present, we ignore it.

Asking because it seemed to be a change in feature requirement (user session is ignored).

adriangohjw: yeah, it should only be triggered via API key

KenLSM (approved):

seaerchin (approved):
lgtm

seaerchin (commented):
## Summary

- Fixes a real auth bypass: `webhookMiddleware` at `apps/studio/src/server/trpc.ts:L181` previously only checked the `x-api-key` header when there was no session, so any authenticated Studio user could invoke `webhookProcedure` routes (currently `webhook.updateCodebuildWebhook` → `apps/studio/src/server/modules/webhook/webhook.router.ts:L7`) via the normal tRPC client without the shared key. The fix removes the `!ctx.session?.userId` guard so the API key is required unconditionally.
- The webhook procedure is also reached from the REST handler at `apps/studio/src/pages/api/webhooks/updateCodebuildWebhook.ts`, which builds context with `session: undefined` (`apps/studio/src/server/webhooks.ts:L24`), so AWS CodeBuild's existing call path is unaffected as long as it sends `x-api-key`.
- Test changes: tightens the env mock (now spreads `actual.env` rather than the whole module export), fixes a stale assertion in `apps/studio/src/schemas/__tests__/webhook.test.ts:L107` where the "missing API key" test was actually sending `"wrong-api-key"`, and adds a regression test in `apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts:L131` for an authenticated caller missing the header.
- Overall: a small, well-targeted security fix. One real follow-up worth considering (timing-safe comparison), the rest are minor.

## Findings

**praise:** `apps/studio/src/server/trpc.ts:L173-L198` — good catch and minimal blast radius. The old comment ("We still check the session ... to allow for easier testing") rationalised the bypass as a convenience, which is exactly the kind of code that needs to be ripped out. Tests still work after the change, so the convenience wasn't load-bearing.

Confidence: high

---

**issue (security, non-blocking):** `apps/studio/src/server/trpc.ts:L191` — `apiKey !== env.STUDIO_SSM_WEBHOOK_API_KEY` is a plain string comparison. It short-circuits on the first differing byte, so it leaks timing information about how many leading bytes matched. The attack is impractical over the public internet (jitter dominates the signal) but trivial from a co-located caller. Since this file is already being touched for an auth-bypass fix, it's the right moment to switch to a constant-time compare:

```ts
import { timingSafeEqual } from "node:crypto"

const provided = Array.isArray(apiKey) ? apiKey[0] : apiKey
const expected = env.STUDIO_SSM_WEBHOOK_API_KEY
if (
  !provided ||
  provided.length !== expected.length ||
  !timingSafeEqual(Buffer.from(provided), Buffer.from(expected))
) {
  throw new TRPCError({ code: "UNAUTHORIZED", message: "Invalid Webhook API key provided" })
}
```

Pre-existing issue, but worth fixing alongside the bypass rather than in a follow-up.

Confidence: medium

---

**nitpick (maintainability):** `apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts:L51-L59` — `getCallerWithMockGrowthbook` now hardcodes `method: "GET"` on the mock request, but the underlying handler is POST-only (`apps/studio/src/pages/api/webhooks/updateCodebuildWebhook.ts:L10`). The tRPC caller doesn't care about the HTTP verb, so this doesn't cause a bug — it just confuses a future reader who searches for "POST" while debugging. Drop the `method` override and let the helper default apply, or set `"POST"` for symmetry with reality.

Confidence: low

---

**suggestion (if-minor):** `apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts:L46-L59` — the `apiKey: string | null` sentinel where `null` means "omit the header" is a bit subtle. If you end up touching this again, an explicit `omitApiKey?: boolean` or `{ apiKey?: string }` (where `undefined` means omit) reads more clearly at the call site than the magic `null`.

Confidence: low

---

**praise:** `apps/studio/src/schemas/__tests__/webhook.test.ts:L13-L24` and `apps/studio/src/server/modules/webhook/__test__/webhook.router.test.ts:L30-L38` — the corrected `vi.importActual<{ env: typeof env }>(...)` + `...actual.env` form is right. The previous `...actual` shape was silently producing `{ env: { env: {real values}, ...overrides } }`, which meant any test that read a non-overridden env var was getting `undefined`. Cleaning this up while you're in here is a good incidental fix.

Confidence: high

---

**question:** the PR is marked as a breaking change because authenticated callers must now provide `x-api-key`. I couldn't find any frontend / internal tRPC caller of `trpc.webhook.updateCodebuildWebhook` — `grep` for `trpc.webhook` and `api.webhook` in `apps/studio/src` returned nothing. Is the "breaking change" purely theoretical (i.e., no caller actually relied on the bypass), or is there a non-obvious caller (e.g., an admin tool, a script, an e2e harness) that needs updating in the same PR? Worth confirming before merge — if nothing relied on the bypass, the breaking-change tag could be dropped, which simplifies release messaging.

Confidence: medium

PR #2222, opened 2026-05-28
Fix client IP extraction precedence

adriangohjw:
## Problem

`getIP()` used the leftmost `x-forwarded-for` value, allowing clients to spoof IP addresses recorded in audit and structured logs when a reverse proxy appends to an existing XFF header.

## Solution

**Breaking Changes**

**Features**:

- None

**Improvements**:

- Prefer `cf-connecting-ip`, then `socket.remoteAddress`, before falling back to `x-forwarded-for`.
- Parse repeated and comma-separated IP headers consistently.
- Add inline comments explaining the header trust order and why XFF is treated as a last-resort fallback.

**Bug Fixes**:

- Use the rightmost `x-forwarded-for` fallback value instead of the attacker-controlled leftmost value.
- Avoid returning malformed array XFF values with trailing proxy IPs.

## Before & After Screenshots

**BEFORE**:

Not applicable.

**AFTER**:

Not applicable.

## Tests

- [x] `pnpm --filter isomer-studio exec oxlint src/utils/getClientIp.ts`
- [x] `pnpm --filter isomer-studio exec oxlint src/utils/getClientIp.ts src/utils/__tests__/getClientIp.test.ts`
- [x] `pnpm --filter isomer-studio exec dotenv -e .env.test -- vitest run --config ./vitest.getClientIp.config.ts src/utils/__tests__/getClientIp.test.ts` (temporary config used locally to bypass unavailable testcontainer runtime)
- [ ] `pnpm --filter isomer-studio test:unit -- src/utils/__tests__/getClientIp.test.ts` (blocked in this cloud runner: no working container runtime for the repo global Vitest setup)
- [ ] `pnpm --filter isomer-studio typecheck` (blocked by existing `@prisma/client` export errors unrelated to this change)

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

<div><a href="https://cursor.com/agents/bc-90707a1e-d9af-483f-aef4-b89f7206f4d0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-90707a1e-d9af-483f-aef4-b89f7206f4d0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/studio/src/utils/getClientIp.ts
+++ b/apps/studio/src/utils/getClientIp.ts
@@ -13,6 +13,8 @@ const getFirstIp = (header: string | string[] | null | undefined) =>
 
 const getLastIp = (header: string | string[] | null | undefined) => {
   const ips = getIpList(header)
+  // Proxies append to X-Forwarded-For, so the leftmost value can be supplied
+  // by the client. Use the rightmost value only as a last-resort fallback.
   return ips[ips.length - 1]
 }
 
@@ -30,6 +32,8 @@ export default function getIP(request: Request | NextApiRequest) {
     ? request.headers.get("x-forwarded-for")
     : request.headers["x-forwarded-for"]
 
+  // Prefer values set by our edge/proxy infrastructure before consulting
+  // X-Forwarded-For, which may contain client-supplied entries.
   return (
     getFirstIp(cfConnectingIp) ??
     remoteAddress ??
--- a/apps/studio/src/utils/getClientIp.ts
+++ b/apps/studio/src/utils/getClientIp.ts
@@ -1,14 +1,6 @@
 import { type NextApiRequest } from "next"
-import { env } from "~/env.mjs"
 
 const LOCALHOST_IP = "127.0.0.1"
-const DEFAULT_TRUSTED_PROXY_HOPS = 1
-
-const getTrustedProxyHopCount = (configuredHopCount?: number) => {
-  return Number.isInteger(configuredHopCount) && configuredHopCount > 0
-    ? configuredHopCount
-    : DEFAULT_TRUSTED_PROXY_HOPS
-}
 
 const getIpList = (header: string | string[] | null | undefined) =>
   (Array.isArray(header) ? header : [header])
@@ -19,28 +11,15 @@ const getIpList = (header: string | string[] | null | undefined) =>
 const getFirstIp = (header: string | string[] | null | undefined) =>
   getIpList(header)[0]
 
-const getTrustedXForwardedForIp = (
-  header: string | string[] | null | undefined,
-  trustedProxyHops?: number,
-) => {
+const getLastIp = (header: string | string[] | null | undefined) => {
   const ips = getIpList(header)
-
-  if (ips.length === 0) {
-    return undefined
-  }
-
-  return ips[
-    Math.max(ips.length - getTrustedProxyHopCount(trustedProxyHops), 0)
-  ]
+  return ips[ips.length - 1]
 }
 
 const isRequest = (request: Request | NextApiRequest): request is Request =>
   typeof Request !== "undefined" && request instanceof Request
 
-export default function getIP(
-  request: Request | NextApiRequest,
-  options: { trustedProxyHops?: number } = {},
-) {
+export default function getIP(request: Request | NextApiRequest) {
   const cfConnectingIp = isRequest(request)
     ? request.headers.get("cf-connecting-ip")
     : request.headers["cf-connecting-ip"]
@@ -54,10 +33,7 @@ export default function getIP(
   return (
     getFirstIp(cfConnectingIp) ??
     remoteAddress ??
-    getTrustedXForwardedForIp(
-      xForwardedFor,
-      options.trustedProxyHops ?? env.CLIENT_IP_TRUSTED_PROXY_HOPS,
-    ) ??
+    getLastIp(xForwardedFor) ??
     LOCALHOST_IP
   )
 }
--- a/apps/studio/src/utils/getClientIp.ts
+++ b/apps/studio/src/utils/getClientIp.ts
@@ -1,14 +1,10 @@
 import { type NextApiRequest } from "next"
+import { env } from "~/env.mjs"
 
 const LOCALHOST_IP = "127.0.0.1"
 const DEFAULT_TRUSTED_PROXY_HOPS = 1
 
-const getTrustedProxyHopCount = () => {
-  const configuredHopCount = Number.parseInt(
-    process.env.CLIENT_IP_TRUSTED_PROXY_HOPS ?? "",
-    10,
-  )
-
+const getTrustedProxyHopCount = (configuredHopCount?: number) => {
   return Number.isInteger(configuredHopCount) && configuredHopCount > 0
     ? configuredHopCount
     : DEFAULT_TRUSTED_PROXY_HOPS
@@ -25,6 +21,7 @@ const getFirstIp = (header: string | string[] | null | undefined) =>
 
 const getTrustedXForwardedForIp = (
   header: string | string[] | null | undefined,
+  trustedProxyHops?: number,
 ) => {
   const ips = getIpList(header)
 
@@ -32,13 +29,18 @@ const getTrustedXForwardedForIp = (
     return undefined
   }
 
-  return ips[Math.max(ips.length - getTrustedProxyHopCount(), 0)]
+  return ips[
+    Math.max(ips.length - getTrustedProxyHopCount(trustedProxyHops), 0)
+  ]
 }
 
 const isRequest = (request: Request | NextApiRequest): request is Request =>
   typeof Request !== "undefined" && request instanceof Request
 
-export default function getIP(request: Request | NextApiRequest) {
+export default function getIP(
+  request: Request | NextApiRequest,
+  options: { trustedProxyHops?: number } = {},
+) {
   const cfConnectingIp = isRequest(request)
     ? request.headers.get("cf-connecting-ip")
     : request.headers["cf-connecting-ip"]
@@ -52,7 +54,10 @@ export default function getIP(request: Request | NextApiRequest) {
   return (
     getFirstIp(cfConnectingIp) ??
     remoteAddress ??
-    getTrustedXForwardedForIp(xForwardedFor) ??
+    getTrustedXForwardedForIp(
+      xForwardedFor,
+      options.trustedProxyHops ?? env.CLIENT_IP_TRUSTED_PROXY_HOPS,
+    ) ??
     LOCALHOST_IP
   )
 }
--- a/apps/studio/src/utils/getClientIp.ts
+++ b/apps/studio/src/utils/getClientIp.ts
@@ -1,10 +1,58 @@
 import { type NextApiRequest } from "next"
 
-export default function getIP(request: Request | NextApiRequest) {
-  const xff =
-    request instanceof Request
-      ? request.headers.get("x-forwarded-for")
-      : request.headers["x-forwarded-for"]
+const LOCALHOST_IP = "127.0.0.1"
+const DEFAULT_TRUSTED_PROXY_HOPS = 1
 
-  return xff ? (Array.isArray(xff) ? xff[0] : xff.split(",")[0]) : "127.0.0.1"
+const getTrustedProxyHopCount = () => {
+  const configuredHopCount = Number.parseInt(
+    process.env.CLIENT_IP_TRUSTED_PROXY_HOPS ?? "",
+    10,
+  )
+
+  return Number.isInteger(configuredHopCount) && configuredHopCount > 0
+    ? configuredHopCount
+    : DEFAULT_TRUSTED_PROXY_HOPS
+}
+
+const getIpList = (header: string | string[] | null | undefined) =>
+  (Array.isArray(header) ? header : [header])
+    .flatMap((value) => value?.split(",") ?? [])
+    .map((value) => value.trim())
+    .filter((value) => value.length > 0)
+
+const getFirstIp = (header: string | string[] | null | undefined) =>
+  getIpList(header)[0]
+
+const getTrustedXForwardedForIp = (
+  header: string | string[] | null | undefined,
+) => {
+  const ips = getIpList(header)
+
+  if (ips.length === 0) {
+    return undefined
+  }
+
+  return ips[Math.max(ips.length - getTrustedProxyHopCount(), 0)]
+}
+
+const isRequest = (request: Request | NextApiRequest): request is Request =>
+  typeof Request !== "undefined" && request instanceof Request
+
+export default function getIP(request: Request | NextApiRequest) {
+  const cfConnectingIp = isRequest(request)
+    ? request.headers.get("cf-connecting-ip")
+    : request.headers["cf-connecting-ip"]
+  const remoteAddress = isRequest(request)
+    ? undefined
+    : request.socket.remoteAddress?.trim() || undefined
+  const xForwardedFor = isRequest(request)
+    ? request.headers.get("x-forwarded-for")
+    : request.headers["x-forwarded-for"]
+
+  return (
+    getFirstIp(cfConnectingIp) ??
+    remoteAddress ??
+    getTrustedXForwardedForIp(xForwardedFor) ??
+    LOCALHOST_IP
+  )
 }
--- a/apps/studio/src/utils/getClientIp.ts
+++ b/apps/studio/src/utils/getClientIp.ts
@@ -13,13 +13,11 @@ const getFirstIp = (header: string | string[] | null | undefined) =>
 
 const getLastIp = (header: string | string[] | null | undefined) => {
   const ips = getIpList(header)
-  // Proxies append to X-Forwarded-For, so the leftmost value can be supplied
-  // by the client. Use the rightmost value only as a last-resort fallback.
   return ips[ips.length - 1]
 }
 
 const isRequest = (request: Request | NextApiRequest): request is Request =>
-  typeof Request !== "undefined" && request instanceof Request
+  request instanceof Request
 
 export default function getIP(request: Request | NextApiRequest) {
   const cfConnectingIp = isRequest(request)
@@ -33,11 +31,13 @@ export default function getIP(request: Request | NextApiRequest) {
     : request.headers["x-forwarded-for"]
 
   // Prefer values set by our edge/proxy infrastructure before consulting
-  // X-Forwarded-For, which may contain client-supplied entries.
+  // X-Forwarded-For, which may contain client-supplied entries. Rightmost XFF
+  // (appended by our L7 proxy/LB) is preferred over socket.remoteAddress,
+  // which in a containerised deployment is the load-balancer IP, not the client.
   return (
     getFirstIp(cfConnectingIp) ??
-    remoteAddress ??
     getLastIp(xForwardedFor) ??
+    remoteAddress ??
     LOCALHOST_IP
   )
 }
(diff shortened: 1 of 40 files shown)


KenLSM (approved):

seaerchin (commented):
## Summary

- Replaces the old `xff[0]`/leftmost-`X-Forwarded-For` fallback in `getIP` with a layered precedence: `cf-connecting-ip` → `socket.remoteAddress` → rightmost-XFF → `127.0.0.1`. Implementation in `apps/studio/src/utils/getClientIp.ts:24-43`, helpers at `:5-19`, narrowing at `:21-22`. Test coverage added in `apps/studio/src/utils/__tests__/getClientIp.test.ts`.
- Two consumers benefit: `lib/logger.ts:21` (structured logs) and `server/modules/auth/auth.router.ts:34` (audit-log `ip` field on logout). The original spoofing risk callout — that an attacker could pin a chosen IP into audit logs via XFF — is addressed for *those* sinks.
- Overall: the fix to `getIP` is directionally right and the test additions are welcome. However, the more security-critical callsites that motivated this kind of fix (OTP rate-limit fingerprint, generic rate-limit fingerprint) still use the original `cf-connecting-ip ?? socket.remoteAddress ?? xff[0]` pattern verbatim. The PR also moves `socket.remoteAddress` ahead of XFF, which in your deployment topology (Docker behind an LB, per `apps/studio/Dockerfile`) likely degrades non-Cloudflare traffic from "spoofable real client IP" to "load-balancer IP". Worth a second pass before merge.

## Findings

**issue (security):** `apps/studio/src/server/modules/rate-limit/utils.ts:18-21` — `getRateLimitFingerprint` still derives the IP using `(forwarded[0]).split(',')[0]`, i.e. the leftmost, client-supplied XFF value. This is the rate-limiter's *primary key*. An attacker can rotate `X-Forwarded-For: <random>` on every request and trivially bypass per-IP rate limits — a strictly higher-impact bypass than the audit-log spoof this PR closes. The same `getIP` helper (or its inner `getFirstIp`/`getLastIp` building blocks) should be reused here instead of duplicating the old logic. At minimum, this callsite has to learn that the leftmost XFF entry is attacker-controlled.

Confidence: high

---

**issue (security):** `apps/studio/src/server/modules/auth/email/utils.ts:9-21` — `getOtpFingerPrint` uses the same spoofable pattern: `cf-connecting-ip ?? socket.remoteAddress ?? xff` followed by `originIp.join(", ")` when it sees an array, then uses that as the OTP rate-limit fingerprint at `auth.service.ts:21,43` and `email.router.ts:66,74,112`. Because the fingerprint feeds OTP request/verify rate limiting, spoofing XFF yields unlimited OTP attempts per email — a credential-stuffing / OTP-brute-force vector. This needs the same fix as `getIP`, and arguably the three call sites should converge on a single helper.

Confidence: high

---

**issue:** `apps/studio/src/utils/getClientIp.ts:28-30, 37-42` — placing `socket.remoteAddress` **above** `getLastIp(xForwardedFor)` is wrong for a containerised deployment behind an L7 proxy (which this app is, per `Dockerfile`). In that topology `socket.remoteAddress` is the load-balancer's IP, while the rightmost XFF value is what the LB just appended — i.e. the actual client (or the closest upstream proxy). After this change, requests that lack `cf-connecting-ip` (anything not flowing through Cloudflare, including health checks, internal traffic, ALB-direct hits) will record the LB IP in logs and on the audit `ip` column. The reverted "Configure trusted proxy IP hops" commit was closer to correct — for a single trusted proxy you want rightmost XFF, and `socket.remoteAddress` is only meaningful when there is no proxy at all (dev/local). Suggest reordering to `cf-connecting-ip → getLastIp(xForwardedFor) → socket.remoteAddress → 127.0.0.1`, or making the XFF hop count an env-driven knob the way the earlier commit did.

Confidence: medium (depends on prod topology — high confidence that the order is wrong if Cloudflare/ALB sit in front; medium that this is your actual deployment shape)

---

**issue:** PR description states "this PR is backwards compatible" but IP semantics change for any user not behind Cloudflare: previously the recorded IP was the (spoofable) leftmost XFF, now it is `socket.remoteAddress` if present (the proxy) or the rightmost XFF. Downstream consumers of audit-log `ip` and the `clientIp` log field will see different values for the same client post-merge. Worth calling this out explicitly so anyone doing forensics on historic vs new entries isn't confused.

Confidence: high

---

**suggestion:** `apps/studio/src/utils/getClientIp.ts:21-22` — `typeof Request !== "undefined" && request instanceof Request` is defensive against an environment where `Request` is unavailable, which isn't the case in any Node runtime Next.js targets. The `instanceof Request` half is sufficient; the `typeof` guard suggests uncertainty that isn't warranted and complicates the predicate. Non-blocking.

Confidence: high

---

**suggestion:** `apps/studio/src/utils/__tests__/getClientIp.test.ts` — coverage gaps that are easy wins given how this util gates security decisions:
- No assertion that an attacker-supplied leftmost XFF is **not** returned when `cf-connecting-ip` or `socket.remoteAddress` is present (the regression you're guarding against).
- No `Request`-instance test for `cf-connecting-ip` precedence — only `NextApiRequest`.
- No test for `cf-connecting-ip: ""` falling through (the `filter((v) => v.length > 0)` is doing real work and should be pinned).
- No test for `socket.remoteAddress: "   "` falling through (the `.trim() || undefined` is similarly load-bearing).

Confidence: medium

---

**question:** `apps/studio/src/utils/getClientIp.ts:25-27` — is Cloudflare actually in front of this app, and is the origin locked down so a client can't reach it directly bypassing CF? If the origin is reachable without going through Cloudflare, `cf-connecting-ip` is no more trustworthy than XFF (a client can set it themselves) and this entire precedence collapses. If CF *is* the edge and origin-direct is blocked at the LB/SG layer, this is fine — but it's worth confirming since the whole security argument of the PR hinges on it.

Confidence: medium

---

**nitpick:** `apps/studio/src/utils/getClientIp.ts:5-19` — three of the four helpers (`getIpList`, `getFirstIp`, `getLastIp`) are reasonable, but `getLastIp` mixes "parse" and "policy" (the comment about leftmost-being-spoofable lives inside it). If you do consolidate the rate-limit and OTP callsites onto a shared helper, consider moving the "rightmost XFF only as last resort" comment up to `getIP` where the precedence decision actually lives — `getLastIp` is just `arr[arr.length - 1]`. Minor.

Confidence: low

seaerchin (commented):
## Summary

- Replaces the old `xff[0]`/leftmost-`X-Forwarded-For` fallback in `getIP` with a layered precedence: `cf-connecting-ip` → `socket.remoteAddress` → rightmost-XFF → `127.0.0.1`. Implementation in `apps/studio/src/utils/getClientIp.ts:24-43`, helpers at `:5-19`, narrowing at `:21-22`. Test coverage added in `apps/studio/src/utils/__tests__/getClientIp.test.ts`.
- Two consumers benefit: `lib/logger.ts:21` (structured logs) and `server/modules/auth/auth.router.ts:34` (audit-log `ip` field on logout). The original spoofing risk callout — that an attacker could pin a chosen IP into audit logs via XFF — is addressed for *those* sinks.
- Overall: the fix to `getIP` is directionally right and the test additions are welcome. However, the more security-critical callsites that motivated this kind of fix (OTP rate-limit fingerprint, generic rate-limit fingerprint) still use the original `cf-connecting-ip ?? socket.remoteAddress ?? xff[0]` pattern verbatim. The PR also moves `socket.remoteAddress` ahead of XFF, which in your deployment topology (Docker behind an LB, per `apps/studio/Dockerfile`) likely degrades non-Cloudflare traffic from "spoofable real client IP" to "load-balancer IP". Worth a second pass before merge.

## Findings

**issue (security):** `apps/studio/src/server/modules/rate-limit/utils.ts:18-21` — `getRateLimitFingerprint` still derives the IP using `(forwarded[0]).split(',')[0]`, i.e. the leftmost, client-supplied XFF value. This is the rate-limiter's *primary key*. An attacker can rotate `X-Forwarded-For: <random>` on every request and trivially bypass per-IP rate limits — a strictly higher-impact bypass than the audit-log spoof this PR closes. The same `getIP` helper (or its inner `getFirstIp`/`getLastIp` building blocks) should be reused here instead of duplicating the old logic. At minimum, this callsite has to learn that the leftmost XFF entry is attacker-controlled.

Confidence: high

---

**issue (security):** `apps/studio/src/server/modules/auth/email/utils.ts:9-21` — `getOtpFingerPrint` uses the same spoofable pattern: `cf-connecting-ip ?? socket.remoteAddress ?? xff` followed by `originIp.join(", ")` when it sees an array, then uses that as the OTP rate-limit fingerprint at `auth.service.ts:21,43` and `email.router.ts:66,74,112`. Because the fingerprint feeds OTP request/verify rate limiting, spoofing XFF yields unlimited OTP attempts per email — a credential-stuffing / OTP-brute-force vector. This needs the same fix as `getIP`, and arguably the three call sites should converge on a single helper.

Confidence: high

---

**issue:** `apps/studio/src/utils/getClientIp.ts:28-30, 37-42` — placing `socket.remoteAddress` **above** `getLastIp(xForwardedFor)` is wrong for a containerised deployment behind an L7 proxy (which this app is, per `Dockerfile`). In that topology `socket.remoteAddress` is the load-balancer's IP, while the rightmost XFF value is what the LB just appended — i.e. the actual client (or the closest upstream proxy). After this change, requests that lack `cf-connecting-ip` (anything not flowing through Cloudflare, including health checks, internal traffic, ALB-direct hits) will record the LB IP in logs and on the audit `ip` column. The reverted "Configure trusted proxy IP hops" commit was closer to correct — for a single trusted proxy you want rightmost XFF, and `socket.remoteAddress` is only meaningful when there is no proxy at all (dev/local). Suggest reordering to `cf-connecting-ip → getLastIp(xForwardedFor) → socket.remoteAddress → 127.0.0.1`, or making the XFF hop count an env-driven knob the way the earlier commit did.

Confidence: medium (depends on prod topology — high confidence that the order is wrong if Cloudflare/ALB sit in front; medium that this is your actual deployment shape)

---

**issue:** PR description states "this PR is backwards compatible" but IP semantics change for any user not behind Cloudflare: previously the recorded IP was the (spoofable) leftmost XFF, now it is `socket.remoteAddress` if present (the proxy) or the rightmost XFF. Downstream consumers of audit-log `ip` and the `clientIp` log field will see different values for the same client post-merge. Worth calling this out explicitly so anyone doing forensics on historic vs new entries isn't confused.

Confidence: high

---

**suggestion:** `apps/studio/src/utils/getClientIp.ts:21-22` — `typeof Request !== "undefined" && request instanceof Request` is defensive against an environment where `Request` is unavailable, which isn't the case in any Node runtime Next.js targets. The `instanceof Request` half is sufficient; the `typeof` guard suggests uncertainty that isn't warranted and complicates the predicate. Non-blocking.

Confidence: high

---

**suggestion:** `apps/studio/src/utils/__tests__/getClientIp.test.ts` — coverage gaps that are easy wins given how this util gates security decisions:
- No assertion that an attacker-supplied leftmost XFF is **not** returned when `cf-connecting-ip` or `socket.remoteAddress` is present (the regression you're guarding against).
- No `Request`-instance test for `cf-connecting-ip` precedence — only `NextApiRequest`.
- No test for `cf-connecting-ip: ""` falling through (the `filter((v) => v.length > 0)` is doing real work and should be pinned).
- No test for `socket.remoteAddress: "   "` falling through (the `.trim() || undefined` is similarly load-bearing).

Confidence: medium

---

**question:** `apps/studio/src/utils/getClientIp.ts:25-27` — is Cloudflare actually in front of this app, and is the origin locked down so a client can't reach it directly bypassing CF? If the origin is reachable without going through Cloudflare, `cf-connecting-ip` is no more trustworthy than XFF (a client can set it themselves) and this entire precedence collapses. If CF *is* the edge and origin-direct is blocked at the LB/SG layer, this is fine — but it's worth confirming since the whole security argument of the PR hinges on it.

Confidence: medium

---

**nitpick:** `apps/studio/src/utils/getClientIp.ts:5-19` — three of the four helpers (`getIpList`, `getFirstIp`, `getLastIp`) are reasonable, but `getLastIp` mixes "parse" and "policy" (the comment about leftmost-being-spoofable lives inside it). If you do consolidate the rate-limit and OTP callsites onto a shared helper, consider moving the "rightmost XFF only as last resort" comment up to `getIP` where the precedence decision actually lives — `getLastIp` is just `arr[arr.length - 1]`. Minor.

Confidence: low

seaerchin (approved):
lgtm, but remote address might be an issue given that our network topology is CF -> ALB -> ECS -> Docker. This means that the remote address seen might actually be the ALB. I'm more in favour for cf ip -> xff -> remoteAddr but i think is ok as is also.

apps/studio/src/utils/getClientIp.ts
```diff
@@ -1,10 +1,43 @@
...
+    ? request.headers.get("x-forwarded-for")
+    : request.headers["x-forwarded-for"]
 
-  return xff ? (Array.isArray(xff) ? xff[0] : xff.split(",")[0]) : "127.0.0.1"
+  // Prefer values set by our edge/proxy infrastructure before consulting
+  // X-Forwarded-For, which may contain client-supplied entries.
+  return (
+    getFirstIp(cfConnectingIp) ??
+    remoteAddress ??
+    getLastIp(xForwardedFor) ??
```
seaerchin, line 40: i realised this is probably wrong. because our network topology is CF -> ALB -> ECS, the chain of xff ips is as follows:

1. client IP (what they declare)
2. CF IP (what cf writes that they see)
3. ALB IP (what alb writes that they see)

taking the last IP for xff would result in us taking the value of the xff IP _recorded by ALB_ - because the preceding hop is from CF, we're essentially recording the CF reverse proxy's IP which is not useful. we should instead update this to `getImmediateNextIp`, which records the second ip instead.

adriangohjw:
Thanks for the thorough review @seaerchin! Addressing your points in the latest commit (7e56c15):

**`getRateLimitFingerprint` and `getOtpFingerPrint` using leftmost XFF**

Both helpers now delegate to `getIP` instead of duplicating the old extraction logic. The leftmost-XFF fallback path is gone from both callsites.

**`socket.remoteAddress` placed above `getLastIp(xForwardedFor)`**

Fixed. The precedence in `getIP` is now `cf-connecting-ip → rightmost XFF → socket.remoteAddress → 127.0.0.1`. In our containerised deployment, `socket.remoteAddress` is the LB IP, so rightmost XFF (appended by the LB) is a better signal for non-CF traffic.

**`typeof Request !== "undefined"` guard**

Removed — `request instanceof Request` is sufficient since Node 18+ (which Next.js 15 requires) always exposes `Request` globally.

**Comment placement in `getLastIp`**

Moved the XFF spoofability comment up to `getIP` where the precedence policy is decided. `getLastIp` is now just a parser with no policy commentary.

**Test coverage gaps**

Added the missing cases:
- Empty `cf-connecting-ip` falling through to XFF
- Whitespace-only `socket.remoteAddress` falling through
- `Request`-instance path for `cf-connecting-ip` precedence
- `socket.remoteAddress` used when XFF is absent

**On the backwards-compatibility note and the CF topology question**

Both are fair observations but outside the scope of code changes — agree they're worth documenting explicitly for anyone doing log forensics across the cutover. The CF + origin-lockdown question is a deployment constraint we can confirm separately.

apps/studio/src/utils/getClientIp.ts
adriangohjw: Good catch on the XFF structure — you're right that in our topology rightmost XFF ends up being the CF IP, not the client. But I think getImmediateNextIp would break things. Here's a walkthrough of all the cases:

**CF proxy ON, normal request**
```
Client (1.2.3.4) → CF → ALB → App

cf-connecting-ip:  1.2.3.4        ← CF stamps this
XFF:               1.2.3.4, 104.x.x.x

Result: cf-connecting-ip → 1.2.3.4 ✓
```

**CF proxy ON, attacker spoofs XFF**
```
Client sends X-Forwarded-For: evil-ip
CF appends:   evil-ip, 1.2.3.4
ALB appends:  evil-ip, 1.2.3.4, 104.x.x.x

cf-connecting-ip:  1.2.3.4        ← CF still stamps this correctly

Result: cf-connecting-ip → 1.2.3.4 ✓
```

**CF proxy OFF, normal request**
```
Client (1.2.3.4) → ALB → App

cf-connecting-ip:  not set
XFF:               1.2.3.4        ← ALB stamps this

Result: rightmost XFF → 1.2.3.4 ✓
```

**CF proxy OFF, attacker spoofs XFF**
```
Client sends X-Forwarded-For: evil-ip
ALB appends:  evil-ip, 1.2.3.4

cf-connecting-ip:  not set
XFF rightmost:     1.2.3.4        ← ALB stamped this, attacker can't fake it
XFF second-right:  evil-ip        ← attacker-controlled

Result: rightmost XFF → 1.2.3.4 ✓  (getImmediateNextIp would give evil-ip ✗)
```

**Local dev (no proxy)**
```
cf-connecting-ip:  not set
XFF:               not set
socket.remoteAddr: 127.0.0.1

Result: socket.remoteAddress → 127.0.0.1 ✓
```

The XFF rightmost = CF IP issue you flagged is real, but it only matters if CF-proxied traffic somehow reaches the XFF branch — which it shouldn't since CF always sets `cf-connecting-ip`. And `getImmediateNextIp` breaks the CF-off spoofing case, where ALB's trustworthy stamp is rightmost and the attacker's injected IP is second from right.

seaerchin: clarification: `getImmediateNextIp` would always take the ip at index one rather than the rightmost. in the case  where you have cf proxy off, attacker spoofing, the xff set looks like: [evil, 1.2.3.4]; `getImmediateNextIp` would then return `1.2.3.4` rather than the first IP that is spoofed. 

the idea is that the client only gets to stamp the left-most IP - if we take the IP immediately after that, it's what our first proxy registers as the client IP. because we're guaranteed to have at least one proxy (alb), i think it's a better fix than using the rightmost.

adriangohjw: @seaerchin 

> clarification: getImmediateNextIp would always take the ip at index one rather than the rightmost. in the case where you have cf proxy off, attacker spoofing, the xff set looks like: [evil, 1.2.3.4]; getImmediateNextIp would then return 1.2.3.4 rather than the first IP that is spoofed.

in this case, isn't the rightmost `1.2.3.4`?

> the idea is that the client only gets to stamp the left-most IP - if we take the IP immediately after that, it's what our first proxy registers as the client IP. because we're guaranteed to have at least one proxy (alb), i think it's a better fix than using the rightmost.

 The client can inject as many XFF entries as they want, not just one

```
Client sends: X-Forwarded-For: evil1, evil2, evil3
ALB appends:  evil1, evil2, evil3, 1.2.3.4

index 1       = evil2        ✗  attacker-controlled
rightmost     = 1.2.3.4      ✓  ALB stamped this
```

seaerchin: oh yes you are right... in this case, i'm ok with this - i misunderstood and thought that client only stamps the left most but actually the initial headers can all be set by client so ignore me

adriangohjw: npnp thanks for scrutinizing haha

PR #2225, opened 2026-05-28
fix(security): prevent stored XSS via unvalidated siteGtmId

adriangohjw:
## Problem

Site admins could set `siteGtmId` to a malicious string (e.g. `');alert(document.cookie);//`) via the integrations settings. That value was interpolated raw into a GTM inline `<script>` on every published page, allowing stored XSS against all visitors of a `.gov.sg` site.

## Solution

Defence in depth: validate at persistence time and escape at render time.

**Breaking Changes**

  - Details ...

**Features**:

- None

**Improvements**:

- Added `GTM_ID_STRING_REGEX` (`^GTM-[A-Z0-9]+$`) and applied it to the `siteGtmId` field in `SimpleIntegrationsSettingsSchema`, so invalid values are rejected by the existing AJV validation on `updateSiteIntegrations` and related config mutations.
- Added `serializeForInlineScript` utility that JSON-stringifies a value and escapes `<` to prevent `</script>` breakout when embedding data in inline script tags.
- Updated `GoogleTagManagerHeader` to embed the GTM ID via `serializeForInlineScript` instead of raw single-quoted interpolation.
- Updated the published-site JSON-LD block in `tooling/template/app/layout.tsx` to use the same utility (replacing duplicated inline logic).

**Bug Fixes**:

- Fixed stored XSS via unvalidated `siteGtmId` in the Google Tag Manager init script on published sites.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Log in as a site admin and open **Settings → Integrations** for a site.
- [ ] Enter a valid GTM ID (e.g. `GTM-ABC123`) and save — confirm the save succeeds and the value persists after reload.
- [ ] Attempt to save an invalid GTM ID (e.g. `');alert(1);//` or `gtm-abc`) — confirm the save is rejected with a validation error.
- [ ] Publish the site with a valid GTM ID, open the live site in a browser, and confirm the GTM init script in page source contains the ID as a JSON-stringified literal (not raw single-quoted interpolation).
- [ ] Confirm JSON-LD structured data still renders correctly in page source on the published site homepage.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

Made with [Cursor](https://cursor.com)


KenLSM (approved):

seaerchin (commented):
## Summary

- Validates `siteGtmId` at persistence time via a `^GTM-[A-Z0-9]+$` AJV `pattern` in `SimpleIntegrationsSettingsSchema` (`packages/components/src/types/site.ts:39`), and escapes at render time via a new shared `serializeForInlineScript` helper (`packages/validators/src/serializeForInlineScript.ts:10-11`) wired into `GoogleTagManagerHeader.tsx:22` and `tooling/template/app/layout.tsx:75`.
- Codeflow: `updateSiteIntegrations` → `isomerSiteConfigValidator` (AJV, `apps/studio/src/schemas/site.ts:80`) → DB write. Render path: site config → `RenderApplicationScripts` → `GoogleTagManagerHeader` (uses `serializeForInlineScript`) and `GoogleTagManagerBody` (still raw — see below).
- Overall: solid defense-in-depth fix. The package extraction is well-motivated. A couple of small things worth a look before merge.

## Findings

**praise:** The "validate at persistence + escape at render" split is exactly the right shape for this class of bug. The decision to lift `serializeForInlineScript` into a standalone `@isomer/validators` workspace (rather than have `tooling/template` depend on the full components bundle) is also a nice architectural choice — keeps the template lightweight.

Confidence: high

---

**issue (security):** `packages/components/src/templates/next/components/internal/GoogleTagManager/GoogleTagManagerBody.tsx:10` still interpolates `siteGtmId` directly into the iframe `src`:

```tsx
src={`https://www.googletagmanager.com/ns.html?id=${siteGtmId}`}
```

For *new* writes this is fine because the regex now restricts the value to `[A-Z0-9]`. But the persistence-time validation does not retroactively clean existing rows. Any pre-fix row that already contains a malicious `siteGtmId` (which is the actual scenario this PR exists to defend against) will still flow through here unsanitized. React escapes the attribute so it isn't HTML-injectable, but a value like `validid&extra=evil` could still inject extra GTM query params, and `validid#anchor` could mangle the URL.

The header path now has defense in depth via `serializeForInlineScript`; the body path doesn't. Suggest either (a) `encodeURIComponent` the ID here, or (b) add a one-off DB scan/migration to confirm no pre-existing row has a value outside the new regex.

Confidence: medium — depends on whether pre-existing bad values exist in the DB, which I can't verify from the diff alone.

---

**issue (robustness, non-blocking):** `packages/validators/src/serializeForInlineScript.ts:10-11` types input as `value: unknown`, but `JSON.stringify(undefined)` returns `undefined` (not the string `"undefined"`), so `.replace` will throw `TypeError`. Same for functions/symbols. Current callsites (`GoogleTagManagerHeader` only renders when `siteGtmId` is truthy; `layout.tsx` passes a literal object) don't trigger this, but `unknown` invites future callers to pass arbitrary values.

Either tighten the type (`value: string | object | number | boolean | null`) or short-circuit early:

```ts
const json = JSON.stringify(value)
if (json === undefined) return "null"
return json.replace(/</g, "\\u003c")
```

Confidence: medium — pure hygiene concern today, not exploitable.

---

**suggestion (test naming):** `packages/validators/src/__tests__/serializeForInlineScript.test.ts:18-28` — the case named *"should escape characters that would break out of single-quoted interpolation"* uses input `');alert(document.cookie);//` and expects output `"');alert(document.cookie);//"`. Nothing is actually escaped here; the safety comes purely from JSON wrapping the value in `"…"` as a string literal. The name overstates what the test verifies. Consider renaming to "should wrap the value as a JS string literal so injection payloads become inert" or adding a case where the input contains a character JSON.stringify actually escapes (e.g. a literal newline or `"`).

Confidence: high

---

**note:** `packages/components/src/utils/validation.ts:217` — the docstring opens with *"start with 'GTM-' followed by 6 to 7 uppercase letters or digits"* but the regex is `^GTM-[A-Z0-9]+$` (1+). The subsequent NOTE acknowledges the intentional loosening "for future proofing", but the opening sentence still reads as the actual constraint. Worth tightening so the comment matches the code.

Confidence: high

---

**question:** Is there a plan / migration to reconcile existing site configs whose `siteGtmId` doesn't match the new regex? `updateSiteIntegrations` validates the *entire* config object (`apps/studio/src/schemas/site.ts:92-98`), not just the changed fields. So a site admin editing an unrelated integration field (e.g. search settings) on a site whose stored `siteGtmId` happens to be invalid under the new regex would now have their save fail with a confusing validation error. Worth checking with a quick DB query — `select id, config->>'siteGtmId' from "Site" where config->>'siteGtmId' !~ '^GTM-[A-Z0-9]+$' and config ? 'siteGtmId'` or equivalent — to confirm no production sites are affected.

Confidence: medium

---

**nitpick (non-blocking):** `packages/components/src/utils/index.ts:18` re-exports `serializeForInlineScript` from `@isomer/validators` through the components utils barrel. Functional but a tiny indirection trap (consumers of `@opengovsg/isomer-components` get the symbol but reading the source they might wonder why it's there). If nothing inside `packages/components` actually needs the re-export anymore, you could drop it and require consumers to import from `@isomer/validators` directly. If it's there for backwards-compat with existing imports, a one-line comment would help.

Confidence: low

seaerchin (approved):

adriangohjw:
@seaerchin i did a search through on isomer's sites and saw that some tags with `G-XXX` and `GT-XXX` work, despite `GTM-XXX` being what google says it supported. Have loosen the regex to support them too in https://github.com/opengovsg/isomer/pull/2225/commits/b52f58ac22886b37e3eb756fdf5bf728c22f27ec but i won't update the copywriting of studio's - i think we still want to encourage sites to use `GTM-XXX` moving forward

PR #2227, opened 2026-05-28
fix(studio): redact sensitive fields in tRPC logger middleware

adriangohjw:
## Problem

The `loggerMiddleware` in `apps/studio/src/server/trpc.ts` logs raw procedure input on every tRPC call. Because this middleware is applied to all procedures via `baseProcedure`, sensitive values such as OTP tokens from `auth.email.verifyOtp` are written to application logs in plaintext. Anyone with log access could harvest a valid OTP within its expiry window and complete sign-in as the corresponding user.

## Solution

Added a `redactLogInput` utility that recursively replaces known-sensitive object keys (case-insensitive) with `[REDACTED]` before logging. The tRPC logger middleware now passes redacted input to both success and error log entries.

Sensitive keys covered: `token`, `otp`, `password`, `apiKey`, `secret`, `authorization`, `accessToken`, `refreshToken`.

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- N/A

**Bug Fixes**:

- Prevent OTP tokens and other sensitive procedure input from being written to application logs

## Before & After Screenshots

N/A — logging-only security fix with no UI changes.

## Tests

- [ ] Trigger an email OTP sign-in flow (request OTP, then submit OTP)
- [ ] Inspect application logs for the `auth.email.verifyOtp` procedure call
- [ ] Confirm the log entry includes the email but shows `token: "[REDACTED]"` instead of the plaintext OTP

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

<div><a href="https://cursor.com/agents/bc-d929f2cd-08cc-4445-9a6c-76d77f1c2806"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d929f2cd-08cc-4445-9a6c-76d77f1c2806"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/studio/src/lib/__tests__/redact-log-input.test.ts
+++ b/apps/studio/src/lib/__tests__/redact-log-input.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "vitest"
+
+import { redactLogInput } from "../redact-log-input"
+
+describe("redactLogInput", () => {
+  it("redacts known sensitive keys at the top level", () => {
+    expect(
+      redactLogInput({
+        email: "user@example.com",
+        token: "123456",
+        password: "hunter2",
+        apiKey: "sk-live",
+        secret: "shh",
+      }),
+    ).toEqual({
+      email: "user@example.com",
+      token: "[REDACTED]",
+      password: "[REDACTED]",
+      apiKey: "[REDACTED]",
+      secret: "[REDACTED]",
+    })
+  })
+
+  it("redacts sensitive keys case-insensitively", () => {
+    expect(redactLogInput({ OTP: "654321", Token: "abcdef" })).toEqual({
+      OTP: "[REDACTED]",
+      Token: "[REDACTED]",
+    })
+  })
+
+  it("redacts sensitive keys in nested objects and arrays", () => {
+    expect(
+      redactLogInput({
+        users: [
+          { email: "a@example.com", token: "one" },
+          { email: "b@example.com", token: "two" },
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
lgtm, was wondering what about stuff like eg: `gbToken` (where it contains a sensitive key as a substring) but i think the correct fix should be to add to the sensitive key list rather than searching for the presence of the substring

adriangohjw:
> lgtm, was wondering what about stuff like eg: `gbToken` (where it contains a sensitive key as a substring) but i think the correct fix should be to add to the sensitive key list rather than searching for the presence of the substring

yeah it depends how we want to avoid logging it. This PR only address is logging in the TRPC middleware. If we want to exclude it from logging in general, we might have to take a look at how the Singpass Auth PR do it which is to manually exclude - or add another mdiddlware to the general logger to redact any sensitive keys automatically

PR #2246, opened 2026-05-29
fix: guard JSON.parse in useCollection against malformed filters param

adriangohjw:
## Problem

Closes https://app.notion.com/p/opengov/unguarded-json-parse-on-user-controlled-filters-param-crashes-collectionclient-36e77dbba78881179ec9d767d355cd93

When a user navigates to a collection page with a malformed `filters` URL param (e.g. `?filters=hello`), `JSON.parse` throws a `SyntaxError` synchronously inside `useMemo`, crashing the entire `CollectionClient` subtree with a React render error. Any user who manually types or shares such a URL would see the page crash.

## Solution

Wraps the `JSON.parse` call in `useCollection.ts` in a `try/catch` so malformed values fall back to an empty array instead of propagating a `SyntaxError`.

**Breaking Changes**

**Features**:

- N/A

**Improvements**:

- Migrated `useQueryParams` and new `useCollection` tests from manual `createRoot`/`HookConsumer` boilerplate to `@testing-library/react`'s `renderHook`, reducing test verbosity significantly.

**Bug Fixes**:

- `useCollection`: unguarded `JSON.parse` on the `filters` URL param no longer crashes `CollectionClient` when the param is not valid JSON.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Navigate to any collection page (e.g. `/resources`)
- [ ] Append `?filters=hello` to the URL and load the page
- [ ] Verify the page renders normally with no filters applied (previously this would crash the collection)
- [ ] Append `?filters=%5B%7B%22id%22%3A%22category%22%2C%22items%22%3A%5B%7B%22id%22%3A%22guides%22%7D%5D%7D%5D` (valid JSON filters) and verify filters are applied correctly

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- `@testing-library/react`: used in `packages/components` for `renderHook`-based hook tests


packages/components/src/templates/next/layouts/Collection/useCollection.ts
```diff
@@ -37,7 +37,12 @@ export const useCollection = ({
     if (isEmpty(filters)) {
       return []
     }
-    return JSON.parse(filters || "[]") as AppliedFilter[]
+    try {
+      return JSON.parse(filters || "[]") as AppliedFilter[]
```
seaerchin: question: should we use `zod.safeParse` here? pattern seems to be that we want to guarantee the shape rather than just blindly validating json

adriangohjw: Done — added a `appliedFiltersSchema` Zod schema in `Filter.ts` and swapped the `as AppliedFilter[]` cast for `safeParse`. Invalid JSON still falls through to the `catch`, and valid-but-wrong-shaped data now also returns `[]`.

seaerchin (approved):
lgtm! i think don't need the `try/catch` loop but happy to keep for security

PR #2248, opened 2026-05-30
fix: use date-fns month token (MM) in slashDateSchema

### LINKED ISSUE #2234: (deepsec) [BUG] slashDateSchema uses date-fns minute token (mm) instead of month (MM)
Reference: https://www.notion.so/opengov/isomer-other-logic-bug-9085357be0-36e77dbba788818898b8c44cf19a7943

**File:** [`[apps/studio/src/schemas/collection.ts](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/schemas/collection.ts#L24-L32)`](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/schemas/collection.ts#L24-L32) (lines 24, 32)
**Project:** isomer
**Severity:** BUG • **Confidence:** high • **Slug:** `other-logic-bug`

## Finding

In `slashDateSchema`, both `parse(d, "dd/mm/yyyy", new Date())` (L24) and `format(d, "dd/mm/yyyy")` (L32) use the date-fns token `mm`, which means *minute* (00-59), not *month*. The intended format — confirmed by the file’s own comments and by every other date-fns format string in the codebase (e.g., `apps/studio/src/server/modules/collection/collection.service.ts:14,36`, `apps/studio/src/features/gazettes/.../CreateGazetteModal.tsx:115`, and gazette.ts’s error message “Date must be dd/MM/yyyy”) — is `dd/MM/yyyy`.

Consequences for `editLinkSchema.date` → `updateCollectionLink`:
1. **Invalid dates are accepted as valid.** Input `"29/02/2023"` (Feb 29 in a non-leap year — invalid) is parsed when the request runs in a month with ≥29 days as: day=29, minute=02, year=2023, month=*current month*. The intermediate Date is valid, so the schema accepts the string, formats it back to `"29/02/2023"`, and stores that bogus date in the collection link’s blob.
2. **Valid dates are rejected depending on the current month.** Input `"31/01/2024"` (Jan 31, valid) when today is February is parsed as day=31, minute=01, year=2024, month=Feb → Feb 31 → Invalid Date → schema rejects. The same value would be accepted any other month.
3. The intermediate Date object’s month is always the server’s current month rather than the user’s intended month. The bug is masked for round-trip because `format` also uses `mm` (minute), so the minute that was just parsed gets re-emitted in the month position — but the underlying Date is wrong, and any downstream code that uses the Date (or re-parses the string with the correct `dd/MM/yyyy`) gets incorrect results.

## Recommendation

Change both `dd/mm/yyyy` literals to `dd/MM/yyyy` (capital `MM` is the month token in date-fns). Consider centralizing the format string in a constant to avoid the lowercase/uppercase mistake recurring.

jiaxinnns:
## Problem

Closes #2234

In `slashDateSchema` ([apps/studio/src/schemas/collection.ts](apps/studio/src/schemas/collection.ts)), both `parse(d, "dd/mm/yyyy", ...)` and `format(d, "dd/mm/yyyy")` use the date-fns token `mm`, which means *minute* (00-59), not *month*.

Also closes https://app.notion.com/p/opengov/wrong-date-fns-format-token-mm-minutes-used-instead-of-mm-months-in-slashdatesch-36e77dbba7888141bfeacc683701b863

## Solution

Change both `dd/mm/yyyy` literals to `dd/MM/yyyy` (capital `MM` is the month token in date-fns), and centralize the format string in a `SLASH_DATE_FORMAT` constant to prevent the lowercase/uppercase mistake from recurring.

**Breaking Changes**

  - Existing stored date strings are unaffected: the previous `mm`/`mm` round-trip emitted the same string the user entered, so values like `"15/03/2024"` are stored identically and now re-parse correctly under the `MM` token.

**Features**:

- NIL

**Improvements**:

- Centralize the slash date format string into a `SLASH_DATE_FORMAT` constant.

**Bug Fixes**:

- Fix `slashDateSchema` using the date-fns minute token (`mm`) instead of the month token (`MM`), which accepted invalid dates and rejected valid ones depending on the server's current month.

## Before & After Screenshots

**BEFORE**:

NIL

**AFTER**:

NIL

## Tests

- [ ] Run the unit tests: `pnpm test:unit -- src/server/modules/collection/__tests__/collection.router.test.ts -t "updateCollectionLink"`

**New scripts**:

NIL

**New dependencies**:

NIL

**New dev dependencies**:

NIL


apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
```diff
@@ -1404,6 +1404,49 @@ describe("collection.router", async () => {
...
+      const result = caller.updateCollectionLink({
+        siteId: site.id,
+        category: "category",
+        ref: "1",
+        linkId: Number(page.id),
+        date: "29/02/2023",
+      })
+
+      // Assert
+      await expect(result).rejects.toThrow()
```
adriangohjw: ask: can specify what's being thrown?

adriangohjw (commented):
nice PR, just a tiny ask, thank you!

adriangohjw (approved):

PR #2251, opened 2026-05-31
fix: match logout events by email in audit log query

adriangohjw:
## Problem

Logout audit log events were silently dropped from monthly CSV exports because the filter compared `al.userId` (a cuid string) against a set of user email addresses — a comparison that can never match, so all logout rows were excluded.

Closes https://app.notion.com/p/opengov/logout-events-for-current-site-users-silently-omitted-userid-compared-against-em-36e77dbba78881ee9903d8e8ceb54443

Closes https://app.notion.com/p/opengov/logout-filter-compares-userid-to-email-strings-dropping-all-logout-events-for-cu-36e77dbba78881e0bb9cf6460b060f50

## Solution

Changed the logout filter in `getAuditLogQuery` to extract the email from `al.delta -> 'before' ->> 'email'` and match it against the `emailsFromUsers` CTE, which is the correct field written by the logout event handler.

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- Extended `vitest.config.ts` include pattern to cover `prisma/scripts/` so script-level tests are picked up by the test runner.
- Exported `getAuditLogQuery` and `getStringifiedValue` from the script and guarded the top-level `await getAuditLogsForSite()` call behind an ESM main check so the module can be safely imported by tests without executing.

**Bug Fixes**:

- `getAuditLogQuery` events query: logout events for site users are now correctly included in audit log exports instead of being silently dropped.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

Added `apps/studio/prisma/scripts/__tests__/getAuditLogs.test.ts` with 53 tests (10 pure unit, 43 integration) covering:
- `getStringifiedValue`: all value types including falsy edge cases
- Users query: column presence, OGP exclusion, soft-delete exclusion, cross-site isolation, date-range ignorance
- Events query descriptions: all 13 event types including both `ResourceUpdate` and `Publish` special paths
- Events filtering: siteId matching, `delta.after.siteId` fallback, excluded event types, Login inclusion/exclusion
- Logout regression (the core fix): site-user logout included, non-site excluded, OGP excluded, null email safe
- Date-range scoping: inclusive boundaries, out-of-range exclusion, `asc` ordering, CTE date scoping
- Edge cases: unknown type throws, malformed `monthYear` throws

### DIFF
--- a/apps/studio/prisma/scripts/__tests__/getAuditLogs.test.ts
+++ b/apps/studio/prisma/scripts/__tests__/getAuditLogs.test.ts
@@ -33,7 +33,7 @@ const insertAuditLog = async ({
   ipAddress?: string | null
   createdAt?: Date
 }) => {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  // oxlint-disable-next-line @typescript-eslint/no-explicit-any
   const values: any = {
     eventType,
     userId,
@@ -1048,7 +1048,7 @@ describe("getAuditLogQuery", () => {
       const buildQuery = () =>
         getAuditLogQuery({
           siteId,
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          // oxlint-disable-next-line @typescript-eslint/no-explicit-any
           type: "unknown" as any,
           monthYear: TEST_MONTH,
         })
--- a/apps/studio/prisma/scripts/__tests__/getAuditLogs.test.ts
+++ b/apps/studio/prisma/scripts/__tests__/getAuditLogs.test.ts
@@ -16,6 +16,41 @@ const MONTH_START = startOfMonth(TEST_DATE)
 const MONTH_END = endOfMonth(TEST_DATE)
 const MID_MONTH = new Date("2026-01-15T12:00:00.000Z")
 
+interface UsersQueryRow {
+  Email: string
+  '"Last login"': Date | null
+  Role: string
+  '"Date added"': Date
+}
+
+interface EventsQueryRow {
+  '"Date and time"': Date
+  '"Event type"': AuditLogEvent
+  '"Account creation date"': Date | null
+  '"Last login date"': Date | null
+  Description: unknown
+  Delta: unknown
+  Email: string | null
+  Metadata: unknown
+  Name: string | null
+}
+
+const getUsersRows = async (params: { siteId: number; monthYear: string }) => {
+  const result = await getAuditLogQuery({
+    ...params,
+    type: "users",
+  }).execute()
+  return result as unknown as UsersQueryRow[]
+}
+
+const getEventsRows = async (params: { siteId: number; monthYear: string }) => {
+  const result = await getAuditLogQuery({
+    ...params,
+    type: "events",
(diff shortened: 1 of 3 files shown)


seaerchin (approved):

adriangohjw (commented):
## Summary

- Fixes a silent bug where all logout events were dropped from monthly audit log CSV exports. The filter at `getAuditLogs.ts:325` compared `al.userId` (a cuid) against the `emailsFromUsers` CTE which contains email addresses — a type mismatch that can never match.
- The fix extracts `al.delta -> 'before' ->> 'email'` and matches it against the email sets. This is grounded correctly in the logout event structure: `auth.router.ts:38-44` stores `delta.before = user` (the full User object), so `delta -> 'before' ->> 'email'` reliably retrieves the email.
- Supporting changes: export two functions for test access, ESM main guard, extend vitest `include` to cover `prisma/scripts/`, add 53 tests (10 unit, 43 integration) including a direct regression test.
- Overall assessment: targeted fix, correct root cause analysis, good test coverage. One dead column to clean up and a pre-existing observation below.

---

**suggestion (non-blocking):** `prisma/scripts/getAuditLogs.ts:L215` — `User.id` is added to the `emailsFromUsers` CTE `SELECT` but is never referenced downstream. Both usages of the CTE (`L318` and `L331`) only reference `emailsFromUsers.email`. This is a dead column that adds noise to an already complex query.

Confidence: high

---

**note:** `prisma/scripts/getAuditLogs.ts:L188-L212` — Pre-existing, not introduced by this PR: `emailsFromPermissionChanges` stores `userId` cuid values in its `email` alias column (extracted from `delta -> 'after' ->> 'userId'` / `delta -> 'before' ->> 'userId'`). Confirmed by `audit.service.ts:L298-L308` where `CreatePermissionDelta.after` and `DeletePermissionDelta.before/after` are `ResourcePermission` objects whose `userId` is a cuid. The UNION with this CTE in both the login (`L317-L323`) and logout (`L330-L336`) filters therefore never produces a match against email-based comparisons. The apparent intent — including events for users whose permissions changed during the period but who no longer appear in `emailsFromUsers` — is not achieved. Out of scope here, but worth a follow-up ticket since recently-removed users may still be silently omitted.

Confidence: high

---

**praise:** `prisma/scripts/getAuditLogs.ts:L327` — Null handling is correct: if `delta -> 'before' ->> 'email'` is NULL, the `IN` comparison evaluates to `FALSE`, quietly excluding the row without crashing. Test coverage for this edge case (`"null email safe"`) is present.

---

**note:** `vitest.config.ts:L6` — Switching from `dir: "src"` to explicit `include` patterns is the right approach for multi-directory discovery. Minor narrowing: the old `dir`-based discovery used vitest's default glob (which also matched `.spec.*`); the new `include` only matches `.test.{ts,tsx}`. Since the codebase consistently uses `.test.ts`, nothing should be missed.

Confidence: medium

PR #2252, opened 2026-06-02
fix: remove orphaned number in Masthead SVG path causing React 19 render error

adriangohjw:
## Problem

The Masthead component's lion head SVG contained an invalid path `d` attribute — an orphaned `0` that left a cubic bezier command short of its required 6 parameters. This produced a **"Expected number"** browser error on every page using the Masthead.

This bug was always latent in the source, but became a surfaced error after the upgrade to Next.js 16 / React 19. React 19 validates and re-serialises SVG path data server-side rather than passing raw strings straight to the DOM, so the malformed segment now throws instead of being silently ignored by the browser.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Removed the orphaned `0` between the last cubic bezier control point and the `h` command in the lion head SVG path (`Masthead.tsx` line 19). The bezier was already complete at 6 numbers; the stray `0` caused the parser to expect 5 more numbers before it hit the `h` command letter, producing the error.

**Why removing the `0` is correct and doesn't affect the shape:**

The SVG cubic bezier command `c` takes exactly 6 numbers per curve (`x1 y1 x2 y2 x y`). In the original path:

```
c0 0 0.704-1.341 0.802-1.341 0h-2.225
```

- `0 0` → control point 1 (x1, y1)
- `0.704-1.341` → control point 2 (x2=-1.341, y2=0.704... wait, parsed as x2=0.704, y2=-1.341)
- `0.802-1.341` → end point (x=0.802, y=-1.341) — bezier complete at 6 numbers

The trailing `0` is a 7th number, which would start an implicit second `c` curve — but only 1 of the required 6 numbers appears before the `h` command letter, causing the parse error.

After the fix (`0.802-1.341h-2.225`), no space is needed before `h`: the letter `h` always terminates the preceding number token, and `-` always starts a new number. The SVG parser reads this unambiguously — `0.802` (bezier x), `-1.341` (bezier y, separated by the `-` sign), then `h` as the next command. The bezier curve itself is completely unchanged.

## Before & After Screenshots

**BEFORE**:

Browser console: `Error: <path> attribute d: Expected number, "…0.802-1.341 0h-2.225…"`

**AFTER**:

No SVG path errors; lion head renders identically (the valid bezier segment is unchanged — only the orphaned trailing `0` was removed).

## Tests

- [ ] Load any Isomer page that renders the Masthead (e.g. any public site page)
- [ ] Open the browser DevTools console and confirm no `<path> attribute d: Expected number` error appears
- [ ] Confirm the Singapore lion head icon in the Masthead banner renders correctly

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None


seaerchin (approved):
it seems like this will mess up the svg? but ok w/ approving if you have tested

adriangohjw:
> it seems like this will mess up the svg? but ok w/ approving if you have tested

yeha it does, but very unnoticeable (as per storeybook) - i will ask SH to approve haha

PR #2303, opened 2026-06-05
fix: validation regex edge cases

### LINKED ISSUE #2229: (deepsec) [BUG] Regex typos: `embed?` and `video.php?` make the preceding character optional instead of escaping `?`
Reference: https://www.notion.so/opengov/isomer-other-regex-typo-9bfac02193-36e77dbba788813bb882fe9a7add3d04

---

**File:** [`[packages/components/src/utils/validation.ts](https://github.com/opengovsg/isomer/blob/main/packages/components/src/utils/validation.ts#L94-L147)`](https://github.com/opengovsg/isomer/blob/main/packages/components/src/utils/validation.ts#L94-L147) (lines 94, 147)
**Project:** isomer
**Severity:** BUG • **Confidence:** high • **Slug:** `other-regex-typo`

## Finding

Two regexes contain a `?` that was likely intended as a literal URL separator but is being interpreted as the quantifier ‘zero or one of the preceding character’. Line 94 `googlemaps: "^https://www\\.google\\.com/maps(?:/d)?/embed?.*$"` — here `embed?` matches `embe` or `embed`, so URLs like `https://www.google.com/maps/embeXYZ` match. Line 147 `fbvideo: "^https://www\\.facebook\\.com/plugins/video.php?.*$"` — here `video.php?` matches `video.ph` or `video.php` (the unescaped `.` is also a regex metacharacter matching any char). These do not lead to a security issue because the runtime helpers (`isValidGoogleMapsEmbedUrl` requires `pathname === "/maps/embed"`/`"/maps/d/embed"`; `parseVideo` only feeds URLs to type-specific helpers) catch the malformed inputs, but they relax JSON schema validation more than intended and can confuse future maintainers about the actual constraint.

## Recommendation

Replace `embed?` with `embed/?` (allow optional trailing slash) or anchor with `embed(/.*)?$` depending on intent. Replace `video.php?` with `video\\.php(?:\\?.*)?$` to escape both the `.` and the `?` query separator. Add unit tests in **tests**/validation.test.ts that assert the typo-introduced edge cases (e.g., `/maps/embe...`) are rejected by the pattern.

## Revalidation

**Verdict:** true-positive

Both regexes are misclassified by the author of the patterns. In line 94 `^https://www\.google\.com/maps(?:/d)?/embed?.*$`, the `?` after `embed` is a quantifier making `d` optional — so the pattern matches `…/embe` as well as `…/embed`. In line 147 `^https://www\.facebook\.com/plugins/video.php?.*$`, the unescaped `.` in `video.php` matches any char and the `?` makes `p` optional — so it matches `video.ph` plus anything. The author correctly notes that the runtime helpers `isValidGoogleMapsEmbedUrl` (strict `pathname === "/maps/embed" || "/maps/d/embed"`) and `isValidVideoUrl` (strict hostname check + VIDEO_EMBED_URL_PATTERN that still has the typo, so this is enforced by the per-video subparsers in `parseVideo`) catch the malformed inputs, so there is no security impact. The finding is a correctness/maintainability bug at high confidence, exactly as described.

### LINKED ISSUE #2231: (deepsec) [BUG] resourceSchema regex is unanchored — accepts arbitrary strings containing a digit
Reference: https://www.notion.so/opengov/isomer-other-weak-validation-f8715a5f62-36e77dbba78881aab2defea3b24ab4d4

**File:** [`[apps/studio/src/schemas/resource.ts](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/schemas/resource.ts#L10-L14)`](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/schemas/resource.ts#L10-L14) (lines 10, 11, 12, 13, 14)
**Project:** isomer
**Severity:** BUG • **Confidence:** high • **Slug:** `other-weak-validation`

## Finding

`resourceSchema` (L10-14) validates with `.regex(/[0-9]+/)`. JavaScript regex literals without `^…$` match anywhere in the string, so inputs like `"abc123"`, `"1abc"`, or `"<script>1"` all pass. The `.refine((s) => !s.startsWith("0"))` only blocks strings beginning with `0`, not non-numeric strings.

Contrast with `bigIntSchema` immediately below (L19-26), which correctly uses `/^[0-9]+$/`. The same author clearly intended numeric-only validation for `resourceSchema`, since the surrounding constants and refinements only make sense for numeric IDs.

`resourceSchema` is used by `deleteResourceSchema.resourceId` and `getParentSchema.resourceId`. Downstream:
- The handlers pass the string into Kysely parameterized queries against the bigint `Resource.id` column, so SQL injection is not possible.
- Permission checks via `bulkValidateUserPermissionsForResources` will throw NOT_FOUND for non-numeric IDs, so authorization isn’t bypassed.
- The realistic impact is that invalid input bypasses the schema’s first line of defense and surfaces as a less helpful database-type error (or a `NOT_FOUND` after a doomed lookup) instead of being rejected as a 400 with a clear validation error.

This is a defense-in-depth/UX bug, not an exploitable vulnerability, but it should be fixed for consistency with `bigIntSchema` and to make the schema actually enforce what it appears to enforce.

## Recommendation

Anchor the regex: `.regex(/^[0-9]+$/)`. While at it, consider replacing `resourceSchema` with `bigIntSchema` (they have effectively identical intent and the latter is already correct) to avoid carrying two near-duplicate validators.

## Revalidation

**Verdict:** true-positive

Confirmed by reading the file: `resourceSchema` (lines 10-14) uses `.regex(/[0-9]+/)` while the adjacent `bigIntSchema` (lines 19-26) correctly uses `.regex(/^[0-9]+$/)`. The unanchored regex matches anywhere in the string, so `abc123`, `1abc`, and `<script>1` all pass; the `.refine((s) => !s.startsWith('0'))` only blocks leading-zero strings, not non-numeric input. Confirmed `resourceSchema` is only used by `deleteResourceSchema.resourceId` and `getParentSchema.resourceId` (resource.router.ts:631, 707). Downstream: the value is passed to `bulkValidateUserPermissionsForResources` and to Kysely `.where('id', '=', resourceId)` (parameterized), so SQL injection is impossible and authorization checks would either error or return NOT_FOUND for a non-numeric ID. The finding itself acknowledges these mitigations and characterizes it as defense-in-depth/UX rather than exploitable, which matches my analysis. BUG severity is appropriate.

Benjam11n:
## Problem

Closes #2229
Closes #2231

Some validation regexes were looser than intended:
- Google Maps and Facebook video embed URL patterns treated `?` as a regex quantifier instead of a literal query separator.
- `resourceSchema` used an unanchored digit regex, allowing strings such as `abc123`, `1abc`, and `<script>1`.

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- None

**Improvements**:

- Reuse the existing stricter `bigIntSchema` for resource IDs to avoid duplicate validation logic.

**Bug Fixes**:

- Require Google Maps embed URLs to match `/maps/embed` or `/maps/d/embed` with an optional query string.
- Require Facebook video URLs to match the literal `/plugins/video.php` path with an optional query string.
- Reject resource IDs that are not entirely numeric.

## Tests

- [ ] In Studio, confirm resource operations using valid numeric resource IDs still work.
- [ ] Confirm malformed resource IDs such as `abc123` are rejected by schema validation.
- [ ] Confirm Google Maps and Facebook video embeds still accept valid embed URLs and reject malformed path variants.


apps/studio/src/schemas/resource.ts
```diff
@@ -25,6 +19,8 @@ const bigIntSchema = z
   .regex(/^[0-9]+$/)
   .refine((v) => v.at(0) !== "0")
```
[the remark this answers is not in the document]
Benjam11n: Addressed by switching the leading-zero check to `!v.startsWith("0")`.

apps/studio/src/schemas/__tests__/resource.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    ["getParentSchema", getParentSchema],
+  ])("%s", (_, schema) => {
+    it("should allow non-zero numeric resource IDs", () => {
+      const result = schema.safeParse({
+        siteId: 1,
+        resourceId: "123",
+      })
+
+      expect(result.success).toBe(true)
+    })
```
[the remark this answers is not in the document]
Benjam11n: Addressed by adding regression cases for `"0"` and `"0123"`.

adriangohjw (commented):
## Summary

- Fixes two regex quantifier bugs: `embed?` (makes `d` optional) and `video.php?` (makes `p` optional and treats `.` as any char) in the URL allow-list patterns used for Google Maps and Facebook Video embeds.
- Consolidates `resourceSchema` by aliasing it to the stricter, already-existing `bigIntSchema`, eliminating the unanchored `/[0-9]+/` regex that accepted strings like `abc123`.
- Adds focused unit tests for the resource schema and extends the embed URL test suite with new rejection cases.

**Code flow**: `validation.ts` exports `MAPS_EMBED_URL_REGEXES` / `VIDEO_EMBED_URL_REGEXES` as string constants that feed both `MAPS_EMBED_URL_PATTERN`/`VIDEO_EMBED_URL_PATTERN` (consumed by JSON Schema validators in components) and the `isValidMapEmbedUrl`/`isValidVideoUrl` helper functions (which add a second layer via `new URL()` object checks). The PR tightens the regex layer while keeping the URL-object layer intact — dual-layer defence is preserved.

Overall this is a clean, well-scoped fix. A few things worth noting below.

---

**praise:** `packages/components/src/utils/validation.ts:L79-L88` — The dual-layer validation approach (URL object check + regex check inside `isValidMapEmbedUrl`) is good defensive design. The PR correctly fixes the regex layer without disturbing the URL-object layer, keeping both guards in place for the function path.

Confidence: high

---

**issue (security):** `packages/components/src/utils/validation.ts:L93-L98` — `MAPS_EMBED_URL_REGEXES` is exported and used standalone as a JSON Schema pattern (via `MAPS_EMBED_URL_PATTERN`) in the components package. In that path there is **no** URL-object layer — only the regex. The old `embed?` pattern meant `https://www.google.com/maps/embeXYZ` or `https://www.google.com/maps/embeSOMETHING` would pass the JSON Schema check. The PR's fix (`embed(?:\\?.*)?`) is correct. Worth calling out explicitly: the new test case `"https://www.google.com/maps/embeXYZ"` in the invalid list directly covers this and is a good regression guard.

Confidence: high

---

**note:** `apps/studio/src/schemas/resource.ts` — The PR description marks this as "backwards compatible", but the `resourceSchema` change from `/[0-9]+/` (unanchored) to `bigIntSchema`'s `/^[0-9]+$/` (fully anchored) is a **tightening** of validation: strings like `abc123` or `1abc` were previously accepted and are now rejected. In practice, IDs from the database are always purely numeric, so no real-world caller breaks — but the compatibility claim is technically inaccurate. If there are any callers constructing resourceIds from mixed-content strings, they would now fail schema validation. Low risk in practice, worth a quick check.

Confidence: medium

---

**suggestion:** `apps/studio/src/schemas/resource.ts:L19-L26` — `bigIntSchema` constraints (`.min(1)` and `.regex(...)`) have no user-facing `message` strings. The schema `CLAUDE.md` convention requires these on every constraint. Now that `resourceSchema = bigIntSchema`, these bare constraints surface in the error paths for `deleteResourceSchema` and `getParentSchema` as well as all the pre-existing callers. This was pre-existing debt, but the PR that aliases `resourceSchema` to `bigIntSchema` is the natural place to add messages.

Example:
```ts
const bigIntSchema = z
  .string()
  .min(1, "Enter a valid resource ID")
  .regex(/^[0-9]+$/, "Resource ID must contain only digits")
  .refine((v) => !v.startsWith("0"), {
    message: "Resource ID must not have leading zeros",
  })
```

Confidence: high

---

**suggestion:** `apps/studio/src/schemas/__tests__/resource.test.ts:L17-L22` — The valid-case test only covers `"123"`. Consider adding `"1"` (single digit, boundary) and `"9007199254740991"` (max safe integer boundary) to verify the schema handles edge-of-range IDs. Minor, but the test file is new so it's worth making the coverage intentional.

Confidence: low

---

**nitpick:** `apps/studio/src/schemas/__tests__/resource.test.ts:L17` — The valid-case test does not include an empty-string check. `bigIntSchema` has `.min(1)` which should reject `""`, but there is no explicit assertion for it. Given the security context (resource ID as a path to a database row), an explicit rejection test for `""` would complete the boundary coverage.

Confidence: medium

apps/studio/src/schemas/__tests__/resource.test.ts
adriangohjw, line 1: good intention! our practice is to usually not write test for schemas directly but rather than the consumer of schema - in this case the routers. Though since these are rather primitive checks, i think we can skip it altogether in this situation

apps/studio/src/schemas/resource.ts
```diff
@@ -23,7 +17,9 @@ const bigIntSchema = z
   .string()
   .min(1)
   .regex(/^[0-9]+$/)
-  .refine((v) => v.at(0) !== "0")
+  .refine((v) => !v.startsWith("0"))
+
+const resourceSchema = bigIntSchema
```
adriangohjw: nit: think we can skip this reassignment - we can just assign `bigIntSchema` directly e.g. in `getMetadataSchema`

apps/studio/src/schemas/__tests__/resource.test.ts
Benjam11n: fixed in latest commit

apps/studio/src/schemas/resource.ts
Benjam11n: fixed in latest commit

adriangohjw (approved):

PR #2304, opened 2026-06-05
chore: add isomer-conventions skill for code smells

karrui:
## Problem

We have no shared, repo-specific record of code smells and blessed conventions. Patterns get re-explained in review, and AFK agents have no convention catalog to check against.

## Solution

Adds an `isomer-conventions` Claude Code skill under `.claude/skills/` — a **growable** catalog of code smells (anti-patterns) and best practices specific to this repo.

**Improvements**:

- `SKILL.md` documents two flows: *applying* conventions during review/authoring (flag deviations as `path:line` with the blessed pattern), and *adding* a new entry. It indexes entries; it intentionally stays lean.
- Detail lives in one file per convention under `conventions/` (mirrors the index + per-entry pattern), with a `TEMPLATE.md` for consistency. Categories are created on the fly as entries land.
- Designed to grow over time — engineers add entries by asking their agent ("code smell: ...", "best practice: ...") or by hand from the template, then PR as normal.

Seed conventions:

- **React** — prefer a new component over overloading props (flag soup / single-caller props)
- **React** — build forms with `useZodForm` + a `~/schemas` zod schema, not per-field `useState`
- **Audit logging** — populate before/after deltas with real DB rows re-read inside the same `tx`, never hand-built objects
- **Readability** — document every regex with a verified what/why comment

Each entry cites the real repo idiom (e.g. `useZodForm` at `apps/studio/src/lib/form.ts`, the move handler's fetch → mutate → re-fetch → log pattern) and carves out sanctioned exceptions so the smells don't misfire.

## Tests

No production code changes — this adds documentation/tooling only (a project skill), so there are no manual verification steps.

### DIFF
--- a/.claude/skills/isomer-conventions/SKILL.md
+++ b/.claude/skills/isomer-conventions/SKILL.md
@@ -1,0 +1,58 @@
+---
+name: isomer-conventions
+description: Repository-specific code smells and blessed conventions for the Isomer Next monorepo. Use when writing or reviewing code in this repo (tRPC routers, Kysely/Prisma access, zod schemas, React components, tests) to catch deviations from team conventions, and when the user wants to record a new code smell or best practice ("remember this convention", "add a code smell", "we always/never do X here").
+---
+
+# Isomer conventions
+
+A growable catalog of repo-specific code smells (patterns to avoid) and best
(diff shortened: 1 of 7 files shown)


seaerchin (approved):
faster merge

PR #2307, opened 2026-06-05
👨🏻‍💻 DX - feat: add Vercel preview environment support

adriangohjw:
## Problem

Reviewers currently need to pull a branch and run the full Docker + Postgres + Mockpass stack locally to exercise CMS changes in a browser. This PR enables per-PR Vercel preview deployments so reviewers can click a link instead.

After merging `main`, preview deployments also crashed on every tRPC request (`ERR_REQUIRE_ESM`) because `asset.service.ts` eagerly imported `jsdom`/`isomorphic-dompurify` at module load — a dependency chain that Turbopack's external loader cannot `require()` on Vercel.

## Solution

All preview-specific behaviour is gated on a single env signal: `NEXT_PUBLIC_APP_ENV === "preview"`. The value auto-derives from `NEXT_PUBLIC_VERCEL_ENV === "preview"` when not explicitly set in Vercel project settings. Adding `"preview"` to the existing enum means the preview path is structurally impossible in production.

**Breaking Changes**

  - Details ...

**Features**:

- **Cloudflare R2 asset storage for preview** (`src/lib/s3.ts`, `src/env.mjs`): preview asset uploads/downloads reuse the exact same `S3Client`-based code path as staging/production, just pointed at R2's S3-compatible endpoint (`https://<account>.r2.cloudflarestorage.com`) with R2 credentials. Which backend to use is derived from whether `R2_ACCOUNT_ID`/`R2_ACCESS_KEY_ID`/`R2_SECRET_ACCESS_KEY` are present (validated together via `env.mjs` superRefine), not from `NEXT_PUBLIC_APP_ENV` — so it isn't coupled to the environment name. Tag-based operations that don't make sense for ephemeral preview data (`deleteFile`, `setAssetAsPublished`, `markScheduledAssetAsCancelled`) are no-ops when R2 is configured. All asset keys stay in the existing `siteId/...` format regardless of backend, so upload, download, and delete all go through the same authorization checks as today.
- **Static OTP bypass** (`DANGEROUSLY_SET_STATIC_OTP`): when set, login stores a fixed 6-character token hash and skips sending an OTP email entirely, so preview reviewers can log in without a Postman key. Named with `DANGEROUSLY_` to make misuse obvious; validated as exactly 6 characters and forbidden outside preview at boot.
- **SingPass bypass for preview** (`NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS`): when set in preview, hides the SingPass login option and suppresses SingPass-off side effects (e.g. login alert emails, email-on-publish). Forbidden outside preview at boot via `env.mjs` superRefine.
- **GitHub Action: Neon branch cleanup** (`.github/workflows/neon_cleanup.yml`): automatically deletes the `preview/<branch>` Neon branch when its PR is closed, preventing stale DB branches from accumulating. Guards against deleting `main`.
- **CodeBuild mocked** (`codebuild.service.ts`): `publishSite` returns early with a log in preview — no AWS calls made.
- **Cron jobs skipped** (`instrumentation.ts`): `initializeCronJobs()` is not called in preview, preventing pg-boss from holding connections on ephemeral Vercel functions.
- **Vercel config** (`vercel.json`): `installCommand` installs from workspace root and builds workspace packages; `buildCommand` runs a new `vercel-build` script that runs Prisma generate + migrations + seed + `next build`.

**Improvements**:

- **Lazy server-side DOM purify** (`src/lib/server-dom-purify.ts`): extracts `jsdom`/`isomorphic-dompurify` into a memoised lazy loader via `createRequire`, so the ESM-only transitive chain is not evaluated at module init. Fixes preview crashes on hot-path tRPC routes (e.g. `me.get`) that mount `assetRouter`; `sanitizeSvg` still runs synchronously on first use.
- **Node.js 24.x**: bumps `.nvmrc`, root `package.json` `engines`, and CI setup action to Node 24; aligns with Vercel's runtime and removes the temporary jsdom 27 pin.
- CSP `connect-src` conditionally includes the R2 endpoint (`https://<R2_ACCOUNT_ID>.r2.cloudflarestorage.com`) only when R2 credentials are configured; `vercel.live` added to `frame-src`/`script-src` for preview only.
- `env.mjs`: `"preview"` added to `NEXT_PUBLIC_APP_ENV` enum; `R2_ACCOUNT_ID`/`R2_ACCESS_KEY_ID`/`R2_SECRET_ACCESS_KEY` and `DANGEROUSLY_SET_STATIC_OTP`/`NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS` added as optional server/client vars with cross-field guards.
- `packages/logging`: `"preview"` added to the env schema so preview logs are accepted by the logging package.
- SingPass visibility now driven by a GrowthBook feature flag (`IS_SINGPASS_ENABLED_FEATURE_KEY`) combined with the preview bypass, instead of a hardcoded flag (`useIsSingpassEnabled`, `src/lib/growthbook.ts`).
- Seed updated with team users and a Sample Site with pages matching the footer and navbar links; seed is skipped if Sample Site already exists (idempotent).
- SingPass client lazy-initialised to avoid DNS lookup failures in preview where SingPass is not configured.
- Neon unpooled connection used for Prisma migrations to avoid pooler incompatibility.

**Bug Fixes**:

- Lockfile regenerated cleanly after merge (no hand-resolved `pnpm-lock.yaml` conflicts).

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Open a PR → confirm Vercel creates a preview deployment with `NEXT_PUBLIC_APP_ENV=preview` (or auto-derived from `NEXT_PUBLIC_VERCEL_ENV`) and a Neon branch DB
- [ ] Check build logs: `prisma generate`, `migrate deploy`, `migrate:custom`, `db:seed`, and `next build` all ran successfully
- [ ] Visit the preview URL; confirm the app boots with no pg-boss/cron errors in function logs
- [ ] Load any authenticated page (e.g. dashboard) → confirm tRPC `me.get` succeeds with no `ERR_REQUIRE_ESM` / jsdom crash in function logs
- [ ] Log in using the static OTP set via `DANGEROUSLY_SET_STATIC_OTP` (no email should be sent; enter the fixed token directly)
- [ ] Confirm SingPass login is hidden when `NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS=true`
- [ ] Create or edit a page and confirm DB writes succeed (Neon branch)
- [ ] Upload a raster image asset → confirm it lands in the Cloudflare R2 bucket and renders back correctly
- [ ] Upload an SVG asset → confirm it sanitizes, uploads to R2, and renders without server errors
- [ ] Delete an uploaded asset → confirm it succeeds (no-op tagging call, no error)
- [ ] Trigger a publish → confirm the UI doesn't error and function logs show the CodeBuild no-op log line
- [ ] Close the PR → confirm the `neon_cleanup` GitHub Action deletes the `preview/<branch>` Neon branch
- [ ] In a non-preview env (`NEXT_PUBLIC_APP_ENV=staging`), confirm SingPass is enabled, S3 upload/delete/publish-lock paths are active, and cron jobs initialise

**New scripts**:

- `vercel-build` (apps/studio): monorepo-aware build pipeline for Vercel — components build, Prisma migrate/seed, preview-tw build, then `next build`

**New dependencies**:

- None

**New dev dependencies**:

- None

This PR adds per-PR Vercel preview deployments to the Isomer Studio CMS. It introduces a new `"preview"` environment mode gated on `NEXT_PUBLIC_APP_ENV === "preview"` (auto-derived from `NEXT_PUBLIC_VERCEL_ENV` when not explicitly set), and wires up a supporting set of preview-only behaviours.

- **Infrastructure**: Cloudflare R2 as the S3-compatible asset backend, a static OTP bypass (`DANGEROUSLY_SET_STATIC_OTP`), a SingPass skip flag (`NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS`), CodeBuild and cron-job no-ops, Neon branch auto-cleanup on PR close, and a `vercel-build` script that runs Prisma migrations + seed + `next build`.
- **ESM fix**: `jsdom`/`isomorphic-dompurify` moved behind a lazy `createRequire` loader (`server-dom-purify.ts`) to prevent `ERR_REQUIRE_ESM` crashes on every tRPC request caused by Turbopack's external module loader.
- **Upload path refactor**: `asset.router.ts` now returns `{ fileKey, uploadConfig }` instead of flat fields; a thin `storage/client.ts` shim (`performUpload`) dispatches the actual HTTP PUT, allowing future storage backends to plug in here.
</details>

Safe to merge. All preview-specific bypasses are structurally gated by env.mjs superRefine rules that forbid them outside preview at boot time, and the ESM fix is additive with no impact on existing code paths.

The env validation guards (superRefine) prevent DANGEROUSLY_* vars from being set in staging or production. The R2 backend selection is driven by credential presence rather than environment name, making it independent of mislabelled envs. The lazy-loading fix for jsdom/DOMPurify is isolated to a new module with no changes to the existing sanitizeSvg contract. Cron skipping and CodeBuild no-ops are cleanly gated. The only finding is a stale field in a test mock with no runtime impact.

No files require special attention. The MSW mock in apps/studio/tests/msw/handlers/asset.ts has a minor stale field worth cleaning up but does not affect any test assertions or runtime behaviour.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/env.mjs | Adds "preview" to the APP_ENV enum, R2 credentials, and DANGEROUSLY_* env vars with cross-field superRefine guards enforcing preview-only use. |
| apps/studio/src/lib/s3.ts | Conditionally points the S3Client at R2 when credentials are present; adds no-op guards on tag-based lifecycle operations for R2. |
| apps/studio/src/lib/server-dom-purify.ts | New lazy-loader for jsdom/isomorphic-dompurify via createRequire; prevents ERR_REQUIRE_ESM on Vercel/Turbopack at module init time. |
| apps/studio/src/server/modules/asset/asset.service.ts | Removes eager jsdom/DOMPurify imports; introduces UploadConfig interface; routes all S3 calls through the shared bucket const; sanitizeSvg now lazily calls getServerDomPurify(). |
| apps/studio/src/server/modules/asset/asset.router.ts | getPresignedPutUrl route now returns { fileKey, uploadConfig } instead of flat fields. |
| apps/studio/src/server/modules/auth/email/email.router.ts | Adds static OTP bypass and gates the login-alert email on getIsSingpassDisabledInNonPreview to avoid spurious alert emails in preview. |
| apps/studio/src/server/modules/auth/singpass/singpass.service.ts | SingPass client lazy-initialised; removes top-level await Issuer.discover() to prevent DNS lookup failures in preview. |
| apps/studio/src/instrumentation.ts | Imports env via env.mjs (so the NEXT_PUBLIC_VERCEL_ENV fallback fires) and skips cron initialisation in preview. |
| apps/studio/src/lib/storage/client.ts | New thin client-side shim that dispatches an S3 presigned PUT via handleAssetUpload and returns the canonical path. |
| .github/workflows/neon_cleanup.yml | New workflow that deletes the preview/<branch> Neon branch on PR close; action pinned to a specific commit hash; guards against deleting main. |
| apps/studio/prisma/seed.ts | Seed replaced with idempotent version using generic role-based emails; adds createPage/createFolder helpers. |
| apps/studio/src/lib/growthbook.ts | Adds getIsSingpassDisabledInNonPreview helper and short-circuits getIsSingpassEnabled to false when NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS is set. |
| apps/studio/tests/msw/handlers/asset.ts | Mock updated to return the new { fileKey, uploadConfig } shape, but uploadConfig includes a stale provider: "s3" field absent from the UploadConfig interface. |

</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
flowchart TD
    A[Request arrives] --> B{NEXT_RUNTIME === nodejs?}
    B -- No --> END1[Skip]
    B -- Yes --> C[initTracer]
    C --> D{APP_ENV === preview?}
    D -- Yes --> END2[Skip cron jobs]
    D -- No --> E[initializeCronJobs]

    subgraph Upload["Asset Upload Flow"]
        F[Client calls getPresignedPutUrl tRPC] --> G["asset.router returns fileKey + uploadConfig"]
        G --> H[performUpload sends HTTP PUT]
        H --> I{isR2Configured?}
        I -- Yes --> J[PUT to R2 via presigned URL]
        I -- No --> K[PUT to AWS S3 via presigned URL]
    end

    subgraph S3Init["S3 Client Init"]
        M[s3.ts module loads] --> N{R2 credentials all present?}
        N -- Yes --> O[S3Client pointed at R2 endpoint]
        N -- No --> P[S3Client pointed at AWS S3]
    end

    subgraph Auth["Auth - Preview"]
        Q[Email login] --> R{DANGEROUSLY_SET_STATIC_OTP set?}
        R -- Yes --> S[Use static token, skip sendMail]
        R -- No --> T["Generate OTP + sendMail"]
        U[OTP verify success] --> V{getIsSingpassDisabledInNonPreview?}
        V -- Yes --> W[sendLoginAlertEmail]
        V -- No --> X[Skip alert email]
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[Request arrives] --> B{NEXT_RUNTIME === nodejs?}
    B -- No --> END1[Skip]
    B -- Yes --> C[initTracer]
    C --> D{APP_ENV === preview?}
    D -- Yes --> END2[Skip cron jobs]
    D -- No --> E[initializeCronJobs]

    subgraph Upload["Asset Upload Flow"]
        F[Client calls getPresignedPutUrl tRPC] --> G["asset.router returns fileKey + uploadConfig"]
        G --> H[performUpload sends HTTP PUT]
        H --> I{isR2Configured?}
        I -- Yes --> J[PUT to R2 via presigned URL]
        I -- No --> K[PUT to AWS S3 via presigned URL]
    end

    subgraph S3Init["S3 Client Init"]
        M[s3.ts module loads] --> N{R2 credentials all present?}
        N -- Yes --> O[S3Client pointed at R2 endpoint]
        N -- No --> P[S3Client pointed at AWS S3]
    end

    subgraph Auth["Auth - Preview"]
        Q[Email login] --> R{DANGEROUSLY_SET_STATIC_OTP set?}
        R -- Yes --> S[Use static token, skip sendMail]
        R -- No --> T["Generate OTP + sendMail"]
        U[OTP verify success] --> V{getIsSingpassDisabledInNonPreview?}
        V -- Yes --> W[sendLoginAlertEmail]
        V -- No --> X[Skip alert email]
    end
```

</a>
</details>

<sub>Reviews (13): Last reviewed commit: ["Merge branch &#39;main&#39; of https://github.co..."](https://github.com/opengovsg/isomer/commit/3d27ff2b8133d43a0811668c823fadb11f8e84ac) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39746793)</sub>


adriangohjw:
@claude /review-pr

adriangohjw:
@claude /review-pr

adriangohjw:
@claude review PR pls

adriangohjw:
@claude nice, fix your HIGH and MEDIUM findings

seaerchin:
@greptileai review this, focus more on security issues

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -49,8 +52,9 @@ export const emailSessionRouter = router({
       // TODO: rate limit this endpoint also
       const expires = new Date(Date.now() + env.OTP_EXPIRY * 1000)
       const expiryMinutes = Math.floor(env.OTP_EXPIRY / 60)
-      const token = createVfnToken()
-      const otpPrefix = createVfnPrefix()
+      const staticOtp = env.DANGEROUSLY_SET_STATIC_OTP
```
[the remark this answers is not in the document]
adriangohjw: yeah its fine, thats the point of `DANGEROUSLY_SET_STATIC_OTP` - and we only use it in vercel so its fine

apps/studio/src/lib/s3.ts
```diff
@@ -74,6 +74,7 @@ export const deleteFile = async ({
   Key,
   Bucket,
 }: Pick<PutObjectTaggingCommandInput, "Key" | "Bucket">) => {
+  if (env.NEXT_PUBLIC_APP_ENV === "preview") return
```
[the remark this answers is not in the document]
adriangohjw: will add comment in incoming push

apps/studio/prisma/seed.ts
```diff
@@ -6,73 +6,264 @@
...
+    .where("id", "=", resourceId)
+    .execute()
+
+  return resourceId
+}
+
+const createFolder = async ({
+  permalink,
+  title,
+  siteId,
```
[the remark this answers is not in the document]
adriangohjw: fixed in incoming push

apps/studio/src/lib/storage/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    try {
+      const jsonResponse = await handleUpload({
+        body: req.body as HandleUploadBody,
+        request: req,
+        onBeforeGenerateToken: () =>
+          Promise.resolve({
+            allowedContentTypes,
+            maximumSizeInBytes: 50 * 1024 * 1024,
+            allowOverwrite: true,
+          }),
```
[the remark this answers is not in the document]
adriangohjw: fixed in incoming push

apps/studio/src/env.mjs
```diff
@@ -25,7 +25,9 @@ const client = z
       "vapt",
       "test",
```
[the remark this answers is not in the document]
adriangohjw: fixed

apps/studio/src/instrumentation.ts
```diff
@@ -9,20 +9,20 @@ export async function register() {
...
     initTracer({ service: process.env.DD_SERVICE ?? "isomer-next" })
 
-    // Import only if runtime is nodejs. This avoids running it on the browser, build time etc.
-    const { initializeCronJobs, stopCronJobs } = await import("~/server/cron")
-
-    // Initialize cron jobs when server starts
-    await initializeCronJobs()
+    // Skip cron on Vercel preview (serverless — no persistent process)
+    // oxlint-disable-next-line node/no-process-env
+    if (process.env.NEXT_PUBLIC_APP_ENV !== "preview") {
```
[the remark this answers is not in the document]
adriangohjw: fixed

apps/studio/prisma/seed.ts
seaerchin, line 1: question: how did we derive these functions?

apps/studio/prisma/seed.ts
```diff
@@ -6,62 +6,230 @@
...
+  db,
+  jsonb,
+} from "../src/server/modules/database"
 import { createSite } from "../src/server/modules/site/site.service"
 import { addUsersToSite } from "./scripts/addUsersToSite"
 
-const EDITOR_USER = "editor"
-const PUBLISHER_USER = "publisher"
+// Mirrors the publish flow in site.service.ts: blob → resource → version → link version back.
+const createPage = async ({
```
seaerchin, line 21: comment: i honestly think these should be proper service functions rather than inlined into a seed script haha

apps/studio/prisma/seed.ts
```diff
@@ -6,62 +6,230 @@
...
+    .values({ resourceId, blobId, publishedBy: userId, versionNum: 1 })
+    .returning("id")
+    .executeTakeFirstOrThrow()
+
+  await db
+    .updateTable("Resource")
+    .set({ draftBlobId: null, publishedVersionId: versionId })
+    .where("id", "=", resourceId)
+    .execute()
+
```
seaerchin, line 83: comment: we shuold also seed w/ unpublished pages

apps/studio/prisma/seed.ts
```diff
@@ -6,62 +6,230 @@
...
+  return resourceId
+}
+
+const createFolder = async ({
```
seaerchin, line 87: comment: we should also create a collectoin

apps/studio/prisma/seed.ts
seaerchin, line 1: comment: non-blocker for this PR, but our seed should create a minimal working reproduction of all our features on prod

1. collections (links + articles) 
2. all types of pages at root (db/article/content)
3. folder w/ those pages
4. navbar/footer etc etc

not important to do now, but i think it's something we should do going forward to ensure our preview is as useful as possible

apps/studio/src/hooks/useIsSingpassEnabled.ts
```diff
@@ -1,11 +1,23 @@
...
 } from "~/lib/growthbook"
 
-export const useIsSingpassEnabled = () =>
-  useFeatureValue<boolean>(
+export const useIsSingpassEnabled = () => {
+  const featureValue = useFeatureValue<boolean>(
     IS_SINGPASS_ENABLED_FEATURE_KEY,
     IS_SINGPASS_ENABLED_FEATURE_KEY_FALLBACK_VALUE,
   )
+  const skipSingpass = env.NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS
```
seaerchin, line 13: comment: vercel exposes an env var called [vercel](https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL) that is set to `1` in vercel deploys.

maybe we should anchor on that instead? clearer and it's automatically true in vercel preview deploys whilst being false everywhere else

apps/studio/src/hooks/useUploadAssetMutation.ts
seaerchin, line 1: comment: need to remember to dup for svg also - seems liek this might cause a regression

apps/studio/src/lib/storage/client.ts
```diff
@@ -0,0 +1,28 @@
+import { upload as vercelUpload } from "@vercel/blob/client"
+import { handleAssetUpload } from "~/hooks/handleAssetUpload"
+
+import type { UploadConfig } from "."
+
+export const performUpload = async (
+  file: File,
+  fileKey: string,
+  config: UploadConfig,
```
seaerchin, line 7: ~~question: this is given by the frontend but i don't think the frontend should dictate this. the current flow is frontend queries backend for the upload config, then uses the config to `performUpload` but the only concrete difference between them is the `provider` and `clientPayload` (for vercel) which we use to check the filekey

wondering if the backend should dictate this flow rather than frontend? probably no security issues but just seems odd to have fe request for a config then have the backend trust said config~~

on second read, i think it feels fine cos the user can't touch this anw? the frontend asks backend for content disp, content type and extra stuff then uploads to s3/vercel direct because there's no backend intermediary so this is required.

feel free to ignore this lol

apps/studio/src/lib/server-dom-purify.ts
seaerchin, line 1: skipped

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -49,8 +52,9 @@ export const emailSessionRouter = router({
...
+      const token = staticOtp ?? createVfnToken()
+      const otpPrefix = staticOtp != null ? "OTP" : createVfnPrefix()
```
seaerchin: comment: prefer `!==`

issue: also, it appears that it's typed as `optional` which might be `undefined` rather than `null`? in my IDE it's `string | undefined` which means this will be true

apps/studio/src/server/modules/auth/email/email.router.ts
```diff
@@ -59,6 +63,8 @@ export const emailSessionRouter = router({
 
       // May have one of them fail,
       // so users may get an email but not have the token saved, but that should be fine.
+      const isStaticOtp = staticOtp != null
```
seaerchin: same here

apps/studio/src/env.mjs
```diff
@@ -25,7 +25,15 @@ const client = z
...
       "uat",
+      "preview",
     ]),
+    NEXT_PUBLIC_STORAGE_PROVIDER: z.enum(["vercel-blob", "s3"]).default("s3"),
+    // WARNING: Setting this bypasses SingPass login entirely. For preview
+    // environments only — never set in staging or production.
+    NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS: z
+      .stringbool()
+      .optional()
+      .default(false),
```
seaerchin, line 43: comment: do you think there's value in shifting these out to a `preview` const which we then `merge` with the `server` schema? 

my thoughts are that these are special values that are conditionally changed based on whether the preview env is set so we should signal this out as much as possible

apps/studio/src/env.mjs
```diff
@@ -64,13 +74,66 @@ const server = z
...
     ALGOLIA_APP_ID: z.string(),
     ALGOLIA_API_KEY: z.string(),
     ALGOLIA_INDEX_NAME: z.string(),
   })
   .merge(s3Schema)
   .merge(singpassSchema)
   .merge(client)
+  .superRefine((data, ctx) => {
+    if (data.NEXT_PUBLIC_STORAGE_PROVIDER === "s3") {
+      if (!data.NEXT_PUBLIC_S3_ASSETS_DOMAIN_NAME) {
```
seaerchin: nit: combine nested `if`

apps/studio/src/env.mjs
```diff
@@ -64,13 +74,66 @@ const server = z
...
+      }
+      if (!data.NEXT_PUBLIC_S3_ASSETS_BUCKET_NAME) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: "Required for non-preview environments",
+          path: ["NEXT_PUBLIC_S3_ASSETS_BUCKET_NAME"],
+        })
+      }
+    }
+    if (data.NEXT_PUBLIC_STORAGE_PROVIDER === "vercel-blob") {
```
seaerchin: nit: combine nested `if`

apps/studio/src/env.mjs
```diff
@@ -64,13 +74,66 @@ const server = z
...
+            "DANGEROUSLY_SET_STATIC_OTP may only be set in preview environments",
+          path: ["DANGEROUSLY_SET_STATIC_OTP"],
+        })
+      }
+    }
+    // Skipping SingPass bypasses the primary authentication mechanism, so
+    // structurally forbid it outside preview — a boot-time failure, not an
+    // operational assumption.
+    if (data.NEXT_PUBLIC_APP_ENV !== "preview") {
+      if (data.NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS) {
```
seaerchin: nit: combine nested `if`

apps/studio/src/server/modules/auth/singpass/singpass.service.ts
```diff
@@ -9,26 +10,46 @@ import {
 } from "./singpass.constants"
 import { extractUuid } from "./singpass.utils"
 
-const singpassIssuer = await Issuer.discover(env.SINGPASS_ISSUER_ENDPOINT)
-const singpassClient = new singpassIssuer.Client(
```
seaerchin, line 13: @dcshzj not sure if this will cause issues in prod, just a double check please! i know signpass requires signing keys and the like so idk if lazy initialising only on first call will work

dcshzj: Yup this is fine, it was originally raised as an issue that we depend on `SINGPASS_ISSUER_ENDPOINT` even if Singpass is disabled, which defeats the purpose since the `SINGPASS_ISSUER_ENDPOINT` might not be able to respond anyway

apps/studio/prisma/seed.ts
adriangohjw: i briefly describe what is needed and claude added it haha

adriangohjw: yeah makes alot of sense

apps/studio/src/hooks/useIsSingpassEnabled.ts
adriangohjw: oooo sounds good, TIL

apps/studio/src/env.mjs
adriangohjw: yeah i think current ENV management is a bit messy - should move to t3-env https://github.com/opengovsg/isomer/pull/2212

apps/studio/src/env.mjs
adriangohjw: Done — flattened the two preview-guard nested `if`s in `env.mjs` (`DANGEROUSLY_SET_STATIC_OTP` and `NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS`) into single `&&` conditions.

Left the S3 / vercel-blob provider checks nested — those are storage-provider guards, not preview guards.

apps/studio/src/hooks/useUploadAssetMutation.ts
adriangohjw: Fixed — SVG uploads now go through `assetStorage` instead of hardcoded S3:

- `putFileDirect` delegates to `assetStorage.putFile()` (S3 → `putObjectDirect`, Vercel Blob → server-side `@vercel/blob` `put`)
- `uploadSvg` returns `{ fileKey, url? }` so preview gets the blob URL
- `useUploadAssetMutation` uses `url ?? \`/${fileKey}\`` for the SVG path, same pattern as non-SVG via `performUpload`

Asset router tests pass (25/25).

apps/studio/src/hooks/useIsSingpassEnabled.ts
adriangohjw: Thanks for the TIL! Sticking with the current approach though:

- `VERCEL=1` is set on **all** Vercel deploys (including production), not just preview
- For preview-only behaviour we already derive `NEXT_PUBLIC_APP_ENV === "preview"` from `NEXT_PUBLIC_VERCEL_ENV` in `env.mjs`
- `NEXT_PUBLIC_DANGEROUSLY_SKIP_SINGPASS` is an explicit opt-in, gated in `superRefine` so it can't be set outside preview — more auditable than auto-detecting from Vercel system vars

apps/studio/src/server/modules/auth/email/email.router.ts
adriangohjw: Keeping `!= null` here — `DANGEROUSLY_SET_STATIC_OTP` is `string | undefined` (Zod `.optional()`), not `string | null`:

- `staticOtp != null` → `false` when `undefined` ✓
- `staticOtp !== null` → `true` when `undefined` ✗ (would wrongly enter static-OTP mode)

`!= null` is the right check for "has a value" when the type is optional/undefined.

apps/studio/src/server/modules/auth/email/email.router.ts
adriangohjw: Same reasoning as above — keeping `!= null` for the optional `string | undefined` type.

apps/studio/src/hooks/useIsSingpassEnabled.ts
adriangohjw: + i think letting users decide means it's more explicit as opposed to it being implicit

apps/studio/src/env.mjs
adriangohjw: fixed

apps/studio/src/env.mjs
adriangohjw: fixed

adriangohjw:
@seaerchin tldr
- fixed all the above (nits + S3/vercel stuff)
- seed enhancement -> i will leave it for future enhancement to not bloat this PR
- will stick to using the dedicated env var like `DANGEROUSLY_SET_STATIC_OTP` and environment being `preview` etc. instead of `VERCEL` -> iMO more explicit and not locked in to vercel

apps/studio/src/hooks/useIsSingpassEnabled.ts
seaerchin: i'm not too opinionated here, jsut that having a preview deploy of production still gated behind VERCEL=1 makes sense?

the fact that we're exposing all these env vars seems extra because all we're really doing is saying if we're on vercel, do x instead of y so i thought linking to a single automatic env var rather than having alot makes sense. conversely, if you feel like we might have a few combinations of these env vars (ie, not all true or all false ubt some mix thereof) then i'm happy to go w/ this too.

apps/studio/src/server/modules/auth/email/email.router.ts
seaerchin: comment: i think this is pretty sus and i'd prefer it to be more explicit especially cos this is security related (statically putting an OTP depending on env var).

could we do !== null and !== undefined?

seaerchin:
as discussed, i think there's value in swapping to a s3 compat provider (such as r2). not a blocker for this PR; however, i do feel like the env var should **not** use coercing comparators (!=) and we should be explicit on the comparison cos it's an auth flow.

happy to approve once done sir

adriangohjw:
@seaerchin done — swapped both `staticOtp != null` checks in `email.router.ts` to `staticOtp !== undefined` (it's `.optional()`, never `null`, so this is the correct explicit form). See cd2958af0.

R2 changes made

apps/studio/src/lib/storage/client.ts
adriangohjw: oh this is mostly refactored as the single source of truth. nothing new introduced here

seaerchin: ya ignore me, i tot might have issues but i re-read and i think it's fine.

seaerchin (approved):
lgtm - i think this PR is good enough to go in; nothing very big on my end, i think maybe the r2 explicit return breaks the abstraction but i think not worth fussing over ba

PR #2308, opened 2026-06-05
fix: use timing-safe webhook api key comparison

### LINKED ISSUE #2232: (deepsec) [BUG] Non-constant-time comparison of webhook API key
Reference: https://www.notion.so/opengov/isomer-other-timing-comparison-981f0de074-36e77dbba78881feab77d0af3cbed960

**File:** [`[apps/studio/src/server/trpc.ts](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/server/trpc.ts#L193)`](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/server/trpc.ts#L193) (lines 193)
**Project:** isomer
**Severity:** BUG • **Confidence:** low • **Slug:** `other-timing-comparison`

## Finding

L193 compares the inbound `apiKey` against `env.STUDIO_SSM_WEBHOOK_API_KEY` using `!==`, which short-circuits on the first differing byte. In theory this allows a timing oracle to recover the key byte-by-byte; in practice the attack is hard to mount over the network due to jitter, JIT, and TLS framing. Still, the comparison is a known anti-pattern for secret matching and trivial to fix.

## Recommendation

Use `crypto.timingSafeEqual(Buffer.from(apiKey), Buffer.from(env.STUDIO_SSM_WEBHOOK_API_KEY))` after first confirming the lengths match, to avoid leaking length and to do a constant-time compare.

## Revalidation

**Verdict:** true-positive

Verified at L193: `apiKey !== env.STUDIO_SSM_WEBHOOK_API_KEY` is the standard `!==` operator which short-circuits on the first differing byte. This is the textbook pattern for a timing-side-channel on secret comparison. Practical exploitability over a network is poor: the secret is a single-shot API key (not bytewise-recoverable like an HMAC for many calls), TLS framing and Node.js event-loop jitter typically dwarf the per-byte timing difference, and TRPCError throwing adds further noise. So I agree with the finding’s low confidence and would adjust severity to BUG / minor — fix is trivial via `crypto.timingSafeEqual` with a length check, but real-world risk is small.

### LINKED ISSUE #2233: (deepsec) [BUG] API key compared with non-constant-time string equality
Reference: https://www.notion.so/opengov/isomer-other-timing-attack-a80bcf51b7-36e77dbba7888182b785dbbf6119d3a5

**File:** [`[apps/studio/src/server/modules/webhook/webhook.router.ts](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/server/modules/webhook/webhook.router.ts#L7)`](https://github.com/opengovsg/isomer/blob/main/apps/studio/src/server/modules/webhook/webhook.router.ts#L7) (lines 7)
**Project:** isomer
**Severity:** BUG • **Confidence:** medium • **Slug:** `other-timing-attack`

## Finding

The webhook API key check in `webhookMiddleware` (apps/studio/src/server/trpc.ts line 193) uses `apiKey !== env.STUDIO_SSM_WEBHOOK_API_KEY`, which is a non-constant-time comparison. JavaScript’s `!==` short-circuits on the first differing byte, making the operation theoretically susceptible to timing-based key recovery. While network jitter usually obscures small timing deltas, the lack of rate limiting on this endpoint (see prior finding) makes statistical attacks more practical because an attacker can issue effectively unlimited probes. The router file is where this check is enforced (via `webhookProcedure`), so flagging here.

## Recommendation

Use `crypto.timingSafeEqual(Buffer.from(apiKey), Buffer.from(env.STUDIO_SSM_WEBHOOK_API_KEY))` after first verifying both strings are equal length (compare lengths separately, with a fixed-length dummy compare if mismatched). Or use a vetted library like `tsscmp`.

## Revalidation

**Verdict:** true-positive

Verified in trpc.ts line 193: `if (!apiKey || apiKey !== env.STUDIO_SSM_WEBHOOK_API_KEY)`. JavaScript’s `!==` short-circuits at the first differing byte, so it is theoretically vulnerable to timing analysis. Practically, remote timing attacks against in-memory string comparison are extremely hard to mount over the network — jitter dominates and modern V8 may not even produce a stable signal. However, two factors make this worth fixing: (1) the endpoint has no rate limiting (prior finding), so an attacker could issue millions of probes for statistical analysis; (2) the trivial fix (`crypto.timingSafeEqual` after a length check) is industry-standard and costs nothing. Severity MEDIUM is generous; I’d argue BUG. Finding is technically correct.

juneha1120:
## Problem

The webhook middleware compares the inbound `x-api-key` header against `STUDIO_SSM_WEBHOOK_API_KEY` using normal string equality. This can short-circuit on the first differing byte, which is a known timing side-channel pattern for secret comparisons.

Closes #2232
Closes #2233

## Solution

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- Replaces the webhook API key string equality check with `crypto.timingSafeEqual`.
- Rejects missing, repeated, or length-mismatched API key headers before comparing equal-length buffers.
- Adds regression coverage for an incorrect API key with the expected length.

## Before & After Screenshots

N/A, backend-only change.

## Tests

- [ ] Send a webhook request with the configured `x-api-key` and verify it succeeds.
- [ ] Send a webhook request with a missing `x-api-key` and verify it returns 401.
- [ ] Send a webhook request with an incorrect same-length `x-api-key` and verify it returns 401.

**New scripts**:

N/A

**New dependencies**:

N/A

**New dev dependencies**:

N/A

### DIFF
--- a/apps/studio/src/schemas/__tests__/webhook.test.ts
+++ b/apps/studio/src/schemas/__tests__/webhook.test.ts
@@ -10,6 +10,13 @@ import { WEBHOOK_X_API_KEY_HEADER } from "~/server/trpc"
 
 import type { codeBuildWebhookSchema } from "../webhook"
 
+const { WEBHOOK_API_KEY, INVALID_WEBHOOK_API_KEY_WITH_EXPECTED_LENGTH } =
+  vi.hoisted(() => ({
+    WEBHOOK_API_KEY: "00000000-0000-4000-8000-000000000000",
+    INVALID_WEBHOOK_API_KEY_WITH_EXPECTED_LENGTH:
+      "11111111-1111-4111-8111-111111111111",
+  }))
+
 vi.mock("~/env.mjs", async () => {
   // Import the real module first to get all default values
   const actual = await vi.importActual<{ env: typeof env }>("~/env.mjs")
@@ -17,7 +24,7 @@ vi.mock("~/env.mjs", async () => {
     env: {
       ...actual.env,
       // override some env variables for testing
-      STUDIO_SSM_WEBHOOK_API_KEY: "test-webhook-api-key",
+      STUDIO_SSM_WEBHOOK_API_KEY: WEBHOOK_API_KEY,
       GROWTHBOOK_CLIENT_KEY: "test-growthbook-client-key",
     },
   }
@@ -31,7 +38,7 @@ vi.mock("~/features/mail/service", () => ({
 
 const createMockRequest = ({
   arn,
-  apiKey = "test-webhook-api-key",
+  apiKey = WEBHOOK_API_KEY,
 }: {
   arn: string
   apiKey?: string | null
@@ -89,6 +96,26 @@ describe("webhook", () => {
       const { req, res } = createMockRequest({
         arn: "build/test-id",
         apiKey: "wrong-api-key",
+      })
+
+      // Act
+      await handler(req, res)
+
+      // Assert
+      expect(res.statusCode).toBe(401)
+    })
+    it("providing an incorrect API key with the expected length causes a 401", async () => {
+      // Arrange
+      const user = await setupUser(createTestUser())
+      await setupCodeBuildJob({
+        userId: user.id,
+        arn: "build/test-id",
+        startedAt: new Date(),
+        isScheduled: true,
+      })
+      const { req, res } = createMockRequest({
+        arn: "build/test-id",
+        apiKey: INVALID_WEBHOOK_API_KEY_WITH_EXPECTED_LENGTH,
       })
 
       // Act
--- a/apps/studio/src/server/trpc.ts
+++ b/apps/studio/src/server/trpc.ts
@@ -9,6 +9,7 @@
  */
 
 import { initTRPC, TRPCError } from "@trpc/server"
+import { timingSafeEqual } from "node:crypto"
 import superjson from "superjson"
 import { ZodError } from "zod"
 import { APP_VERSION_HEADER_KEY } from "~/constants/version"
@@ -179,6 +180,17 @@ const authMiddleware = t.middleware(async ({ next, ctx }) => {
 
 export const WEBHOOK_X_API_KEY_HEADER = "x-api-key"
 
+const isValidWebhookApiKey = (
+  apiKey: string | string[] | undefined,
+  expectedApiKey: string,
+): boolean => {
+  return (
+    typeof apiKey === "string" &&
+    apiKey.length === expectedApiKey.length &&
+    timingSafeEqual(Buffer.from(apiKey), Buffer.from(expectedApiKey))
+  )
+}
+
 const webhookMiddleware = t.middleware(async ({ next, ctx }) => {
   const apiKey = ctx.req.headers[WEBHOOK_X_API_KEY_HEADER]
   // Ensure that the API key is set in the env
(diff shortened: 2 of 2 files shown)


seaerchin (approved):

apps/studio/src/schemas/__tests__/webhook.test.ts
```diff
@@ -97,6 +97,26 @@ describe("webhook", () => {
...
+      const user = await setupUser(createTestUser())
+      await setupCodeBuildJob({
+        userId: user.id,
+        arn: "build/test-id",
+        startedAt: new Date(),
+        isScheduled: true,
+      })
+      const { req, res } = createMockRequest({
+        arn: "build/test-id",
+        apiKey: "wrong-webhook-api-zz",
```
adriangohjw: can we extract `STUDIO_SSM_WEBHOOK_API_KEY: "test-webhook-api-key"` out into a constant and we use that constant to refence here to create an equal length but incorrect API key? reason being if somehow someone updates `test-webhook-api-key` in the future, this test will still fail, which is a false assertion

apps/studio/src/server/trpc.ts
```diff
@@ -179,6 +180,17 @@ const authMiddleware = t.middleware(async ({ next, ctx }) => {
...
+  apiKey: string | string[] | undefined,
+  expectedApiKey: string,
+) => {
+  if (typeof apiKey !== "string" || apiKey.length !== expectedApiKey.length) {
+    return false
+  }
+
+  return timingSafeEqual(Buffer.from(apiKey), Buffer.from(expectedApiKey))
+}
+
```
adriangohjw, line 193: preference: i find this to be more readable since there's less negation
```ts
const isValidWebhookApiKey = (
  apiKey: string | string[] | undefined,
  expectedApiKey: string,
): boolean => {
  return (
    typeof apiKey === "string" &&
    apiKey.length === expectedApiKey.length &&
    timingSafeEqual(
      Buffer.from(apiKey),
      Buffer.from(expectedApiKey),
    )
  )
}
```

adriangohjw (changes requested):

apps/studio/src/schemas/__tests__/webhook.test.ts
juneha1120: Thank you for the comment! addressed in https://github.com/opengovsg/isomer/pull/2308/commits/5ce65fc5dc18c15771cc0d8e427413297de1bcc0 by extracting the test webhook API key into a shared constant and deriving the same length invalid key from it. Used `vi.hoisted` because the constant is referenced inside a `vi.mock` factory.

apps/studio/src/server/trpc.ts
juneha1120: Addressed in https://github.com/opengovsg/isomer/pull/2308/commits/5ce65fc5dc18c15771cc0d8e427413297de1bcc0, thanks

adriangohjw (approved):

adriangohjw (commented):
Actually we can just use uuid for both, so we will know it's actually both the same length and different. Then can skip this whole section and it will be simpler

juneha1120:
> Actually we can just use uuid for both, so we will know it's actually both the same length and different. Then can skip this whole section and it will be simpler

Addressed in dc7a19f29. Updated the test to use two UUID-shaped constants so the valid and invalid API keys are visibly the same length and different, while keeping `vi.hoisted` for the `vi.mock` factory.

adriangohjw (approved):

PR #2321, opened 2026-06-06
fix: resolve dompurify hook leak

### LINKED ISSUE #2230: (deepsec) [BUG] DOMPurify hook is added on every sanitize call without removal
closes https://www.notion.so/opengov/isomer-other-dompurify-hook-leak-7e827d96fb-36e77dbba78881e4903ac38e6c3d7618

**File:** `[packages/components/src/templates/next/components/complex/Iframe/Iframe.tsx](https://github.com/opengovsg/isomer/blob/main/packages/components/src/templates/next/components/complex/Iframe/Iframe.tsx#L37)` (lines 37)
**Project:** isomer
**Severity:** BUG • **Confidence:** high • **Slug:** `other-dompurify-hook-leak`

## Finding

Every call to `getSanitizedIframeWithTitle` (used by this Iframe component on each render) calls `DOMPurify.addHook("beforeSanitizeElements", ...)` (see packages/components/src/utils/getSanitizedIframeWithTitle.ts line 6) without ever calling `DOMPurify.removeHook` or `removeAllHooks`. Hooks accumulate in the singleton DOMPurify instance for the lifetime of the process. On a server-rendered page this means every Iframe render adds another hook that fires on every subsequent DOMPurify.sanitize call across the whole codebase. Functionally the latest hook’s `title` wins (because hooks run in insertion order and overwrite the `title` attribute), so the result is correct, but the memory/perf cost grows unboundedly and the hook will also run for any other future DOMPurify usage that includes iframes, applying a stale `title` from a previous render to those iframes. Security impact is minimal (the hook only mutates attributes on iframe elements, doesn’t introduce XSS), so this is a BUG rather than a vulnerability.

## Recommendation

In getSanitizedIframeWithTitle, either (a) `DOMPurify.removeAllHooks()` before/after the sanitize call, (b) use a local DOMPurify instance via `DOMPurify(window)` scoped to this function, or (c) avoid the hook altogether by setting the title/etc. attributes on the returned DOM fragment AFTER sanitization rather than via a hook.

## Revalidation

**Verdict:** true-positive

Confirmed. `getSanitizedIframeWithTitle` at packages/components/src/utils/getSanitizedIframeWithTitle.ts line 6 calls `DOMPurify.addHook("beforeSanitizeElements", ...)` unconditionally on every invocation. I grepped the entire `packages/components` directory for `removeHook` and `removeAllHooks` and found zero matches, confirming the hook is never removed. The Iframe component (line 37) calls this helper on every render. Since `isomorphic-dompurify` exports a shared singleton, hook callbacks accumulate for the process lifetime in SSR. Functionally the latest-added hook overwrites the `title` attribute last (so the rendered iframe gets the correct title), but: (a) memory grows linearly with renders, (b) every DOMPurify.sanitize call anywhere in the codebase that processes an iframe element will run all accumulated hooks, applying a stale title from a previous render. Security impact is minimal — the hook only sets benign attributes on iframe elements and DOMPurify’s allowed-tags list already restricts content to `iframe` — so BUG severity is appropriate. The proposed fixes (removeAllHooks, scoped DOMPurify instance, or post-sanitization attribute setting) are all valid.

jiaxinnns:
## Problem

`getSanitizedIframeWithTitle.ts` registers a global `beforeSanitizeElements` hook on the shared `isomorphic-dompurify` singleton on every invocation. Because `Iframe.tsx` calls this utility on every render and hooks are never removed, it causes an unbounded memory leak and performance degradation during SSR.

Closes https://github.com/opengovsg/isomer/issues/2230

## Solution

Removed the global DOMPurify hook mechanism entirely. Instead, the function now mutates attributes (title, height, width, class) directly onto the returned HTMLIFrameElement fragment after sanitization completes. This ensures side-effect-free, localized attribute assignment, mirroring the existing post-sanitization pattern used for the loading attribute in Iframe.tsx.

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- Fixed a linear memory leak and unintended global side-effects caused by unremoved DOMPurify hooks during server-side rendering.

## Tests

- [ ] Run unit tests to verify both standard HTML sanitization logic and the regression test confirming that global hooks no longer leak into separate, unrelated DOMPurify operations.

### DIFF
--- a/packages/components/src/templates/next/components/complex/Iframe/Iframe.tsx
+++ b/packages/components/src/templates/next/components/complex/Iframe/Iframe.tsx
@@ -35,6 +35,11 @@ export const Iframe = ({
   shouldLazyLoad = true,
 }: IframeProps) => {
   const sanitizedIframe = getSanitizedIframeWithTitle(content, title)
+
+  if (!sanitizedIframe) {
+    return null
+  }
+
   sanitizedIframe.setAttribute("loading", shouldLazyLoad ? "lazy" : "eager")
 
   const iframeUrl = sanitizedIframe.getAttribute("src")
--- a/packages/components/src/utils/__tests__/getSanitizedIframeWithTitle.test.ts
+++ b/packages/components/src/utils/__tests__/getSanitizedIframeWithTitle.test.ts
@@ -1,0 +1,42 @@
+import DOMPurify from "isomorphic-dompurify"
+import { describe, expect, it } from "vitest"
+
+import { getSanitizedIframeWithTitle } from "../getSanitizedIframeWithTitle"
+
+describe("getSanitizedIframeWithTitle", () => {
+  it("sanitizes iframe and sets accessibility attributes", () => {
+    const iframe = getSanitizedIframeWithTitle(
+      '<iframe src="https://example.com" onload="alert(1)"></iframe>',
+      "My embed",
+    )
+
+    expect(iframe.tagName).toBe("IFRAME")
+    expect(iframe.getAttribute("src")).toBe("https://example.com")
+    expect(iframe.getAttribute("onload")).toBeNull()
+    expect(iframe.getAttribute("title")).toBe("My embed")
+    expect(iframe.getAttribute("height")).toBe("100%")
+    expect(iframe.getAttribute("width")).toBe("100%")
+    expect(iframe.getAttribute("class")).toBe(
+      "absolute top-0 left-0 bottom-0 right-0",
+    )
+  })
+
+  it("does not leak hooks across multiple calls", () => {
+    for (let i = 0; i < 10; i++) {
+      getSanitizedIframeWithTitle(
+        `<iframe src="https://example.com/${i}"></iframe>`,
+        `Title ${i}`,
+      )
+    }
+
+    const unrelated = DOMPurify.sanitize(
+      '<iframe src="https://other.com"></iframe>',
+      {
+        ALLOWED_TAGS: ["iframe"],
+        RETURN_DOM_FRAGMENT: true,
+      },
+    ).firstChild as HTMLIFrameElement
+
+    expect(unrelated.getAttribute("title")).toBeNull()
+  })
+})
--- a/packages/components/src/utils/__tests__/getSanitizedIframeWithTitle.test.ts
+++ b/packages/components/src/utils/__tests__/getSanitizedIframeWithTitle.test.ts
@@ -10,15 +10,33 @@ describe("getSanitizedIframeWithTitle", () => {
       "My embed",
     )
 
-    expect(iframe.tagName).toBe("IFRAME")
-    expect(iframe.getAttribute("src")).toBe("https://example.com")
-    expect(iframe.getAttribute("onload")).toBeNull()
-    expect(iframe.getAttribute("title")).toBe("My embed")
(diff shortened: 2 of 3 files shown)


adriangohjw (commented):
## Summary

- Removes the global `DOMPurify.addHook(\"beforeSanitizeElements\", ...)` registration that accumulated on every call to `getSanitizedIframeWithTitle`, fixing an unbounded hook leak during SSR. Since the `Iframe` component renders on every request and hooks are never removed from DOMPurify's singleton, the old code registered N hooks after N renders — each of which would fire on all subsequent unrelated `DOMPurify.sanitize` calls.
- Fix: sanitize first, then assign `title`, `height`, `width`, and `class` directly on the returned DOM node — a clean, side-effect-free approach consistent with how `Iframe.tsx:38` already sets `loading` post-sanitization.
- New test at `packages/components/src/utils/__tests__/getSanitizedIframeWithTitle.test.ts` directly reproduces the bug (10 calls → unrelated sanitize should not see leaked `title` attribute).

Overall the fix is correct and well-targeted. One pre-existing null safety issue is now surfaced more prominently and is worth fixing in this PR.

---

**issue (non-blocking):** `packages/components/src/utils/getSanitizedIframeWithTitle.ts:L23-L30` — The `RETURN_DOM_FRAGMENT: true` + `.firstChild as HTMLIFrameElement` cast is not null-safe. `firstChild` is `null` if `content` is empty or contains no `<iframe>` (e.g. user pasted plain text into a CMS field). The four new `setAttribute` calls added by this PR will throw `TypeError: Cannot set properties of null` in that case, crashing SSR for that page. The same issue existed for the downstream `Iframe.tsx:38` call, but this PR moves the crash surface earlier and makes it four times more likely to surface during testing.

Suggested fix — use `querySelector` on the fragment and guard:
```ts
const fragment = DOMPurify.sanitize(content, {
  ALLOWED_TAGS: ["iframe"],
  ALLOWED_ATTR: [...],
  RETURN_DOM_FRAGMENT: true,
})
const iframe = fragment.querySelector("iframe")
if (!iframe) return null   // or throw, or return a fallback element
```
Then update `Iframe.tsx` to handle a `null` return. This also sidesteps the leading-whitespace edge case where `firstChild` is a `TextNode`, not the `<iframe>`.

Confidence: high

---

**note:** `packages/components/src/utils/__tests__/getSanitizedIframeWithTitle.test.ts:L24-L41` — The leak test is well designed; running the old code against it would fail (the hook would have set `title` on the \"unrelated\" iframe). One gap: no test covers the empty/no-iframe input case, which would expose the `firstChild as HTMLIFrameElement` null crash described above. Worth adding if you address that issue.

Confidence: high

---

**praise:** The core approach — drop the global hook entirely and do post-sanitization attribute assignment — is the right call. Setting `title`, `height`, `width`, and `class` via `setAttribute` after sanitization is safe (none of these are injection vectors, and the values are either static strings or come from the content schema). The change is minimal and targeted exactly at the reported issue.

adriangohjw:
@jiaxinnns can we improve it based on the issue and note raised by the automated review above? thanks!

adriangohjw:
note: the first issue is also raised in https://github.com/opengovsg/isomer/issues/2359

adriangohjw (approved):

adriangohjw:
thank you!

PR #2322, opened 2026-06-07
fix: handle URL error for searchsg config

### LINKED ISSUE #1956: [DETAIL] Studio: Updating SearchSG site config can crash server when site URL lacks protocol
> [!NOTE]
> From: https://app.detail.dev/oss/opengovsg/isomer/bug_02ef1271-58f1-4c96-8a33-3cedc95226b6 

# Summary
- **Context**: The `updateSearchSGConfig` function in `searchsg.service.ts` is called when site configuration is updated for sites using SearchSG integration.
- **Bug**: `new URL(url)` is called without try-catch handling, and the function is invoked with `void` keyword, causing any URL parsing error to become an unhandled promise rejection.
- **Actual vs. expected**: If `site.config.url` is stored without a protocol (e.g., `"www.example.com"`), the function throws a `TypeError` that propagates as an unhandled promise rejection, crashing the Node.js process.
- **Impact**: In production environments, unhandled promise rejections crash Node.js (v15+ default) without graceful cleanup. The calling mutation completes "successfully" - no user-visible error.

# Code with Bug
```typescript
// apps/studio/src/server/modules/searchsg/searchsg.service.ts
export const updateSearchSGConfig = async (
  props: UpdateSearchSGConfigProps,
  searchsgClientId: string,
  url: string,
) => {
  if (!['production', 'staging'].includes(env.NEXT_PUBLIC_APP_ENV)) {
    return
  }

  const client = await requestSearchSGClient()
  const actualUrl = new URL(url)  // <-- BUG 🔴 throws TypeError for invalid URLs; no handling
  // ...
}
```

```typescript
// apps/studio/src/server/modules/site/site.router.ts
void updateSearchSGConfig(
  { name: siteName, _kind: "name" },
  updatedConfig.search.clientId,
  updatedConfig.url,  // <-- BUG 🔴 promise is discarded; rejection becomes unhandled
)
```

```typescript
// packages/components/src/types/site.ts
url: Type.String({
  title: "Base URL of the site",
  description: "The base URL of the site.",
  format: "hidden",  // <-- BUG 🔴 not a validation format (AJV accepts any string)
})
```

# Explanation
- `updateSiteConfig` persists `updatedConfig.url` to the DB without URI format validation because the TypeBox schema uses `format: "hidden"` (a UI hint), not an AJV-recognized format like `"uri"`.
- When `search.type === "searchSG"`, the mutation triggers a fire-and-forget call: `void updateSearchSGConfig(...)`. Inside `updateSearchSGConfig`, `new URL(url)` throws for values like `"www.example.com"` (no protocol).
- Because the promise is discarded and there is no `.catch(...)` or global `unhandledRejection` handler in the codebase, the rejection is unhandled and can terminate the Node.js process in production/staging.

## Codebase Inconsistency
The codebase explicitly treats URLs without a protocol as invalid:
```typescript
// packages/components/src/utils/__tests__/isUrl.test.ts
it("should return false for invalid strings", () => {
  const invalidUrls = [
    "",
    " ",
    "not a url",
    "example.com",
    "www.example.com",
    "just text",
    "123456",
  ]
  invalidUrls.forEach((url) => {
    expect(isUrl(url)).toBe(false)
  })
})
```
Yet the site config schema allows these values, enabling the crash path.

# Recommended Fix
Add try-catch around URL parsing and log+return on invalid input:
```typescript
let actualUrl: URL
try {
  actualUrl = new URL(url)
} catch (error) {
  logger.error(
    { error, url, ...props },
    `[ERROR] Invalid URL format for SearchSG config update`,
  )
  return
}
```
Also ensure the fire-and-forget call cannot surface as an unhandled rejection:
```typescript
void updateSearchSGConfig(...).catch((error) => logger.error({ error }, "[ERROR] updateSearchSGConfig failed"))
```

# History
This bug was introduced in commit a4eb6aa0. The commit created the entire `searchsg.service.ts` file as part of a feature to allow users to update site name with SearchSG integration. The `new URL(url)` call on line 87 was added without try-catch error handling, and when combined with the `void` keyword at call sites in `site.router.ts`, any invalid URL would cause an unhandled promise rejection that crashes the Node.js process.

jovnc:
## Problem

`updateSearchSGConfig` called `new URL(url)` without error handling, causing unhandled promise rejections when an invalid URL (e.g. `www.example.com`, empty string) was passed.

Closes #1956 

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- NIL

**Improvements**:

- NIL

**Bug Fixes**:

- Wrapped `new URL(url)` in a `try/catch` — logs the error and returns early on invalid URLs.
- Added `.catch()` handlers on both `updateSearchSGConfig()` call sites in `site.router.ts` to prevent unhandled rejections from bubbling up.
- Added unit tests for invalid URL inputs.

## Before & After Screenshots

**BEFORE**:

- NIL

**AFTER**:

- NIL

## Tests

- [ ] Set the site URL to an invalid value (e.g. `www.example.com`) and save. Confirm no 500 error and no unhandled rejection in server logs.
- [ ] Set a valid URL (e.g. `https://example.gov.sg`) and save. Confirm the SearchSG config update proceeds normally.
- [ ] Run unit tests: `pnpm test:unit -- src/server/modules/searchsg/__tests__/searchsg.service.test.ts`

**New scripts**:

- NIL

**New dependencies**:

- NIL

**New dev dependencies**:

- NIL

### DIFF
--- a/apps/studio/src/server/modules/searchsg/__tests__/searchsg.service.test.ts
+++ b/apps/studio/src/server/modules/searchsg/__tests__/searchsg.service.test.ts
@@ -74,4 +74,19 @@ describe("updateSearchSGConfig", () => {
       expect(mockWretch).toHaveBeenCalled()
     })
   })
+
+  describe("URL validation", () => {
+    it.each(["www.example.com", "example.com", "not a url", ""])(
+      "should resolve without rejecting and skip SearchSG config fetch for invalid URL %j",
+      async (invalidUrl) => {
+        // Act
+        await updateSearchSGConfig(PROPS, VALID_UUID, invalidUrl).catch(
+          () => {},
+        )
+
+        // Assert
+        expect(mockWretch).not.toHaveBeenCalled()
+      },
+    )
+  })
 })
--- a/apps/studio/src/server/modules/searchsg/searchsg.service.ts
+++ b/apps/studio/src/server/modules/searchsg/searchsg.service.ts
@@ -118,8 +118,17 @@ export const updateSearchSGConfig = async (
     return
   }
 
+  let actualUrl: URL
+  try {
+    actualUrl = new URL(url)
+  } catch (error) {
+    logger.error(
+      { error, url, ...props },
+      `[ERROR] Invalid URL format for SearchSG config update`,
+    )
+    return
+  }
   const client = await requestSearchSGClient()
-  const actualUrl = new URL(url)
   logger.info(
     { ...props, searchsgClientId, url: actualUrl.host },
     `[INFO] Updating searchsg config for ${url} with searchsg client id: ${searchsgClientId}`,
--- a/apps/studio/src/server/modules/site/site.router.ts
+++ b/apps/studio/src/server/modules/site/site.router.ts
@@ -173,6 +173,8 @@ export const siteRouter = router({
           { name: siteName, _kind: "name" },
           updatedConfig.search.clientId,
           updatedConfig.url,
+        ).catch((error) =>
+          ctx.logger.error({ error }, "[ERROR] updateSearchSGConfig failed"),
         )
 
       return updatedConfig
@@ -324,6 +326,8 @@ export const siteRouter = router({
           { colour: theme.colors.brand.canvas.inverse, _kind: "colour" },
           site.config.search.clientId,
           site.config.url,
+        ).catch((error) =>
+          ctx.logger.error({ error }, "[ERROR] updateSearchSGConfig failed"),
         )
       }
 


adriangohjw (commented):
## Summary

This PR fixes an unhandled promise rejection in `updateSearchSGConfig` when an invalid URL (e.g. `www.example.com`, empty string) was passed to `new URL(url)`. Two changes work together: the service now validates the URL early and returns cleanly on parse failure, and both call sites in the router gain `.catch()` handlers to prevent any remaining upstream errors from becoming unhandled rejections.

**Codeflow:**
- `searchsg.service.ts:121-130` — URL validation moved *before* `requestSearchSGClient()`, which is a nice fail-fast improvement; no auth token is wasted on an unparseable URL.
- `site.router.ts:172,323` — `void updateSearchSGConfig(...).catch(handler)` is the correct pattern here: `.catch()` is chained onto the Promise *before* `void` discards it, so rejections are properly caught.
- `searchsg.service.test.ts:79-93` — unit tests cover the four invalid URL shapes.

**Overall assessment:** Clean, focused fix. No security regressions. Two things worth addressing.

---

**suggestion:** `searchsg.service.test.ts:L82-L89` — The `.catch(() => {})` silently swallows *any* error, including bugs in the implementation. If `updateSearchSGConfig` were to throw for an unrelated reason, this test would still pass as long as `mockWretch` wasn't called.

```ts
// current
await updateSearchSGConfig(PROPS, VALID_UUID, invalidUrl).catch(() => {})
expect(mockWretch).not.toHaveBeenCalled()

// more precise — asserts the function resolves (doesn't reject) and returns undefined
await expect(
  updateSearchSGConfig(PROPS, VALID_UUID, invalidUrl),
).resolves.toBeUndefined()
expect(mockWretch).not.toHaveBeenCalled()
```

This explicitly verifies *both* that the function resolves cleanly *and* that it returns early (the `undefined` from the early `return`), rather than just that `mockWretch` wasn't called.

Confidence: high

---

**note (non-blocking):** `site.router.ts:L173-L176,L324-L327` — For the PUT failure path, errors will be logged twice. `searchsg.service.ts:L146-L153` already logs with full context (`url`, `searchsgClientId`) and re-throws; the router's `.catch()` will then log the same error a second time with minimal context (`{ error }` only). Not a bug, but can produce confusing duplicate log entries in production. Consider whether the router-level catch needs to log at all given the service already does — or pass through context so the router log is additive rather than redundant.

Confidence: medium (depends on whether your log aggregation deduplicates by correlation ID)

adriangohjw (approved):

PR #2399, opened 2026-06-09
fix(asset): cap deleteAssets fan-out to paid S3 calls

### LINKED ISSUE #2354: (warden) unbounded-filekeys-array-in-deleteassets-enables-uncapped-paid-s3-api-calls
# Unbounded fileKeys array in deleteAssets enables uncapped paid S3 API calls

**Skill:** security-review

**Severity:** 🟡 MEDIUM

**ID:** `D4F-HKZ`

**Confidence:** high

**Location:** `apps/studio/src/server/modules/asset/asset.router.ts:106-108`

## Description

The `deleteAssetsSchema` places no upper bound on the `fileKeys` array (verified: `z.array(z.string())` with no `.max()`), and the router fans out all keys in parallel via `Promise.allSettled`. Each key triggers two paid S3 operations in `markFileAsDeleted` (a `GetObjectTaggingCommand` followed by a `PutObjectTaggingCommand`), as confirmed in `src/lib/s3.ts`. Any authenticated user holding delete permission on a single site resource can submit a single mutation with thousands of keys, generating an unbounded number of concurrent S3 API calls that incurs direct financial cost and can exhaust server-side connection resources without any per-request quota control.

## Verification

Checked deleteAssetsSchema in src/schemas/asset.ts — fileKeys is z.array(z.string()) with no .max() constraint. Confirmed deleteFile in src/lib/s3.ts performs GetObjectTaggingCommand + PutObjectTaggingCommand per key (two paid S3 calls). Confirmed Promise.allSettled in the hunk fans out all keys concurrently with no concurrency cap. Checked validateUserPermissionsForAsset in asset.service.ts — only site/resource-level auth, no rate or quota guard.

adriangohjw:
Closes #2354. Implements A+B+D from the feature plan: per-request `.max(100)` cap on `fileKeys`, a 10/min rate limit on `deleteAssets`, and an idempotent skip of the expensive `PutObjectTagging` when a key is already soft-deleted.

Generated with [Claude Code](https://claude.ai/code)


adriangohjw:
@claude what does `meta({ rateLimitOptions: { max: 10, windowMs: 60 * 1000 } })` mean? and what is it based on? is it arbitrary and is it sufficient in our use case?

adriangohjw:
@claude let's play safe and update this rate limit to 20 per 1 minute. add a comment that this is practically on the generous and also arbitrary, and we should adjust if needed. do not touch the rate limit for other part of the codebase

adriangohjw:
@claude fix failing CI tests. once done, run "pnpm format:fix" before committing

adriangohjw:
@claude fix failing unit test

apps/studio/src/server/modules/asset/asset.router.ts
```diff
@@ -89,6 +89,12 @@ export const assetRouter = router({
 
   deleteAssets: protectedProcedure
     .input(deleteAssetsSchema)
+    // Bounds abuse frequency on this S3-hitting mutation. Combined with the
+    // per-request cap in deleteAssetsSchema, this limits the total number of
+    // paid S3 tagging calls a single caller can drive within the window.
+    // 20/min is practically generous (well above normal editor usage) and
+    // arbitrary — adjust if legitimate usage is blocked or abuse persists.
+    .meta({ rateLimitOptions: { max: 20, windowMs: 60 * 1000 } })
```
dcshzj: question: all our agency editors are behind a single IP address as they are using remote browser isolation to access Studio, and this rate limit would apply to everyone. I think we can drop this rate limit since we also validate permissions so the risk of spam actually resulting in S3 calls would be quite low?

adriangohjw: good point, maybe we should eventually add device ID as part of rate limit fingerprint. for now i will remove it (@claude remove the rate limit for this endpoint)

adriangohjw: @claude remove the rate limit for this endpoint

adriangohjw:
@claude /pr-review

dcshzj (approved):

PR #2400, opened 2026-06-09
feat(components): widen site.search union with egazette-algolia variant

seaerchin:
## Problem

`packages/components` ships an `IsomerSiteProps` whose `site.search` is a TypeBox union of `LocalSearchSchema | SearchSGSearchSchema`. There is no way for a published Isomer site to declare an Algolia-powered search page. The egazette site currently runs on a legacy Jekyll template — porting it to Isomer Next requires the schema to accept a third variant.

This PR is the foundation of the egazette-algolia stack: it widens the union and lands the Algolia client deps, but does not yet render anything. The renderer ships in the upstack PR.

### New schema variant

`packages/components/src/interfaces/internal/EgazetteAlgoliaSearchInputBox.ts` adds `EgazetteAlgoliaSearchSchema`:

- `type: "egazette-algolia"` discriminator
- `appId`, `searchApiKey`, `indexName` — Algolia connection (search-only key, ships in the browser)
- `categories[]` — ordered taxonomy with display labels + nested sub-categories, replacing the Jekyll JS's hardcoded `CATEGORY_INTERNAL_MAPPING`

Exported from `interfaces/internal/index.ts`.

### Union widening

`packages/components/src/types/site.ts:48` — the `site.search` `Type.Union` now lists all three variants. **Backward compatible**: existing sites with `type: "searchSG"` or `type: "localSearch"` continue to validate; the discriminator on `type` means consumers can narrow safely.

### Downstream type fixes the widening surfaced

Widening the union exposed two pre-existing consumers that hadn't been written to be exhaustive:

- `templates/next/components/complex/Hero/HeroSearchbar/SearchbarContent.tsx` — added the new case (returns `null`; egazette search runs on its own page, not the Hero searchbar)
- `templates/next/layouts/Skeleton/Skeleton.tsx:40` — `Navbar`'s `search` prop only accepts `LocalSearchProps | NavbarSearchSGInputBoxProps`, so the egazette-algolia case is narrowed to `undefined` before passing through (the Algolia search has no navbar input box equivalent)

### Dependencies

`pnpm-workspace.yaml` adds two catalog entries:

- `algoliasearch: ^5.53.0`
- `react-instantsearch: ^7.35.0`

`packages/components/package.json` references them via `catalog:` (matches the dep-catalog convention in `.claude/skills/isomer-conventions/conventions/pnpm-catalog-references.md`).

Bundle impact is ~50–60 KB gzipped, **only paid by sites that opt into the new variant in the upstack PR** — this PR alone changes no rendering paths.

**Breaking Changes**

- [ ] Yes
- [x] No — purely additive schema widening

**Features**:

- `IsomerSiteProps['search']` can now carry the `egazette-algolia` shape (no consumer renders it yet)

**Improvements**:

- N/A

**Bug Fixes**:

- N/A

## Tests

- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — 0 new warnings on changed files
- [x] `pnpm --filter @opengovsg/isomer-components test:unit` — 313/313 pass
- [x] `pnpm --filter @opengovsg/isomer-components build` — publishable tarball produced

**New dependencies**:

- `algoliasearch: catalog:` (5.53.0 in workspace catalog)
- `react-instantsearch: catalog:` (7.35.0 in workspace catalog)


seaerchin:
* **#2401** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2401?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2400** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2400?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2400?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/components/src/interfaces/internal/EgazetteAlgoliaSearchInputBox.ts
```diff
@@ -0,0 +1,61 @@
...
+  displayLabel: Type.String({
+    title: "Category label shown in the UI",
+  }),
+  subCategories: Type.Optional(
+    Type.Array(EgazetteAlgoliaSubCategorySchema, {
+      title: "Sub-categories under this category",
+    }),
+  ),
+})
+
```
seaerchin: updated downstream to be hard-coded, safe to ignore

seaerchin (commented):
fyi

packages/components/package.json
adriangohjw, line 1: before we release, can we do a check on the bundle size of a non-egazette site thats not using these packages?

packages/components/src/templates/next/layouts/Skeleton/Skeleton.tsx
```diff
@@ -37,7 +37,11 @@ export const Skeleton = ({
           logoUrl={site.logoUrl}
           logoAlt={site.siteName}
           layout={layout}
-          search={site.search}
+          search={
+            // Navbar only renders LocalSearch + SearchSG input boxes.
+            // Egazette Algolia search lives on its own page, not in the navbar.
+            site.search?.type === "egazette-algolia" ? undefined : site.search
+          }
```
adriangohjw, line 44: non-blocking product question: how do the user land on the search page ah? isit a CTA in the hero?  if that's the case, isit more intuitive that we still have the same Search Icon but it rediects to the search page?

pnpm-workspace.yaml
```diff
@@ -56,6 +56,7 @@ patchedDependencies:
 
 catalog:
   ajv: ^8.20.0
+  algoliasearch: ^5.53.0
```
adriangohjw: question: saw that we are using v4 in studio - there's no compat issue if we use v5 here right?  _(most likely no but just wanna double confirm)_

adriangohjw (commented):
By Devin

**Search layout page does not handle egazette-algolia**

The `SearchLayout` at `packages/components/src/templates/next/layouts/Search/Search.tsx:7-8` only checks for `searchSG` type. If a site has `egazette-algolia` configured and a user navigates to the search page, the Skeleton will hide the search icon from the navbar (correct), and the search page body will render nothing since neither `searchSG` condition matches. This is presumably intentional per the PR comments ('Egazette Algolia search lives on its own page'), but it means the `/search` route would render an empty page for egazette-algolia sites. The dedicated egazette search page presumably lives elsewhere, but this route may need a redirect or informational message.

https://github.com/opengovsg/isomer/blob/a7bb2f67ac07f04a2616fdba57e66ab50e0ab05f/packages/components/src/templates/next/layouts/Search/Search.tsx#L6-L17

adriangohjw (commented):
FYI - i think we might also need to update studio's backend to prevent users from switching to/from this new search variance - similar to what was raised in https://github.com/opengovsg/isomer/pull/2242

dcshzj:
> FYI - i think we might also need to update studio's backend to prevent users from switching to/from this new search variance - similar to what was raised in #2242

Hmm I think it is ok? Right now site admins cannot set the search type on the site config level, that is isolated to just Isomer admins. This new search variant would also not be accessible to end users, only Isomer admins to modify themselves.

Thinking to keep our Isomer admin config more flexible in case we need to make adjustments quickly.

packages/components/package.json
seaerchin: Good call. There's a CI check being set up for exactly this on `ci/components-package-size` — it builds the components package and compares its size against `main`, running only when the package changes.

Runtime-wise: `algoliasearch` + `react-instantsearch` are only pulled in via the `EgazetteAlgoliaSearch` client island, which `Search.tsx` renders only when `site.search.type === "egazette-algolia"`. It's a static import though, so it lands in the Search-layout chunk rather than being tree-shaken away for non-egazette sites that use that layout. If the size check shows the delta is material, the follow-up is to `dynamic()`-import the egazette island so non-egazette sites don't download it. I'll confirm against the size numbers before we release.

pnpm-workspace.yaml
seaerchin: No compat issue. studio doesn't depend on `algoliasearch` at all — nothing in `apps/studio/package.json`, its source, or the lockfile — so the v5 catalog entry is new and consumed only by `components`. `react-instantsearch@7` is built against `algoliasearch@5`, and since `components` bundles its own deps there's nothing for it to clash with. (The v4 is the legacy Jekyll egazette CDN script — `algoliasearch@4.20.0` — not studio.)

packages/components/src/templates/next/layouts/Skeleton/Skeleton.tsx
seaerchin: Egazette search is its own page (matches the legacy site), so the navbar search box is intentionally suppressed for this variant. The entry point is the hero search bar — it redirects to the search page carrying the query as `?q=`, which `routing.ts` deserializes back into the Algolia search box. We deliberately *don't* embed the Algolia searchbox in the navbar, since that'd pull `react-instantsearch` onto every page and bloat non-search bundles.

adriangohjw:
> > FYI - i think we might also need to update studio's backend to prevent users from switching to/from this new search variance - similar to what was raised in #2242
> 
> Hmm I think it is ok? Right now site admins cannot set the search type on the site config level, that is isolated to just Isomer admins. This new search variant would also not be accessible to end users, only Isomer admins to modify themselves.
> 
> Thinking to keep our Isomer admin config more flexible in case we need to make adjustments quickly.

> site admins cannot set the search type on the site config level

they cannot from the UI but they can via the API, which was what #2242 was tackling

also, the endpoint is for `updateSiteConfig` while the endpoint for isomer admin config is `setSiteConfigByAdmin` so iit will not affect Isomer Admin's.

apps/studio/src/server/modules/site/site.service.ts
```diff
@@ -39,6 +40,43 @@ export const validateUserPermissionsForSite = async ({
...
+ * `updateSiteConfig` / `updateSiteIntegrations` (cf. the SearchSG `clientId`
+ * protections in #2242).
+ *
+ * Returns the search config that should be persisted: when the site is already
+ * on `egazette-algolia`, the DB value is preserved verbatim and the incoming
+ * value is ignored; otherwise the incoming value is returned unchanged. Throws
+ * a `BAD_REQUEST` when the caller attempts to switch the search type to or from
+ * `egazette-algolia`.
+ */
+export const resolveEgazetteAlgoliaSearchConfig = (
```
adriangohjw, line 61: nice, i like this way of encapsulating the logic!

adriangohjw (approved):

PR #2403, opened 2026-06-09
refactor(form-builder): extract shared array control primitives

seaerchin:
## Problem

Four jsonforms array control renderers — `JsonFormsArrayControl`, `JsonFormsCategoryOptionsControl`, `JsonFormsTagCategoryControl`, `JsonFormsTagCategoryOptionsControl` — each carried ~150 lines of duplicated state, drag-and-drop glue, and nested-drawer prop wiring. Customisation flowed through a wide `JsonFormsArrayControlProps` interface (icon, subtitle, trailing, error caption, padding, empty-state, etc.) where every consumer used a different subset — a classic premature abstraction.

`DraggableTagButton` had the same shape problem: nine optional rendering props where each consumer cherry-picked a few.

## Solution

Pulled the shared glue out into focused primitives, then inlined each consumer's specific UI directly so the renderers read top-to-bottom without a prop-passing maze.

**Breaking Changes**

Internal refactor only. No public API or schema changes.

**Improvements**:

- New `useArray(args)` hook (`form-builder/hooks/useArray.ts`) owns the previously-duplicated state and callbacks: `selectedIndex`, `isAddItemDisabled`, `isRemoveItemDisabled`, `childUiSchema`, `handleRemoveItem`, `onDragEnd`.
- New `NestedDrawerProvider` wraps each renderer's body so the `selectedIndex !== undefined → render <ComplexEditorNestedDrawer/>` branch lives in one place instead of being copy-pasted four times.
- `DraggableTagButton` is now a compound component (`Root`, `Handle`, `Body`, `Icon`, `Content`, `Label`, `Subtitle`, `ErrorCaption`, `Trailing`). Consumers compose row contents instead of passing nine optional slot props.
- `DraggableDrawerButton` is re-implemented as a thin composition over `DraggableTagButton.*` (~135 LOC → ~45 LOC).
- Extracted shared primitives into `form-builder/components/`: `AddItemButton`, `ComplexEditorNestedDrawer`, `DeleteConfirmModal`, `DuplicateLabelError`, `EmptyArray`, `EmptyCategory`, `TagRowActionsMenu`.
- Validation utilities split out with tests: `hasBlankOptionLabel`, `indicesWithDuplicateLabels`.
- Deleted `JsonFormsArrayControlProps` extension interface and `JsonFormsArrayControlView` — each of the four renderers now expresses its unique header / add-button / empty-state / row layout / error / delete-modal pieces directly.

**Bug Fixes**:

- `NestedDrawerProvider`'s `selectedIndex ?` truthy check would have rendered children (not the drawer) when the user clicked the very first row (index `0`); switched to `selectedIndex !== undefined`.

## Before & After Screenshots

No visual changes intended — all four controls render the same headers, row layouts, empty states, and nested-drawer flows as before.

**BEFORE**:

**AFTER**:

## Tests

Existing unit tests for the validation utilities (`__tests__/hasBlankOptionLabel.test.ts`, `__tests__/indicesWithDuplicateLabels.test.ts`) cover the extracted helpers. `pnpm typecheck` and `pnpm format` are clean on the touched files.

Manual smoke for each of the four controls:

- Add an item; observe `maxItems` disables the Add button.
- Drag to reorder.
- Click a row → nested drawer opens (including row index `0`, which the truthy check would have broken).
- Previous / Next inside the drawer.
- Delete from the drawer — confirm `selectedIndex` adjusts (closes if removed row, decrements if a row above is removed).
- Delete via row actions menu (tag / category controls) — confirm modal flow.
- Trigger duplicate-label state on tag controls — confirm error caption + critical stripe still render.

**New scripts**:

- None.

**New dependencies**:

- None.

**New dev dependencies**:

- None.


apps/studio/src/features/editing-experience/components/form-builder/components/ComplexEditorNestedDrawer.tsx
seaerchin, line 1: extracted out as is, no change

apps/studio/src/features/editing-experience/components/form-builder/components/DraggableDrawerButton.tsx
seaerchin, line 1: redefined per tag button components, no change but happy to remove this

apps/studio/src/features/editing-experience/components/form-builder/components/EmptyArray.tsx
seaerchin, line 1: extracted from existing call sites, no change

apps/studio/src/features/editing-experience/components/form-builder/components/EmptyCategory.tsx
seaerchin, line 1: extracted from existing call sites, no change

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
seaerchin, line 1: wrapper component that provides the early return if we have a `selectedIndex`, no change in code

apps/studio/src/features/editing-experience/components/form-builder/hooks/useArray.ts
seaerchin, line 1: extracted from existing call sites into hook

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsArrayControl.tsx
seaerchin, line 1: reverted to `main`

apps/studio/src/features/editing-experience/components/form-builder/components/TagRowActionsMenu.tsx
seaerchin, line 1: taken from other commit

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
```diff
@@ -0,0 +1,47 @@
...
+          label={label}
+          setSelectedIndex={setSelectedIndex}
+          isRemoveItemDisabled={isRemoveItemDisabled}
+          handleRemoveItem={handleRemoveItem(path, selectedIndex)}
+          selectedIndex={selectedIndex}
+          maxIndex={data - 1}
+        />
+      ) : (
+        children
+      )}
```
adriangohjw: nit nit: can switch the ordering around so no need negative negation

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
```diff
@@ -0,0 +1,47 @@
+import { composePaths, type ArrayLayoutProps } from "@jsonforms/core"
+
+import type { UseArrayReturn } from "../hooks/useArray"
+import { ComplexEditorNestedDrawer } from "./ComplexEditorNestedDrawer"
+
+type NestedDrawerProviderProps = ArrayLayoutProps &
+  UseArrayReturn & {
+    children: React.ReactNode
```
adriangohjw: nit nit: should import ReactNode as type

adriangohjw (approved):
@claude /pr-review

adriangohjw:
@claude /pr-review

seaerchin:
@claude /pr-review

seaerchin:
wrong branch in checkout i htink 
image

seaerchin:
image

seaerchin (commented):
## Review

### Summary

- This PR dissolves the wide `JsonFormsArrayControlProps` / `JsonFormsArrayControlView` abstraction: shared state moves into `useArray` (`apps/studio/src/features/editing-experience/components/form-builder/hooks/useArray.ts`), the drawer-or-list branch into `NestedDrawerProvider` (`form-builder/components/NestedDrawerProvider.tsx`), and row rendering into a compound `DraggableTagButton` (`form-builder/components/DraggableTagButton.tsx`). Each of the four renderers now inlines its own header/empty-state/row markup and composes these primitives.
- Codeflow: each renderer destructures `ArrayLayoutProps` → calls `useArray` → spreads `{...props} {...arrayResult}` into `NestedDrawerProvider`, which renders `ComplexEditorNestedDrawer` when `selectedIndex !== undefined`, otherwise the list. Row deletion via the actions menu goes through `useDeleteTarget` → `DeleteConfirmModal`; drawer deletion goes through `useArray.handleRemoveItem` which also adjusts `selectedIndex`.
- Overall: this is a clear improvement. The four renderers now read top-to-bottom, the index-0 truthy fix in `NestedDrawerProvider.tsx:31` is a genuine bug catch, and I verified the deleted `addItemLabel` schema extension and `JsonFormsArrayControlView` have zero remaining references anywhere in the repo. Findings below are mostly about behavior that quietly changed despite the "no visual changes" claim.

### Findings

**issue (non-blocking):** `form-builder/components/DraggableTagButton.tsx:L260-L294` — `Handle` and `Body` hardcode `py="0.5rem"`, but the old defaults for consumers that did *not* pass `listItemContentProps` were `py: isError ? "0.75rem" : "1.25rem"` (handle) and `py: isError ? "0.75rem" : "1rem"` (body). The tag/category controls all passed `py: "0.5rem"` so they're unchanged, but every schema hitting the generic `JsonFormsArrayControl` (e.g. InfoCards `cards`, KeyStatistics `statistics`) now renders visibly shorter rows. The PR description says "No visual changes intended — all four controls render the same … row layouts", so either this is unintended, or the generic control was deliberately tightened and the description should say so. If unintended, the generic control's rows need the larger padding restored (e.g. a `py` prop on `Handle`/`Body`).

Confidence: high

**note:** `renderers/controls/JsonFormsArrayControl.tsx` — the new generic control no longer renders `description` (the old view had `{description && <Text textStyle="body-2">…}`). I checked: every array schema with an array-level `description` today is handled by a dedicated renderer or hidden (`Navbar.ts:60` uses `format: "navbar"`, `ChildrenPages.ts:60` uses `format: "childrenPagesOrdering"`, `dgs.ts:31` is `format: "hidden"`), so nothing user-visible changes right now. But a schema author adding a `description` to a plain object/primitive array will silently get nothing. Worth either restoring the two-line render or accepting this consciously.

Confidence: medium

**suggestion (non-blocking):** `form-builder/hooks/useArray.ts:L51-L64` — `handleRemoveItem` keeps the `selectedIndex === undefined → return` guard from the old drawer-only code, but it's now exported under a general-purpose name and passed as `removeItem` to `DraggableTagButton.Label` in all four renderers. Called outside the drawer flow it silently no-ops, which is a trap for the next consumer (row-level delete already had to route around it via `useDeleteTarget`). Consider renaming it to something like `handleRemoveSelectedItem`, or documenting that it's the nested-drawer delete specifically.

Confidence: medium

**nitpick (non-blocking):** `form-builder/components/NestedDrawerProvider.tsx` — "Provider" implies React context, but this is a conditional overlay swap (`drawer if selected, else children`). Something like `NestedDrawerSwitch` / `WithNestedDrawer` would read truer. Relatedly, consumers feed it via `{...props} {...arrayResult}` double-spread; that's convenient, but the component only uses ~10 of those fields, and the spread makes it harder to see what the drawer actually depends on.

Confidence: medium

**nitpick (if-minor):** `renderers/controls/TagRowActionsMenu.tsx` — this shared piece stayed in `renderers/controls/` while its siblings (`DeleteConfirmModal`, `DuplicateLabelError`, `EmptyArray`, …) moved to `components/`; meanwhile `components/DraggableLinkButton.tsx:168` reaches back into `renderers/controls/utils/parseHref`. The two-directional dependency between `components/` and `renderers/controls/` blurs the layering this refactor otherwise establishes. Moving `TagRowActionsMenu` (and its `ROW_ACTIONS_MENU_BUTTON_PROPS` constant) over would finish the job.

Confidence: high

**praise:** `form-builder/components/NestedDrawerProvider.tsx:31` — catching the `selectedIndex ?` truthy bug (index 0 would have rendered the list instead of the drawer) before it shipped, and calling it out explicitly in the PR description, is exactly how a refactor PR should surface incidental fixes.

**praise:** Deleting the nine-optional-prop `DraggableTagButton` and the `renderListItem`/`emptyState`/`mapNewArrayItem` escape hatches in favor of inlined composition is the right call — I verified there are no remaining references to `JsonFormsArrayControlProps`, `JsonFormsArrayControlView`, or the `addItemLabel` schema extension anywhere in the repo, so the deletion is complete, not just relocated.

**chore:** The Vercel deployment for this PR is failing (and the `@claude /pr-review` runs hit a wrong-branch checkout per the PR comments). I couldn't run `pnpm typecheck` locally to rule out a real build break. Since this is stacked on `feat/collection-category-management`, worth confirming the failure is the stacked-base/CI issue and not something in this diff before merge.

Confidence: low

seaerchin:
@adriangohjw i'll fix the prop issue then merge in

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
```diff
@@ -0,0 +1,47 @@
...
+  path,
+  label,
+  setSelectedIndex,
+  isRemoveItemDisabled,
+  handleRemoveItem,
+  data,
+}: NestedDrawerProviderProps) => {
+  return (
+    <>
+      {selectedIndex !== undefined ? (
```
dcshzj: nit: can do early return so we don't use this ternary operator unnecessarily

```ts
if (selectedIndex === undefined) {
    return children
}

return ...
```

seaerchin:
Addressed the review findings in 95a50a525:

- **issue (row padding):** `DraggableTagButton.Handle`/`.Body` now accept an optional `py` (default `0.5rem`), and the generic `JsonFormsArrayControl` passes the original values back (`0.75rem` in error / `1.25rem` handle, `1rem` body) — so InfoCards, KeyStatistics, etc. keep their pre-refactor row heights while the tag/category controls are unchanged.
- **note (description):** restored array-level `description` rendering and the original header/list spacing in the generic `JsonFormsArrayControl`, matching the other controls' layout.
- **suggestion (`handleRemoveItem`):** renamed to `handleRemoveSelectedItem` with a doc comment spelling out the drawer-only guard and pointing row-level deletes at `useDeleteTarget`.
- **nitpick (naming):** `NestedDrawerProvider` → `NestedDrawerSwitch`.
- **nitpick (layering):** `TagRowActionsMenu` moved to `components/` with `ROW_ACTIONS_MENU_BUTTON_PROPS` inlined (its only remaining consumer), and `JsonFormsCategoryOptionsControl`'s identical inline menu now reuses it. `parseHref` moved to `LinkEditor/utils` next to `getLinkHrefType`, so `components/` no longer imports from `renderers/`.
- **chore (failing build):** local `pnpm typecheck` shows zero errors in files touched by this PR; the remaining failures (`JsonFormsNavbarControl/utils.ts`, `JsonFormsWidgetIntegrationControl.tsx`) are pre-existing on the base branch `feat/collection-category-management` and untouched here.

Verified with `pnpm lint` (clean), `pnpm format:fix`, and the `LinkEditor/__tests__/utils.test.ts` suite (6/6 passing).

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerProvider.tsx
```diff
@@ -0,0 +1,47 @@
+import { composePaths, type ArrayLayoutProps } from "@jsonforms/core"
+
+import type { UseArrayReturn } from "../hooks/useArray"
+import { ComplexEditorNestedDrawer } from "./ComplexEditorNestedDrawer"
+
+type NestedDrawerProviderProps = ArrayLayoutProps &
+  UseArrayReturn & {
+    children: React.ReactNode
+  }
+export const NestedDrawerProvider = ({
```
dcshzj: nit: the name of this is a bit confusing, it's not a Provider since it doesn't have context

dcshzj (approved):

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
seaerchin: Fixed in e67bba85b — replaced with an early return on `selectedIndex === undefined`, so the negated check is gone.

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
seaerchin: Fixed in e67bba85b — now `import type { ReactNode } from "react"`.

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerSwitch.tsx
seaerchin: Fixed in e67bba85b — early return of `children` when no row is selected, ternary removed.

apps/studio/src/features/editing-experience/components/form-builder/components/NestedDrawerProvider.tsx
seaerchin: Renamed to `NestedDrawerSwitch` in 95a50a525 (pushed just before this comment, so the diff was outdated).

PR #2412, opened 2026-06-10
[DO NOT MERGE] feat: migrate redirects from infra to database table

dcshzj:
> [!CAUTION]
> **DO NOT MERGE — this PR will never be merged.** It exists only to review and record the one-off script used to migrate redirects from isomer-next-infra into the Studio database. The script is executed manually against the target environment as part of the redirects migration, and this PR will be closed once the migration is complete. Because the branch never merges, the script carries an inline copy of `createRedirectSchema` (from `feat/redirects-management-backend`) instead of importing it.

## Problem

Redirects for Isomer Next sites currently live as static rules in `redirects.production.csv` inside the isomer-next-infra repository. With redirects management moving into Studio (backed by the new `Redirect` table), the existing CSV rules need to be migrated into the database so sites keep their redirects once the new feature takes over.

There is currently no tooling to perform this migration.

Fixes ISOM-2274.

## Solution

**Breaking Changes**

**Features**:

- Adds `apps/studio/prisma/scripts/importRedirectsFromCsv.ts`, a one-off data script that imports rows from a `domainName,source,target` CSV into the `Redirect` table:
  - Maps each CSV domain to a site by matching against `Site.config.url` (comparison ignores case, scheme, and trailing path). Rows for domains with no matching site are skipped and reported with counts.
  - Ignores rows containing wildcards (`*`) or query parameters (`?`) in the source or target, since these are infra publishing features with no equivalent in the `Redirect` table's exact-match semantics.
  - Validates and normalises every row with the same rules as the Studio publish flow (inline copy of `createRedirectSchema`), so imported sources are normalised identically to redirects created through Studio. Invalid rows (e.g. empty, `http://`, or `mailto:` destinations) are skipped and reported with reasons.
  - Keeps the first occurrence when two CSV sources normalise to the same value, and reports the dropped duplicates.
  - Is idempotent: inserts use `ON CONFLICT (siteId, source)` to update destinations in place and revive soft-deleted rows, mirroring the publish flow. Inserts run in chunks of 1,000 inside a single transaction.
  - Runs as a dry run by default, printing a full summary (sites matched, rows to import, ignored/invalid/unmatched rows) without writing.
- Adds 11 integration tests (`prisma/scripts/__tests__/importRedirectsFromCsv.test.ts`) covering domain matching, wildcard/query-parameter/invalid row handling, source normalisation, duplicate handling, idempotency and soft-delete revival, dry-run behaviour, and header validation.
- Documents the script in `prisma/scripts/README.md`.

## Before & After Screenshots

N/A — this PR adds a manually-run data script with no UI changes.

## Tests

- Run `pnpm test:unit prisma/scripts/__tests__/importRedirectsFromCsv.test.ts` from `apps/studio` (requires `pnpm setup:test` for the test database). All 11 tests pass.

**Manual Verification Steps** (this doubles as the migration runbook):

- [ ] From `apps/studio`, follow `prisma/scripts/README.md` to tunnel to the target environment and set `DATABASE_URL`
- [ ] Set `csvPath` at the bottom of the script to the path of `redirects.production.csv` from isomer-next-infra and leave `dryRun = true`
- [ ] Run `source .env && pnpm exec tsx prisma/scripts/importRedirectsFromCsv.ts` and review the summary: sites matched, rows to import, and the reported unmatched domains, wildcard/query-parameter rows, and invalid rows
- [ ] Confirm the unmatched domains are expected (sites not yet on Studio) before proceeding
- [ ] Set `dryRun = false` and re-run the script with another engineer pairing (production requirement per `prisma/scripts/README.md`)
- [ ] Spot-check a few imported rows in the `Redirect` table: sources have a single leading slash and no trailing slash, destinations are unchanged
- [ ] Re-run the script once more and confirm the row count is unchanged (idempotency)
- [ ] Close this PR once the migration is complete

**New scripts**:

- `prisma/scripts/importRedirectsFromCsv.ts` : imports redirect rules from the isomer-next-infra `redirects.production.csv` into the `Redirect` table; dry-run by default, idempotent on re-run

**New dependencies**:

- None

**New dev dependencies**:

- None

Adds two new one-off scripts and their tests for a data migration that moves static redirect rules from the infra CSV into the `Redirect` database table. The import script is idempotent, dry-run-by-default, and resolves internal destinations to `[resource:...]` references; the companion verification script probes the live CloudFront distribution to confirm each migrated redirect returns a 301.

- `importRedirectsFromCsv.ts` — parses and validates the CSV with an intentionally lenient schema subset, bulk-inserts in 1 000-row chunks inside a single transaction, and writes a `migrated-redirects.csv` for the verifier.
- `verifyRedirectsLive.ts` — reads the same CSV, fires concurrent HTTP probes with abort-signal timeout, and reports wrong-status / wrong-location failures to a CSV.
- 16 integration tests cover domain matching, wildcard/query-param skipping, invalid-row reporting, source normalisation, duplicate handling, idempotency, soft-delete revival, dry-run correctness, `updatedAt` advancement, empty-path guard, and header validation.
</details>

Safe to proceed — the script is a manually-executed, dry-run-by-default migration with no schema changes, and an idempotent ON CONFLICT path ensures re-runs are harmless.

The migration logic is well-guarded: every code path that previously caused concern (stale updatedAt, parentId cycles, empty csvPath) has been addressed. The single gap is that destinationSchema omits the control-char, backslash, and '..' guards the inline comment says are intended — a real-CSV destination with those characters would be stored verbatim — but the mandatory dry-run review step provides a practical safety net before any write occurs.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts — destinationSchema is missing the control-char, backslash, and '..' guards present on sourceSchema.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/prisma/scripts/importRedirectsFromCsv.ts | New one-off migration script: parses a domain/source/target CSV, resolves internal destinations to [resource:...] references, and bulk-inserts via ON CONFLICT; addressed feedback on updatedAt, cycle guard, and empty-csvPath guard from prior rounds. |
| apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts | 16 integration tests covering domain matching, wildcard/query-param/invalid rows, source normalisation, duplicate handling, idempotency, soft-delete revival, dry-run, updatedAt bump, empty-csvPath guard, and header validation. |
| apps/studio/prisma/scripts/verifyRedirectsLive.ts | New verification script: reads the same CSV, probes each redirect over HTTP with a concurrency pool, compares status + Location, and writes failures to a CSV. No database access needed. |
| apps/studio/prisma/scripts/README.md | README extended with runbook steps for import and verification, including dry-run → review → real-run flow and pointers to the migrated-redirects output file. |

</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
flowchart TD
    A[redirects.production.csv] --> B[importRedirectsFromCsv.ts]
    B --> C{Parse & validate rows}
    C -->|wildcard / query-param| D[Skip — log wildcardRowCount / queryParamRowCount]
    C -->|no matching Site.config.url| E[Skip — log unmatchedDomains]
    C -->|fails lenient schema| F[Skip — log invalidRows]
    C -->|duplicate after normalisation| G[Skip — log duplicateRows]
    C -->|valid| H[buildPermalinkToResourceId]
    H --> I{Destination type}
    I -->|https:// external| J[Store verbatim]
    I -->|internal path — live page/folder found| K[Store as resource:siteId:id reference]
    I -->|internal path — no live page| L[Store as literal path — log unresolvedDestinations]
    J & K & L --> M{dryRun?}
    M -->|yes| N[Print summary — no DB writes]
    M -->|no| O[INSERT in 1k-row chunks — ON CONFLICT doUpdateSet + updatedAt]
    O --> P[Write migrated-redirects.csv]
    P --> Q[verifyRedirectsLive.ts]
    Q --> R[Probe each redirect over HTTP]
    R --> S{301 + correct Location?}
    S -->|yes| T[OK]
    S -->|wrong status or location| U[Write redirect-verification-failures.csv]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[redirects.production.csv] --> B[importRedirectsFromCsv.ts]
    B --> C{Parse & validate rows}
    C -->|wildcard / query-param| D[Skip — log wildcardRowCount / queryParamRowCount]
    C -->|no matching Site.config.url| E[Skip — log unmatchedDomains]
    C -->|fails lenient schema| F[Skip — log invalidRows]
    C -->|duplicate after normalisation| G[Skip — log duplicateRows]
    C -->|valid| H[buildPermalinkToResourceId]
    H --> I{Destination type}
    I -->|https:// external| J[Store verbatim]
    I -->|internal path — live page/folder found| K[Store as resource:siteId:id reference]
    I -->|internal path — no live page| L[Store as literal path — log unresolvedDestinations]
    J & K & L --> M{dryRun?}
    M -->|yes| N[Print summary — no DB writes]
    M -->|no| O[INSERT in 1k-row chunks — ON CONFLICT doUpdateSet + updatedAt]
    O --> P[Write migrated-redirects.csv]
    P --> Q[verifyRedirectsLive.ts]
    Q --> R[Probe each redirect over HTTP]
    R --> S{301 + correct Location?}
    S -->|yes| T[OK]
    S -->|wrong status or location| U[Write redirect-verification-failures.csv]
```

</a>
</details>

<sub>Reviews (4): Last reviewed commit: ["feat(redirect): write migrated rows to C..."](https://github.com/opengovsg/isomer/commit/f02f1a9bc92772e29edd9aa08fd85ad232568a33) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39746724)</sub>

**Context used:**

- Context used - apps/studio/prisma/CLAUDE.md ([source](https://app.greptile.com/opengovsg/github/opengovsg/isomer/-/custom-context?memory=511e297b-11a8-4965-aa9a-a7db76a58b1c))

### DIFF
--- a/apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
+++ b/apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
@@ -174,6 +174,54 @@ describe("importRedirectsFromCsv", () => {
     expect(summary.unresolvedDestinations).toHaveLength(1)
   })
 
+  it("writes the migrated rows to migratedCsvPath with the expected Location as target", async () => {
+    // Arrange — one external, one reference (live page), one literal (no page)
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const user = await setupUser({ email: "import-migrated@open.gov.sg" })
+    await setupPageResource({
+      siteId,
+      resourceType: ResourceType.Page,
+      permalink: "new-page",
+      parentId: null,
+      state: ResourceState.Published,
+      userId: user.id,
+    })
+    const csvPath = writeCsv([
+      "www.alpha.gov.sg,/ext,https://example.gov.sg/page",
+      "www.alpha.gov.sg,/old-page,/New-Page",
+      "www.alpha.gov.sg,/gone,/missing-page",
+    ])
+    const migratedCsvPath = path.join(tmpDir, "migrated.csv")
+
+    // Act
+    await importRedirectsFromCsv({ csvPath, dryRun: false, migratedCsvPath })
+
+    // Assert — header plus a row per migrated redirect; the reference target is
+    // the resolved (lowercased) permalink, not the [resource:...] reference
+    const lines = fs.readFileSync(migratedCsvPath, "utf8").trim().split(/\r?\n/)
+    expect(lines[0]).toBe("domainName,source,target")
+    expect(lines.slice(1).sort()).toEqual(
+      [
+        "www.alpha.gov.sg,/ext,https://example.gov.sg/page",
+        "www.alpha.gov.sg,/old-page,/new-page",
+        "www.alpha.gov.sg,/gone,/missing-page",
+      ].sort(),
+    )
+  })
+
+  it("does not write a migrated CSV on a dry run", async () => {
+    // Arrange
+    await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv(["www.alpha.gov.sg,/old-page,/new-page"])
+    const migratedCsvPath = path.join(tmpDir, "should-not-exist.csv")
+
+    // Act
+    await importRedirectsFromCsv({ csvPath, dryRun: true, migratedCsvPath })
+
+    // Assert
+    expect(fs.existsSync(migratedCsvPath)).toBe(false)
+  })
+
   it("skips rows for domains with no matching site and reports them", async () => {
     // Arrange
     const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
--- a/apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
+++ b/apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
@@ -354,6 +354,41 @@ describe("importRedirectsFromCsv", () => {
     ])
   })
 
+  it("bumps updatedAt when a re-run changes a destination in place", async () => {
+    // Arrange — Kysely bypasses Prisma's @updatedAt, so the conflict update
+    // must set it explicitly or it stays frozen at the original insert time
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const firstCsvPath = writeCsv(["www.alpha.gov.sg,/old-page,/first"])
+    await importRedirectsFromCsv({ csvPath: firstCsvPath, dryRun: false })
+    const [before] = await db
+      .selectFrom("Redirect")
+      .select(["updatedAt"])
+      .where("siteId", "=", siteId)
+      .execute()
+    const secondCsvPath = writeCsv(["www.alpha.gov.sg,/old-page,/second"])
+
+    // Act
+    await importRedirectsFromCsv({ csvPath: secondCsvPath, dryRun: false })
+
+    // Assert
+    const [after] = await db
+      .selectFrom("Redirect")
+      .select(["updatedAt"])
+      .where("siteId", "=", siteId)
+      .execute()
+    expect(after?.updatedAt.getTime()).toBeGreaterThan(
+      before?.updatedAt.getTime() ?? 0,
+    )
+  })
+
+  it("throws a clear error when csvPath is empty", async () => {
+    // Act
+    const run = () => importRedirectsFromCsv({ csvPath: "", dryRun: true })
+
+    // Assert
+    await expect(run()).rejects.toThrow("csvPath is empty")
+  })
+
   it("writes nothing in dry-run mode but still reports the summary", async () => {
     // Arrange
     const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
--- a/apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
+++ b/apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
@@ -1,0 +1,250 @@
+import fs from "node:fs"
+import os from "node:os"
+import path from "node:path"
+import { db, jsonb } from "~/server/modules/database"
+
+import { resetTables } from "../../../tests/integration/helpers/db"
+import { setupSite } from "../../../tests/integration/helpers/seed"
+import { importRedirectsFromCsv } from "../importRedirectsFromCsv"
+
+let tmpDir: string
+let csvCounter = 0
+
+beforeAll(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "import-redirects-test-"))
+})
+
+afterAll(() => {
+  fs.rmSync(tmpDir, { recursive: true, force: true })
+})
+
+const writeCsv = (
+  rows: string[],
+  { header = "domainName,source,target" }: { header?: string } = {},
+) => {
+  csvCounter += 1
+  const csvPath = path.join(tmpDir, `redirects-${csvCounter}.csv`)
+  fs.writeFileSync(csvPath, [header, ...rows].join("\n"))
+  return csvPath
+}
+
+// setupSite seeds config.url as "" - point it at a domain so the import
+// script's domain matching can find the site
+const setupSiteWithUrl = async (url: string) => {
+  const { site } = await setupSite()
+  await db
+    .updateTable("Site")
+    .set({ config: jsonb({ ...site.config, url }) })
+    .where("id", "=", site.id)
+    .execute()
+  return site.id
+}
+
+const getRedirects = (siteId: number) =>
+  db
+    .selectFrom("Redirect")
+    .select(["source", "destination", "deletedAt"])
+    .where("siteId", "=", siteId)
+    .orderBy("source")
+    .execute()
+
+describe("importRedirectsFromCsv", () => {
+  beforeEach(async () => {
+    await resetTables("Redirect", "Navbar", "Footer", "Site")
+  })
+
+  it("imports rows for the site whose config.url matches the CSV domain", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv([
+      "www.alpha.gov.sg,/old-page,/new-page",
+      "www.alpha.gov.sg,/external,https://example.gov.sg/page",
+    ])
+
+    // Act
+    const summary = await importRedirectsFromCsv({ csvPath, dryRun: false })
+
+    // Assert
+    const redirects = await getRedirects(siteId)
+    expect(redirects).toEqual([
+      {
+        source: "/external",
+        destination: "https://example.gov.sg/page",
+        deletedAt: null,
+      },
+      { source: "/old-page", destination: "/new-page", deletedAt: null },
+    ])
+    expect(summary.sitesMatched).toBe(1)
+    expect(summary.importedCount).toBe(2)
+  })
+
+  it("skips rows for domains with no matching site and reports them", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv([
+      "www.alpha.gov.sg,/old-page,/new-page",
+      "www.unknown.gov.sg,/a,/b",
+      "www.unknown.gov.sg,/c,/d",
+    ])
+
+    // Act
+    const summary = await importRedirectsFromCsv({ csvPath, dryRun: false })
+
+    // Assert
+    const redirects = await getRedirects(siteId)
+    expect(redirects).toHaveLength(1)
+    expect(summary.unmatchedDomains.get("www.unknown.gov.sg")).toBe(2)
+  })
+
+  it("matches domains ignoring case, scheme and trailing path", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("https://WWW.Alpha.gov.sg/")
+    const csvPath = writeCsv(["www.alpha.gov.sg,/old-page,/new-page"])
+
+    // Act
+    await importRedirectsFromCsv({ csvPath, dryRun: false })
+
+    // Assert
+    const redirects = await getRedirects(siteId)
+    expect(redirects).toHaveLength(1)
+  })
+
+  it("ignores rows with wildcards in source or target", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv([
+      "www.alpha.gov.sg,/old-section/*,/new-section",
+      "www.alpha.gov.sg,/old-page,/new-section/*",
+      "www.alpha.gov.sg,/kept,/destination",
+    ])
+
+    // Act
+    const summary = await importRedirectsFromCsv({ csvPath, dryRun: false })
+
+    // Assert
+    const redirects = await getRedirects(siteId)
+    expect(redirects.map((row) => row.source)).toEqual(["/kept"])
+    expect(summary.wildcardRowCount).toBe(2)
+  })
+
+  it("ignores rows with query parameters in source or target", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv([
+      "www.alpha.gov.sg,/old-page?lang=en,/new-page",
+      "www.alpha.gov.sg,/news,/articles/?page=1",
+      "www.alpha.gov.sg,/kept,/destination",
+    ])
+
+    // Act
+    const summary = await importRedirectsFromCsv({ csvPath, dryRun: false })
+
+    // Assert
+    const redirects = await getRedirects(siteId)
+    expect(redirects.map((row) => row.source)).toEqual(["/kept"])
+    expect(summary.queryParamRowCount).toBe(2)
+  })
+
+  it("skips rows that fail redirect validation and reports the reason", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv([
+      "www.alpha.gov.sg,/empty-destination,",
+      "www.alpha.gov.sg,/insecure,http://example.com",
+      "www.alpha.gov.sg,/mail,mailto:hello@example.gov.sg",
+      "www.alpha.gov.sg,/kept,/destination",
+    ])
+
+    // Act
+    const summary = await importRedirectsFromCsv({ csvPath, dryRun: false })
+
+    // Assert
+    const redirects = await getRedirects(siteId)
+    expect(redirects.map((row) => row.source)).toEqual(["/kept"])
+    expect(summary.invalidRows).toHaveLength(3)
+    expect(summary.invalidRows[0]?.reason).toContain(
+      "Destination must start with '/' or 'https://'",
+    )
+  })
+
+  it("normalises sources the same way as the publish flow", async () => {
+    // Arrange
+    const siteId = await setupSiteWithUrl("www.alpha.gov.sg")
+    const csvPath = writeCsv(["www.alpha.gov.sg,//old//page/,/new-page"])
+
+    // Act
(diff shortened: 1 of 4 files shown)


apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,242 @@
+import fs from "node:fs"
+import { fileURLToPath } from "node:url"
+import { z } from "zod"
+import { db } from "~/server/modules/database"
```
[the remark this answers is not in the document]
dcshzj: Done in ccabf1e39 — rows are now parsed with `Papa.parse` (already a dependency via `getAuditLogs.ts`), so standard CSV quoting and CRLF are handled.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,242 @@
...
+    const parts = line.split(",")
+    if (parts.length !== 3) {
+      invalidRows.push({
+        lineNumber,
+        line,
+        reason: `Expected 3 comma-separated fields, got ${parts.length}`,
+      })
+      return
+    }
+    const [domainName = "", source = "", target = ""] = parts
```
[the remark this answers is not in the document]
dcshzj: Done in ccabf1e39 — `line.split(',')` is gone; each row comes from `Papa.parse` and fields are trimmed before validation.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,242 @@
...
+}: ImportRedirectsFromCsvProps) => {
+  const [header = "", ...dataLines] = fs
+    .readFileSync(csvPath, "utf8")
+    .split("\n")
+    .map((line) => line.replace(/\r$/, ""))
+    .filter((line) => line.length > 0)
+
+  if (header !== "domainName,source,target") {
+    throw new Error(`Unexpected CSV header: ${header}`)
+  }
```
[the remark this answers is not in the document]
dcshzj: Done in ccabf1e39 — a leading UTF-8 BOM is stripped and header fields are trimmed before the comparison.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,242 @@
...
+// Only run when executed directly, not when imported by tests
+const isMain = process.argv[1] === fileURLToPath(import.meta.url)
+if (isMain) {
+  // NOTE: Update the CSV path and set dryRun to false before executing!
+  const csvPath = ""
+  const dryRun = true
+
+  await importRedirectsFromCsv({ csvPath, dryRun })
+  await db.destroy()
+}
```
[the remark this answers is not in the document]
dcshzj: Done in ccabf1e39 — the main-mode execution is wrapped in try/finally so `db.destroy()` always runs.

apps/studio/prisma/scripts/__tests__/importRedirectsFromCsv.test.ts
```diff
@@ -0,0 +1,250 @@
...
+        deletedAt: null,
+      },
+      { source: "/old-page", destination: "/new-page", deletedAt: null },
+    ])
+    expect(summary.sitesMatched).toBe(1)
+    expect(summary.importedCount).toBe(2)
+  })
+
+  it("skips rows for domains with no matching site and reports them", async () => {
+    // Arrange
```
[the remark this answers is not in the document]
dcshzj: Done in ccabf1e39 — added regression tests for a quoted destination containing a comma, and for a BOM'd header with whitespace around fields. Also re-ran the dry run against the real 38k-row production CSV: counts are identical to the previous parser, with no Papa parse warnings.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,385 @@
...
+        .values(rowsToInsert.slice(i, i + INSERT_CHUNK_SIZE))
+        // Re-running the import updates destinations in place and revives
+        // soft-deleted rows, mirroring the publish flow - the script is
+        // idempotent
+        .onConflict((oc) =>
+          oc.columns(["siteId", "source"]).doUpdateSet((eb) => ({
+            destination: eb.ref("excluded.destination"),
+            deletedAt: null,
+          })),
+        )
```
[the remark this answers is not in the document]
dcshzj: Done — the ON CONFLICT update now sets `updatedAt: new Date()` explicitly, since Kysely bypasses Prisma's `@updatedAt`. Added a test asserting updatedAt advances when a re-run changes a destination.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,385 @@
...
+    { skipEmptyLines: true },
+  )
+  for (const error of parseErrors) {
+    console.warn(`CSV parse warning at row ${error.row}: ${error.message}`)
+  }
+
+  const [header = [], ...dataRows] = rows
+  if (
+    header.map((field) => field.trim()).join(",") !== "domainName,source,target"
+  ) {
```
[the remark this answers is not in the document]
dcshzj: Done — `fullPermalinkOf` now threads a visited-set and short-circuits on a revisit, so a parentId cycle in corrupt data resolves to `/` instead of overflowing the stack.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,385 @@
...
+
+export const importRedirectsFromCsv = async ({
+  csvPath,
+  dryRun,
+}: ImportRedirectsFromCsvProps) => {
+  // Papa handles standard CSV quoting (fields containing commas) and CRLF
+  // line endings; the BOM is stripped up front so spreadsheet-exported files
+  // don't fail the header check below.
+  const { data: rows, errors: parseErrors } = Papa.parse<string[]>(
+    fs.readFileSync(csvPath, "utf8").replace(/^\uFEFF/, ""),
```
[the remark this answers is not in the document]
dcshzj: Done — added a guard that throws a clear "csvPath is empty" error before the readFileSync, with a test covering it.

mvincentong:
I think the inline redirect schema has drifted from the current Studio schema enough to matter for this migration. The branch already has `apps/studio/src/schemas/redirect.ts`, where `createRedirectSchema` lowercases sources, caps source length at 900, rejects reserved prefixes like `/_next`, rejects full-URL sources / spaces, normalises internal destinations, rejects backslashes / `..` path segments / internal anchors, validates `https://` destinations with a real public host, and rejects source == destination after normalisation.

The importer copy only does the smaller checks at the top of `importRedirectsFromCsv.ts` and then writes `parsed.data.source` / `parsed.data.destination` directly. So rows that Studio would reject, e.g. `/_next/static/chunk.js,/new-page`, `/old,/foo/../bar`, or a normalised self-redirect, can still be imported into `Redirect` despite the README/comment saying the script validates with the same rules as Studio.

Could we import `createRedirectSchema` / `normalizeRedirectPath` from `~/schemas/redirect` now, validate with `{ siteId, source, destination: target }`, and use the schema-normalised destination before the resource-reference lookup? A small regression test for a reserved source and a self-redirect should lock this down.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,399 @@
...
+  const cache = new Map<string, string>()
+  // Guards against a parentId cycle in corrupt data (A -> B -> A), which would
+  // otherwise recurse until the call stack overflows.
+  const fullPermalinkOf = (id: string, visited = new Set<string>()): string => {
+    const cached = cache.get(id)
+    if (cached !== undefined) return cached
+    if (visited.has(id)) return "/"
+    visited.add(id)
+    const resource = byId.get(id)
+    if (!resource) return "/"
```
seaerchin, line 163: question: how come it's safe to return "/" when the resource can't be found?

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,399 @@
...
+    const segment =
```
seaerchin: comment: no impact since this is a one off and as long as you can read it im g but this is more of `isIndexOrRoot`

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,399 @@
...
+          resource.publishedVersionId !== null &&
+          resource.parentId !== null,
+      )
+      .map((resource) => String(resource.parentId)),
+  )
+
+  const cache = new Map<string, string>()
+  // Guards against a parentId cycle in corrupt data (A -> B -> A), which would
+  // otherwise recurse until the call stack overflows.
+  const fullPermalinkOf = (id: string, visited = new Set<string>()): string => {
```
seaerchin, line 157: comment: do we not need to normalise permalinks before writing to cache?

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,399 @@
...
+  // Surface a clear message instead of an opaque ENOENT when csvPath is unset.
+  if (!csvPath) {
+    throw new Error(
+      "csvPath is empty - set the csvPath variable at the bottom of the script before running.",
+    )
+  }
+  // Papa handles standard CSV quoting (fields containing commas) and CRLF
+  // line endings; the BOM is stripped up front so spreadsheet-exported files
+  // don't fail the header check below.
+  const { data: rows, errors: parseErrors } = Papa.parse<string[]>(
```
seaerchin, line 218: comment: in my editor, there are some `^M` line endings which you might wanna sanitise 

comment: we do the read synchronously and hold it all in memory - this might be quite expensive and it feels off to me given that we process the db tx in chunks. not a big deal if we have enough memory, just raising as a potential issue because the size of teh csv is pretty large (40k rows)

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
```diff
@@ -0,0 +1,399 @@
...
+
+const destinationSchema = z
+  .string()
+  .min(1, { message: "Destination is required" })
+  .max(MAX_REDIRECT_PATH_LENGTH, { message: "Destination is too long" })
+  // Destinations must be a path on the same site ("/...") or an external
+  // https URL — anything else (http://, javascript:, ...) is rejected.
+  .refine((value) => value.startsWith("/") || value.startsWith("https://"), {
+    message: "Destination must start with '/' or 'https://'",
+  })
```
seaerchin, line 67: comment: this will fail a few rows in our existing csv. what we're doing now is migrate -> report which i don't feel quite fits a prod migration 

i'd be more comfortable doing the following:

1. parse csv
2. partition into [fit, don't fit]
3. for all don't fit, write to disk and report
4. we validate don't fit then either rerun or delete those rows from the actual csv
5. run only when the size of `don't fit` is 0 (ie, we fixed all or deleted from the migration csv)

seaerchin (commented):
i think this script should:

1. include a verification script where we check for the 301 redirect from cloudfront
2. be ran against staging and we verify existence of the redirection

the worst case that i can think of is that a commonly used redirect is broken, which might be the case if google serves the old URL

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
dcshzj: This stays migrate-then-report on purpose. We decided to keep the import lenient rather than apply Studio's stricter validation to legacy data — the goal is to preserve as many existing redirects as possible — so very little is rejected on validation now. The remaining skips are by-design: wildcard/query-param rules (no equivalent in the `Redirect` table) and domains not yet on Studio. Neither can be "fixed" to reach a zero unfit-count, so a hard gate that only runs when unfit == 0 would never pass. The dry run already reports every skipped row by category. I can additionally dump the rejected rows to a CSV in `output/` so they're a reviewable artifact rather than console-only — would that cover it, or did you specifically want the import to refuse to run while any genuinely-invalid rows remain?

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
dcshzj: Fixed. The map keys are built from individual permalink segments — already lowercase slugs with no slashes — so the keys themselves are normalised. The real gap was the lookup: it used `normalizeRedirectPath`, which collapses slashes but doesn't lowercase, so a mixed-case CSV target (e.g. `/About-Us`) would miss the `/about-us` key and fall back to a literal path. Added `normalizeRedirectSource` (= `normalizeRedirectPath` + lowercase, mirroring the schema) and used it for both the map key and the destination lookup.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
dcshzj: It's a defensive fallback for a dangling `parentId` (corrupt FK). In a consistent DB it's unreachable — every resource and its ancestors are loaded into `byId` and the parent FK is enforced — so returning `/` instead of throwing just keeps the one-off resilient; the only consequence is a resource with a broken parent chain resolving to root. Note this is only used while building the live-permalink map — it never becomes a stored redirect destination (an unresolved destination keeps the original CSV path). Can make it skip such a resource instead if you'd prefer it not contribute a `/` key.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
dcshzj: Done — extracted an `isIndexOrRoot(permalink)` helper.

apps/studio/prisma/scripts/importRedirectsFromCsv.ts
dcshzj: Two things here: the `^M`/CRLF endings are handled — `Papa.parse` is CRLF-aware and the BOM is stripped before the header check, so line endings don't affect parsing. On the in-memory read: the CSV is ~38k rows / a few MB, which `readFileSync` + parse holds comfortably; the chunking bounds the INSERT statement/parameter count, not memory, so streaming wouldn't buy us anything for a one-off. Left as-is.

dcshzj:
@seaerchin on the verification ask — added `verifyRedirectsLive.ts` (in 80fbfc1, runbook in `prisma/scripts/README.md`).

It reads the same `domainName,source,target` CSV and, for each row, requests `https://<domain>/<source>` with a browser `User-Agent` (so the WAF doesn't 403 the probe) and asserts a **301**; a mismatched `Location` is reported as a softer finding. Dry-run by default, `domainFilter` to verify one site at a time, `sampleSize` for a quick check vs all, and failures are written to `output/redirect-verification-failures.csv`.

Lol this Claude wrote some nonsense. Anyway the intent is that we will migrate to DB -> remove from redirects.production.csv -> trigger CodeBuild build -> invalidate -> verify redirects remain

dcshzj:
@mvincentong you're right that the inline copy has drifted, and that the original "can't import, that branch isn't merged" rationale is stale now that `~/schemas/redirect.ts` is on this branch.

This is a deliberate decision rather than an oversight, though. It's a one-off migration of legacy redirects from infra, so the goal is to preserve as many existing rules as possible rather than retroactively enforce Studio's stricter validation on data that predates it. Applying `createRedirectSchema` in full — reserved-prefix rejection, the source character whitelist, lowercasing, self-redirect rejection — would silently drop legacy redirects that are live and working today. Those rules apply going forward: anything created or edited through Studio after the migration is validated by `createRedirectSchema` as normal. We're just not rewriting history with them.

Where you're definitely right is that the comment and README overclaimed parity ("same rules as the Studio publish flow"). I've corrected both to describe the lenient subset and the rationale instead.

seaerchin:
faster close

dcshzj:
Closing this since the migration has completed

PR #2414, opened 2026-06-10
fix(gazette): single deletion email to Datadog with site admins cc'd

seaerchin:
## Problem

Deleting a gazette fanned out one notification email per site admin — N separate Postman sends per deletion. This had three issues:

1. **Mass emails**: every deletion triggered a burst of identical emails (one per admin), each a separate external API call.
2. **Wrong audience**: the recipient query included Isomer internal team members (`IsomerAdmin` users) who hold admin permissions on the site, when the notification is meant for the agency's own site admins (IMDA) only.
3. **No ops visibility**: there was no way for the Isomer team to be alerted (e.g. in Slack) when a gazette is deleted.

## Solution

Send a **single email per deletion**: the primary recipient is a Datadog events email address (`event-xxx@dtdg.co`), which turns every deletion into a Datadog event that can drive a monitor/Slack alert, and all agency site admins are **cc'd** on the same email.

**Breaking Changes**

**Features**:

- `sendMail` (`lib/mail.ts`) now supports `cc: string[]`. Non-whitelisted cc addresses are dropped with a warning (mirroring Postman's own behaviour of ignoring blacklisted cc recipients) instead of failing the whole send; the primary recipient still hard-fails if not whitelisted. `cc` is only included in the Postman payload when non-empty.
- `cc` is available on `BaseEmailTemplateData`, so any templated email can use it.
- New `DD_DELETION_EMAIL` constant — the Datadog events inbox used as the primary recipient for deletion notifications.

**Improvements**:

- Gazette deletion now sends exactly one email instead of one per admin, with admin emails deduped (a user can hold multiple Admin permission rows, and Postman rejects duplicate cc entries).
- The admins query excludes `IsomerAdmin` users via a `NOT EXISTS` subquery, so internal team members no longer receive agency-facing notifications.
- The email greeting is no longer personalised per recipient (`Hi everyone,`) since one email now goes to the whole group.

**Bug Fixes**:

- N/A

## Before & After Screenshots

**BEFORE**:

N/A — backend-only change.

**AFTER**:

N/A — backend-only change.

## Tests

Run `pnpm test:unit -- src/server/modules/gazette/__tests__/gazette.router.test.ts` (requires `pnpm setup:test` for the local Postgres). All 28 tests pass, including two new ones:

- `sends a single deletion email to Datadog with all site admins cc'd` — asserts `sendGazetteDeletionEmail` is called exactly once with `recipientEmail = DD_DELETION_EMAIL` and every site admin in `cc`.
- `excludes Isomer admins from the deletion email` — asserts a site admin who also has an `IsomerAdmin` row is not cc'd.

The mail service is mocked alongside the other external-service mocks in the delete suite, so tests never hit the whitelist/Postman path.

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A


apps/studio/src/constants/gazette.ts
```diff
@@ -1 +1,3 @@
 export const ALLOWED_GAZETTE_DELETION_TIMEFRAME_IN_MINUTES = 15
+
+export const DD_DELETION_EMAIL = "event-pqck7p7o@dtdg.co"
```
adriangohjw: question/issue(?): is this supposed to be a secret? wondering if anyone can just send email to this?

apps/studio/src/features/mail/service.ts
```diff
@@ -39,11 +39,22 @@ async function sendEmailWithTemplate({
...
 
+  // Drop malformed cc addresses rather than failing the send for everyone
+  const validCc = data.cc?.filter((email) => {
+    if (isValidEmail(email)) return true
+    logger.error({
+      error: "Invalid cc email format",
+      email,
+    })
+    return false
+  })
```
adriangohjw, line 50: question: this is defence-in-depth for future ya? since this current PR means we got the email from the User table

apps/studio/src/server/modules/gazette/gazette.router.ts
```diff
@@ -897,18 +900,32 @@ export const gazetteRouter = router({
...
-            recipientEmail: email,
-          })
-        }),
-      )
+      // Send a single email: the Datadog events address is the primary
+      // recipient (so deletions always alert ops, even with zero admins) and
+      // every site admin is cc'd. Dedupe is required: a user can hold
+      // multiple Admin permission rows, and Postman rejects duplicate cc
+      // entries.
+      const cc = [...new Set(admins.map(({ email }) => email))]
```
adriangohjw, line 920: nit: why need this? shouldn't it be unique based on the query above? or defence-in-depth?

apps/studio/src/lib/mail.ts
```diff
@@ -18,13 +19,38 @@ export const sendMail = async (params: SendMailParams): Promise<void> => {
...
+      subject: params.subject,
+    })
+  }
+  const cc = ccResults.filter((r) => r.isWhitelisted).map((r) => r.email)
+  const payload = {
+    recipient: params.recipient,
+    subject: params.subject,
+    body: params.body,
+    ...(cc.length > 0 && { cc }),
+  }
```
adriangohjw: suggestion: can use lodash's partition, might be more readable

example
```ts
import { partition } from "lodash-es"

const ccResults = await Promise.all(
  (params.cc ?? []).map(async (email) => ({
    email,
    isWhitelisted: await isEmailWhitelisted(email),
  })),
)

const [whitelistedCc, droppedCc] = partition(
  ccResults,
  (r) => r.isWhitelisted,
)

if (droppedCc.length > 0) {
  logger.warn({
    error: "Dropping non-whitelisted cc recipients",
    cc: droppedCc.map((r) => r.email),
    subject: params.subject,
  })
}

const cc = whitelistedCc.map((r) => r.email)
```

or even combine
```ts
const [cc, droppedCc] = partition(
  await Promise.all(
    (params.cc ?? []).map(async (email) => ({
      email,
      isWhitelisted: await isEmailWhitelisted(email),
    })),
  ),
  (r) => r.isWhitelisted,
)

if (droppedCc.length > 0) {
  logger.warn({
    error: "Dropping non-whitelisted cc recipients",
    cc: droppedCc.map((r) => r.email),
    subject: params.subject,
  })
}
```

apps/studio/src/constants/gazette.ts
seaerchin: err technically not a secret - it's supposed to be kept for internal alerting though, if you think this should be kept as an env var, i'm also happy to migrate over to infra

adriangohjw: But how does this work ah? I assume anyone can email this and it will trigger alert? Similar to uptimerobot heartbeat monirot

seaerchin: it just posts to the #isomer-alarm channel, doesn't trigger on call

apps/studio/src/server/modules/gazette/gazette.router.ts
seaerchin: more of defence in depth cos we hold a `unique` on `ResourcePermission` over the `resource.id`; happy to remove if you want it gone

adriangohjw: No need,l, can keep, just wanna clarify. Thanks!

apps/studio/src/constants/gazette.ts
adriangohjw: Okay then I think we should treat this as a secret haha need to generate one and pass it in via SSM, don't want off-chance of ppl spamming this hahaha

seaerchin: sounds good, lemme do that instead

adriangohjw (approved):

PR #2415, opened 2026-06-10
fix(studio): prevent page type from being changed to RootPage via updateSettings

adriangohjw:
## Problem

A VAPT finding (low severity) identified that any page could be escalated to `RootPage` type by calling `page.updateSettings` with `type: \"RootPage\"` in the request payload. Because the handler spread the full input `settings` (which included the `type` field) directly into the Kysely `.set()` call, an authenticated user with edit permissions could mutate the `type` column of any page resource.

## Solution

Destructure `type` out of the `settings` spread in the `updateSettings` handler so it is never passed to the database update. The resource type is now immutable via this endpoint — only `title` and `permalink` (where applicable) are written.

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- N/A

**Bug Fixes**:

- `page.router.ts` — exclude `type` from the `.set()` payload in `updateSettings` so a caller cannot change a resource's type to `RootPage` (or any other type).
- `page.router.test.ts` — add regression test asserting that calling `updateSettings` with `type: \"RootPage\"` on a regular `Page` resource leaves its type unchanged.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] As an admin user, open any regular page in Studio and capture its `pageId` and `siteId` from the network tab.
- [ ] Using a tool such as Postman or browser DevTools, call the `page.updateSettings` tRPC mutation with `type: "RootPage"` and the captured IDs.
- [ ] Verify the response `type` field is **not** `RootPage` and matches the original page type.
- [ ] Confirm no `RootPage`-type duplicates appear in the site's resource list.

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

### DIFF
--- a/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
+++ b/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
@@ -2240,6 +2240,33 @@ describe("page.router", async () => {
       expect(result).toMatchObject(expectedSettings)
     })
 
+    it("should not allow changing a page type to RootPage", async () => {
+      // Arrange
+      const { site, page } = await setupPageResource({
+        resourceType: "Page",
+      })
+      await setupAdminPermissions({
+        userId: session.userId ?? undefined,
+        siteId: site.id,
+      })
+
+      // Act
+      await caller.updateSettings({
+        siteId: site.id,
+        pageId: Number(page.id),
+        type: "RootPage",
+        title: "Attempted RootPage",
+      })
+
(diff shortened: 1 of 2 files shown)


seaerchin (approved):
comment: we use a discriminated union to strictly type the input but wondering if it's possible to run a superRefine post parse to strip the `type` key for backend ingestion

PR #2439, opened 2026-06-13
fix: aria-hidden for info box icon

### LINKED ISSUE #2363: (warden) aria-hidden prop silently dropped from InfoBoxIcon, leaving decorative icon exposed to screen readers
# aria-hidden prop silently dropped from InfoBoxIcon, leaving decorative icon exposed to screen readers

**Skill:** code-review

**Severity:** 🟡 MEDIUM

**ID:** `WTK-PNP`

**Confidence:** high

**Location:** `packages/components/src/templates/next/components/complex/InfoCols/InfoCols.tsx:110`

## Description

At line 110, `aria-hidden="true"` is passed to `InfoBoxIcon`, but the component’s props type only declares `{ icon?: SupportedIconName; hasLink: boolean }`. The component destructures only those two props and renders `<Icon className={...} />` with no spread of additional attributes, so the `aria-hidden` prop is discarded at runtime. The underlying `react-icons` SVG will be rendered without `aria-hidden`, causing screen readers to potentially announce a decorative icon that the author intended to hide. This is also a TypeScript compile error since `aria-hidden` is not part of the declared props interface, which will fail `pnpm typecheck`.

## Verification

Confirmed by reading InfoBoxIcon (lines 68–86): props destructure only `icon` and `hasLink`; the `<Icon>` at line 80 only receives `className`. The aria-hidden attribute is not declared in the props type and is never forwarded. Checked `packages/components/src/common/icons.ts` to confirm `SUPPORTED_ICONS_MAP` values are react-icons `IconType` components that render SVG elements.

jiaxinnns:
## Problem

Decorative icons in InfoCols were intended to be hidden from screen readers via aria-hidden="true", but the prop was passed to InfoBoxIcon without being forwarded to the underlying <svg>. The attribute was dropped at runtime and caused a TypeScript error.

Closes #2363 

## Solution

Moved aria-hidden onto the <Icon> element inside InfoBoxIcon and removed the invalid prop from the call site. 

**Breaking Changes**

  - Details ...

**Features**:

NIL

**Improvements**:

NIL

**Bug Fixes**:

- Fix aria-hidden being silently dropped on InfoBoxIcon in InfoCols, restoring intended screen-reader behavior and fixing the TypeScript error.

## Before & After Screenshots

**BEFORE**:

Isomer-2

**AFTER**:

Isomer-1

## Tests

- [ ] Open Storybook → Next/Components/InfoCols
- [ ] Inspect an info-box icon <svg> in DevTools and confirm aria-hidden="true" is present

**New scripts**:

NIL

**New dependencies**:

NIL

**New dev dependencies**:

NIL

### DIFF
--- a/packages/components/src/templates/next/components/complex/InfoCols/InfoCols.tsx
+++ b/packages/components/src/templates/next/components/complex/InfoCols/InfoCols.tsx
@@ -107,9 +107,7 @@ const InfoBoxes = ({
               className={compoundStyles.infoBox()}
               isExternal={isExternalLink}
             >
-              {icon && (
-                <InfoBoxIcon icon={icon} hasLink={hasLink} />
-              )}
+              {icon && <InfoBoxIcon icon={icon} hasLink={hasLink} />}
 
               <h3
                 className={compoundStyles.infoBoxTitle({
--- a/packages/components/src/templates/next/components/complex/InfoCols/InfoCols.tsx
+++ b/packages/components/src/templates/next/components/complex/InfoCols/InfoCols.tsx
@@ -78,6 +78,7 @@ const InfoBoxIcon = ({
 
   return (
     <Icon
+      aria-hidden
       className={compoundStyles.infoBoxIcon({
         hasLink,
       })}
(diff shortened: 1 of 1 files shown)


adriangohjw (approved):

adriangohjw:
@jiaxinnns can you help to fix the formatting error raised during CI? should be fixable by `pnpm format:fix` - thanks!

PR #2466, opened 2026-06-16
feat(redirect): add create validation, preflight warnings, and loop guard

dcshzj:
## Problem

Studio users can create individual redirects, but until now nothing stopped them from creating redirects that are broken on arrival: a redirect pointing a path to itself, a pair of redirects that trap visitors in a loop, a destination that is a full external-looking URL stuffed into the source field, or a destination pointing at a page that doesn't exist (or isn't published) yet. There was also no feedback loop warning users about destinations that themselves redirect elsewhere.

Closes [ISOM-2263](https://linear.app/ogp/issue/ISOM-2263) — Add validation to prevent a redirect source from being created when a page exists, and vice versa.
Closes [ISOM-2264](https://linear.app/ogp/issue/ISOM-2264) — Add validation logic for creating redirects.

## Solution

Adds a validation layer for creating redirects, shared between the client form and the server so the two can never drift, plus a non-blocking preflight that surfaces "this might not do what you expect" warnings before the user commits.

**Blocking errors** (reject creation):
- Invalid source/destination paths — control characters, backslashes, `..` traversal segments, and a source that is a full URL (contains `://`).
- Recreating a redirect that already exists for the same (normalised) source.
- A redirect whose source and destination are the same path.
- A redirect loop up to 1 level deep (e.g. `/a → /b` while `/b → /a` already exists).

**Non-blocking warnings** (surfaced inline, don't block):
- The destination is itself the source of another redirect, so visitors end up somewhere else.
- The destination is an internal path that doesn't exist, or exists but isn't published yet.

**Breaking Changes**

  - Adds a new read-only `redirect.validate` query and tightens `redirect.create`. Existing redirects are unaffected.

**Features**:

- New `redirect.validate` tRPC query: a read-only preflight (gated on `read`) that returns blocking `errors` and non-blocking `warnings` for a would-be redirect without mutating anything.
- `getResourceByFullPermalink` in the resource service: resolves a full permalink (e.g. `/foo/bar`) to the resource that serves it, walking permalink segments from the site root and resolving Folder/Collection containers to their IndexPage child. Used to check whether a redirect destination is a real, published page.
- The Add Redirect form now fetches warnings on blur once inputs are sync-valid and renders them as inline warning infoboxes under the destination field.

**Improvements**:

- Path normalisation (`normalizeRedirectPath`) is now exported and reused for both sources and internal destinations, so equivalent inputs (`/foo`, `foo/`, `/foo//`) canonicalise identically — keeping the `(siteId, source)` unique constraint meaningful and making same-source/destination comparison reliable. External `https://` destinations are left verbatim (trailing slash, query, fragment are meaningful off-site).
- The create schema is split into a refinement-free `createRedirectObjectSchema` (so callers can still `.omit()` / `.extend()`) plus a `superRefine`d `createRedirectSchema`. The form reuses the object schema and re-applies the same cross-field refinement, so client and server validate identically.
- Redirect validation copy is centralised in `REDIRECT_MESSAGES` / `RedirectValidationCode`, shared by the preflight, the create-time guards, and the form, so server and client messages can't drift.
- Create-time errors (`CONFLICT` for already-exists, `UNPROCESSABLE_CONTENT` for loop) are now surfaced inline on the relevant field instead of as a generic toast.

**Bug Fixes**:

- Destination validation now rejects control characters (incl. CR/LF/NUL) and backslashes for both internal paths and external `https://` URLs. A destination is persisted verbatim and later emitted into the published site's redirect rules (S3 object metadata → CloudFront `Location` header), so a CR/LF could otherwise reach a live response header.

## Tests

`pnpm test:unit` (from `apps/studio`) covers the new schema rules and service logic:
- `src/schemas/__tests__/redirect.test.ts` — source/destination validation, normalisation, and the source-equals-destination cross-field refinement.
- `src/server/modules/redirect/__tests__/redirect.router.test.ts` — `validate` preflight (errors + warnings) and the create-time loop re-enforcement.

The `Redirects.stories.tsx` Storybook stories exercise the inline error and warning states in the Add Redirect form.

- [ ] Go to a site's **Settings → Redirects** page.
- [ ] In **Add redirect**, enter a source and a destination set to the same path (e.g. source `/contact`, destination `/contact`) and confirm the form blocks with "You can't redirect a URL to itself."
- [ ] Enter a full URL in the source field (e.g. `https://example.com/foo`) and confirm it is rejected with a message to enter the path only.
- [ ] Create a redirect `/a → /b`, then attempt `/b → /a` and confirm it is blocked with the redirect-loop message.
- [ ] Attempt to recreate a redirect for a source that already has a live redirect and confirm it is blocked with "This page is already being redirected."
- [ ] Enter a destination pointing to an internal path that does not exist (or exists but is unpublished), blur the field, and confirm a non-blocking warning infobox appears under the destination — and that creation is still allowed.
- [ ] Enter a destination that is itself the source of another redirect, blur, and confirm the "this page already redirects to …" warning appears.
- [ ] Confirm a valid, real, published internal destination (and a valid `https://` external destination) shows no warnings and creates successfully, publishing immediately.

### DIFF
--- a/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
@@ -21,12 +21,12 @@ import {
 } from "@opengovsg/design-system-react"
 import { useState } from "react"
 import { BiLinkAlt, BiPlus, BiRightArrowAlt } from "react-icons/bi"
+import { REDIRECT_MESSAGES } from "~/constants/redirect"
 import {
   BRIEF_TOAST_SETTINGS,
   SETTINGS_TOAST_MESSAGES,
 } from "~/constants/toast"
 import { useZodForm } from "~/lib/form"
-import { REDIRECT_MESSAGES } from "~/schemas/redirect"
 
 import { useCreateRedirect, useValidateRedirect } from "../api"
 import { addRedirectSchema, type AddRedirectInput } from "../types"
--- a/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
@@ -7,6 +7,7 @@ import {
   Input,
   InputGroup,
   InputLeftAddon,
+  Stack,
   Text,
   useDisclosure,
 } from "@chakra-ui/react"
@@ -14,17 +15,20 @@ import {
   Button,
   FormErrorMessage,
   FormLabel,
+  Infobox,
   Link,
   useToast,
 } from "@opengovsg/design-system-react"
+import { useState } from "react"
 import { BiLinkAlt, BiPlus, BiRightArrowAlt } from "react-icons/bi"
 import {
   BRIEF_TOAST_SETTINGS,
   SETTINGS_TOAST_MESSAGES,
 } from "~/constants/toast"
 import { useZodForm } from "~/lib/form"
+import { REDIRECT_MESSAGES } from "~/schemas/redirect"
 
-import { useCreateRedirect } from "../api"
+import { useCreateRedirect, useValidateRedirect } from "../api"
 import { addRedirectSchema, type AddRedirectInput } from "../types"
 import { SelectDestinationPageModal } from "./SelectDestinationPageModal"
 
@@ -39,7 +43,10 @@ export const AddRedirectCard = ({
     register,
     handleSubmit,
     reset,
+    setError,
+    clearErrors,
     watch,
+    getValues,
     setValue,
     formState: { errors },
   } = useZodForm<typeof addRedirectSchema>({
@@ -57,6 +64,30 @@ export const AddRedirectCard = ({
   const [source, destination] = watch(["source", "destination"])
   const isAddDisabled = !source?.trim() || !destination?.trim()
 
+  // Non-blocking warnings (e.g. the destination doesn't exist / isn't
+  // published) are fetched on blur once the inputs are sync-valid.
+  const [validateInput, setValidateInput] = useState<AddRedirectInput | null>(
+    null,
+  )
+  const { warnings } = useValidateRedirect(siteId, validateInput)
+
+  const checkForWarnings = () => {
+    const parsed = addRedirectSchema.safeParse(getValues())
+    setValidateInput(parsed.success ? parsed.data : null)
+  }
+  // Drop any shown warning the moment a field changes, so it never lingers
+  // against edited input. (reset() doesn't fire onChange, so a successful add
+  // clears it explicitly in onSuccess below.)
+  const clearWarnings = () => setValidateInput(null)
+  // On edit, also clear any server-set error on that field — otherwise an
+  // inline "already redirected" / "loop" error (set via setError, which the
+  // onSubmit-mode form doesn't revalidate) would linger over freshly-typed
+  // input.
+  const clearFieldFeedback = (field: keyof AddRedirectInput) => () => {
+    clearWarnings()
+    clearErrors(field)
+  }
+
   const onSubmit = ({ source, destination }: AddRedirectInput) => {
     // source arrives normalised (leading slash, no trailing slash) by the
     // shared schema's transform. An internal-path destination is converted to a
@@ -66,27 +97,33 @@ export const AddRedirectCard = ({
       {
         onSuccess: () => {
           reset()
+          clearWarnings()
           toast({ ...SETTINGS_TOAST_MESSAGES.success, status: "success" })
         },
-        // The inputs are left untouched on error so the user can adjust
-        // the source instead of retyping everything
+        // The inputs are left untouched on error so the user can fix the
+        // offending field. The schema covers the synchronous rules; these are
+        // the DB-level checks the server re-enforces on create — surface them
+        // inline on the relevant field rather than as a toast. redirect.create
+        // only throws these codes; anything else is unexpected.
         onError: (error) => {
           switch (error.data?.code) {
             case "CONFLICT":
-              toast({
-                title: "A redirect already exists for this path",
-                description: `Delete the redirect for ${source} first if you want to change where it points to.`,
-                status: "error",
+              setError("source", { message: REDIRECT_MESSAGES.alreadyExists })
+              break
+            case "PRECONDITION_FAILED":
+              setError("source", {
+                message: REDIRECT_MESSAGES.sourceIsExistingPage,
               })
               break
+            // An internal-path destination is resolved to a page reference on
+            // create, which 404s when no published page lives at that path.
             case "NOT_FOUND":
-              toast({
-                title: "That page doesn't exist",
-                description:
-                  "Check the destination, or pick a page with “Link to a page”.",
-                status: "error",
+              setError("destination", {
+                message: REDIRECT_MESSAGES.destinationNotLive,
               })
               break
+            case "UNPROCESSABLE_CONTENT":
+              setError("destination", { message: REDIRECT_MESSAGES.loop })
             default:
               toast({
                 title: "Failed to add redirect",
@@ -131,7 +168,13 @@ export const AddRedirectCard = ({
             >
               /
             </InputLeftAddon>
-            <Input placeholder="redirect-from" {...register("source")} />
+            <Input
+              placeholder="redirect-from"
+              {...register("source", {
+                onBlur: checkForWarnings,
+                onChange: clearFieldFeedback("source"),
+              })}
+            />
           </InputGroup>
           <FormErrorMessage>{errors.source?.message}</FormErrorMessage>
         </FormControl>
@@ -154,7 +197,10 @@ export const AddRedirectCard = ({
           <Input
             placeholder="/redirect-to"
             size="xs"
-            {...register("destination")}
+            {...register("destination", {
+              onBlur: checkForWarnings,
+              onChange: clearFieldFeedback("destination"),
+            })}
           />
           <FormErrorMessage>{errors.destination?.message}</FormErrorMessage>
           <Link
@@ -168,6 +214,15 @@ export const AddRedirectCard = ({
             <Icon as={BiLinkAlt} mr="0.25rem" />
             Link to a page
           </Link>
+          {warnings.length > 0 && (
+            <Stack spacing="0.5rem" mt="0.5rem">
+              {warnings.map((warning) => (
+                <Infobox key={warning.code} variant="warning" size="sm">
+                  {warning.message}
+                </Infobox>
+              ))}
+            </Stack>
+          )}
         </FormControl>
 
         <Box flexShrink={0}>
--- a/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
@@ -115,13 +115,6 @@ export const AddRedirectCard = ({
                 message: REDIRECT_MESSAGES.sourceIsExistingPage,
               })
               break
-            // An internal-path destination is resolved to a page reference on
-            // create, which 404s when no published page lives at that path.
-            case "NOT_FOUND":
-              setError("destination", {
-                message: REDIRECT_MESSAGES.destinationNotLive,
-              })
-              break
             case "UNPROCESSABLE_CONTENT":
               setError("destination", { message: REDIRECT_MESSAGES.loop })
               break
--- a/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
@@ -124,6 +124,7 @@ export const AddRedirectCard = ({
               break
             case "UNPROCESSABLE_CONTENT":
               setError("destination", { message: REDIRECT_MESSAGES.loop })
+              break
             default:
               toast({
                 title: "Failed to add redirect",
--- a/apps/studio/src/schemas/redirect.ts
+++ b/apps/studio/src/schemas/redirect.ts
@@ -4,11 +4,19 @@ import { z } from "zod"
 import { generateBigIntSchema } from "./common"
 import { offsetPaginationSchema } from "./pagination"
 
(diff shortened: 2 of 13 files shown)


apps/studio/src/server/modules/redirect/__tests__/redirect.router.test.ts
seaerchin, line 1: chonky file

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -57,6 +64,30 @@ export const AddRedirectCard = ({
...
+  const clearWarnings = () => setValidateInput(null)
+  // On edit, also clear any server-set error on that field — otherwise an
+  // inline "already redirected" / "loop" error (set via setError, which the
+  // onSubmit-mode form doesn't revalidate) would linger over freshly-typed
+  // input.
+  const clearFieldFeedback = (field: keyof AddRedirectInput) => () => {
+    clearWarnings()
+    clearErrors(field)
+  }
+
```
seaerchin, line 90: question: not entirely convinced this loop makes sense - why not just submit then set that error as the message? we centralise the validity check where it's actually required and save this state management loop

```ts
        onError: (error) => {
          if (error.data?.code === "CONFLICT") {
            formMethods.setError(
              "permalink",
              { message: error.message },
              { shouldFocus: true },
            )
          } else {
            console.error(error)
          }
        },
```

this is on `CreatePageWizardContext.tsx`

seaerchin: (can add to conventions if we this makes sense lol)

apps/studio/src/schemas/redirect.ts
```diff
@@ -28,61 +45,188 @@ const INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/
...
-  // Collapse a leading run of slashes on an internal path so a protocol-relative
-  // "//evil.com" can't pass as an open redirect ("//evil.com" -> "/evil.com").
+  // Control characters (incl. CR/LF/NUL) and backslashes are never valid in a
+  // redirect destination — internal path or external https URL alike. This is
+  // NOT gated on internal paths: a destination is persisted verbatim and later
+  // emitted into the published site's redirect rules (S3 object metadata and
+  // ultimately the CloudFront Location header), so a CR/LF in an https URL
+  // could otherwise reach a live response header.
+  .refine((value) => !INVALID_PATH_CHARS_REGEX.test(value), {
+    message: "Add a valid URL.",
```
seaerchin: comment: this copy isn't very helpful to our end user and sounds very cold - wondering if we should maek it clearer?

apps/studio/src/schemas/redirect.ts
```diff
@@ -28,61 +45,188 @@ const INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/
...
+  })
+  // "../" path traversal only has meaning for an internal path; an external
+  // https URL may legitimately contain ".." in its own path, so scope this one
+  // to internal destinations.
+  .refine(
+    (value) =>
+      !value.startsWith("/") || !trimSlashes(value).split("/").includes(".."),
```
seaerchin: comment: if we maek the external url the site's own url, this comment doesn't hold anymore. we should just ban `..` straight out imo cos it's just for path traversal _which the user can already control on their end for redirects_

ie, the meaning of `..` in a url is to go up a level, so `/a/b/../c` would lead to `/a/c` - the user can already set the URL directly here (and in an external url, there's little to no point exposing this cos the site itself would collapse the path) so i think it doesn't make sense to have this

apps/studio/src/schemas/redirect.ts
```diff
@@ -28,61 +45,188 @@ const INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/
...
+// Messages that interpolate the path (the loop detail and the chain warning)
+// are produced in the service since they aren't reused by the client.
+export const REDIRECT_MESSAGES = {
+  alreadyExists: "This page is already being redirected.",
+  loop: "This will trap visitors in a never-ending loop.",
+  destinationNotLive:
+    "This page doesn't exist on your site yet. Make sure the page is live before publishing this redirect.",
+  sourceIsExistingPage:
+    "A live page already uses this URL. The redirect would hide it. Move or unpublish that page first.",
+} as const
```
seaerchin: comment: prefer to have these live in `constants`

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -106,6 +117,178 @@ const getByUser = async (byUserId: string) =>
...
+  // doesn't involve `source` (which would otherwise spin until the hop cap).
+  const visited = new Set([normalizedDestination])
+  let current = target
+  let isLoop = closesLoop(current)
+  let hops = 1
+  while (!isLoop && current.startsWith("/") && hops < MAX_REDIRECT_CHAIN_HOPS) {
+    const normalizedCurrent = normalizeRedirectSource(current)
+    if (visited.has(normalizedCurrent)) {
+      break
+    }
```
seaerchin: question: how come we don't need to set `isLoop: true` when we find a loop here?

seaerchin: answer: cos we set the `isLoop = closesLoop(current)` in the prev cycle

seaerchin (approved):
not entirely convinced by allowing `..` and the error message handling seems abit off but i'm happy to approve as neither of these are blockers

seaerchin (commented):
## Summary

This PR adds a shared redirect-validation layer (client + server) plus a read-only `redirect.validate` preflight. The schema (`apps/studio/src/schemas/redirect.ts`) is split into a refinement-free `createRedirectObjectSchema` + a `superRefine`d `createRedirectSchema`, with `normalizeRedirectPath`/`normalizeRedirectSource` exported so source/destination canonicalise identically on both sides. The server adds `validateRedirect` (`redirect.service.ts:224`) returning blocking `errors` + non-blocking `warnings`, backed by `getChainedRedirect` (`redirect.service.ts:168`, multi-hop loop detection with a `visited` set and a 10-hop cap) and `getResourceByFullPermalink` (`resource.service.ts:719`, walks permalink segments from the root page and resolves Folder/Collection → IndexPage). `createRedirect` (`redirect.service.ts:330`) re-enforces the loop guard (`UNPROCESSABLE_CONTENT`) and the source-shadows-published-page guard (`PRECONDITION_FAILED`) inside the tx, and now stores a literal path for not-yet-live internal destinations instead of throwing `NOT_FOUND`. The form fetches warnings on blur and maps server error codes to inline field errors.

Overall: solid, well-commented, well-tested work. The security-sensitive bits (CRLF/control-char rejection on destinations, open-redirect collapse of `//evil.com`, the tightened external-URL check rejecting `https://https://...`) are correct and covered by tests. Findings below are mostly correctness/maintainability nits; nothing I'd block on. The previously-raised Copilot threads (switch fall-through, `NOT_FOUND`-vs-warning divergence) are already addressed in later commits.

## Findings

**praise:** The destination CRLF/control-char refinement (`redirect.ts:119-127`) is applied to *all* destinations, not just internal paths, with a comment tracing the exact sink (S3 metadata → CloudFront `Location` header). That is the right call and the response-header-injection test cases (`redirect.test.ts:456`) lock it in. The `//evil.com` → `/evil.com` open-redirect collapse via `normalizeRedirectPath` is also a nice catch. Confidence: high

**issue: (non-blocking)** The loop `description` is misleading for multi-hop loops. In `validateRedirect` (`redirect.service.ts:261`) the copy is `${chained.normalizedDestination} already redirects to ${source}`, but for an A→B→C→A loop, `normalizedDestination` is `/b` and `/b` does *not* redirect to `/a` — it redirects to `/c`. The user is told "/b already redirects to /a", which is factually wrong and will confuse anyone trying to find the offending redirect to delete. Since `getChainedRedirect` already walks the full chain, consider either describing the actual closing hop or using generic copy that doesn't assert a direct `dest → source` link. Confidence: high

**issue: (non-blocking, performance)** `getResourceByFullPermalink` (`resource.service.ts:740-759`) issues one sequential DB query per path segment (N+1), and `getChainedRedirect` (`redirect.service.ts:201-217`) issues up to 10 sequential round-trips, each potentially followed by another `getResourceFullPermalinks` query inside `resolveStoredDestination`. `validateRedirect` then runs these helpers sequentially (`redirect.service.ts:234,245,256`) even though the already-exists, source-shadow, and chain checks are independent and could be `Promise.all`'d. The source cap is 900 chars, so a deep destination path is a lot of serial queries on a `read`-gated endpoint that fires on every blur. Not exploitable (bounded, authed), but worth a `Promise.all` on the independent preflight checks and/or noting the segment-walk cost. Confidence: medium

**question:** In `getResourceByFullPermalink` (`resource.service.ts:740`), the segment walk excludes `IndexPage`/`FolderMeta`/`CollectionMeta` from matching a path segment, but does it correctly handle a Collection's *child* pages (CollectionLink / CollectionPage) at depth >1? The tests cover Folder→leaf and Folder→IndexPage but I don't see a CollectionPage-under-Collection case. If a published collection item's permalink should resolve, a missing test could hide a wrong "destination not found" warning. Non-blocking since warnings don't block creation, but worth confirming the intended resolution semantics for collection items. Confidence: low

**note:** The create-time loop and source-shadow guards (`redirect.service.ts:363-389`) run inside the tx but are plain reads, so the two-admin race is acknowledged in the comments and deemed recoverable — agreed, that's a reasonable trade-off given the already-exists check is still constraint-backed. No change requested; flagging that the transaction here buys atomicity for the insert/audit-log but not mutual exclusion for these guards, which the comments already state. Confidence: high

**nitpick:** `MAX_REDIRECT_CHAIN_HOPS` (`redirect.service.ts:156`) and `RESERVED_SOURCE_PREFIXES` (`redirect.ts:67`) are inline consts; seaerchin's existing comment already asked to move validation constants to a `constants` location. Low priority and consistent with the rest of the file, but worth folding in if you touch this again. Confidence: low

dcshzj:
* **#2508** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2466** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2466?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2490** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/studio/src/schemas/redirect.ts
dcshzj: @sehyunidaaa could you weigh in on this copy? It's the message shown when a destination isn't a valid target. Proposed warmer version: **"Enter a page on this site (like /contact-us) or a full https:// link"** — happy to take whatever wording you prefer. One note: this same string is currently reused by three checks (invalid scheme, control characters, and `..` segments), so if you'd rather they read differently we can split them into distinct messages.

apps/studio/src/schemas/redirect.ts
dcshzj: Went with your preference — `..` is now rejected outright for **all** destinations (internal path and external https alike), and the internal-only scoping + comment are gone. Added a schema test covering both `/foo/../bar` and `https://www.example.gov.sg/foo/../bar`.

apps/studio/src/schemas/redirect.ts
dcshzj: Done — moved `REDIRECT_MESSAGES`, `RedirectValidationCode`, `RESERVED_SOURCE_PREFIXES`, and the validation-result contract (`RedirectValidationIssue`/`RedirectValidationResult`) out to `~/constants/redirect`. The schema, service, and form now all import from there. `MAX_REDIRECT_CHAIN_HOPS` is gone entirely, since loop detection is now single-hop.

apps/studio/src/schemas/redirect.ts
sehyunidaaa: @seaerchin Thanks for the feedback, agree it's a bit cold!

@dcshzj For the message, can we use: "Enter a valid path (/path-to-page) or external URL (www.google.com)."

Can we handle the 3 checks on our end instead of communicating each through an error message? 
- Invalid scheme - are we rejecting destinations starting with `http://` or `www.`? What about those without www. (google.com)? Better experience to accept all and append `https://` to those without a specified protocol, and switch http:// to https://. 
- Control characters - can we strip them off?
- .. segments - can we resolve them and add the clean destination to the table? [Edit - saw below we aren't accepting these anymore. What will happen?]

For all three cases, we could fix on submission, because I believe in all 3 cases the user intention is unambiguous and getting users to fix small issues like this is very frustrating for them, feels like those strict password setting flows. Wdyt?

dcshzj: @sehyunidaaa have updated the wording! For the 3 cases:
1. We previously reject everything that does not start with a `/` or `https://`, but have updated it to allow for `http://`. I think handling the case without the `www.` in front would be quite hard to do, cos it can be a bit ambiguous (e.g. `https://test.zip` vs `/test.zip`), so would prefer to surface it as an error for the user. I have implemented the check to automatically change `http://` to `https://`.
2. Implemented this as well.
3. For this, it is only applicable for internal links and is more of a security defence. It is very unlikely for users to be putting this as an input on their side. For external links, we are not touching as this is more for the external site to handle.

Lmk what you think of this approach!

sehyunidaaa: Okie! Sgtm. 

Then should we say: "Enter a valid path (/path-to-page) or URL (starts with www., http://, or https://.)."

dcshzj: I think just `Enter a valid path (/path-to-page) or URL (starts with http://, or https://).`? Or are you looking at also automatically adding the `https://` if the user adds the `www.` in front?

But actually ok ah I think is valid, will add the auto-add part for `https://`

PR #2468, opened 2026-06-16
fix(components): filter CKAN internal columns from DGS searchable table

adriangohjw:
## Summary

- CKAN datastore auto-generates `_id` (row counter) and `_full_text` (PostgreSQL tsvector) on every dataset ingested into data.gov.sg
- These were being included in `extractColumnMetadata`, causing `_id` to appear as the first column header (it always has `index: "0"`)
- This also shifted user columns out of their expected positions when `MAX_NUMBER_OF_COLUMNS` trimming was in play

## Test plan

- [ ] Verify a DGS searchable table using dataset `d_21359ecc93954aea7b8c0d48157b11d9` — first column should now show `Course_Id` instead of `_id`
- [ ] Check that column count matches the number of user-defined columns in the source dataset (no phantom `_id` column)

### DIFF
--- a/packages/components/src/utils/dgs/fetchDgsMetadata.ts
+++ b/packages/components/src/utils/dgs/fetchDgsMetadata.ts
@@ -64,15 +64,23 @@ const extractColumnMetadata = (
   data: FetchDgsMetadataResponse,
 ): FetchDgsMetadataOutput["columnMetadata"] => {
   try {
-    return Object.values(data.data.columnMetadata.metaMapping)
-      .sort((a, b) => Number(a.index) - Number(b.index))
-      .reduce<NonNullable<FetchDgsMetadataOutput["columnMetadata"]>>(
-        (acc, mapping) => {
-          acc.push([mapping.name, mapping.columnTitle])
-          return acc
-        },
-        [],
-      )
+    return (
+      Object.values(data.data.columnMetadata.metaMapping)
+        // CKAN datastore auto-generates _id (row counter) and _full_text
+        // (PostgreSQL tsvector for full-text search) on every dataset — they are
+        // never part of the source data and should not be shown as columns.
+        .filter(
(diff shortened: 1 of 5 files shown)


packages/components/src/utils/dgs/fetchDgsMetadata.ts
```diff
@@ -64,15 +64,23 @@ const extractColumnMetadata = (
...
-        [],
-      )
+    return (
+      Object.values(data.data.columnMetadata.metaMapping)
+        // CKAN datastore auto-generates _id (row counter) and _full_text
+        // (PostgreSQL tsvector for full-text search) on every dataset — they are
+        // never part of the source data and should not be shown as columns.
+        .filter(
+          (mapping) => mapping.name !== "_id" && mapping.name !== "_full_text",
+        )
```
seaerchin: question: should we check with dgs if they have fields beginning iwth `_`? if don't have, we can make this filter more permissive just in case

seaerchin (approved):

packages/components/src/utils/dgs/fetchDgsMetadata.ts
adriangohjw: claude points me to a doc. that says these 2 are the only reserved fields

https://github.com/ckan/ckan/blob/0d2f995f3e5b70934793eceb12fa345485e8bf3f/ckanext/datastore/backend/postgres.py#L970-L973

seaerchin: not ckan itself, but dgs in general cos these _ prefixed fields feel like they aren't meant for consumpotion anyways (similar to . prefixed on linux)

adriangohjw: im not sure if there's anything stopping users/DGS from using `_` prefix - IMO i prefer to opt for a safer approach to avoid filtering out potentially actual fields

PR #2469, opened 2026-06-16
Add support for rendering Facebook Reels in portrait aspect ratio

adriangohjw:
## Problem

Facebook Reels are vertical (9:16) videos, but the Video component was rendering them in the default landscape (16:9) aspect ratio box, which clips the content and makes them unusable.

## Solution

Added detection for Facebook Reel embed URLs and conditional rendering logic to display them in a portrait aspect ratio instead of landscape.

**Features**:

- Added `isFacebookReelEmbedUrl()` utility function that detects whether a Facebook video embed URL points to a Reel by checking if the `href` query parameter contains a `/reel/` path
- Updated `ParsedVideo` type to include an `isReel` boolean flag for Facebook videos
- Modified the Video component to:
  - Detect portrait-oriented videos (Facebook Reels)
  - Apply a 9:16 aspect ratio (177.78% padding) for Reels instead of the default 16:9 (56.25%)
  - Cap the width of portrait videos to 20.3125rem to maintain sensible dimensions on wider screens
- Refactored Video component to move `parseVideo()` call outside the `RenderedVideo` function for cleaner logic flow

**Improvements**:

- Added comprehensive test coverage for `isFacebookReelEmbedUrl()` including valid Reels, regular Facebook videos, missing href params, non-Facebook URLs, and invalid inputs
- Added Storybook story for Facebook Reel to demonstrate the portrait rendering

**Breaking Changes**:

The `ParsedVideo` type change is internal to the component and does not affect the public API.

## Tests

Added comprehensive unit tests in `__tests__/utils.test.ts`:
- Tests for valid Facebook Reel embed URLs (returns true)
- Tests for regular Facebook video URLs (returns false)
- Tests for missing href parameter (returns false)
- Tests for non-Facebook URLs (returns false)
- Tests for invalid URLs (returns false)

All existing tests continue to pass. The new Storybook story (`FacebookReel`) provides visual verification of the portrait aspect ratio rendering.

- [ ] View the FacebookReel story in Storybook and verify it renders in portrait orientation with appropriate width constraints
- [ ] Verify regular Facebook videos still render in landscape (16:9) aspect ratio
- [ ] Verify YouTube and Vimeo videos are unaffected

https://claude.ai/code/session_01GTTv9LANYopBgYjjkXaLGW

### DIFF
--- a/packages/components/src/templates/next/components/complex/Video/utils.ts
+++ b/packages/components/src/templates/next/components/complex/Video/utils.ts
@@ -60,6 +60,32 @@ export const getYouTubeVideoId = (url: string): string | null => {
   }
 }
 
+/**
+ * Detects whether a Facebook video embed URL points to a Reel.
+ *
+ * Reels are embedded through the same plugins/video.php endpoint as regular
+ * Facebook videos, with the reel's URL passed in the `href` query param, e.g.
+ * https://www.facebook.com/plugins/video.php?href=https%3A%2F%2Fwww.facebook.com%2Freel%2F123
+ *
+ * Reels are vertical (9:16) videos, so we need to detect them to render them
+ * in a portrait aspect ratio instead of the default 16:9 box (which clips them).
+ */
+export const isFacebookReelEmbedUrl = (url: string): boolean => {
+  try {
+    const urlObject = new URL(url)
+    if (!VALID_VIDEO_DOMAINS.fbvideo.includes(urlObject.hostname)) {
+      return false
+    }
+    const href = urlObject.searchParams.get("href")
+    if (!href) {
+      return false
+    }
+    return new URL(href).pathname.startsWith("/reel/")
+  } catch {
+    return false
+  }
+}
+
 // NOTE: We are setting a do-not-track attribute on Vimeo embeds
 // Ref: https://developer.vimeo.com/api/oembed/videos
 export const getPrivacyEnhancedVimeoEmbedUrl = (url: string): string => {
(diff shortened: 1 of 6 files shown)


packages/components/src/templates/next/components/complex/Video/utils.ts
```diff
@@ -60,6 +60,35 @@ export const getYouTubeVideoId = (url: string): string | null => {
...
+export const isFacebookReelEmbedUrl = (url: string): boolean => {
+  try {
+    const urlObject = new URL(url)
+    if (
+      !VALID_VIDEO_DOMAINS.fbvideo.includes(urlObject.hostname) ||
+      urlObject.pathname !== "/plugins/video.php"
+    ) {
+      return false
+    }
+    const href = urlObject.searchParams.get("href")
```
dcshzj, line 82: question: why not perform a calculation based on the `height` and `width` that is provided in the URL to determine if it is more suited for a 16:9, 9:16 or even a 4:3 ratio?

adriangohjw: The URL is valid even without height and width params + IMO its slightly more complicated and might introduce more edge cases. Given that this embed is not a huge use case, im happy to keep it simple for now

dcshzj: Fair, ok with not building that capability for now

dcshzj (approved):

PR #2449, opened 2026-06-15
refactor: consolidate DraggableTagButton into compound component

adriangohjw:
## Problem

Two parallel implementations of `DraggableTagButton` existed side-by-side:

- `renderers/controls/DraggableTagButton.tsx` — monolithic component with a prop-based API, also exporting a `TagDraggableButton` convenience variant
- `form-builder/components/DraggableTagButton.tsx` — compound component (`Root`, `Handle`, `Body`, `Content`, `Label`, `Subtitle`, `ErrorCaption`, `Trailing`, `Icon`)

The monolithic version was getting incremental fixes (e.g. moving `pl`/`pr` padding into merged-props objects) that were duplicating work already done in the compound version, making maintenance harder.

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [x] No - this PR is backwards compatible

**Improvements**:

- Deleted `renderers/controls/DraggableTagButton.tsx` (monolithic + `TagDraggableButton` export).
- Migrated all three callers to the compound `form-builder/components/DraggableTagButton`:
  - `JsonFormsArrayControl` — default fallback renderer now uses the compound pattern, preserving the original `isError`-conditional `py` values on `Handle` and `Body`.
  - `JsonFormsTagCategoryControl` — replaced `TagDraggableButton` spread with explicit `Root / Handle / Body / Icon / Content / Label / Subtitle / ErrorCaption / Trailing` composition.
  - `JsonFormsTagCategoryOptionsControl` — same as above without the icon/subtitle slots.
- Also fixes two schema issues in `getCategoryOptionUsageCountSchema` caught in PR review:
  - Drops redundant `.int()` from `siteId`/`pageId` (codebase convention is `z.number().min(1)` for DB IDs).
  - Replaces deprecated `z.string().uuid()` with `z.uuid()` (Zod v4 standalone type).

## Before & After Screenshots

**BEFORE**:

N/A — internal refactor, no visible UI change.

**AFTER**:

N/A

## Tests

**Manual Verification Steps**:

- [ ] Open a collection with category filters configured. Confirm filter rows render and are draggable.
- [ ] Open a collection page/link with a Category ID field. Confirm the category options dropdown renders correctly.
- [ ] Trigger a duplicate-label error on a filter or option row. Confirm the error caption still appears inline.
- [ ] Confirm the delete confirmation modal for a category option still opens and closes correctly.

PR #2471, opened 2026-06-16
Refactor/components

seaerchin:
duplicate of #2449; reopened as the changes got overwritten


seaerchin (commented):
## Review — refactor of form-builder array controls

**What this does:** Pulls the shared bits of the four array-style JSON Forms controls out of `renderers/controls/` into `form-builder/components/` and `form-builder/hooks/`. The previously god-prop'd `DraggableTagButton` (a single `forwardRef` with ~10 render-prop/`listItem*` callbacks) becomes a compound component (`Root`/`Handle`/`Body`/`Icon`/`Content`/`Label`/`Subtitle`/`ErrorCaption`/`Trailing`), and the per-control logic (selection, add/remove guards, drag reorder, delete-target, duplicate-label detection) moves into `useArray`, `useDeleteTarget`, `useDuplicateLabels`. The old shared `JsonFormsArrayControlView` is dropped; each control now inlines its own drag-and-drop list and composes the shared pieces.

**Codeflow:** Each control destructures `ArrayLayoutProps` → `useArray(...)` → renders `<NestedDrawerSwitch {...props} {...arrayResult}>` which short-circuits to `ComplexEditorNestedDrawer` when a row is selected (`components/NestedDrawerSwitch.tsx:22`) and otherwise renders the list children. Row deletes from the list go through `useDeleteTarget` → `removeItems` directly; deletes from inside the nested drawer go through the guarded `useArray.handleRemoveSelectedItem` (`hooks/useArray.ts:672`).

**Overall:** Solid, mechanical refactor. CI is green across typecheck/build/lint/format/unit/E2E, no dangling imports to any moved file, and I verified the behavioral details below are preserved. The compound-component split is a real readability win over the old `listItem*` prop soup. My only substantive comment is about list-scaffolding duplication. Nothing blocking.

---

**praise:** The `DraggableTagButton` compound refactor is a clear improvement — the old `listItemContentProps` / `listItemTrailing` / `renderListItemErrorCaption` callback explosion was hard to follow, and the new namespaced parts read much better at the call sites. Padding semantics (`py: hasError ? 0.75 : 1.25/1` for the generic control, flat `0.5rem` default for the tag controls) are preserved exactly.

**thought (non-blocking):** `renderers/controls/JsonFormsArrayControl.tsx`, `JsonFormsCategoryOptionsControl.tsx:L134-L150`, `JsonFormsTagCategoryControl.tsx` and `JsonFormsTagCategoryOptionsControl.tsx` now each repeat the same ~40-line `<Box><DragDropContext><Droppable>{({droppableProps,innerRef,placeholder}) => <VStack>...{[...Array(data).keys()].map(...)}` scaffold. The refactor traded one over-parameterized shared view for four near-identical copies of the list shell. That's arguably the right call given the old prop soup, and the per-row content genuinely differs — but if a fifth array control shows up, consider a thin `<DraggableList data renderRow>` that owns just the DnD shell while the compound parts stay at the call site. Flagging the duplication consciously rather than asking for a change.

Confidence: high

**note:** The generic `JsonFormsArrayControl` dropped support for the schema-level `addItemLabel` extension (old `arraySchemaWithExtensions.addItemLabel`) and the `mapNewArrayItem`/`belowDescription`/`emptyState` props, hardcoding `"Add item"`. I grepped `apps/studio` and `packages` — nothing sets `addItemLabel` and nothing else consumes those props, so this is dead-feature removal, not a regression. Calling it out only so it's a deliberate record.

Confidence: high

**question:** `components/DraggableTagButton.tsx` `Label` forwards `removeItem={handleRemoveSelectedItem}` (all four call sites), but `Label` no-ops `handleSelect` and only renders the title text — `removeItem` is never invoked, and its `(path, index) => () => void` shape doesn't match the master-list `removeItem` contract anyway. Is it needed to satisfy `withJsonFormsMasterListItemProps`, or is it leftover from the old spread? If unused, dropping it from `LabelProps` would remove a misleading wire.

Confidence: medium

**note:** 26 Chromatic stories need baseline acceptance and the Vercel preview deploy shows `Error` (though the GitHub `build` job passed, so this looks like a preview-env issue rather than a code break). Since there are no unit/visual tests over these four controls, the Chromatic diffs are the main safety net for this refactor — worth eyeballing the row padding/empty-state/error-caption rendering on each control before merging rather than blind-accepting.

Confidence: medium

**nitpick (non-blocking):** `hooks/useDuplicateLabels.ts` calls `useJsonForms()` in controls that already call `useJsonForms()` for `core?.data`, so there are now two context reads per render. Negligible, and the hook is cleaner — not worth changing unless you're already touching it.

Confidence: high

PR #2473, opened 2026-06-16
ci: comment Template bundle size by category (PR vs main) when components change

seaerchin:
## Problem

We have no visibility into how a PR affects the size of the JavaScript/CSS that published Isomer sites actually ship. The `@opengovsg/isomer-components` package is consumed by the Template (`next build --webpack`), which **tree-shakes, bundles, minifies and gzips** it — so the bytes a visitor downloads are very different from the package's raw source. A new dependency or a heavier component can grow every published site's bundle (and hurt Lighthouse) with no signal in review.

Closes [insert issue #]

## Solution

Add a **report-only** CI check that measures the Template's built bundle on the PR vs `main` and comments the per-category delta.

The job builds the Template (`isomer-base-template`, `build:template` = `next build --webpack`, `output: "export"`) against its committed sample content, consuming the PR's freshly-built components package. It then measures the gzipped `out/_next/static` output, broken down by category, repeats the build from `main`, and posts a sticky PR comment comparing the two.

Categories (gzipped):
- **🧩 Vendor + shared** — components & their dependencies after tree-shaking. This is the signal.
- **🎨 CSS** — the components' Tailwind output.
- **📄 Page / route** — per-route app chunks.
- **⚙️ Framework & runtime** — Next.js + React + polyfills. Fixed overhead, reported separately so it doesn't mask the signal.

Why measure the bundle and not the `pnpm pack` tarball (the first iteration of this PR): the tarball is *everything* in the package before tree-shaking, so it doesn't reflect what ships. webpack is the bundler production uses (`next build --webpack`), so that's what we measure.

A note surfaced while building this: the render dispatch (`renderComponent`/`renderLayout`) is a static `switch` over statically-imported components, so webpack keeps the whole block/layout registry in the shared vendor chunks regardless of which blocks a page renders (per-route chunks are ~300 bytes). Tree-shaking therefore only drops never-referenced code (the `classic` template, unused barrel exports, unreferenced icons). Upside for this check: the bundle number is stable and content-independent, and adding a new block shows up as a vendor-chunk increase.

Implementation notes:
- Runs only on PRs (needs a `main` baseline) and only when `packages/components/**` changed — detected by a `changes` job using `dorny/paths-filter` (the same pattern `chromatic.yml` uses). A new dependency edits `packages/components/package.json`, which is covered by that path.
- Gated by the existing Graphite CI optimizer, consistent with the test jobs.
- Posts the comment via the preinstalled `gh` CLI (sticky: `--edit-last --create-if-none`), so no new third-party action needs to be added and SHA-pinned.
- The measurement and comment scripts are staged to `$RUNNER_TEMP` before the `main` checkout, since they only exist on this branch.

**Breaking Changes**

  - Details ...

**Features**:

- New `components-bundle-size` CI job: comments the Template's per-category gzipped bundle-size delta (PR vs `main`) when the components package changes.

**Improvements**:

- Reviewers get an at-a-glance, ships-to-users measure of a PR's bundle impact, with the components-driven "vendor + shared" category isolated from fixed framework overhead.

**Bug Fixes**:

- None.

## Before & After Screenshots

**BEFORE**:

No bundle-size signal on PRs.

**AFTER**:

A sticky comment on the PR, e.g.:

> 📈 **Template bundle size (gzipped)**
>
> | Category | main | this PR | Δ |
> | --- | ---: | ---: | ---: |
> | 🧩 Vendor + shared — components & deps | 221.26 KB | 225.26 KB | +4.00 KB (+1.81%) |
> | 🎨 CSS | 17.23 KB | 17.73 KB | +0.50 KB (+2.90%) |
> | 📄 Page / route | 1.37 KB | 1.37 KB | +0.00 KB (+0.21%) |
> | ⚙️ Framework & runtime (fixed) | 124.57 KB | 124.57 KB | ±0.00 KB (+0.00%) |
> | **Total** | 364.41 KB | 368.92 KB | +4.50 KB (+1.24%) |

## Tests

This change is CI configuration. Verified locally by:
- Building the Template against the committed sample content and confirming `next build --webpack` emits a static export to `tooling/template/out/_next/static` (24 JS chunks + CSS).
- Running `measure-template-bundle.sh` against that output and confirming the category sums are correct (JS categories sum to the exact total measured independently).
- Running `comment-template-bundle-size.sh` with fabricated and zero-delta inputs to confirm the table, per-category math, signs (📈/📉/±), and headline emoji.

Full validation happens when the job runs on this PR — its bundle-size comment is the end-to-end test.

**New scripts**:

- `.github/scripts/measure-template-bundle.sh` : categorizes and gzip-measures the Template's `_next/static` output.
- `.github/scripts/comment-template-bundle-size.sh` : formats and posts (sticky) the per-category PR-vs-`main` comparison via the `gh` CLI.

**New dependencies**:

- None.

**New dev dependencies**:

- None.


adriangohjw (commented):
i think the overall zip size isn't as meaningful because Template does treeshaking, so we would need to measure the build output's bundle size to know impact on the lighthouse performance

for example, the same components package will produce different bundle size for Next.js when we build it with webpack v.s. turbopack -> which is what we care about

seaerchin:
@adriangohjw — agreed, the tarball isn't the right metric. Reworked in f8184710f to measure the actual build output instead.

The check now builds the Template (`isomer-base-template`, `next build --webpack` → static export) against its committed sample content, consuming this PR's components package, and compares the gzipped `_next/static` output to `main` — broken down by category:

- **Vendor + shared** (components & their deps after tree-shaking) — the signal
- **CSS**
- **Page / route**
- **Framework & runtime** (Next + React + polyfills) — fixed overhead, called out separately so it doesn't drown the signal

On the webpack-vs-turbopack point: `build:template` is `next build --webpack`, so webpack is what production uses — that's what we measure.

One thing worth flagging from digging into this: the render dispatch (`renderComponent`/`renderLayout`) is a static `switch` over statically-imported components, so webpack can't prove any arm unreachable — the whole block/layout registry lands in the shared vendor chunks regardless of which blocks a page uses (per-route chunks are ~300 bytes). So tree-shaking here only drops things never referenced at all (the `classic` template, unused barrel exports, unreferenced icons), not per-page components. Practically that's good for this check: the bundle number is stable and content-independent, and adding a new block shows up as a vendor-chunk increase. If we ever want real per-page shaking, the lever is converting the dispatch to `next/dynamic`.

Still report-only — never fails the build.

dcshzj (approved):

PR #2474, opened 2026-06-16
feat(gazette): push gazette documents to Algolia (SearchSG behind flag)

seaerchin:
## Summary

Push gazette documents to **Algolia** by default, reusing the existing `PushDocumentJob` table + per-minute cron. **SearchSG** is retained behind a GrowthBook flag (`enable-searchsg-gazette-ingestion`, default-off → Algolia) as a swap-over escape hatch.

- Records are chunked egazette-style (~7000 chars/record) into the **shared egazette Algolia index**: `objectGroup` = S3 key, `objectID = ${objectGroup}-text-${idx}`, Asia/Singapore dates, `lexiNotificationNum = notificationNum.padStart(10,'0')`.
- The cron push path is **add-only**; deletion is human-driven (grace-period delete in the router) via `deleteBy` on `objectGroup`. Push and delete both branch on the same flag with the same polarity.
- Design is recorded in `docs/adr/0001-algolia-default-searchsg-behind-flag.md`, `docs/adr/0002-chunked-records-in-shared-egazette-algolia-index.md`, and `apps/studio/src/server/modules/gazette/CONTEXT.md`.

## ⚠️ Required before merge (and to make CI green)

CI typecheck is currently **red** on purpose — `env.mjs` is permission-locked to the agent, so the `ALGOLIA_*` vars aren't wired yet.

- [x] Add `ALGOLIA_APP_ID`, `ALGOLIA_API_KEY`, `ALGOLIA_INDEX_NAME` to `apps/studio/src/env.mjs` (server schema + `runtimeEnv`), and dummy values to `.env.test`. This clears the 3 typecheck errors in `lib/algolia.ts`.
- [x] In **each Algolia environment**, add `filterOnly(objectGroup)` to `attributesForFaceting` in the dashboard before enabling deletes (otherwise `deleteBy` silently matches nothing).
- [x] Create the `enable-searchsg-gazette-ingestion` flag in GrowthBook if the SearchSG escape-hatch should be togglable.

## Test plan

- [x] 58/58 unit/integration tests pass: `gazette.service`, `gazette.router`, `schedulePushDocumentJob` — covering both flag branches for push **and** delete, plus record-building edges (chunk boundaries, SG dates, zero-padding, `lexiNotificationNum` present/absent, empty-text, failure isolation, no 50k truncation on the Algolia path).
- [ ] After `env.mjs` is updated: `pnpm typecheck` green in CI.

### DIFF
--- a/apps/studio/.env.test
+++ b/apps/studio/.env.test
@@ -28,6 +28,9 @@ SEARCHSG_API_KEY=test-searchsg-api-key
 EGAZETTE_DOCUMENT_INDEX="very-cool-index"
 S3_GAZETTE_BUCKET_NAME="cool-bucket"
 S3_GAZETTE_DOMAIN_NAME="cool-domain"
+ALGOLIA_APP_ID="test-algolia-app-id"
+ALGOLIA_API_KEY="test-algolia-api-key"
+ALGOLIA_INDEX_NAME="test-algolia-index"
 
 # Datadog email for deletion alerts
 DD_DELETION_EMAIL="alerts@dd.com"
--- a/apps/studio/src/lib/algolia.ts
+++ b/apps/studio/src/lib/algolia.ts
@@ -1,0 +1,32 @@
+// NOTE: algoliasearch v4 is used deliberately — the legacy egazette integration
+// uses the v4 client API (algoliasearch(appId, key).initIndex(name),
+// index.saveObjects, index.deleteBy). v5 changed the client API; do not
+// upgrade without reviewing egazette compatibility.
+import algoliasearch from "algoliasearch"
+import { env } from "~/env.mjs"
+
+const index = algoliasearch(env.ALGOLIA_APP_ID, env.ALGOLIA_API_KEY).initIndex(
+  env.ALGOLIA_INDEX_NAME,
+)
+
+/**
+ * Upsert records into the search index. Each record must carry an `objectID`.
+ * Calling this with the same `objectID` overwrites the existing record cleanly.
+ */
+export const saveObjectsToSearchIndex = (
+  objects: readonly ({ objectID: string } & Record<string, unknown>)[],
+) => {
+  return index.saveObjects(objects as { objectID: string }[])
+}
+
+/**
+ * Delete all records matching the given Algolia filter expression.
+ * Example: `deleteObjectsFromSearchIndexByFilter('objectGroup:"year/cat/sub/file.pdf"')`.
+ *
+ * NOTE: `objectGroup` must be registered as `filterOnly(objectGroup)` in
+ * `attributesForFaceting` in the Algolia dashboard before this is used —
+ * otherwise deleteBy silently matches nothing.
+ */
+export const deleteObjectsFromSearchIndexByFilter = (filters: string) => {
+  return index.deleteBy({ filters })
+}
--- a/apps/studio/src/lib/algolia.ts
+++ b/apps/studio/src/lib/algolia.ts
@@ -13,10 +13,10 @@ const index = algoliasearch(env.ALGOLIA_APP_ID, env.ALGOLIA_API_KEY).initIndex(
  * Upsert records into the search index. Each record must carry an `objectID`.
  * Calling this with the same `objectID` overwrites the existing record cleanly.
  */
-export const saveObjectsToSearchIndex = (
+export const saveObjectsToSearchIndex = async (
   objects: readonly ({ objectID: string } & Record<string, unknown>)[],
 ) => {
-  return index.saveObjects(objects as { objectID: string }[])
+  await index.saveObjects(objects)
 }
 
 /**
@@ -27,6 +27,6 @@ export const saveObjectsToSearchIndex = (
  * `attributesForFaceting` in the Algolia dashboard before this is used —
  * otherwise deleteBy silently matches nothing.
  */
-export const deleteObjectsFromSearchIndexByFilter = (filters: string) => {
-  return index.deleteBy({ filters })
+export const deleteObjectsFromSearchIndexByFilter = async (filters: string) => {
+  await index.deleteBy({ filters })
 }
--- a/apps/studio/src/lib/growthbook.ts
+++ b/apps/studio/src/lib/growthbook.ts
@@ -15,6 +15,10 @@ export const IS_HOMEPAGE_ANTI_SCAM_BANNER_ENABLED_FEATURE_KEY =
   "homepage-antiscam-banner-enabled"
 export const EGAZETTE_INFO_FEATURE_KEY = "egazette-info"
 export const IS_REDIRECTIONS_ENABLED_FEATURE_KEY = "is-redirections-enabled"
+// Default-off: Algolia is the live document-index destination when this flag is off.
+// Turn on to route gazette ingestion (push + delete) to SearchSG instead.
+export const ENABLE_SEARCHSG_GAZETTE_INGESTION =
+  "enable-searchsg-gazette-ingestion"
 
 export const IS_SINGPASS_ENABLED_FEATURE_KEY_FALLBACK_VALUE = true
 
--- a/apps/studio/src/lib/growthbook.ts
+++ b/apps/studio/src/lib/growthbook.ts
@@ -15,8 +15,8 @@ export const IS_HOMEPAGE_ANTI_SCAM_BANNER_ENABLED_FEATURE_KEY =
   "homepage-antiscam-banner-enabled"
 export const EGAZETTE_INFO_FEATURE_KEY = "egazette-info"
 export const IS_REDIRECTIONS_ENABLED_FEATURE_KEY = "is-redirections-enabled"
-// Default-off: Algolia is the live document-index destination when this flag is off.
-// Turn on to route gazette ingestion (push + delete) to SearchSG instead.
+// When OFF (default): gazette ingestion targets Algolia directly.
+// When ON: gazette ingestion is routed to SearchSG instead.
 export const ENABLE_SEARCHSG_GAZETTE_INGESTION =
   "enable-searchsg-gazette-ingestion"
 
--- a/apps/studio/src/server/cron/jobs/__test__/schedulePushDocumentJob.test.ts
+++ b/apps/studio/src/server/cron/jobs/__test__/schedulePushDocumentJob.test.ts
@@ -162,6 +162,11 @@ describe("schedulePushDocumentJobHandler", async () => {
   let user: User
 
   beforeEach(async () => {
+    // clearAllMocks resets call counts / return-value overrides on the
+    // module-level vi.mock("~/lib/algolia") auto-mock (which restoreAllMocks
+    // would destroy, breaking vi.mocked(algoliaLib.*) calls below).
+    // restoreAllMocks then cleans up the vi.spyOn stubs re-registered each
+    // tick so they don't bleed across tests.
     vi.clearAllMocks()
     vi.restoreAllMocks()
     MockDate.set(FIXED_NOW)
@@ -299,19 +304,23 @@ describe("schedulePushDocumentJobHandler", async () => {
         })
         .execute()
 
+      // Spy on buildGazetteSearchRecords to capture its typed SearchRecord[]
+      // return value — text is string there, so no unsafe cast is needed.
+      const buildSpy = vi.spyOn(gazetteService, "buildGazetteSearchRecords")
+
       // Act
       await schedulePushDocumentJobHandler()
 
       // Assert — records were built from the full text (>1 chunk because the
       // text exceeds one 7 000-char chunk), and no 50k truncation was applied.
       expect(algoliaLib.saveObjectsToSearchIndex).toHaveBeenCalledTimes(1)
-      const [records] = vi.mocked(algoliaLib.saveObjectsToSearchIndex).mock
-        .calls[0]!
-      expect(records.length).toBeGreaterThan(1)
+      const builtRecords: gazetteService.SearchRecord[] =
+        buildSpy.mock.results[0]!.value
+      expect(builtRecords.length).toBeGreaterThan(1)
       // The combined text length across all chunks equals the full text, not
       // the 50 000-char truncated version.
-      const combinedLength = records.reduce(
-        (acc, r) => acc + (r as unknown as { text: string }).text.length,
+      const combinedLength = builtRecords.reduce(
+        (acc, r) => acc + r.text.length,
         0,
       )
       expect(combinedLength).toBe(longText.length)
--- a/apps/studio/src/server/cron/jobs/__test__/schedulePushDocumentJob.test.ts
+++ b/apps/studio/src/server/cron/jobs/__test__/schedulePushDocumentJob.test.ts
@@ -1,3 +1,4 @@
+import type * as serverContextType from "~/server/context"
 import type { User } from "~prisma/generated/selectableTypes"
 import { addMinutes } from "date-fns"
 import MockDate from "mockdate"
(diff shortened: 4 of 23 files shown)


apps/studio/src/lib/algolia.ts
```diff
@@ -0,0 +1,32 @@
+// NOTE: algoliasearch v4 is used deliberately — the legacy egazette integration
+// uses the v4 client API (algoliasearch(appId, key).initIndex(name),
+// index.saveObjects, index.deleteBy). v5 changed the client API; do not
+// upgrade without reviewing egazette compatibility.
+import algoliasearch from "algoliasearch"
+import { env } from "~/env.mjs"
+
+const index = algoliasearch(env.ALGOLIA_APP_ID, env.ALGOLIA_API_KEY).initIndex(
+  env.ALGOLIA_INDEX_NAME,
+)
```
seaerchin, line 10: TODO: add these env vars

apps/studio/src/lib/growthbook.ts
```diff
@@ -15,6 +15,10 @@ export const IS_HOMEPAGE_ANTI_SCAM_BANNER_ENABLED_FEATURE_KEY =
   "homepage-antiscam-banner-enabled"
 export const EGAZETTE_INFO_FEATURE_KEY = "egazette-info"
 export const IS_REDIRECTIONS_ENABLED_FEATURE_KEY = "is-redirections-enabled"
+// When OFF (default): gazette ingestion targets Algolia directly.
+// When ON: gazette ingestion is routed to SearchSG instead.
+export const ENABLE_SEARCHSG_GAZETTE_INGESTION =
+  "enable-searchsg-gazette-ingestion"
```
seaerchin, line 21: TODO: add this growth book

apps/studio/src/lib/algolia.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Upsert records into the search index. Each record must carry an `objectID`.
+ * Calling this with the same `objectID` overwrites the existing record cleanly.
+ */
+export const saveObjectsToSearchIndex = async (
+  objects: readonly ({ objectID: string } & Record<string, unknown>)[],
+) => {
+  await index.saveObjects(objects)
+}
```
seaerchin, line 20: matches https://github.com/opengovsg/isomer-egazette/blob/76937273034f48f3423a7b12cdd6ae2277a0f1f3/src/server/modules/gazette/selfPublishingServices/SearchService.ts#L72

apps/studio/src/lib/algolia.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Delete all records matching the given Algolia filter expression.
+ * Example: `deleteObjectsFromSearchIndexByFilter('objectGroup:"year/cat/sub/file.pdf"')`.
+ *
+ * NOTE: `objectGroup` must be registered as `filterOnly(objectGroup)` in
+ * `attributesForFaceting` in the Algolia dashboard before this is used —
+ * otherwise deleteBy silently matches nothing.
+ */
+export const deleteObjectsFromSearchIndexByFilter = async (filters: string) => {
+  await index.deleteBy({ filters })
+}
```
seaerchin, line 32: this is new functionality that we put in to allow deletion of **search records**

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -52,112 +156,139 @@ export const schedulePushDocumentJob = async () => {
...
+
+          if (records.length === 0) {
+            logger.warn(
+              { resourceId },
+              "No search records built; skipping save",
+            )
+            continue
+          }
+
+          await saveObjectsToSearchIndex(records)
```
seaerchin, line 276: taken with reference to: https://github.com/opengovsg/isomer-egazette/blob/76937273034f48f3423a7b12cdd6ae2277a0f1f3/src/server/modules/gazette/egazette-utils.ts#L174

we (presumably) don't use a `map` to avoid rate limits - that's my best guess but cc @alexanderleegs to check

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -37,6 +42,105 @@ const collectionIndexPageContentSchema = z.object({
...
+  // reduce the tag category options into a single array then we find
+  const options =
+    tagCategories?.map((category) => category.options).flat() ?? []
+  const subcategory = options.find(
+    (option) => option.id === parsed.data.page.tagged[0],
+  )
+
+  const pdfTextContent = await parseFullTextFromPDF(blob)
+
+  return {
```
seaerchin, line 134: refer here for what's required for Algolia: https://github.com/opengovsg/isomer-egazette/blob/main/src/server/modules/gazette/selfPublishingServices/SearchService.ts#L26

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -37,6 +42,105 @@ const collectionIndexPageContentSchema = z.object({
...
+      { content, resourceId },
+      "Invalid content structure for push document",
+    )
+    return null
+  }
+
+  const ref = parsed.data.page.ref
+  // objectGroup is the S3 key (no leading slash), matching egazette's
+  // objectKey convention.
+  const objectGroup = ref.slice(1)
```
seaerchin, line 89: refere here: https://github.com/opengovsg/isomer-egazette/blob/76937273034f48f3423a7b12cdd6ae2277a0f1f3/src/server/modules/gazette/egazette-utils.ts#L199

this is 

```ts
const objectKey = `${publishYear}/${category}/${subcategory}/${fileNumber}.pdf`
```

in our case here, the `ref` stores the internal reference so it matches our format

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -306,6 +307,154 @@ export interface PushDocument {
...
+
+  // lexiNotificationNum is notificationNum left-padded to 10 digits.
+  // 10 = egazette's MAX_NOTIFICATION_NUMBER_LENGTH; must match for consistent
+  // sort ordering in Algolia.
+  const lexiNotificationNum = notificationNum
+    ? notificationNum.padStart(10, "0")
+    : undefined
+
+  return chunks.map((chunk, idx) => ({
+    objectID: `${objectGroup}-text-${idx}`,
```
seaerchin, line 416: refer to egazette repo: https://github.com/opengovsg/isomer-egazette/blob/76937273034f48f3423a7b12cdd6ae2277a0f1f3/src/server/modules/gazette/selfPublishingServices/SearchService.ts#L64

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -306,6 +307,154 @@ export interface PushDocument {
...
+  const [day, month, year] = publishDate.split("/").map(Number) as [
+    number,
+    number,
+    number,
+  ]
+
+  // lexiNotificationNum is notificationNum left-padded to 10 digits.
+  // 10 = egazette's MAX_NOTIFICATION_NUMBER_LENGTH; must match for consistent
+  // sort ordering in Algolia.
+  const lexiNotificationNum = notificationNum
```
seaerchin, line 411: refer here: https://github.com/opengovsg/isomer-egazette/blob/76937273034f48f3423a7b12cdd6ae2277a0f1f3/src/server/modules/gazette/egazette-utils.ts#L141

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -306,6 +307,154 @@ export interface PushDocument {
...
+ * Deletes by objectGroup filter rather than individual objectIDs because the
+ * number of chunks is not known at delete time (Algolia's deleteBy handles the
+ * fan-out). resourceId is not needed for Algolia — objectGroup is the unique
+ * dedup key for all of a gazette's records.
+ *
+ * The objectGroup value is wrapped in double quotes in the filter expression
+ * because it contains forward slashes and a dot (e.g. "2026/cat/sub/file.pdf")
+ * which Algolia's filter parser would otherwise mis-tokenise.
+ */
+export const removeGazetteFromAlgolia = async (ref: string): Promise<void> => {
```
seaerchin, line 445: all gazettes with the same base document **have an identical object group** - they are then further identified out by their `objectID`, which is comprised of objectgroup and the index

hence, this filter would match any published pdf

pnpm-workspace.yaml
```diff
@@ -55,6 +55,10 @@ patchedDependencies:
   next@16.2.6: patches/next@16.2.6.patch
 
 catalog:
+  # v4 is pinned deliberately: the legacy egazette integration uses the v4 client API
+  # (algoliasearch(appId, key).initIndex(name), index.saveObjects, index.deleteBy).
+  # v5 changed the client API — do not "upgrade" this without reviewing egazette compatibility.
+  algoliasearch: ^4.24.0
```
adriangohjw, line 61: ask: can add it to exclude from dependabot for major version? thanks!

apps/studio/.env.test
```diff
@@ -28,6 +28,9 @@ SEARCHSG_API_KEY=test-searchsg-api-key
 EGAZETTE_DOCUMENT_INDEX="very-cool-index"
 S3_GAZETTE_BUCKET_NAME="cool-bucket"
 S3_GAZETTE_DOMAIN_NAME="cool-domain"
+ALGOLIA_APP_ID="test-algolia-app-id"
+ALGOLIA_API_KEY="test-algolia-api-key"
+ALGOLIA_INDEX_NAME="test-algolia-index"
```
adriangohjw, line 33: ask: add to .env.example too

docs/adr/0001-algolia-default-searchsg-behind-flag.md
```diff
@@ -0,0 +1,14 @@
+# Algolia is the default document-index backend; SearchSG sits behind a GrowthBook flag
+
+The gazette document index can be served by either Algolia or SearchSG. We push to **Algolia by default**; a GrowthBook flag, evaluated in the push cron (via `createGrowthBookContext`) and in the router delete flow, swaps the destination to **SearchSG** when needed. The two backends are **mutually exclusive** per tick ("swap over"), not written in parallel.
+
+## Considered Options
+
+- **Push to both backends every tick** — keeps the indexes in lockstep, but doubles push cost and failure modes and contradicts the intent of being able to *swap* destinations.
```
adriangohjw, line 7: question: how to swap destination when we want to, if we don't push to both? since it's not parity in terms of documents

apps/studio/src/lib/algolia.ts
adriangohjw: abit sus that `filters` is a string haha but ok

apps/studio/src/server/cron/jobs/__test__/schedulePushDocumentJob.test.ts
```diff
@@ -9,6 +10,25 @@ import * as gazetteService from "~/server/modules/gazette/gazette.service"
...
+// the remote GrowthBook CDN.
+vi.mock("~/server/context", async (importOriginal) => {
+  const actual = await importOriginal<typeof serverContextType>()
+  return {
+    ...actual,
+    createGrowthBookContext: vi.fn(),
+  }
+})
+
+import * as algoliaLib from "~/lib/algolia"
```
adriangohjw, line 29: question: we mock `~/lib/algolia` but also import it - how does it work?

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -52,112 +156,139 @@ export const schedulePushDocumentJob = async () => {
...
+            title,
+            category: parsedPage.category,
+            subCategory: subcategoryLabel ?? "",
+            notificationNum: parsedPage.description,
+            fileUrl,
+            scheduledAt,
+          })
+
+          if (records.length === 0) {
+            logger.warn(
```
adriangohjw: question: should this be error instead of warn? since it meant some kind of chunking issue?

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
```diff
@@ -52,112 +156,139 @@ export const schedulePushDocumentJob = async () => {
...
+    } else {
+      // --- Algolia path (flag OFF, default) ---
+      let savedCount = 0
+      for (const {
+        scheduledAt,
+        resourceId,
+        title,
+        parentId,
+        content,
+      } of scheduledResources) {
```
adriangohjw, line 240: question: why not do the same Promise.all(...) as searchsg?

docs/adr/0001-algolia-default-searchsg-behind-flag.md
seaerchin: searchsg can ingest algolia's NDJSON (see [message on slack](https://opengovproducts.slack.com/archives/C0770B5RU03/p1781174062864579?thread_ts=1779325201.069909&cid=C0770B5RU03))

concretely, this means that we'll export Algolia's record as NDJSON then pipe it into searchsg for ingestion!

apps/studio/src/server/modules/gazette/gazette.service.ts
```diff
@@ -306,6 +307,154 @@ export interface PushDocument {
...
+    await deleteObjectsFromSearchIndexByFilter(`objectGroup:"${objectGroup}"`)
+  } catch (error) {
+    logger.warn({ error, objectGroup }, "Failed to remove gazette from Algolia")
+    throw new TRPCError({
+      message: "Failed to remove gazette from search index",
+      code: "PRECONDITION_FAILED",
+    })
+  }
+}
+
```
adriangohjw, line 457: by Devin

**Algolia deleteBy silently no-ops if objectGroup facet is not configured**

The `removeGazetteFromAlgolia` function at `apps/studio/src/server/modules/gazette/gazette.service.ts:445-456` calls `deleteBy({ filters: 'objectGroup:"..."' })`. Algolia's `deleteBy` returns HTTP 200 even when the filter matches zero records — including the case where `objectGroup` is not registered as `filterOnly(objectGroup)` in `attributesForFaceting`. The code comment in `apps/studio/src/lib/algolia.ts:26-28` documents this prerequisite correctly. This is a deployment-time concern, not a code bug, but a missing facet configuration would make gazette deletion silently ineffective. The ADR at `docs/adr/0002-chunked-records-in-shared-egazette-algolia-index.md:15` also documents this as a manual rollout prerequisite.

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: >taken with reference to: https://github.com/opengovsg/isomer-egazette/blob/76937273034f48f3423a7b12cdd6ae2277a0f1f3/src/server/modules/gazette/egazette-utils.ts#L174
> 
> we (presumably) don't use a map to avoid rate limits - that's my best guess but cc @alexanderleegs to check

i think it's cos of rate-limiting on Algolia's end since 1 pdf can be alot of records as we chunk to 7kB limit and some of these PDFs are like 20/30MB; will ping alex on slack to confirm

apps/studio/src/server/modules/gazette/gazette.service.ts
seaerchin: it's already configured for `staging`

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: fair enough, will update to `logger.error`; not throwing an error here cos idw to jam the flow

apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts
seaerchin: confirmed w/ alex on slack that it's rate limiting pepehands
image

adriangohjw: saddd, can add as comments so future-ppl dont refactor this by accident?

apps/studio/src/server/cron/jobs/__test__/schedulePushDocumentJob.test.ts
seaerchin: what `vi.mock` does is to **replace the export as a whole** with a [mocked factory function](https://vitest.dev/guide/mocking/modules.html#mocking-a-module), which we can then spy on and assert against later. the mock calls are hoisted (ie, they appear at top of file automatically when executing tests) 

the most common usage pattern (like now lol) is when we want to test against that import _without_ actually using that import. contrast this against `spyOn`, where we actually still want to call that function; here, we just use it to avoid a side-effect on the import

apps/studio/.env.test
seaerchin: shit we cannot enforce this on isomer-conventions :sadge:

adriangohjw (approved):
approved to be non-blocking tho can add https://github.com/opengovsg/isomer/pull/2474/changes#r3426374328? 

hahaha thanks sir

apps/studio/src/lib/algolia.ts
seaerchin: done in infra

PR #2490, opened 2026-06-17
feat(redirect): store internal destinations as page references

dcshzj:
Addresses the blocking review comment on #2093 (destinations stored as raw strings break when a page's permalink changes): internal-path destinations are now stored as `[resource:siteId:resourceId]` references and resolved to the live permalink at publish time, so a redirect automatically follows the page if its permalink changes.

Fixes ISOM-2270.

## What changed

- **Resolvers** (`resource.service`): batched `getResourceFullPermalinks` (reference → permalink) and `getResourceIdByPermalink` (path → resourceId, reverse segment-walk).
- **Service** (`redirect.service`): on create, an internal path with no query/hash suffix is converted to a reference; external `https://` URLs and query-suffixed paths are stored verbatim; `resolveRedirectReferences` powers display.
- **Router/schema**: `resolveReferences` query; destination accepts the reference form.
- **Publishing**: references resolve to the page's current permalink at build time; the trailing `_index` segment is stripped (matching `getConvertedPermalink`), and an unresolved reference (deleted or unpublished page) drops the redirect rather than emitting `/`.
- **Frontend**: `ResourceSelector` page picker in the add form; references render as the page's current permalink in the table and delete modal.

## Behaviour notes

- Resolution restricts to **published targets only** — a draft/unpublished page has no live URL.
- The site root `/` resolves to the `RootPage`.
- Internal destinations are stored as references, so audit-log entries show the raw `[resource:...]` reference (the stable record of what was configured) rather than a permalink.

## Tests

- Router/schema: reference conversion, nested paths, `_index`, root `/`, query-suffix passthrough, unpublished/non-existent → 404.
- DB-backed publishing tests: reference → permalink, `_index` stripping, root, and dropping references to deleted/unpublished pages.

### DIFF
--- a/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
@@ -8,6 +8,7 @@ import {
   InputGroup,
   InputLeftAddon,
   Text,
+  useDisclosure,
 } from "@chakra-ui/react"
 import {
   Button,
@@ -16,7 +17,6 @@ import {
   Link,
   useToast,
 } from "@opengovsg/design-system-react"
-import { useState } from "react"
 import { BiLinkAlt, BiPlus, BiRightArrowAlt } from "react-icons/bi"
 import {
   BRIEF_TOAST_SETTINGS,
@@ -48,7 +48,11 @@ export const AddRedirectCard = ({
   })
   const toast = useToast(BRIEF_TOAST_SETTINGS)
   const { mutate: createRedirect, isPending } = useCreateRedirect()
-  const [isPageModalOpen, setIsPageModalOpen] = useState(false)
+  const {
+    isOpen: isPageModalOpen,
+    onOpen: onPageModalOpen,
+    onClose: onPageModalClose,
+  } = useDisclosure()
 
   const [source, destination] = watch(["source", "destination"])
   const isAddDisabled = !source?.trim() || !destination?.trim()
@@ -159,7 +163,7 @@ export const AddRedirectCard = ({
             variant="standalone"
             p="0"
             mt="0.25rem"
-            onClick={() => setIsPageModalOpen(true)}
+            onClick={onPageModalOpen}
           >
             <Icon as={BiLinkAlt} mr="0.25rem" />
             Link to a page
@@ -185,7 +189,7 @@ export const AddRedirectCard = ({
       <SelectDestinationPageModal
         isOpen={isPageModalOpen}
         siteId={siteId}
-        onClose={() => setIsPageModalOpen(false)}
+        onClose={onPageModalClose}
         onSelect={(permalink) =>
           setValue("destination", permalink, {
             shouldValidate: true,
--- a/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
@@ -13,9 +13,11 @@ import {
   Button,
   FormErrorMessage,
   FormLabel,
+  Link,
   useToast,
 } from "@opengovsg/design-system-react"
-import { BiPlus, BiRightArrowAlt } from "react-icons/bi"
+import { useState } from "react"
+import { BiLinkAlt, BiPlus, BiRightArrowAlt } from "react-icons/bi"
 import {
   BRIEF_TOAST_SETTINGS,
   SETTINGS_TOAST_MESSAGES,
@@ -24,6 +26,7 @@ import { useZodForm } from "~/lib/form"
 
 import { useCreateRedirect } from "../api"
 import { addRedirectSchema, type AddRedirectInput } from "../types"
+import { SelectDestinationPageModal } from "./SelectDestinationPageModal"
 
 interface AddRedirectCardProps {
   siteId: number
@@ -37,6 +40,7 @@ export const AddRedirectCard = ({
     handleSubmit,
     reset,
     watch,
+    setValue,
     formState: { errors },
   } = useZodForm<typeof addRedirectSchema>({
     schema: addRedirectSchema,
@@ -44,13 +48,15 @@ export const AddRedirectCard = ({
   })
   const toast = useToast(BRIEF_TOAST_SETTINGS)
   const { mutate: createRedirect, isPending } = useCreateRedirect()
+  const [isPageModalOpen, setIsPageModalOpen] = useState(false)
 
   const [source, destination] = watch(["source", "destination"])
   const isAddDisabled = !source?.trim() || !destination?.trim()
 
   const onSubmit = ({ source, destination }: AddRedirectInput) => {
     // source arrives normalised (leading slash, no trailing slash) by the
-    // shared schema's transform
+    // shared schema's transform. An internal-path destination is converted to a
+    // page reference server-side, which 404s if the page doesn't exist.
     createRedirect(
       { siteId, source, destination },
       {
@@ -60,23 +66,34 @@ export const AddRedirectCard = ({
         },
         // The inputs are left untouched on error so the user can adjust
         // the source instead of retyping everything
-        onError: (error) =>
-          toast(
-            error.data?.code === "CONFLICT"
-              ? {
-                  title: "A redirect already exists for this path",
-                  description: `Delete the redirect for ${source} first if you want to change where it points to.`,
-                  status: "error",
-                }
-              : {
-                  title: "Failed to add redirect",
-                  // Surface the server's message (e.g. validation rejections).
-                  // Client-side zod validation catches malformed input before
-                  // submit, so what reaches here is a clean TRPCError message.
-                  description: error.message,
-                  status: "error",
-                },
-          ),
+        onError: (error) => {
+          switch (error.data?.code) {
+            case "CONFLICT":
+              toast({
+                title: "A redirect already exists for this path",
+                description: `Delete the redirect for ${source} first if you want to change where it points to.`,
+                status: "error",
+              })
+              break
+            case "NOT_FOUND":
+              toast({
+                title: "That page doesn't exist",
+                description:
+                  "Check the destination, or pick a page with “Link to a page”.",
+                status: "error",
+              })
+              break
+            default:
+              toast({
+                title: "Failed to add redirect",
+                // Surface the server's message (e.g. validation rejections).
+                // Client-side zod validation catches malformed input before
+                // submit, so what reaches here is a clean TRPCError message.
+                description: error.message,
+                status: "error",
+              })
+          }
+        },
       },
     )
   }
@@ -136,6 +153,17 @@ export const AddRedirectCard = ({
             {...register("destination")}
           />
           <FormErrorMessage>{errors.destination?.message}</FormErrorMessage>
+          <Link
+            as="button"
+            type="button"
+            variant="standalone"
+            p="0"
+            mt="0.25rem"
+            onClick={() => setIsPageModalOpen(true)}
+          >
+            <Icon as={BiLinkAlt} mr="0.25rem" />
+            Link to a page
+          </Link>
         </FormControl>
 
         <Box flexShrink={0}>
@@ -153,6 +181,18 @@ export const AddRedirectCard = ({
           </Button>
         </Box>
       </HStack>
+
+      <SelectDestinationPageModal
+        isOpen={isPageModalOpen}
+        siteId={siteId}
+        onClose={() => setIsPageModalOpen(false)}
+        onSelect={(permalink) =>
+          setValue("destination", permalink, {
+            shouldValidate: true,
+            shouldDirty: true,
+          })
+        }
+      />
     </Box>
   )
 }
--- a/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
@@ -4,6 +4,7 @@ import {
   HStack,
   Icon,
   IconButton,
+  Skeleton,
   Stack,
   Text,
   Tooltip,
@@ -37,11 +38,48 @@ import {
   useCountRedirects,
   useDeleteRedirect,
   useListRedirects,
+  useResolveRedirectReferences,
 } from "../api"
-import { formatAddedAt } from "../utils"
+import {
+  formatAddedAt,
+  getDestinationLabel,
+  isReferenceDestination,
+  MISSING_PAGE_LABEL,
+} from "../utils"
 import { DeleteRedirectModal } from "./DeleteRedirectModal"
 
 const columnsHelper = createColumnHelper<RedirectRow>()
+
+// Renders a redirect destination. Reference destinations are shown as the
+// page's current permalink (resolved server-side); while that resolution is in
+// flight we show a skeleton, and a reference whose page no longer exists is
+// flagged rather than leaking the raw "[resource:...]" string.
+function DestinationCell({
+  destination,
+  permalinkByReference,
+}: {
+  destination: string
+  permalinkByReference: Map<string, string | null>
+}): JSX.Element {
+  const label = getDestinationLabel(destination, permalinkByReference)
+  if (label === null) {
+    return <Skeleton height="1.25rem" width="60%" />
+  }
+
+  const isMissing = label === MISSING_PAGE_LABEL
+  return (
+    <Tooltip label={label} openDelay={500} placement="top">
+      <Text
+        textStyle="body-2"
+        color={isMissing ? "utility.feedback.critical" : "base.content.strong"}
+        noOfLines={1}
+        wordBreak="break-all"
+      >
+        {label}
+      </Text>
+    </Tooltip>
+  )
+}
 
 function SortableHeader({
   label,
@@ -81,7 +119,10 @@ function SortableHeader({
   )
 }
 
-const getColumns = (onDeleteClick: (row: RedirectRow) => void) => [
+const getColumns = (
+  onDeleteClick: (row: RedirectRow) => void,
+  permalinkByReference: Map<string, string | null>,
+) => [
   columnsHelper.accessor("source", {
     minSize: 250,
     enableSorting: true,
@@ -148,16 +189,10 @@ const getColumns = (onDeleteClick: (row: RedirectRow) => void) => [
       />
     ),
     cell: ({ getValue }) => (
-      <Tooltip label={getValue()} openDelay={500} placement="top">
-        <Text
-          textStyle="body-2"
-          color="base.content.strong"
-          noOfLines={1}
-          wordBreak="break-all"
-        >
-          {getValue()}
-        </Text>
-      </Tooltip>
+      <DestinationCell
+        destination={getValue()}
+        permalinkByReference={permalinkByReference}
+      />
     ),
   }),
   columnsHelper.accessor("publishedAt", {
@@ -215,8 +250,6 @@ export const RedirectsTable = ({
     null,
   )
 
-  const columns = useMemo(() => getColumns(setRedirectToDelete), [])
-
   const { data: totalRowCount, isLoading: isCountLoading } =
     useCountRedirects(siteId)
 
@@ -235,6 +268,36 @@ export const RedirectsTable = ({
     sortBy: (sorting[0]?.id ?? "publishedAt") as RedirectSortField,
     sortDirection: (sorting[0]?.desc ?? true) ? "desc" : "asc",
   })
+
+  // Resolve the reference destinations on the visible page to permalinks for
+  // display, in one batched request rather than per row
+  const referenceDestinations = useMemo(
+    () =>
+      Array.from(
+        new Set(
+          redirects
+            .map((redirect) => redirect.destination)
+            .filter(isReferenceDestination),
+        ),
+      ),
+    [redirects],
+  )
+  const { data: resolvedReferences } = useResolveRedirectReferences(
+    siteId,
+    referenceDestinations,
+  )
+  const permalinkByReference = useMemo(() => {
+    const map = new Map<string, string | null>()
+    resolvedReferences.forEach(({ reference, permalink }) =>
+      map.set(reference, permalink),
+    )
+    return map
+  }, [resolvedReferences])
+
+  const columns = useMemo(
+    () => getColumns(setRedirectToDelete, permalinkByReference),
+    [permalinkByReference],
+  )
 
   // Changing the sort order reshuffles rows across pages, so jump back to
   // the first page to avoid showing a stale slice
@@ -300,6 +363,14 @@ export const RedirectsTable = ({
       />
       <DeleteRedirectModal
         redirect={redirectToDelete}
+        destinationLabel={
+          redirectToDelete
+            ? (getDestinationLabel(
+                redirectToDelete.destination,
+                permalinkByReference,
+              ) ?? redirectToDelete.destination)
+            : ""
+        }
         isPending={isPending}
         onClose={() => setRedirectToDelete(null)}
         onDelete={handleDelete}
--- a/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
@@ -365,10 +365,12 @@ export const RedirectsTable = ({
         redirect={redirectToDelete}
         destinationLabel={
           redirectToDelete
-            ? (getDestinationLabel(
+            ? // Fall back to "" (not the raw destination) while a reference is
+              // still resolving, so the modal never shows the "[resource:...]" token.
+              (getDestinationLabel(
                 redirectToDelete.destination,
                 permalinkByReference,
-              ) ?? redirectToDelete.destination)
+              ) ?? "")
             : ""
         }
         isPending={isPending}
--- a/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
@@ -33,6 +33,7 @@ import {
 import { useTablePagination } from "~/hooks/useTablePagination"
 
 import type { RedirectRow, RedirectSortField } from "../types"
+import type { DestinationDisplay } from "../utils"
 import {
   REDIRECTS_PAGE_SIZE,
   useCountRedirects,
@@ -42,31 +43,32 @@ import {
 } from "../api"
 import {
   formatAddedAt,
-  getDestinationLabel,
+  getDestinationDisplay,
   isReferenceDestination,
-  MISSING_PAGE_LABEL,
 } from "../utils"
 import { DeleteRedirectModal } from "./DeleteRedirectModal"
 
 const columnsHelper = createColumnHelper<RedirectRow>()
 
-// Renders a redirect destination. Reference destinations are shown as the
-// page's current permalink (resolved server-side); while that resolution is in
-// flight we show a skeleton, and a reference whose page no longer exists is
-// flagged rather than leaking the raw "[resource:...]" string.
+// Copy shown when a reference destination's page has since been deleted. Kept
+// here (the render site) so it can change without touching the resolution logic.
+const MISSING_PAGE_LABEL = "Page no longer exists"
+
+// Renders a pre-resolved redirect destination. Reference destinations arrive as
+// the page's current permalink; while that resolution is in flight we show a
+// skeleton, and a reference whose page no longer exists is flagged rather than
+// leaking the raw "[resource:...]" string.
 function DestinationCell({
-  destination,
-  permalinkByReference,
+  display,
 }: {
-  destination: string
-  permalinkByReference: Map<string, string | null>
+  display: DestinationDisplay
 }): JSX.Element {
-  const label = getDestinationLabel(destination, permalinkByReference)
-  if (label === null) {
+  if (display.status === "resolving") {
     return <Skeleton height="1.25rem" width="60%" />
   }
 
-  const isMissing = label === MISSING_PAGE_LABEL
+  const isMissing = display.status === "missing"
+  const label = isMissing ? MISSING_PAGE_LABEL : display.label
   return (
     <Tooltip label={label} openDelay={500} placement="top">
       <Text
@@ -79,6 +81,20 @@ function DestinationCell({
       </Text>
     </Tooltip>
   )
+}
+
+// Plain-text destination label for contexts without a skeleton (e.g. the delete
+// modal): resolved → permalink/path, missing → the missing-page copy, and still
+// resolving → "" so the raw "[resource:...]" token never shows.
+const destinationLabelFor = (display: DestinationDisplay): string => {
+  switch (display.status) {
+    case "resolving":
+      return ""
+    case "missing":
+      return MISSING_PAGE_LABEL
+    case "resolved":
+      return display.label
+  }
 }
 
 function SortableHeader({
@@ -190,8 +206,7 @@ const getColumns = (
     ),
     cell: ({ getValue }) => (
       <DestinationCell
-        destination={getValue()}
-        permalinkByReference={permalinkByReference}
+        display={getDestinationDisplay(getValue(), permalinkByReference)}
       />
     ),
   }),
@@ -365,12 +380,12 @@ export const RedirectsTable = ({
         redirect={redirectToDelete}
         destinationLabel={
           redirectToDelete
-            ? // Fall back to "" (not the raw destination) while a reference is
-              // still resolving, so the modal never shows the "[resource:...]" token.
-              (getDestinationLabel(
-                redirectToDelete.destination,
-                permalinkByReference,
-              ) ?? "")
+            ? destinationLabelFor(
+                getDestinationDisplay(
+                  redirectToDelete.destination,
+                  permalinkByReference,
+                ),
+              )
             : ""
         }
         isPending={isPending}
--- a/apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
@@ -1,0 +1,105 @@
+import {
+  Modal,
+  ModalBody,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+} from "@chakra-ui/react"
+import {
+  Button,
+  ModalCloseButton,
+  useToast,
+} from "@opengovsg/design-system-react"
+import { useState } from "react"
+import { ResourceSelector } from "~/components/ResourceSelector"
+import { BRIEF_TOAST_SETTINGS } from "~/constants/toast"
+import { trpc } from "~/utils/trpc"
+
+interface SelectDestinationPageModalProps {
+  isOpen: boolean
+  siteId: number
+  onClose: () => void
+  // Receives the selected page's current full permalink. The redirect stores
+  // this as a reference server-side, so the permalink is only what we show.
+  onSelect: (permalink: string) => void
+}
+
+export const SelectDestinationPageModal = ({
+  isOpen,
+  siteId,
+  onClose,
+  onSelect,
+}: SelectDestinationPageModalProps): JSX.Element => {
+  const toast = useToast(BRIEF_TOAST_SETTINGS)
+  const utils = trpc.useUtils()
+  const [selectedResourceId, setSelectedResourceId] = useState<string | null>(
+    null,
+  )
+  const [isResolving, setIsResolving] = useState(false)
+
+  const handleClose = () => {
+    setSelectedResourceId(null)
+    onClose()
+  }
+
+  const handleConfirm = async () => {
+    if (!selectedResourceId) return
+    setIsResolving(true)
+    try {
+      // Resolve the picked resource to its current permalink so the user sees a
+      // real path in the input; conversion back to a reference happens on save.
+      const permalink = await utils.page.getFullPermalink.fetch({
+        siteId,
+        pageId: Number(selectedResourceId),
+      })
+      onSelect(permalink)
+      handleClose()
+    } catch {
+      toast({
+        title: "Couldn't select this page",
+        description: "Please try again.",
+        status: "error",
+      })
+    } finally {
+      setIsResolving(false)
+    }
+  }
+
+  return (
+    <Modal isOpen={isOpen} onClose={handleClose} size="xl">
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader mr="3.5rem">Link to a page</ModalHeader>
+        <ModalCloseButton size="lg" />
+
+        <ModalBody>
+          <ResourceSelector
+            interactionType="link"
+            siteId={siteId}
+            onChange={setSelectedResourceId}
+            selectedResourceId={selectedResourceId ?? undefined}
+          />
+        </ModalBody>
+
+        <ModalFooter>
+          <Button
+            variant="clear"
+            colorScheme="neutral"
+            onClick={handleClose}
+            mr="1rem"
+          >
+            Cancel
+          </Button>
+          <Button
+            isDisabled={!selectedResourceId}
+            isLoading={isResolving}
+            onClick={() => void handleConfirm()}
+          >
+            Use this page
+          </Button>
+        </ModalFooter>
+      </ModalContent>
+    </Modal>
+  )
+}
--- a/apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
@@ -6,15 +6,9 @@ import {
   ModalHeader,
   ModalOverlay,
 } from "@chakra-ui/react"
-import {
-  Button,
-  ModalCloseButton,
-  useToast,
-} from "@opengovsg/design-system-react"
+import { Button, ModalCloseButton } from "@opengovsg/design-system-react"
 import { useState } from "react"
 import { ResourceSelector } from "~/components/ResourceSelector"
-import { BRIEF_TOAST_SETTINGS } from "~/constants/toast"
-import { trpc } from "~/utils/trpc"
 
 interface SelectDestinationPageModalProps {
   isOpen: boolean
@@ -31,39 +25,25 @@ export const SelectDestinationPageModal = ({
   onClose,
   onSelect,
 }: SelectDestinationPageModalProps): JSX.Element => {
-  const toast = useToast(BRIEF_TOAST_SETTINGS)
-  const utils = trpc.useUtils()
   const [selectedResourceId, setSelectedResourceId] = useState<string | null>(
     null,
   )
-  const [isResolving, setIsResolving] = useState(false)
+  // ResourceSelector already builds the selected resource's full permalink, so
+  // we read it straight off onChange instead of re-fetching it on confirm.
+  const [selectedPermalink, setSelectedPermalink] = useState("")
 
   const handleClose = () => {
     setSelectedResourceId(null)
+    setSelectedPermalink("")
     onClose()
   }
 
-  const handleConfirm = async () => {
+  const handleConfirm = () => {
     if (!selectedResourceId) return
-    setIsResolving(true)
-    try {
-      // Resolve the picked resource to its current permalink so the user sees a
-      // real path in the input; conversion back to a reference happens on save.
-      const permalink = await utils.page.getFullPermalink.fetch({
-        siteId,
-        pageId: Number(selectedResourceId),
-      })
-      onSelect(permalink)
-      handleClose()
-    } catch {
-      toast({
-        title: "Couldn't select this page",
-        description: "Please try again.",
-        status: "error",
-      })
-    } finally {
-      setIsResolving(false)
-    }
+    // ResourceSelector's permalink has no leading slash; destinations are stored
+    // as rooted paths, and conversion to a reference happens on save.
+    onSelect(`/${selectedPermalink}`)
+    handleClose()
   }
 
   return (
@@ -77,7 +57,10 @@ export const SelectDestinationPageModal = ({
           <ResourceSelector
             interactionType="link"
             siteId={siteId}
-            onChange={setSelectedResourceId}
+            onChange={(resourceId, fullPermalink) => {
+              setSelectedResourceId(resourceId)
+              setSelectedPermalink(fullPermalink)
+            }}
             selectedResourceId={selectedResourceId ?? undefined}
           />
         </ModalBody>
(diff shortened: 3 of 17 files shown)


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,159 @@ export const getResourceFullPermalink = async (
...
+    leaf = match
+    parentId = String(match.id)
+  }
+
+  // Only resolve to a published target (the walk above traverses unpublished
+  // folders as path segments, but the destination page itself must be live).
+  if (leaf === null || leaf.publishedVersionId === null) {
+    return null
+  }
+  return Number(leaf.id)
```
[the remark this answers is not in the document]
dcshzj: Fixed in this PR's latest revision: `getResourceIdByPermalink` now resolves a Folder/Collection to the container id when its IndexPage is published, and `GET_REDIRECTS` resolves container references via that published IndexPage child. We reference the container (not the index page) because the build remaps the index page's id to the folder/collection id.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,159 @@ export const getResourceFullPermalink = async (
...
+
+  const rows = await db
+    .withRecursive("Ancestors", (eb) =>
+      eb
+        // Base case: the resources we want permalinks for, each tagged as its
+        // own ancestry seed
+        .selectFrom("Resource")
+        .where("Resource.siteId", "=", siteId)
+        .where("Resource.id", "in", resourceIds.map(String))
+        .select([
```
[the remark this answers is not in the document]
dcshzj: By design. The live path (`GET_REDIRECTS`) already drops references to unpublished/missing targets, so an unpublished reference never publishes a redirect. `getResourceFullPermalinks` feeds the Studio table display and the loop guard, where resolving the resource's current permalink regardless of publish state is intentional (and filtering it would weaken loop detection). Not a correctness issue on the published site.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -26,6 +33,67 @@ const SORT_FIELD_TO_COLUMN = {
...
+  { reference: string; permalink: string | null }[]
+> => {
+  const parsed = references.map((reference) => ({
+    reference,
+    resourceId: getResourceIdFromReferenceLink(reference),
+  }))
+  const resourceIds = parsed
+    .filter(({ resourceId }) => resourceId !== "")
+    .map(({ resourceId }) => Number(resourceId))
+  const permalinks = await getResourceFullPermalinks(siteId, resourceIds)
```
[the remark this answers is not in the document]
dcshzj: Fixed. `resolveRedirectReferences` now parses the reference with an anchored regex and only resolves it when the embedded siteId matches the requested site; a cross-site reference resolves to null. Added a test.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -26,6 +33,67 @@ const SORT_FIELD_TO_COLUMN = {
...
+    return destination
+  }
+
+  const resourceId = await getResourceIdByPermalink(siteId, destination)
+  if (resourceId === null) {
+    throw new TRPCError({
+      code: "NOT_FOUND",
+      message: `The page "${destination}" does not exist`,
+    })
+  }
```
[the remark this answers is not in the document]
dcshzj: Resolved downstream in #2466: `resolveDestinationForStorage` no longer throws here — an internal path with no live page is stored as a literal path (with a non-blocking preflight warning), so the misleading "does not exist" message is gone.

apps/studio/src/schemas/redirect.ts
```diff
@@ -46,11 +61,21 @@ const destinationSchema = z
...
+  // suffix are converted to a reference server-side on create.
+  .refine(
+    (value) =>
+      value.startsWith("/") ||
+      value.startsWith("https://") ||
+      REFERENCE_DESTINATION_REGEX.test(value),
+    {
+      message: "Destination must start with '/' or 'https://'",
+    },
+  )
```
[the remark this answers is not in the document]
dcshzj: Reworded downstream in #2466 to a generic "Add a valid URL." References are produced by the page picker, never typed by users, so we intentionally don't surface the internal `[resource:...]` form in user-facing copy.

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -300,6 +363,14 @@ export const RedirectsTable = ({
...
       <DeleteRedirectModal
         redirect={redirectToDelete}
+        destinationLabel={
+          redirectToDelete
+            ? (getDestinationLabel(
+                redirectToDelete.destination,
+                permalinkByReference,
+              ) ?? redirectToDelete.destination)
+            : ""
+        }
```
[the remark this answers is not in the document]
dcshzj: Fixed. The delete modal now falls back to an empty label (not the raw destination) while a reference is still resolving, honouring the modal's no-raw-token contract.

tooling/build/scripts/publishing/queries.ts
```diff
@@ -61,6 +61,49 @@ export const GET_CONFIG = `
...
+)
+SELECT source, destination FROM (
+    SELECT
+        redirect.source AS source,
+        CASE
+            WHEN redirect.destination ~ '^\\[resource:\\d+:\\d+\\]$'
+            -- "||" (not CONCAT) so an unresolved reference stays NULL and is
+            -- dropped below; CONCAT would coerce NULL to '' and emit "/"
+            THEN '/' || regexp_replace(rp."fullPermalink", '(^|/)_index$', '')
+            ELSE redirect.destination
```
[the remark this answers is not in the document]
dcshzj: Fixed. `GET_REDIRECTS` now strips a trailing `_meta` as well as `_index`, matching `getConvertedPermalink`. In practice a redirect reference can never resolve to a `_meta` path (references only target pages/containers), so this is purely defensive parity.

tooling/build/scripts/publishing/queries.ts
```diff
@@ -61,6 +61,49 @@ export const GET_CONFIG = `
...
+        END AS destination
+    FROM public."Redirect" redirect
+    LEFT JOIN "resourcePath" rp
+        ON redirect.destination ~ '^\\[resource:\\d+:\\d+\\]$'
+        AND rp.id = CAST(substring(redirect.destination from '\\[resource:\\d+:(\\d+)\\]') AS bigint)
+        AND rp."publishedVersionId" IS NOT NULL
+    WHERE redirect."siteId" = $1 AND redirect."deletedAt" IS NULL
```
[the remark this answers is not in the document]
dcshzj: Fixed. Added a join predicate that the reference's embedded siteId matches `$1`, so a mismatched reference is dropped. (The `resourcePath` CTE was already scoped to the site, so there was never a cross-site leak — but it now also rejects a reference whose siteId is wrong even when the id coincidentally exists here.) Added a test.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -26,6 +33,67 @@ const SORT_FIELD_TO_COLUMN = {
...
+  destination: string,
+): Promise<string> => {
+  if (
+    REFERENCE_DESTINATION_REGEX.test(destination) ||
+    !destination.startsWith("/") ||
+    destination.includes("?") ||
+    destination.includes("#")
+  ) {
+    return destination
+  }
```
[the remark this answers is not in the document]
dcshzj: References are minted only by the page picker (always same-site + a real resource). Both resolvers now reject a non-matching reference — resolve returns null, publish drops it — when it isn't a published same-site resource, so a stale/hand-edited reference can't produce a wrong live redirect. We rely on that fail-safe rather than create-time validation.

apps/studio/src/features/settings/Redirects/utils.ts
```diff
@@ -8,3 +9,29 @@ export const formatAddedAt = (date: Date): string => {
...
   return format(date, "d MMM yyyy")
 }
+
+// Shown in place of a reference destination whose page has since been deleted.
+export const MISSING_PAGE_LABEL = "Page no longer exists"
+
+// Internal-page destinations are stored as "[resource:siteId:resourceId]"
+// references; literal paths and external URLs are not.
+export const isReferenceDestination = (destination: string): boolean =>
+  REFERENCE_LINK_REGEX.test(destination)
```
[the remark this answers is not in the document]
dcshzj: Fixed. `isReferenceDestination` now tests an anchored form of the shared `REFERENCE_LINK_REGEX`, so an external URL that merely contains the `[resource:...]` substring is no longer misclassified as a reference. Added tests.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -60,20 +70,31 @@ export const AddRedirectCard = ({
...
-                  description: `Delete the redirect for ${source} first if you want to change where it points to.`,
-                  status: "error",
-                }
-              : {
-                  title: "Failed to add redirect",
-                  description: "Something went wrong. Please try again.",
-                  status: "error",
-                },
-          ),
+        onError: (error) => {
```
seaerchin, line 73: comment: this is most likely the wrong pattern here. throwing out ideas - trpc allows the use of middleware to handle errors + `errorFormatter` to handle them nicely

should we do a **per module** transformer so that errors out to users are sensible? i think we can do a 2 fold approch here:

1. we extend TRPC error and define a method for transforming the message to something user-facing
2. at construction of, for example. `IsomerError`, call `logger.error` that captures full context of the error
3. at middleware or router level, call `IsomerError.toUserError` if available otherwise (this means we didn't throw this and hence, is something totally unexpected) just log and return generic error messages based on the class of error

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
```diff
@@ -150,6 +182,18 @@ export const AddRedirectCard = ({
...
       </HStack>
+
+      <SelectDestinationPageModal
+        isOpen={isPageModalOpen}
+        siteId={siteId}
+        onClose={onPageModalClose}
+        onSelect={(permalink) =>
+          setValue("destination", permalink, {
+            shouldValidate: true,
+            shouldDirty: true,
```
seaerchin, line 196: comment: i didn't try this but guessing that it's just to trigger validation and register the field as being edited

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
seaerchin: comment: if you're ok w/ ^, i can get claude to do it lol

apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+  onSelect: (permalink: string) => void
+}
+
+export const SelectDestinationPageModal = ({
+  isOpen,
+  siteId,
+  onClose,
+  onSelect,
+}: SelectDestinationPageModalProps): JSX.Element => {
+  const toast = useToast(BRIEF_TOAST_SETTINGS)
```
seaerchin: comment: might as well just call it `useBriefToast` at this point lmao

apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+      onSelect(permalink)
+      handleClose()
+    } catch {
+      toast({
+        title: "Couldn't select this page",
+        description: "Please try again.",
+        status: "error",
+      })
+    } finally {
+      setIsResolving(false)
```
seaerchin: comment: don't want to hold this PR but this feels abit off; the `ResourceSelector` already builds the full permalink via `useResourceStack` so i think the correct thing to do would be to make that extendable

some suggestions on my end:

1. for resource selector, pass also `fullPermalink` to the `onChange` 
2. make `ResourceSelector` write to an atom and you can call that `atom` (though this atom has to have a deeply nested structure lol)

imo 1 is cleanest

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
seaerchin, line 1: don't quite grok this haha, will relook when i review backend

seaerchin (commented):
frontend only review, but i think we should:

1. separate concerns by layer - our frontend here is transforming backend errors into frontend user facing messages, which feels off. i think we should just make the error itself handle this lol
2. we're recomputing permalink just for view (as we will pass the resourceId, this is functionally only used for user facing display); this isn't a big issue to me honestly, but idk how bad load times are for deeply nested URLs and the change seems easy to make so might as well

will review bacskend sometime later

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -37,12 +38,49 @@ import {
...
+  destination,
+  permalinkByReference,
+}: {
+  destination: string
+  permalinkByReference: Map<string, string | null>
+}): JSX.Element {
+  const label = getDestinationLabel(destination, permalinkByReference)
+  if (label === null) {
+    return <Skeleton height="1.25rem" width="60%" />
+  }
```
seaerchin: question: not important but why not just get the caller to pass `label` in so this component takes `label?`

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -24,6 +30,69 @@ const SORT_FIELD_TO_COLUMN = {
   "source" | "destination" | "createdAt"
 >
 
+// Anchored [resource:siteId:resourceId] reference, capturing siteId/resourceId.
+// Such a destination (or an external URL) is stored verbatim; only a bare
+// internal path becomes a reference.
+const REFERENCE_DESTINATION_REGEX = /^\[resource:(\d+):(\d+)\]$/
```
seaerchin: comment: we seem to have something similar in `components`: `packages/components/src/utils/getResourceIdFromReferenceLink.ts` - wonder if we should reuse?

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -24,6 +30,69 @@ const SORT_FIELD_TO_COLUMN = {
...
+// query/hash stays literal; references and external URLs are stored verbatim.
+const resolveDestinationForStorage = async (
+  siteId: number,
+  destination: string,
+): Promise<string> => {
+  if (
+    REFERENCE_DESTINATION_REGEX.test(destination) ||
+    !destination.startsWith("/") ||
+    destination.includes("?") ||
+    destination.includes("#")
```
seaerchin: question: i don't quite understand why we'd test for the symbols. my understanding is that if `destination` is an internal link, we can store as is? conversely, any external URL (any valid URL) we can also keep it?

the only time where we need to `resolve` is when we submit a page resourceId for this?

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -24,6 +30,69 @@ const SORT_FIELD_TO_COLUMN = {
...
+// Such a destination (or an external URL) is stored verbatim; only a bare
+// internal path becomes a reference.
+const REFERENCE_DESTINATION_REGEX = /^\[resource:(\d+):(\d+)\]$/
+
+// Resolves a destination to its stored form. An internal path with no query/hash
+// becomes a [resource:...] reference (so it follows page renames); a path with a
+// query/hash stays literal; references and external URLs are stored verbatim.
+const resolveDestinationForStorage = async (
+  siteId: number,
+  destination: string,
```
seaerchin, line 65: comment: actually what we really want here isn't a `string` but `InternalUrl | ResourceId | ExternalUrl`, which would make it clear to reviewers (and agents) what this is actually resolving

that said, i think upstream (at zod.parse stage) might be able to give this granularity or if cmi then just cast it later on HAHA

apps/studio/src/server/modules/redirect/redirect.service.ts
seaerchin: comment: actually we _do_ have valid use cases for ? (sharing collections with filters applied) and # (jumping to section on page) and i think we should allow for them down the line. this might be over-eager, especially since just using a check for whether `const url = new URL(destination)` would work equally well haha

apps/studio/src/schemas/redirect.ts
```diff
@@ -14,6 +18,16 @@ export const MAX_REDIRECT_PAGE_SIZE = 100
...
 
+// Matches ASCII control characters (0x00-0x1f, 0x7f) and backslashes. These are
+// never valid in a destination and can corrupt the generated redirect rules on
+// the published site. (Source paths use the stricter whitelist above instead.)
+const INVALID_PATH_CHARS_REGEX = /[\x00-\x1f\x7f\\]/
+
+// Anchored form of the [resource:siteId:resourceId] reference. Internal-page
+// destinations are stored in this form (converted from a path on create, see
+// redirect.service) so the redirect follows the page if its permalink changes.
+const REFERENCE_DESTINATION_REGEX = /^\[resource:\d+:\d+\]$/
```
seaerchin: nit: we already export this from components

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,179 @@ export const getResourceFullPermalink = async (
...
+              "Ancestors.seedId as seedId",
+              "Resource.id as id",
+              "Resource.permalink as permalink",
+              "Resource.parentId as parentId",
+            ]),
+        ),
+    )
+    .selectFrom("Ancestors")
+    .select(["Ancestors.seedId", "Ancestors.id", "Ancestors.permalink"])
+    .select("Ancestors.parentId")
```
seaerchin: supernit: can combine?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,179 @@ export const getResourceFullPermalink = async (
...
+export const getResourceFullPermalinks = async (
+  siteId: number,
+  resourceIds: number[],
+): Promise<Map<number, string>> => {
+  if (resourceIds.length === 0) {
+    return new Map()
+  }
+
+  const rows = await db
+    .withRecursive("Ancestors", (eb) =>
```
seaerchin: comment: the term `Ancestors` is actually confusing here cos we start from the leaf node and work upwards by joining the leaf's to Resource via leaf.parentId = Resource.id

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,179 @@ export const getResourceFullPermalink = async (
...
+      new Map<string, { permalink: string; parentId: string | null }>()
+    nodes.set(String(row.id), {
+      permalink: row.permalink,
+      parentId: row.parentId === null ? null : String(row.parentId),
+    })
+    nodesBySeed.set(seedId, nodes)
+  }
+
+  const result = new Map<number, string>()
+  for (const [seedId, nodes] of nodesBySeed) {
```
seaerchin: comment: i think we don't need this? shouldn't we just be walking by `resourceIds` since those are the leafs? 

additionally, i think we can use a `map` here!

apps/studio/src/features/settings/Redirects/utils.ts
```diff
@@ -8,3 +9,32 @@ export const formatAddedAt = (date: Date): string => {
   if (isYesterday(date)) return "yesterday"
   return format(date, "d MMM yyyy")
 }
+
+// Shown in place of a reference destination whose page has since been deleted.
+export const MISSING_PAGE_LABEL = "Page no longer exists"
```
seaerchin: comment: not in favour of doing this because we can't change the copy next time ._. i suggest using a sentinel string, that we then transform into a user copy on the frontend, what do you think?

main benefit is that the sentinel doesn't change, but we can always update the copy

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,179 @@ export const getResourceFullPermalink = async (
...
+        ),
+    )
+    .selectFrom("Ancestors")
+    .select(["Ancestors.seedId", "Ancestors.id", "Ancestors.permalink"])
+    .select("Ancestors.parentId")
+    .execute()
+
+  // Index every fetched node by seed so each seed's chain can be walked from
+  // the leaf (the seed itself) up to the root in memory
+  const nodesBySeed = new Map<
```
seaerchin: comment: this is abit hard to difficult + we create and mutate which is easy to miss; could we extract this (and the following if required but idts) into helper methods that makes it clear that this `extractsNodeBySeed` and does the map generation?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,179 @@ export const getResourceFullPermalink = async (
...
+  }
+  return result
+}
+
+// Reverse of getResourceFullPermalink: resolves a full permalink path back to
+// the live resource it points at (for storing a redirect destination as a
+// [resource:...] reference), or null if nothing live matches. One query fetches
+// every resource matching a path segment, then walks the parent chain in memory
+// — the (siteId, parentId, permalink) constraint makes each step unambiguous.
+export const getResourceIdByPermalink = async (
```
seaerchin, line 802: comment: i think this is off. the frontend passes a resourceId in `@apps/studio/src/features/settings/Redirects/components/SelectDestinationPag
  eModal.tsx :L52` which we then call `getFullPermalink` for to translate into a full permalink to save.

then, at the backend, we call this method to translate it back into the full permalink. instead of doing this, we should just submit the resourceId as is haha

tooling/build/scripts/publishing/queries.ts
```diff
@@ -61,6 +61,57 @@ export const GET_CONFIG = `
...
+// resolve each reference to the target's current full permalink by walking the
+// resource tree. A Folder/Collection reference resolves via its published
+// IndexPage child (the container has no version of its own); the trailing
+// "_index" is stripped to match getConvertedPermalink in index.ts. Non-reference
+// destinations (literal paths, external https URLs) pass through untouched. A
+// reference resolves to NULL — and the redirect is dropped — when nothing live
+// matches (deleted/unpublished target), since that redirect would be broken.
 export const GET_REDIRECTS = `
-SELECT source, destination FROM public."Redirect" WHERE "siteId" = $1 AND "deletedAt" IS NULL;
+WITH RECURSIVE "resourcePath" (id, "publishedVersionId", "parentId", "fullPermalink", "type") AS (
```
seaerchin, line 74: didn't check this, please take extra caution lol

seaerchin (commented):
some stuff is abit odd and i'd like those to be refactored please!

seaerchin (commented):
## Summary

Stores internal redirect destinations as stable `[resource:siteId:resourceId]` references (instead of literal paths) so a redirect follows its target page when the permalink changes. Internal paths are converted to references on create (`redirect.service.ts:36` `resolveDestinationForStorage` → `resource.service.ts:866` `getResourceIdByPermalink`); references are resolved back to live permalinks for display via a batched `resolveReferences` query (`redirect.service.ts:54` → `resource.service.ts:716` `getResourceFullPermalinks`) and at publish time by the recursive CTE in `tooling/build/scripts/publishing/queries.ts:64` (`GET_REDIRECTS`). The Studio table renders references as resolved permalinks (`RedirectsTable.tsx:48` `DestinationCell`), and a new page-picker modal (`SelectDestinationPageModal.tsx`) lets users pick a destination.

The author (dcshzj) has already addressed all 10 Copilot inline comments in later commits, and I verified the fixes hold: `isReferenceDestination` is anchored (`utils.ts:25`), both the resolve path and the publish CTE reject cross-site references (`redirect.service.ts:69`, `queries.ts:101`), the delete modal no longer leaks the raw token (`RedirectsTable.tsx:366`), and `_index`/`_meta` are both stripped at publish. Folder/collection destinations correctly resolve to the container id (matching the build's id remapping). Overall this is solid, well-tested (router integration + DB-backed publishing tests covering page/index/folder/root/unpublished/deleted/cross-site), and the security-sensitive paths (open-redirect, cross-site reference) are handled. Findings below are minor.

## Findings

**praise:** The publishing CTE comments at `tooling/build/scripts/publishing/queries.ts:81-83` are excellent — explaining `||` vs `CONCAT` (NULL coercion to `''` would emit `/`) is exactly the kind of non-obvious SQL trap that earns a comment. The cross-site siteId guard added in both resolution paths plus a regression test for each is the right defensive posture.
Confidence: high

**suggestion: (non-blocking)** `resolveDestinationForStorage` (`redirect.service.ts:39`) does the path→reference read *before* opening the create transaction, and `getResourceIdByPermalink` issues 2 queries (candidates + possible index-page lookup) on the global `db`. That's correct (it's a read, kept out of the tx as the comment notes), but the resulting reference is never re-validated inside the tx — between resolve and insert the target page could be deleted. The publish-time CTE drops broken references anyway, so the worst case is a stored reference that resolves to NULL and is silently dropped. Acceptable given the fail-safe; worth a one-line comment that the create-time check is best-effort and publish is the source of truth.
Confidence: medium

**issue: (non-blocking)** Display/publish divergence for *unpublished* (not deleted) targets. `getResourceFullPermalinks` (`resource.service.ts:716`) resolves a resource's permalink regardless of publish state, so the Studio table shows an unpublished reference as a normal black permalink — but `GET_REDIRECTS` requires `publishedVersionId IS NOT NULL` and drops it, so the redirect is silently absent from the live site. A page can become a valid reference at create (it was published) and later be unpublished; the table then misrepresents it as a working redirect. The author flagged this as by-design for loop detection on the singular path, but for the *display* column specifically it's misleading. Consider surfacing unpublished references distinctly (the `MISSING_PAGE_LABEL` treatment, or a "not published" variant) so the table matches what actually ships.
Confidence: medium

**question:** `getResourceIdByPermalink` (`resource.service.ts:875-887`) walks segments against a flat `candidates` set keyed on `(permalink, parentId)`, but never checks that intermediate (non-leaf) segments are containers (Folder/Collection) rather than e.g. a Page that happens to have a child. In practice the data model likely forbids a Page having children, so this is probably fine — but is that invariant enforced? If a Page can ever have a child resource, a crafted path could walk through it. Not a security issue (still scoped to siteId and published-leaf), just correctness robustness.
Confidence: low

**nitpick: (non-blocking)** `SelectDestinationPageModal.tsx:33` keeps an `isResolving` `useState` flag that the author's own PR thread acknowledges could just be the mutation/`useDisclosure` pattern, and seaerchin's review asked to push `fullPermalink` through `ResourceSelector.onChange` rather than re-fetching the permalink on confirm (`getFullPermalink.fetch` at `:53`). Both are explicitly deferred-by-agreement in the threads, so non-blocking — flagging only so they aren't lost: the extra `getFullPermalink` round-trip duplicates work `ResourceSelector`/`useResourceStack` already did.
Confidence: medium

**note:** seaerchin's main open thread (`AddRedirectCard.tsx:73`) — frontend translating backend error codes (`NOT_FOUND`/`CONFLICT`) into user copy via a `switch`, rather than a per-module tRPC error transformer — is a real architectural point but a cross-cutting refactor beyond this PR's scope. The current `switch` is readable and matches how the rest of this feature already handles `CONFLICT`. Reasonable to land as-is and track the transformer separately.
Confidence: high

adriangohjw (commented):
## Summary

This PR stores internal redirect destinations as `[resource:siteId:resourceId]` references, resolves them to live permalinks at publish time, and adds Studio UI to pick pages and display resolved paths. The core design is sound: path→reference conversion on create, batched resolution for the table, publish-time SQL that drops broken references instead of emitting `/`, and cross-site reference hardening.

**Overall:** Good to merge after addressing one correctness gap on the create path. No blocking security issues found.

---

## Findings

**issue:** `apps/studio/src/server/modules/redirect/redirect.service.ts:L45-L52` — `resolveDestinationForStorage` returns a `[resource:...]` destination verbatim without validating that the embedded `siteId` matches the request, the `resourceId` exists, or the target is published. Path-based input is validated via `getResourceIdByPermalink`, but a client can bypass that by POSTing a reference string directly (the schema explicitly allows it). That permits creating redirects to non-existent resources (stored forever as dead rows), unpublished pages (shown in Studio but dropped at publish), or references with a mismatched embedded `siteId` (always unresolved). Recommend validating references the same way as paths — at minimum assert `match[1] === siteId`, resource exists on that site, and the target satisfies the same published/live rules as `getResourceIdByPermalink`.

Confidence: high

---

**suggestion:** `apps/studio/src/server/modules/redirect/redirect.service.ts:L719-L799` — `getResourceFullPermalinks` (used by `resolveReferences` for Studio display) does not filter on `publishedVersionId`, while `getResourceIdByPermalink` and `GET_REDIRECTS` require published targets. A redirect to an unpublished page therefore shows a normal permalink in the table but is silently omitted from `redirects.json` until the page is published. If intentional (preview of future target), a note in the UI would help; otherwise align display resolution with the publish-time published-only rule.

Confidence: medium

---

**suggestion:** `apps/studio/src/schemas/redirect.ts:L123-L127` — `resolveRedirectReferencesSchema` accepts any string in `references` (only the array length is capped). Consider refining items to the anchored reference regex so malformed input is rejected at the trust boundary rather than silently returning `null` permalinks.

Confidence: low

---

**thought:** `apps/studio/src/server/modules/redirect/redirect.service.ts:L24-L27` — sorting by the `destination` column will now order rows lexically on `[resource:…]` tokens rather than resolved permalinks. Minor UX regression when users sort the destination column; may be acceptable given references are an implementation detail.

Confidence: medium

---

**praise:** Cross-site reference handling is thorough — `resolveRedirectReferences` checks embedded `siteId` (`redirect.service.ts:L81-L82`), `GET_REDIRECTS` enforces the same in SQL (`tooling/build/scripts/publishing/queries.ts:L100-L102`), and both paths drop unresolved references instead of falling back to `/`. The anchored `isReferenceDestination` guard and delete-modal empty-string fallback prevent leaking raw tokens in the UI.

Confidence: high

---

## Security

- **AuthZ:** All new/changed procedures use `protectedProcedure` with `validateUserPermissionsForSite`. No IDOR — `siteId` is checked on every call and reference parsing rejects mismatched embedded site IDs.
- **Injection / XSS:** Reference format is a strict numeric regex; destinations are rendered as React text nodes. No new XSS surface.
- **Open redirect:** External `https://` destinations remain supported (pre-existing); protocol-relative `//` paths are still normalized by the schema transform.
- **Info disclosure:** `resolveReferences` requires site `read` (any site member). That is consistent with `list`/`count` and equivalent to existing `page.getFullPermalink` access patterns for editors — not a new privilege escalation.
- **Audit:** Create/delete still log redirect and publish events; storing references in audit deltas is the correct stable record.

✅ No blocking security findings.

adriangohjw (commented):
## Summary

This PR changes internal redirect destinations from literal paths to `[resource:siteId:resourceId]` references so redirects follow page renames. The flow is coherent end-to-end: create resolves `/path` → reference (server-side), the table batches `resolveReferences` for display, and publish-time SQL re-resolves references to live permalinks (dropping broken ones).

Auth is wired correctly on all touched endpoints (`protectedProcedure` + `validateUserPermissionsForSite`). Cross-site references are guarded at resolve (`redirect.service.ts:L75-L83`) and publish (`queries.ts:L99-L102`). Test coverage is strong — router integration tests, schema tests, utils tests, and publishing e2e for reference resolution edge cases.

**Overall:** Solid design with good security posture. One correctness gap on the create path when destinations are submitted as pre-formed references.

---

## Findings

**praise:** `redirect.service.ts:L75-L83` + `queries.ts:L99-L112` — Cross-site reference handling is done in both the Studio read path and the publish path. A reference whose embedded `siteId` does not match the current site resolves to `null` / is dropped rather than leaking another site's permalink.

Confidence: high

**praise:** `redirect.service.ts:L69-L94` + `resource.service.ts:L719-L800` — Batched permalink resolution avoids N+1 queries on the redirects table page. The recursive CTE with `seedId` tagging is a clean approach for multi-resource ancestor walks.

Confidence: high

**praise:** `redirect.ts:L57-L59` — Adding `INVALID_PATH_CHARS_REGEX` on destinations closes a class of publish-time rule corruption (control chars / backslashes) that the source whitelist already covered.

Confidence: high

**issue:** `redirect.service.ts:L45-L51` — When the destination is already a `[resource:...]` reference, `resolveDestinationForStorage` returns it verbatim without checking that (a) the embedded `siteId` matches the input `siteId`, (b) the `resourceId` exists on that site, or (c) the target is published. The normal UI flow converts a permalink first (which does validate), but a direct API call can persist a broken or cross-site reference that only surfaces as "Page no longer exists" in the table and is silently dropped at publish. Consider validating references the same way as path resolution (existence + site match + published), or rejecting mismatched `siteId` with `BAD_REQUEST`.

Confidence: high

**suggestion (non-blocking):** `redirect.router.test.ts` — There is a create test for folder references (`L684-L709`) and a resolve test for index-page references (`L796-L816`), but no `resolveReferences` test for a folder-id reference. Publishing e2e covers this (`publishing.test.ts:L385-L388`); a router-level test would lock Studio display behaviour to the same contract.

Confidence: medium

**thought:** `resource.service.ts:L719-L800` — `getResourceFullPermalinks` (used for Studio display) does not filter on `publishedVersionId`, while publish-time resolution does. If a page is unpublished after a redirect is created, the admin table may still show its permalink even though the redirect will be dropped on the next publish. Probably acceptable for an admin-only view, but worth being aware of.

Confidence: medium

**nitpick (if-minor):** `redirect.ts:L68` — The destination refine error message says "must start with '/' or 'https://'" but references are also valid after this change. Misleading only if zod client-side validation surfaces it.

Confidence: high

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -24,6 +30,69 @@ const SORT_FIELD_TO_COLUMN = {
...
+): Promise<string> => {
+  if (
+    REFERENCE_DESTINATION_REGEX.test(destination) ||
+    !destination.startsWith("/") ||
+    destination.includes("?") ||
+    destination.includes("#")
+  ) {
+    return destination
+  }
```
adriangohjw: security (non-blocking/good-to-fix): if destination matches REFERENCE_DESTINATION_REGEX but is pointing to a site ID of another site, or pointing to a non-existent resource ID, that might be an issue. Not quite possible currently with the happy path from UI, but possible via direct TRPC query so thought to point it out. Concern for pointing to another site ID might be unintended IDOR, should a bug be introduced in the future

dcshzj:
* **#2508** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2466** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2466?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2490** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2490?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2490?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Agreed this is the cleaner long-term shape. I checked the codebase first: we have no error-transformer / user-facing-error infrastructure today — `errorFormatter` only enriches Zod errors, and the ~6 places that map `error.data?.code` → copy (this card, `DefaultTrpcError`, `EmailInput`, `VerificationInput`, `CreatePageWizardContext`, `IndexpageRow`) all do it ad-hoc. So `errorFormatter` + an `IsomerError.toUserError()` would be net-new, cross-cutting infra. I'll do it in a separate PR rather than expand this one, and keep the `switch` here for now.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Yep — `shouldValidate` re-runs the destination validation and `shouldDirty` marks the field edited, so the form state is correct after the picker fills it.

apps/studio/src/features/settings/Redirects/components/AddRedirectCard.tsx
dcshzj: Let's split it into its own PR — it touches error handling repo-wide, not just redirects. Keeping the `switch` here for now.

apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
dcshzj: Good call, but `BRIEF_TOAST_SETTINGS` is used in many places, so a `useBriefToast` hook is an unrelated repo-wide change — I'll do it as a separate PR.

apps/studio/src/features/settings/Redirects/components/SelectDestinationPageModal.tsx
dcshzj: Done — went with option 1. `ResourceSelector.onChange` now emits `(resourceId, fullPermalink)`, computed from the freshly-built stack inside `useResourceSelector` (so it isn't a render behind). The modal uses that directly, dropping the `getFullPermalink.fetch` round-trip and the `isResolving` state.

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
dcshzj: Done — `DestinationCell` is now presentational and takes the resolved display; the caller computes it via `getDestinationDisplay`.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Done — both the schema and the service now derive the anchored regex from the shared `REFERENCE_LINK_REGEX` (`^${REFERENCE_LINK_REGEX.source}$`) instead of redefining the pattern, and `resolveStoredDestination` reuses `getResourceIdFromReferenceLink`.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: The symbol check decides whether a destination is *converted into a reference*, not whether it's accepted. A bare internal path maps to exactly one resource, so we convert it to a `[resource:...]` reference so the redirect follows the page on rename. A path carrying a query (`/search?q=x`) can't be expressed as a single resource reference, so we keep it literal. (On `resourceId` vs permalink — we keep the permalink as the wire format; see the other thread.)

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Done — modelled as a tagged discriminated union (`ReferenceDestination | ExternalDestination | InternalPathDestination`) with a `parseDestination` classifier, so the storage branches are explicit.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Kept the simple checks rather than `new URL()`. On `?` vs `#`: I checked downstream — the validation PR (#2466) doesn't reject `#` and deliberately allows fragments on external https URLs. So I scoped it here — internal-path destinations now reject `#` (we don't support anchor redirects), `?` stays allowed (kept literal, since a query can't map to one resource), and external URLs keep both.

apps/studio/src/schemas/redirect.ts
dcshzj: Done — reusing `REFERENCE_LINK_REGEX` from components (anchored for the destination form).

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Done — combined into a single `.select()`.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Done — renamed the CTE to `PermalinkChain`.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Done — dropped the per-seed nested map. A node's permalink/parentId are intrinsic to its id, so one id-keyed map lets each requested id walk to the root; removed the `seedId` tagging from the query entirely.

apps/studio/src/features/settings/Redirects/utils.ts
dcshzj: Done — `getDestinationDisplay` now returns a status sentinel (`resolving | missing | resolved`); the "Page no longer exists" copy lives at the render site, so it can change without touching the resolution logic.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Done as part of the simplification above — the reconstruction is now a single id-keyed map plus a per-id walk, with no create-and-mutate nested map.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: We're keeping the permalink as the wire format rather than the resourceId — the destination field is a permalink (typed or picked), and the backend resolves it to a reference on save. What I did remove is the redundant frontend round-trip: `ResourceSelector` now hands us the permalink it already built, so there's no extra `getFullPermalink.fetch` (see the other thread).

apps/studio/src/schemas/redirect.ts
```diff
@@ -93,3 +127,16 @@ export const countRedirectsSchema = z.object({
...
+
+// Resolves stored [resource:...] destinations back to display permalinks. Kept
+// off the list endpoint so the read path stays a plain query; the table calls
+// this once per page to render references as the page's current permalink.
+export const resolveRedirectReferencesSchema = z.object({
+  siteId: z.number().min(1),
+  references: z
+    .array(z.string())
+    .max(MAX_REDIRECT_REFERENCES, { message: "Too many references" }),
+})
```
adriangohjw, line 139: nit (non-blocking): might wanna bake in the `[resource:...]` validation here

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -710,6 +710,174 @@ export const getResourceFullPermalink = async (
...
+    .execute()
+
+  let parentId: string | null = null
+  let leaf: (typeof candidates)[number] | null = null
+  for (const segment of segments) {
+    const match = candidates.find(
+      (candidate) =>
+        candidate.permalink === segment &&
+        (candidate.parentId === null ? null : String(candidate.parentId)) ===
+          parentId,
```
adriangohjw: readability(non-blocking): `candidate.parentId === parentId` should be sufficient?

tooling/build/scripts/publishing/queries.ts
```diff
@@ -61,6 +61,57 @@ export const GET_CONFIG = `
...
+// Internal destinations are stored as "[resource:siteId:resourceId]" references
+// so a redirect follows the page when its permalink changes. At publish time we
+// resolve each reference to the target's current full permalink by walking the
+// resource tree. A Folder/Collection reference resolves via its published
+// IndexPage child (the container has no version of its own); the trailing
+// "_index" is stripped to match getConvertedPermalink in index.ts. Non-reference
+// destinations (literal paths, external https URLs) pass through untouched. A
+// reference resolves to NULL — and the redirect is dropped — when nothing live
+// matches (deleted/unpublished target), since that redirect would be broken.
 export const GET_REDIRECTS = `
```
adriangohjw, line 73: performance (non-blocking): given that this is now doing a recursive CTE, it might be slow for sites with large redirects (e.g. those large ministral sites with many legacy redirects). Off the top of my head fetching resources can be ~30 seconds so this might add another ~30 seconds to site build.

if its really an issue, can consider updating `tooling/build/scripts/publishing/index.ts` to concurrent these operations by switching from a single PG client to a PG pool
```ts
    // Fetch and write navbar, footer, and config JSONs
    await fetchAndWriteSiteData(client)

    // Fetch and write redirects
    await fetchAndWriteRedirects(client)

    // Fetch all resources and their full permalinks
    const resources = await getAllResourcesWithFullPermalinks(client)
```

non-blocking and happy to wait until this really is an issue or if we migrate to using `isomer/db` package one day instead of raw sql haha

adriangohjw (approved):

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: Good catch, and agreed it's worth being explicit about. The create path does store a reference verbatim, but both resolution paths already gate on the embedded siteId, so a cross-site or dangling reference can never produce a live redirect:

- **Display resolve** drops it — `resolveRedirectReferences` only keeps a match when `Number(match[1]) === siteId`, otherwise the permalink resolves to `null`.
- **Publish** drops it — the `GET_REDIRECTS` join requires `substring(destination ... siteId) = $1` **and** `publishedVersionId IS NOT NULL`, and an unmatched reference stays `NULL` (via `||`) so the row is filtered out.

So a hand-crafted `[resource:<otherSiteId>:<id>]` submitted via direct tRPC resolves to `null` and the redirect is dropped at publish — no cross-site read or IDOR, just a stored value that never serves. We're deliberately relying on that fail-safe rather than create-time validation (same call as the earlier line-57 thread).

Happy to add create-time siteId/existence validation as defense-in-depth if you'd rather the junk never gets stored in the first place — let me know.

apps/studio/src/schemas/redirect.ts
dcshzj: The destination schema already anchor-tests the reference shape — `REFERENCE_DESTINATION_REGEX = ^${REFERENCE_LINK_REGEX.source}$` — so a malformed `[resource:...]` value is rejected at parse time.

The deeper checks (reference's siteId matches the redirect's site, and the target resource exists + is published) are the same concern as the IDOR thread above. They're enforced at resolve + publish as a fail-safe rather than in the schema, because the schema is shared with the client-side form and has no DB or site context to validate against. If we want a stale/hand-edited reference rejected *earlier*, the natural home is a create-time service check — I'll fold that into the same follow-up as the IDOR hardening if we decide to do it.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Done in c3acd6f. `Resource.parentId` is already `string | null` in the generated types, and the `parentId` accumulator is built from `String(match.id)`, so the `=== null ? null : String(...)` wrapper was a no-op. Simplified to `candidate.parentId === parentId`.

tooling/build/scripts/publishing/queries.ts
dcshzj: Agreed, and thanks for the detailed suggestion. I'm keeping it out of this PR: it's an orthogonal change to the publisher's connection model (single `Client` → `Pool`) on a prod-critical path, and the win is currently unmeasured — both heavy CTEs (redirects and resources) walk the same tree and hit the same tables, so contention may eat some of the parallel gain. I'd rather land and benchmark it on its own.

Tracking as a follow-up: switch `publishing/index.ts` to a `Pool` and run `fetchAndWriteSiteData` / `fetchAndWriteRedirects` / `getAllResourcesWithFullPermalinks` concurrently. Will prioritise it if build times actually regress on the large legacy sites (or when we move to the `isomer/db` package).

PR #2506, opened 2026-06-17
feat(dgs): client-load full dataset with chunked fetch and 20MB cap

adriangohjw:
## Problem

DGS (data.gov.sg) is deprecating the `q` full-text-search parameter on its `datastore_search` API with no replacement for partial-string matching. The existing `DynamicDGSSearchableTable` path (used for datasets ≥ 4MB) relied on `q` for server-side search — without it, the over-4MB branch cannot support search at all.

## Solution

Collapse the two-path architecture (static < 4MB / dynamic ≥ 4MB) into a single client-load path: always fetch the entire dataset in parallel 4MB chunks and search, filter, paginate locally. A 20MB hard cap prevents runaway memory use. See [ADR-0001](docs/adr/0001-dgs-searchable-table-client-load.md) for the full rationale.

**Breaking Changes**

  - Details ...

**Features**:

- `fetchAllRecordsInChunks` — probes `total` with a `limit=1` call, then fans out `ceil(datasetSize / 4MB)` parallel `datastore_search` calls with `sort=_id` for stable ordering across concurrent offset/limit chunks. Each chunk retries twice with backoff; persistent failure rejects the whole load (no partial data).
- `useAllDgsRecords` — React hook wrapping the helper with loading state, an `enabled` flag, and cancellation.
- **20MB dataset cap** enforced in two places:
  - `DGSSearchableTable` renders an explicit error state (via `SearchableTableClient`'s `isError` prop) if `metadata.size > 20MB`.
  - `JsonFormsDgsDatasetIdControl` in Studio rejects datasets over 20MB in the editor with a human-readable error (names actual vs. max size via `formatBytes`), blocking Save before publish.

**Improvements**:

- `DGSSearchableTable` collapsed to a single component; `DynamicDGSSearchableTable.tsx` and `StaticDGSSearchableTable.tsx` deleted.
- `useDgsMetadata` state management streamlined.
- `DgsDatasetIdModal` form handling migrated to `react-hook-form` Controller for consistent validation and debounce.
- `CKAN_ROW_ID_COLUMN` / `CKAN_FULL_TEXT_COLUMN` constants centralised in `utils/dgs/constants.ts`, replacing magic strings in `isCkanInternalColumn`.

**Bug Fixes**:

- Fixed loading state not releasing when `useAllDgsRecords` is disabled.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Open the Studio editor and add/edit a DGS Searchable Table block. Enter a valid dataset ID for a CSV dataset under 20MB — confirm the block saves successfully and search works on the published page.
- [ ] In the Studio editor, enter a dataset ID for a dataset whose `metadata.size` exceeds 20MB — confirm an error message appears (showing actual vs. 20MB max) and the Save button remains disabled.
- [ ] On a published page with a DGS Searchable Table, type a search term and confirm results filter correctly (client-side partial match).
- [ ] Simulate a slow/chunked load by using a dataset large enough to require multiple chunks — confirm the table renders fully and search is not limited to a subset of rows.
- [ ] Verify the `OverDatasetSizeCap` Storybook story renders the error fallback, and `NoSearchResults` shows the empty state correctly.

**New scripts**:

None.

**New dependencies**:

None.

**New dev dependencies**:

None.


adriangohjw:
a single uncached API can be as long as 18 seconds -> i don't think this is an acceptable UX and users will likely churn.

it was a good experiment but nah

PR #2507, opened 2026-06-18
feat(redirect): warn on permalink edit, block publish, and cascade-delete redirects

dcshzj:
## Problem

Redirects and pages are managed independently, so they can silently collide on the published site. A redirect takes precedence over a page at the same path (the redirect "shadows" the page), so a page can end up live-but-unreachable, and deleting a page can leave redirects pointing at nothing. Studio gave editors no warning in any of these cases.

Closes ISOM-2266.

## Solution

Adds guard rails at the three points where a page and a redirect can collide.

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible
  - No DB schema or API-shape changes. Adds two read-only queries and new guards; existing flows are unaffected unless they actually hit a page↔redirect collision.

**Features**:

- **Warn on permalink edit** — in the page Settings modal, if the page's URL is already a redirect source, a non-blocking warning shows where visitors would be sent instead ("This URL already redirects to … Visitors will end up there instead"). Saving is still allowed, since the page may be in preparation.
- **Block publishing onto a redirect** — publishing a page **for the first time** at a URL occupied by a live redirect is blocked with an actionable error ("Can't publish — a redirect already exists at `/x`. Remove it on the Redirections page first"). The guard lives in `publishPageResource`, so it covers both the Publish button and the scheduled-publish cron (the cron emails the publisher on failure).
- **Cascade-delete redirects on page delete** — deleting a page (or a folder and its subtree) soft-deletes the redirects whose destination points to it, in the same transaction, and the delete modal warns how many will be removed.

**Improvements**:

- New read-only tRPC queries `redirect.getBySource` and `redirect.countByDestinationResource`, both gated on `read`.
- The publish guard fires only on the first publish (going live); re-publishing an already-live page is unaffected.
- "Points to a page" is reference-only and descendant-aware: it counts redirects stored as `[resource:…]` references to the page or any descendant, matching exactly what gets removed.

**Bug Fixes**:

- Deleting a page no longer leaves dangling redirects that resolve to nothing on the published site.

## Before & After Screenshots

Storybook (Studio, `:6007`):

- `Components/PageSettingsModal` → `UrlIsRedirectSource`
- `Components/DeleteResourceModal` → `HasRedirectsPointingHere`

**BEFORE**: no warning when a page URL collides with a redirect; publishing onto a redirect path silently produced an unreachable page; deleting a destination page left broken redirects.

**AFTER**:

## Tests

[...]

PR #2508, opened 2026-06-18
feat(redirect): auto-create a redirect when a page's permalink changes

dcshzj:
## Problem

When a published page's permalink changes — either by moving it to another folder or editing its URL in settings — its old URL stops working. Existing bookmarks, inbound links, and search-engine results that point at the old path break with no way for an editor to preserve them.

Closes ISOM-2265

## Solution

Add an optional **"Redirect page automatically"** option to both permalink-change flows (the Move modal and the Edit-page Settings modal). When it's checked and a published `Page`/`CollectionPage` actually changes URL, the server creates a redirect from the old URL to the page — in the same transaction as the move/rename, with the single existing site publish covering it.

Key technical points:

- **Schema** — a `shouldCreateRedirect` boolean is added to the `move` and `updateSettings` inputs (optional, defaults `false` server-side so no caller mints a redirect implicitly).
- **Server** — a publish-free, in-transaction helper inserts the redirect. The source is the page's own (already-validated) old permalink and the destination is a **page reference** (`[resource:siteId:resourceId]`), so the redirect follows the page on any *future* move. It runs only when `flag ∧ published ∧ permalink-actually-changed`.
- **Auto-clear** — when a permalink change lands a page on a path that already has a redirect pointing back at that same page (e.g. the page is moved back onto its own earlier auto-redirect), that now self-referential redirect is soft-deleted as the page reclaims its URL.
- **Frontend** — the checkbox (default on, shown only when the URL genuinely changes) renders a live "from → to" preview. The Move modal also gains the destination-shadow warning the Settings modal already had; both modals suppress that warning when the existing redirect points back at the page being edited (it gets auto-cleared anyway).

**Breaking Changes**

  - The new input field is optional and defaults off; existing callers are unaffected.

**Features**:

- "Redirect page automatically" checkbox in the Move modal and the Edit-page Settings modal (`Page`/`CollectionPage` only; default on; shown only when the URL actually changes).
- On move or URL change of a published page, a redirect from the old URL to the page is created automatically, stored as a page reference so it survives future moves.

**Improvements**:

- The Move modal now warns when the chosen destination URL is already a redirect source (parity with the Settings modal).
- Both modals suppress that shadow warning when the existing redirect points back at the page being edited.
- Moving/renaming a page onto a path that redirected to it now clears the redundant redirect instead of leaving a self-shadowing loop.

**Bug Fixes**:

- None.

## Tests

7 tRPC integration tests against a real Postgres (`createCallerFactory`) cover both flows: auto-create on move and on URL edit, the gating (flag off / unpublished / unchanged URL), and the reclaim auto-clear. Storybook stories and MSW handlers cover the checkbox, the destination-shadow warning, and its suppression in both modals.

- [ ] As a site admin, publish a page (e.g. `/newsroom/contact`). Open its **⋯ → Move to…**, pick a destination folder, leave **Redirect page automatically** checked, and click **Move here**. Confirm a redirect from `/newsroom/contact` to the page now appears on the Redirections page (and resolves on the published site after the build).
- [ ] Open a published page's **Settings**, change the **URL**, confirm the **Redirect page automatically** option appears (checked), and **Publish**. Confirm a redirect from the old URL is created.
- [ ] Repeat a move or URL edit with the checkbox **unchecked** — confirm **no** redirect is created.
- [ ] Move or edit the URL of an **unpublished (draft)** page — confirm the checkbox does **not** appear and no redirect is created.
- [ ] In Settings, change **only the title** (leave the URL untouched) — confirm no redirect is created.
- [ ] Move a page into a folder where a redirect already points back at that page — confirm the redundant redirect is **removed** after the move.
- [ ] Move or edit a page onto a URL that already redirects elsewhere — confirm the amber "This URL already redirects to …" warning is shown.

**New scripts**:

- None.

**New dependencies**:

- None.

**New dev dependencies**:

- None.

This PR adds an optional "Redirect page automatically" checkbox to both the Move modal and the Edit-page Settings modal. When enabled for a published `Page` or `CollectionPage` whose URL genuinely changes, the server creates a redirect from the old permalink to the page (stored as a resource reference) inside the same transaction, covering the implicit publish already present in both flows.

- **Server (`redirect.service.ts`)**: Three new helpers (`createRedirectForPermalinkChange`, `assertPermalinkNotShadowed`, `clearReclaimedRedirect`) are orchestrated by `applyPermalinkChangeRedirects`, which is called from both `page.router.ts` (`updateSettings`) and `resource.router.ts` (`move`). The upsert revives soft-deleted rows and throws `CONFLICT` on a live collision, matching the existing publish-block guard.
- **Schema (`resource.ts`, `page.ts`)**: `shouldCreateRedirect` is added as an optional field; however, both schemas use `.default(true)`, meaning any caller that omits the field will implicitly create redirects—contradicting the PR description's backwards-compatibility claim of "defaults off."
- **Frontend**: Both modals gate the checkbox on `publishedVersionId !== null` and URL actually changing; `getRedirectBySource` now returns `destinationResourceId` so each modal can suppress the shadow warning when the existing redirect already points back at the page being edited.
</details>

Safe to merge — the redirect creation, shadow-assertion, and reclaim-clear logic are all inside the mutation's existing transaction, so a failure rolls back cleanly; the new checkbox is gated on publication status and URL change on both client and server.

The core orchestration is correct: shadow-assert runs before reclaim-clear, which runs before redirect-create, and the whole sequence is transactional. Thirteen new integration tests against a real Postgres database cover the key scenarios (create, flag-off, unpublished, reclaim, shadow block, upsert revival, CollectionPage). The only notable discrepancy is a documentation-level mismatch: the schema defaults `shouldCreateRedirect` to `true` while the PR description says 'defaults off.' In practice this only matters if a caller built against the previous schema sends a move/settings request without the field — it would unexpectedly create redirects — but the frontend (the only consumer) has been updated to always send the field explicitly.

apps/studio/src/schemas/resource.ts and apps/studio/src/schemas/page.ts — the `shouldCreateRedirect` default of `true` contradicts the PR's backwards-compatibility claim; worth aligning the default with the stated intent.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/redirect/redirect.service.ts | Adds three well-scoped helpers plus the orchestrating `applyPermalinkChangeRedirects`. Ordering is correct (shadow-assert → reclaim-clear → optional create). Upsert logic safely distinguishes live from soft-deleted rows. `getRedirectBySource` now returns `destinationResourceId` with a cross-site siteId guard. |
| apps/studio/src/server/modules/resource/resource.router.ts | Permalink reads for old/destination URLs happen before the UPDATE (pre-move committed state), which is correct. `applyPermalinkChangeRedirects` is gated on Page/CollectionPage with a non-null full permalink. `publishedVersionId` is now returned from `getMetadataById`. |
| apps/studio/src/server/modules/page/page.router.ts | Calls `applyPermalinkChangeRedirects` inside the mutation's transaction before `publishResource`. `isPublished` correctly uses the post-UPDATE `publishedVersionId` (pre-publish), so the shadow-assert is skipped for first-time publications (handled by the existing publish-block guard). |
| apps/studio/src/schemas/resource.ts | Adds `shouldCreateRedirect: z.boolean().optional().default(true)` to `moveSchema`. The server-side default is `true`, inconsistent with the PR description's "defaults off" backwards-compatibility claim. |
| apps/studio/src/schemas/page.ts | Adds `shouldCreateRedirect: z.boolean().optional().default(true)` to `basePageSettingsSchema`. Same default-value concern as `resource.ts`. |
| apps/studio/src/features/dashboard/components/PageSettingsModal.tsx | Correctly gates the checkbox on `publishedVersionId !== null` and URL actually changing. Shadow warning suppressed when `destinationResourceId` matches the current page ID. |
| apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx | Correct publication-status guard using `publishedVersionId` from `getMetadataById`. `isDestinationResolved` properly defers the redirect option until the destination permalink is known. |
| apps/studio/src/server/modules/resource/resource.service.ts | Adds optional `trx` parameter to `getResourcePermalinkTree` and `getResourceFullPermalink`. Correctly passes `tx` in `publishPageResource` for the publish-time shadow guard; move/settings callers use their own connections for pre-UPDATE reads, which is safe and intentional. |

</details>

```mermaid
sequenceDiagram
    participant FE as Frontend (Move/Settings modal)
    participant R as Router (move / updateSettings)
    participant S as applyPermalinkChangeRedirects
    participant DB as Database (tx)

    FE->>R: "move/updateSettings { shouldCreateRedirect, ... }"
    R->>DB: UPDATE Resource (parentId / permalink)
    R->>S: "applyPermalinkChangeRedirects(tx, { old, new, isPublished, shouldCreateRedirect })"

    alt URL unchanged
        S-->>R: no-op (early return)
    else URL changed
        alt isPublished
            S->>DB: assertPermalinkNotShadowed(new URL) throws CONFLICT if live redirect elsewhere
        end
        S->>DB: clearReclaimedRedirect(new URL) soft-delete if redirect points back at this page
        alt shouldCreateRedirect AND isPublished
            S->>DB: createRedirectForPermalinkChange upsert (old URL to resource reference)
        end
    end

    R->>DB: publishResource / publishSite
    R-->>FE: updated resource
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant FE as Frontend (Move/Settings modal)
    participant R as Router (move / updateSettings)
    participant S as applyPermalinkChangeRedirects
    participant DB as Database (tx)

    FE->>R: "move/updateSettings { shouldCreateRedirect, ... }"
    R->>DB: UPDATE Resource (parentId / permalink)
    R->>S: "applyPermalinkChangeRedirects(tx, { old, new, isPublished, shouldCreateRedirect })"

    alt URL unchanged
        S-->>R: no-op (early return)
    else URL changed
        alt isPublished
            S->>DB: assertPermalinkNotShadowed(new URL) throws CONFLICT if live redirect elsewhere
        end
        S->>DB: clearReclaimedRedirect(new URL) soft-delete if redirect points back at this page
        alt shouldCreateRedirect AND isPublished
            S->>DB: createRedirectForPermalinkChange upsert (old URL to resource reference)
        end
    end

    R->>DB: publishResource / publishSite
    R-->>FE: updated resource
```

</a>
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(test): include publishedVersionId in..."](https://github.com/opengovsg/isomer/commit/07e2346b19d3352c302cff59a9be6eca01c5c45a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39746626)</sub>

### DIFF
--- a/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
@@ -251,8 +251,7 @@ const PageSettingsModalContent = ({
               {existingRedirect &&
                 existingRedirect.destinationResourceId !== Number(pageId) && (
                   <Infobox my="0.5rem" variant="warning" size="sm">
-                    This URL already redirects to {existingRedirect.destination}
-                    . Visitors will end up there instead.
+                    {`This URL already redirects to ${existingRedirect.destination}. Visitors will end up there instead.`}
                   </Infobox>
                 )}
               {showRedirectOption && (
--- a/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
@@ -40,7 +40,6 @@ import {
   MAX_PAGE_URL_LENGTH,
   MAX_TITLE_LENGTH,
 } from "~/schemas/page"
-import { getReferenceLink } from "~/utils/link"
 import { trpc } from "~/utils/trpc"
 import { ResourceType } from "~prisma/generated/generatedEnums"
 
@@ -131,10 +130,6 @@ const PageSettingsModalContent = ({
   )
 
   const originalPermalink = permalinkTree[permalinkTree.length - 1] ?? ""
-  const selfReference = getReferenceLink({
-    siteId: String(siteId),
-    resourceId: String(pageId),
-  })
   // Offer the redirect only when a Page/CollectionPage URL actually changes.
   const showRedirectOption =
     (type === ResourceType.Page || type === ResourceType.CollectionPage) &&
@@ -254,7 +249,7 @@ const PageSettingsModalContent = ({
               {/* Suppressed when the redirect points back at this page —
                   saving auto-clears it, so it won't actually shadow. */}
               {existingRedirect &&
-                existingRedirect.destination !== selfReference && (
+                existingRedirect.destinationResourceId !== Number(pageId) && (
                   <Infobox my="0.5rem" variant="warning" size="sm">
                     This URL already redirects to {existingRedirect.destination}
                     . Visitors will end up there instead.
--- a/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
@@ -55,10 +55,11 @@ const PageSettingsModalContent = ({
   onClose,
 }: PageSettingsState & { onClose: () => void }) => {
   const { siteId } = useQueryParse(editSettingsSchema)
-  const [{ title: originalTitle }] = trpc.page.readPage.useSuspenseQuery({
-    pageId: Number(pageId),
-    siteId,
-  })
+  const [{ title: originalTitle, publishedVersionId }] =
+    trpc.page.readPage.useSuspenseQuery({
+      pageId: Number(pageId),
+      siteId,
+    })
   const [permalinkTree] = trpc.page.getPermalinkTree.useSuspenseQuery({
     pageId: Number(pageId),
     siteId,
@@ -130,10 +131,13 @@ const PageSettingsModalContent = ({
   )
 
   const originalPermalink = permalinkTree[permalinkTree.length - 1] ?? ""
-  // Offer the redirect only when a Page/CollectionPage URL actually changes.
+  // Offer the redirect only when a published Page/CollectionPage URL actually
+  // changes — an unpublished page has no live URL to preserve, so the server
+  // skips redirect creation for it anyway.
   const showRedirectOption =
     (type === ResourceType.Page || type === ResourceType.CollectionPage) &&
-    permalink !== originalPermalink
+    permalink !== originalPermalink &&
+    publishedVersionId !== null
   const oldFullPermalink = `${permalinksToRender.parentPermalinks}${originalPermalink}`
 
   const utils = trpc.useUtils()
--- a/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
+++ b/apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
@@ -1,5 +1,6 @@
 import type { PageSettingsState } from "~/features/dashboard/atoms"
 import {
+  Box,
   chakra,
   FormControl,
   Input,
@@ -15,6 +16,7 @@ import {
 } from "@chakra-ui/react"
 import {
   Button,
+  Checkbox,
   FormErrorMessage,
   FormHelperText,
   FormLabel,
@@ -38,6 +40,7 @@ import {
   MAX_PAGE_URL_LENGTH,
   MAX_TITLE_LENGTH,
 } from "~/schemas/page"
+import { getReferenceLink } from "~/utils/link"
 import { trpc } from "~/utils/trpc"
 import { ResourceType } from "~prisma/generated/generatedEnums"
 
@@ -81,6 +84,7 @@ const PageSettingsModalContent = ({
     defaultValues: {
       title: originalTitle,
       permalink: permalinkTree[permalinkTree.length - 1] || "",
+      shouldCreateRedirect: true,
     },
   })
 
@@ -125,6 +129,17 @@ const PageSettingsModalContent = ({
         debouncedPermalink !== "/",
     },
   )
+
+  const originalPermalink = permalinkTree[permalinkTree.length - 1] ?? ""
+  const selfReference = getReferenceLink({
+    siteId: String(siteId),
+    resourceId: String(pageId),
+  })
+  // Offer the redirect only when a Page/CollectionPage URL actually changes.
+  const showRedirectOption =
+    (type === ResourceType.Page || type === ResourceType.CollectionPage) &&
+    permalink !== originalPermalink
+  const oldFullPermalink = `${permalinksToRender.parentPermalinks}${originalPermalink}`
 
   const utils = trpc.useUtils()
   const toast = useToast(BRIEF_TOAST_SETTINGS)
@@ -236,11 +251,48 @@ const PageSettingsModalContent = ({
                 {MAX_PAGE_URL_LENGTH - permalink.length} characters left
               </FormHelperText>
               <FormErrorMessage>{errors.permalink?.message}</FormErrorMessage>
-              {existingRedirect && (
-                <Infobox my="0.5rem" variant="warning" size="sm">
-                  This URL already redirects to {existingRedirect.destination}.
-                  Visitors will end up there instead.
-                </Infobox>
+              {/* Suppressed when the redirect points back at this page —
+                  saving auto-clears it, so it won't actually shadow. */}
+              {existingRedirect &&
+                existingRedirect.destination !== selfReference && (
+                  <Infobox my="0.5rem" variant="warning" size="sm">
+                    This URL already redirects to {existingRedirect.destination}
+                    . Visitors will end up there instead.
+                  </Infobox>
+                )}
+              {showRedirectOption && (
+                <Controller
+                  control={control}
+                  name="shouldCreateRedirect"
+                  render={({ field: { value, onChange, ref, ...field } }) => (
+                    <Box
+                      mt="0.75rem"
+                      w="full"
+                      bg="utility.feedback.info-subtle"
+                      borderRadius="0.5rem"
+                      p="1rem"
+                    >
+                      <Checkbox
+                        alignItems="flex-start"
+                        size="1.25rem"
+                        isChecked={!!value}
+                        onChange={(e) => onChange(e.target.checked)}
+                        ref={ref}
+                        {...field}
+                      >
+                        <VStack align="flex-start" spacing="0.25rem">
+                          <Text textStyle="subhead-2">
+                            Redirect page automatically
+                          </Text>
+                          <Text textStyle="body-2" color="base.content.medium">
+                            Check this box to redirect visitors from{" "}
+                            {oldFullPermalink} to {candidateFullPermalink}.
+                          </Text>
+                        </VStack>
+                      </Checkbox>
+                    </Box>
+                  )}
+                />
               )}
             </FormControl>
           )}
--- a/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
@@ -52,7 +52,7 @@ const MoveResourceContent = withSuspense(
     >(undefined)
     const { siteId } = useQueryParse(sitePageSchema)
     const setMovedItem = useSetAtom(moveResourceAtom)
-    const [{ title, type, permalink: movedSlug }] =
+    const [{ title, type, permalink: movedSlug, publishedVersionId }] =
       trpc.resource.getMetadataById.useSuspenseQuery({
         siteId: Number(siteId),
         resourceId,
@@ -128,9 +128,11 @@ const MoveResourceContent = withSuspense(
       { enabled: !!curResourceId },
     )
 
-    // Only Page/CollectionPage have a URL worth preserving.
+    // Only published Page/CollectionPage have a live URL worth preserving — the
+    // server skips redirect creation for unpublished pages, so don't offer it.
     const isRedirectableType =
-      type === ResourceType.Page || type === ResourceType.CollectionPage
+      (type === ResourceType.Page || type === ResourceType.CollectionPage) &&
+      publishedVersionId !== null
     const oldFullPermalink = normalizeRedirectPath(movedFullPermalink)
     const newFullPermalink = normalizeRedirectPath(
       `${curResourceId && destination ? destination.fullPermalink : ""}/${movedSlug}`,
--- a/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
+++ b/apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
@@ -1,4 +1,5 @@
 import {
+  Box,
   Button,
   Modal,
   ModalBody,
@@ -8,9 +9,10 @@ import {
   ModalHeader,
   ModalOverlay,
   Skeleton,
+  Text,
   VStack,
 } from "@chakra-ui/react"
-import { Infobox, useToast } from "@opengovsg/design-system-react"
+import { Checkbox, Infobox, useToast } from "@opengovsg/design-system-react"
 import { useAtom, useAtomValue, useSetAtom } from "jotai"
 import { useState } from "react"
 import { ResourceSelector } from "~/components/ResourceSelector/ResourceSelector"
@@ -19,7 +21,10 @@ import { usePermissions } from "~/features/permissions"
 import { withSuspense } from "~/hocs/withSuspense"
 import { useQueryParse } from "~/hooks/useQueryParse"
 import { sitePageSchema } from "~/pages/sites/[siteId]"
+import { normalizeRedirectPath } from "~/schemas/redirect"
+import { getReferenceLink } from "~/utils/link"
 import { trpc } from "~/utils/trpc"
+import { ResourceType } from "~prisma/generated/generatedEnums"
 
 import { moveResourceAtom } from "../../atoms"
 
@@ -48,10 +53,11 @@ const MoveResourceContent = withSuspense(
     >(undefined)
     const { siteId } = useQueryParse(sitePageSchema)
     const setMovedItem = useSetAtom(moveResourceAtom)
-    const [{ title }] = trpc.resource.getMetadataById.useSuspenseQuery({
-      siteId: Number(siteId),
-      resourceId,
-    })
+    const [{ title, type, permalink: movedSlug }] =
+      trpc.resource.getMetadataById.useSuspenseQuery({
+        siteId: Number(siteId),
+        resourceId,
+      })
     const ability = usePermissions()
     const utils = trpc.useUtils()
     const toast = useToast({ status: "success" })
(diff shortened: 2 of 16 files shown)


apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
```diff
@@ -244,11 +254,48 @@ const PageSettingsModalContent = ({
...
-                  Visitors will end up there instead.
-                </Infobox>
+              {/* Suppressed when the redirect points back at this page —
+                  saving auto-clears it, so it won't actually shadow. */}
+              {existingRedirect &&
+                existingRedirect.destinationResourceId !== Number(pageId) && (
+                  <Infobox my="0.5rem" variant="warning" size="sm">
+                    This URL already redirects to {existingRedirect.destination}
+                    . Visitors will end up there instead.
+                  </Infobox>
```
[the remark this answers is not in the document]
dcshzj: Fixed — the message is now a single string expression, so the period stays attached to the destination (oxfmt was wrapping the text and pushing `.` onto its own line, which is what produced the stray space).

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -112,6 +117,34 @@ const MoveResourceContent = withSuspense(
...
+
+    // Only Page/CollectionPage have a URL worth preserving; show the option
+    // once a destination is picked.
+    const showRedirectOption =
+      curResourceId !== undefined &&
+      (type === ResourceType.Page || type === ResourceType.CollectionPage)
+    const oldFullPermalink = normalizeRedirectPath(movedFullPermalink)
+    const newFullPermalink = normalizeRedirectPath(
+      `${curResourceId && destination ? destination.fullPermalink : ""}/${movedSlug}`,
+    )
```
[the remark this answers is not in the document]
dcshzj: Good catch — the option now shows only once the destination resolves AND the computed full permalink differs from the current one, so a same-parent move no longer renders `/x → /x`.

seaerchin (commented):
## Review — auto-create a redirect when a page's permalink changes

### Summary

This PR adds an editor-controlled "Redirect page automatically" option to both permalink-change flows (the Move modal and the Edit-page Settings modal). When a published `Page`/`CollectionPage` actually changes URL, the server creates an old-URL→page redirect inside the same transaction as the move/rename.

- **Backend orchestration** lives in `resource.router.ts` `move` (~L547-587) and `page.router.ts` `updateSettings` (~L863-896). Each runs the same three steps when the resource is a `Page`/`CollectionPage`: `assertPermalinkNotShadowed` (block a published page landing on a live redirect that points elsewhere), `clearReclaimedRedirect` (soft-delete a self-pointing redirect the page reclaims), then `createRedirectForPermalinkChange` (gated on `shouldCreateRedirect ∧ published ∧ URL-changed`).
- **Service layer** (`redirect.service.ts` L462-702) holds the three helpers, all transaction-scoped and publish-free, plus `getRedirectBySource` now returning `destinationResourceId` so the modals can suppress the shadow warning when the redirect points back at the page being edited.
- **Frontend** adds the checkbox + live "from → to" preview to both modals, shown only when the URL genuinely changes.

**Overall assessment:** Well-constructed and defensively designed. The "redirect creation can't fail by construction" reasoning genuinely holds — `createRedirect` (`redirect.service.ts:249`) blocks any redirect whose source is a published page's live URL, so a published page is never left on a live redirect source and the old path carries at most a soft-deleted row the upsert revives. Test coverage is good (both flows: create, gating, shadow-block, reclaim). The findings below are minor.

---

### Findings

**suggestion (maintainability):** `resource.router.ts:L547-587` and `page.router.ts:L863-896` — the redirect-orchestration block (type guard → `assertPermalinkNotShadowed` → `clearReclaimedRedirect` → conditional `createRedirectForPermalinkChange`) is duplicated near-verbatim across the two routers, differing only in how `old`/`new` full permalinks are derived. Both copies must stay in lockstep for correctness (e.g. the published-only gate on the shadow check, the reclaim-before-create ordering). Consider a single `applyPermalinkChangeRedirects(tx, { siteId, oldFullPermalink, newFullPermalink, resourceId, isPublished, shouldCreateRedirect, byUserId })` in `redirect.service.ts` that owns the ordering, leaving each router to compute the two permalinks. Reduces the chance the two flows drift apart later.

Confidence: medium

**issue (correctness, non-blocking):** `getResourceIdFromReferenceLink.ts:L3-11` — `getRedirectBySource` derives `destinationResourceId` from `getResourceIdFromReferenceLink`, which returns `match[2]` (the resourceId) and **discards `match[1]` (the siteId)**. The frontend then suppresses the shadow warning on `destinationResourceId === Number(pageId)`. If a redirect's stored reference ever embedded a *different* site's id with a colliding resourceId, the warning would be wrongly suppressed. In practice redirects are siteId-scoped and references are minted with the same siteId, so this is benign today — but the resolver throwing away the site context is a latent footgun. Worth either asserting the embedded siteId matches or noting the invariant. (`getRedirectBySource` is queried per-site, so the collision requires same-resourceId across sites — unlikely but not impossible.)

Confidence: low

**suggestion (tests):** `page.router.test.ts` / `resource.router.test.ts` — the new tests exercise `Page` only, but both routers explicitly special-case `CollectionPage` (`type === ResourceType.Page || type === ResourceType.CollectionPage`). A `CollectionPage` case in at least one flow would lock in that branch. Also, no test exercises the soft-deleted-redirect *revive* path in `createRedirectForPermalinkChange` (the `onConflict ... doUpdateSet` where a prior redirect at the old source was soft-deleted) — the reasoning that "the upsert revives it" is asserted by the comment but not by a test.

Confidence: medium

**note (correctness):** `resource.router.ts:L~575` and `page.router.ts:L~880` — the comment "Block before mutating anything further" precedes `assertPermalinkNotShadowed`, but by this point the `UPDATE Resource` (move/rename) has already run earlier in the transaction. Correctness still holds because the throw rolls back the whole transaction — but the comment implies an ordering guarantee the code doesn't have. Either move the check above the `UPDATE`, or soften the comment to "the throw rolls the move back."

Confidence: medium

**nitpick (non-blocking):** `PageSettingsModal.tsx:L~73` and `MoveResourceModal.tsx:L~208` — `<Checkbox size="1.25rem" ...>`. Chakra/design-system `Checkbox` `size` expects a token (`sm`/`md`/`lg`); a raw `"1.25rem"` typechecks via the `string & {}` escape hatch but won't map to a defined size and will silently fall back to the default. If a larger checkbox was intended, this is a no-op; if the default is fine, drop the prop. There's no other `Checkbox size=` usage in the repo to copy from.

Confidence: low

**praise:** `redirect.service.ts:L455-461` — the comment block on `createRedirectForPermalinkChange` that spells out *why* the existing-page/loop/format guards can be skipped (server-derived source+destination, published page never on a live source) is exactly the kind of invariant-documentation that makes this safe to maintain. Same for the publish-block mirror reasoning on `assertPermalinkNotShadowed`.

Confidence: high

dcshzj:
* **#2508** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2508?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2508?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2466** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2466?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2490** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/studio/src/schemas/page.ts
```diff
@@ -124,6 +124,10 @@ export const getRootPageSchema = z.object({
 
 export const basePageSettingsSchema = basePageSchema.extend({
   title: pageTitleSchema,
+  // Create a redirect from the page's old URL when its permalink changes (acted
+  // on only for Page/CollectionPage). On the base so the union destructures
+  // cleanly. Defaults off.
+  shouldCreateRedirect: z.boolean().optional().default(false),
```
seaerchin: question: why is the backend taking a different opinion from frontend? frontend says that the default is `true` but the backend defaults to `false`. i lean slightly towards both being `true` but i don't feel very strongly haha

apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
```diff
@@ -126,6 +129,13 @@ const PageSettingsModalContent = ({
     },
   )
 
+  const originalPermalink = permalinkTree[permalinkTree.length - 1] ?? ""
+  // Offer the redirect only when a Page/CollectionPage URL actually changes.
+  const showRedirectOption =
+    (type === ResourceType.Page || type === ResourceType.CollectionPage) &&
```
seaerchin, line 138: question: do we not need to normalise permalink here? i'm assuming this is because the URL returned by `permalinkTree` is already normalised?

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -112,6 +117,39 @@ const MoveResourceContent = withSuspense(
...
+        siteId: Number(siteId),
+        resourceId,
+      })
+    const { data: destination } = trpc.resource.getWithFullPermalink.useQuery(
+      { siteId: Number(siteId), resourceId: curResourceId ?? "" },
+      { enabled: !!curResourceId },
+    )
+
+    // Only Page/CollectionPage have a URL worth preserving.
+    const isRedirectableType =
```
seaerchin, line 133: question: why don't we do this for folders/collections too? i was under the impression, from the previous PRs, that we are doing mass redirects when a folder URL changes. is this scheduled in a future PR? 

ok w/ not doing it now and just maintaining the 1-1 redirect on page move but just curiou

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -867,6 +875,33 @@ export const pageRouter = router({
...
+              updatedResource.permalink !== resource.permalink
+            ) {
+              const parentFullPermalink = resource.parentId
+                ? await getResourceFullPermalink(
+                    siteId,
+                    Number(resource.parentId),
+                  )
+                : null
+              const oldFullPermalink = `${parentFullPermalink ?? ""}/${resource.permalink}`
+              const newFullPermalink = `${parentFullPermalink ?? ""}/${updatedResource.permalink}`
```
seaerchin, line 892: comment: reads abit convoluted - i was thinking we should just compare `parentId` and your `permalink`?

though i think that will miss the case where the parentId changes (although this does too at present)

seaerchin (commented):
i think this looks generally correct; have a few questions below but going to loop in @greptileai to help review too

seaerchin:
@greptileai

apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
```diff
@@ -126,6 +129,13 @@ const PageSettingsModalContent = ({
...
+    permalink !== originalPermalink
```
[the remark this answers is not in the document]
dcshzj: Fixed — `showRedirectOption` now also gates on `publishedVersionId !== null` (destructured from `trpc.page.readPage`), so the checkbox no longer shows for draft pages.

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -129,6 +167,44 @@ const MoveResourceContent = withSuspense(
               existingResource={movedItem ?? undefined}
               onChange={(resourceId) => setCurResourceId(resourceId)}
             />
+            {showRedirectOption && (
+              <VStack alignItems="flex-start" spacing="0.5rem" w="full">
+                {/* Suppressed when the redirect points back at this page —
+                    the move auto-clears it, so it won't actually shadow. */}
+                {existingRedirect &&
```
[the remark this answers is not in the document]
dcshzj: Fixed — `getMetadataById` now returns `publishedVersionId`, and the move modal gates `showRedirectOption` on it, so the checkbox is hidden for unpublished pages.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -511,21 +723,37 @@ export const deleteRedirect = async ({
...
   if (!redirect) {
     return null
   }
+  // Resolve via the full reference so we can confirm the embedded siteId matches
+  // this site. getResourceIdFromReferenceLink discards the siteId, and the caller
+  // suppresses the shadow warning on a resourceId match — a reference embedding a
+  // different site's id with a colliding resourceId would suppress it wrongly. The
+  // lookup is already siteId-scoped so the ids should match; if they don't, fall
+  // back to null (show the warning) rather than trust a cross-site id.
+  const match = REFERENCE_LINK_REGEX.exec(redirect.destination)
```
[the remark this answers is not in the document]
dcshzj: Fixed — switched to the file-local anchored `REFERENCE_DESTINATION_REGEX` here.

apps/studio/src/schemas/page.ts
dcshzj: Good point — aligned the backend default to `true` to match the checkbox's default-checked state, in both `basePageSettingsSchema` and `moveSchema`.

apps/studio/src/features/dashboard/components/PageSettingsModal.tsx
dcshzj: Right — no normalisation needed here. This compares single slug segments (the last `permalinkTree` entry vs the form `permalink`), both already normalised/slug-validated. `normalizeRedirectPath` is only needed in the Move modal, where we concatenate full parent paths.

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
dcshzj: Yeah, intentionally scoped to the 1-1 page redirect for now. Bulk/descendant redirects when a folder or collection URL changes is separate (future) work — keeping this PR to the page case.

apps/studio/src/server/modules/page/page.router.ts
dcshzj: In `updateSettings` the parent can't change — moves go through the `move` mutation — so detecting on `permalink` alone is correct and complete here. The `getResourceFullPermalink` call isn't change detection; it builds the redirect *source* (the old full URL). So the `parentId`-change case is structurally handled by the move flow, not missed.

seaerchin (approved):
lgtm; no blockers from me

PR #2512, opened 2026-06-18
🔖 feat: render categoryId correctly on published collection pages

adriangohjw:
> [!NOTE]
> this works for both pre/post migration of category 

## Summary

- **Bug**: Published collection pages were not rendering categories for items that used the new UUID-based `categoryId` field introduced in the category management backend. Items defaulted to \"Others\" because the rendering pipeline only read the legacy `category` string field.
- **Fix**: Threads `categoryId` (on child sitemap nodes) and `categoryOptions` (on the collection sitemap node) through `injectTagMappings` → `IsomerBaseSitemap` / `IsomerCollectionPageSitemap`, mirroring the existing `tagged`/`tagCategories` pattern.
- **UUID → label resolution**: `getCollectionItems` now looks up `categoryId` against `categoryOptions` and resolves it to the human-readable label, falling back to the legacy `category` string, then to `\"Others\"`.
- **Admin-defined ordering**: `getCategoryFilter` respects the order of `categoryOptions` when building the filter sidebar, rather than always sorting alphabetically. Categories not present in `categoryOptions` are appended at the end.
- **Storybook**: Added a `WithCategoryIdItems` story to `Collection.stories.tsx` so the resolved-label flow is visible in isolation.
- **Unit tests**: New test cases in `getCollectionItems.test.ts` and `getCategoryFilter.test.ts` cover UUID resolution, legacy fallback, unknown-UUID fallback, and ordering logic.

Closes https://linear.app/ogp/issue/ISOM-2302/feat-render-categoryid-correctly-on-published-collection-pages

## Test plan

- [ ] Run unit tests: `pnpm test:unit -- src/templates/next/layouts/Collection/utils/__tests__/getCollectionItems.test.ts` and `getCategoryFilter.test.ts` — all new cases pass
- [ ] Open Storybook (`pnpm storybook`) and view the **Collection → With categoryId items** story — filter sidebar shows \"Policy\", \"Research\", \"News\" with correct counts
- [ ] On a local environment, create a collection page with items assigned to categories via the category management UI, then visit the published page and confirm the category filter sidebar shows the correct labels in the admin-defined order
- [ ] Confirm that existing collection pages with the legacy `category` string field still render correctly (no regression)
- [ ] Confirm that items with neither `categoryId` nor `category` still fall back to \"Others\"

### DIFF
--- a/packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
+++ b/packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
@@ -35,11 +35,11 @@ export const getCategoryFilter = (
       label.toLowerCase(),
     )
 
-    const known = categoryFilterItems.filter(
-      (item) => categoryOptionLabels.indexOf(item.id) !== -1,
+    const known = categoryFilterItems.filter((item) =>
+      categoryOptionLabels.includes(item.id),
     )
     const unknown = categoryFilterItems.filter(
-      (item) => categoryOptionLabels.indexOf(item.id) === -1,
+      (item) => !categoryOptionLabels.includes(item.id),
     )
 
     known.sort(
--- a/packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
+++ b/packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
@@ -1,10 +1,12 @@
 import type { ProcessedCollectionCardProps } from "~/interfaces"
 import type { Filter } from "~/templates/next/types/Filter"
+import type { CollectionPageCategoryOption } from "~/types/page"
 
 import { FILTER_ID_CATEGORY } from "./constants"
 
 export const getCategoryFilter = (
   items: ProcessedCollectionCardProps[],
+  categoryOptions?: CollectionPageCategoryOption[],
 ): Filter => {
   const categories: Record<string, number> = {}
 
@@ -16,15 +18,31 @@ export const getCategoryFilter = (
     }
   })
 
-  const categoryFilterItems = Object.entries(categories)
-    .map(([label, count]) => ({
+  const categoryFilterItems = Object.entries(categories).map(
+    ([label, count]) => ({
       id: label.toLowerCase(),
       label: label.charAt(0).toUpperCase() + label.slice(1),
       count,
-    }))
-    .sort((a, b) =>
+    }),
+  )
+
+  if (categoryOptions && categoryOptions.length > 0) {
+    const categoryOptionLabels = categoryOptions.map(({ label }) => label)
+    categoryFilterItems.sort((a, b) => {
+      const indexA = categoryOptionLabels.indexOf(a.label)
+      const indexB = categoryOptionLabels.indexOf(b.label)
+
+      if (indexA === -1 && indexB === -1) return 0
+      if (indexA === -1) return 1
+      if (indexB === -1) return -1
+
+      return indexA - indexB
+    })
+  } else {
+    categoryFilterItems.sort((a, b) =>
       a.label.localeCompare(b.label, undefined, { numeric: true }),
     )
+  }
 
   return {
     id: FILTER_ID_CATEGORY,
--- a/packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
+++ b/packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
@@ -27,23 +27,39 @@ export const getCategoryFilter = (
   )
 
   if (categoryOptions && categoryOptions.length > 0) {
-    // Sort by the editor-defined order in categoryOptions; unknown categories fall to the end
-    const categoryOptionLabels = categoryOptions.map(({ label }) => label)
-    categoryFilterItems.sort((a, b) => {
-      const indexA = categoryOptionLabels.indexOf(a.label)
-      const indexB = categoryOptionLabels.indexOf(b.label)
+    // Split into two groups: items whose id matches a categoryOption (known) and those that don't
+    // (unknown). Known items are ordered by the editor-defined categoryOptions order; unknown items
+    // are appended alphabetically. Matching is done on the lowercased id (= lowercased label) so
+    // it is case-insensitive and stable even if the display label is later re-cased.
+    const categoryOptionLabels = categoryOptions.map(({ label }) =>
+      label.toLowerCase(),
+    )
 
-      if (indexA === -1 && indexB === -1) return 0 // both unknown, preserve order
-      if (indexA === -1) return 1 // a unknown, push after b
-      if (indexB === -1) return -1 // b unknown, push after a
+    const known = categoryFilterItems.filter(
+      (item) => categoryOptionLabels.indexOf(item.id) !== -1,
+    )
+    const unknown = categoryFilterItems.filter(
+      (item) => categoryOptionLabels.indexOf(item.id) === -1,
+    )
 
-      return indexA - indexB
-    })
-  } else {
-    categoryFilterItems.sort((a, b) =>
+    known.sort(
+      (a, b) =>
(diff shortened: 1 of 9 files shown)


adriangohjw:
* **#2512** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2512?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2524](https://github.com/opengovsg/isomer/pull/2524) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2524?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2526](https://github.com/opengovsg/isomer/pull/2526) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2526?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
```diff
@@ -16,15 +18,32 @@ export const getCategoryFilter = (
...
+    const categoryOptionLabels = categoryOptions.map(({ label }) => label)
+    categoryFilterItems.sort((a, b) => {
+      const indexA = categoryOptionLabels.indexOf(a.label)
+      const indexB = categoryOptionLabels.indexOf(b.label)
+
+      if (indexA === -1 && indexB === -1) return 0 // both unknown, preserve order
+      if (indexA === -1) return 1 // a unknown, push after b
+      if (indexB === -1) return -1 // b unknown, push after a
+
+      return indexA - indexB
```
seaerchin: comment: sense-check my understanding to make sure i'm getting this correct please! we're sorting the filter items by whether they are inside the option labels array or not; if they are, we use the user defined order and otherwise, stick them at the back based off their position in the existing array

issue: i didn't quite understand why we're using the `indexOf(a.label)` rather than `indexOf(a.id)` since the labels are, by definition, changeable. i'm assuming i'm missing context **but if i'm not**, please make the change prior to merge 

suggestion: i noticed the sorting here is via a `if/else`, which fails if, for example, the user has only 1 item in the new array but many items in the old array. this would mean that the old items are no longer sorted which isn't ideal. why don't we try the following:

1. split the array into those inside `cateogoryOptionLabels` and those not in
2. for those inside, sort via their existing order
3. for those outside, follow the existing `localeCompare`
4. merge the two arrays

this would preserve existing behaviour + simplify the logic here. not opionionated though, doesn't seem very important to me since this behaviour only kicks in if we have a migration failure which is presumably an edge case.

packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
```diff
@@ -128,7 +133,12 @@ export const getCollectionItems = ({
       id: item.permalink,
       date,
       lastModified: item.lastModified,
-      category: item.category || CATEGORY_OTHERS,
+      category:
```
seaerchin, line 136: suggestion: what do we feel about

```ts
        (item.categoryId && categoryOptions && (categoryOptions.find((opt) => opt.id === item.categoryId)?.label) ||
            (item.category ??
            CATEGORY_OTHERS))
```

seaerchin (approved):
no issues with this PR, some minor suggestions but overall not blocking. however, i don't see the publishing portion being done here (ie, how is the categoryOptions etc being passed to the built site) and i'm assuming this is being done in a PR down the line. if that assumption is not true, please dismiss this review

seaerchin (commented):
## Summary

Threads the new UUID-based category model through the published-collection rendering pipeline:

- **`sitemap.ts`** (`apps/studio/src/utils/sitemap.ts:273-356`) — `injectTagMappings` now also passes `categoryId` (child) and `categoryOptions` (collection index) into `_injectTagMappings`, mirroring the existing `tagged`/`tagCategories` threading. The extracted `childPageProps`/`collectionPageProps` locals are a nice readability win over the inline casts.
- **`getCollectionItems.ts:136-141`** — resolves `categoryId` → label via `categoryOptions`, with a fallback chain to legacy `category` then `"Others"`.
- **`getCategoryFilter.ts:29-46`** — orders filter items by the admin-defined `categoryOptions` order when present, else alphabetical.
- **`Collection.tsx` / `getAvailableFilters.ts`** — plumb `categoryOptions` from `page` down into the two utils.
- Good unit coverage of the fallback chain and ordering, plus a Storybook story.

Overall solid and backward-compatible. One correctness concern in the ordering join, and one architectural confirmation (already raised by @seaerchin).

---

## Findings

**issue (correctness):** `getCategoryFilter.ts:24` & `:33` — the ordering join is between two differently-cased strings, so known categories can be silently pushed to the end.

The filter item's `label` is display-transformed (`label.charAt(0).toUpperCase() + label.slice(1)`), but `categoryOptionLabels` holds the *raw* option labels. `categoryOptionLabels.indexOf(a.label)` therefore only matches when the admin's label already happens to start with an uppercase letter and is otherwise identical. The schema (`page.ts:106`) permits any trimmed non-empty string, so labels like `"policy"`, `"eGov updates"`, or `"POLICY"` won't match — `indexOf` returns `-1` and the category is treated as "unknown" and shoved to the end, even though it's a first-class option. The existing tests only use already-capitalized labels (`"Policy"`, `"Research"`, `"News"`), so this slips through.

The same case mismatch can also split one logical category into two filter buckets during migration: a `categoryId`-resolved item (label `"Policy"`) and a legacy item whose `category` string is `"POLICY"` produce two entries, since both bucketing and ordering key off the cased string.

Suggest joining on a normalized form — e.g. compare `id`s (`getCategoryFilter` already computes `id: label.toLowerCase()`) against `categoryOptions` labels lowercased, rather than the display label. That makes both the ordering and (separately worth considering) the bucketing robust to casing.

Confidence: medium — depends on whether admin-entered labels are reliably title-cased in practice; the schema does not enforce it.

**note (non-blocking, architecture):** `apps/studio/src/utils/sitemap.ts` — the `injectTagMappings` change only affects the **preview** sitemap (`resource.service.ts:592`, "inject tag mappings for the preview"). The published site is built by a separate pipeline, which must independently emit `categoryId` on items and `categoryOptions` on the collection index into the published sitemap JSON for the `packages/components` rendering changes to take effect live. This matches @seaerchin's open question — flagging only to confirm the publish-side wiring is a tracked follow-up, not dropped.

Confidence: high.

**thought (non-blocking):** `getCollectionItems.ts:137-141` — the resolution ternary is doing a fair bit (presence check + lookup + double fallback). Pulling it into a small `resolveCategory(item, categoryOptions)` helper would read more clearly and be directly unit-testable, though it's already covered indirectly. Optional.

Confidence: low.

**praise:** the fallback chain in `getCollectionItems` (categoryId → option label → legacy `category` → `"Others"`) is exactly right for the mixed pre/post-migration state, and the test cases enumerate each branch explicitly. Mirroring the established `tagged`/`tagCategories` plumbing keeps the change easy to follow.

packages/components/src/templates/next/layouts/Collection/utils/getCollectionItems.ts
adriangohjw: im neutral, since this will be removed in the cleanup post-release (after the migration)

packages/components/src/templates/next/layouts/Collection/utils/getCategoryFilter.ts
adriangohjw: sense-check: yep, that's correct!

indexOf label vs id: fixed in 5100145 — matching is done on `a.id` (lowercased label) against `categoryOptions.map(label => label.toLowerCase())`, so it's case-insensitive and stable.

sort suggestion: also fixed in 5100145 — switched to the split-sort-merge approach, with a new test covering multiple unknown categories sorted alphabetically.

PR #2513, opened 2026-06-18
🔖 fix: suppress legacy category free-text field when categoryId flag is on

adriangohjw:
> [!NOTE]
> Both feature flags will be gone once the feature goes live 

## Summary

- When `CATEGORY_ID_DROPDOWN_FEATURE_KEY` is enabled for a site, `JsonFormsCategoryControl` (the legacy free-text category field) now returns `null`, so editors only see the new `categoryId` dropdown — not both fields simultaneously.
- The legacy `category` value stored in the blob is preserved unchanged; the component simply hides itself rather than clearing any data.
- Adds a `createCategoryIdDropdownGbParameters` helper to `stories/utils/growthbook.ts` to keep Storybook GrowthBook setup consistent with existing helpers.
- Adds `FlagOff` and `FlagOn` Storybook stories for `JsonFormsCategoryControl` to document and visually verify both states.

Closes https://linear.app/ogp/issue/ISOM-2304/feat-suppress-legacy-category-free-text-field-when-categoryid-flag-is

## Test plan

- [ ] With `CATEGORY_ID_DROPDOWN_FEATURE_KEY` **disabled** for the site: the legacy free-text category field renders as expected (no regression).
- [ ] With `CATEGORY_ID_DROPDOWN_FEATURE_KEY` **enabled** for the site: the legacy free-text category field is hidden; only the `categoryId` dropdown is visible.
- [ ] Existing `category` value in the blob is not cleared when the flag is turned on — data is preserved.
- [ ] Storybook: `FlagOff` story shows the legacy free-text field; `FlagOn` story renders nothing (empty canvas).

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryControl.tsx
@@ -8,7 +8,10 @@ import Suspense from "~/components/Suspense"
 import { JSON_FORMS_RANKING } from "~/constants/formBuilder"
 import { collectionItemSchema } from "~/features/editing-experience/schema"
 import { useQueryParse } from "~/hooks/useQueryParse"
-import { CATEGORY_DROPDOWN_FEATURE_KEY } from "~/lib/growthbook"
+import {
+  CATEGORY_DROPDOWN_FEATURE_KEY,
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
question: should we also do this on the backend?

adriangohjw:
> question: should we also do this on the backend?

@seaerchin IMO no need since there's no security issue in allowing it - it just that we want to gate the UI to reduce confusion. Also mostly defence-in-depth - will be removed during launch (post-migration)

PR #2514, opened 2026-06-18
🔖 fix: surface tag category and categoryId UX for collection item editors

adriangohjw:
## Summary

- **Preserve `categoryId` in blob write**: `updateCollectionLink` now persists `categoryId` so it is not silently dropped on save
- **Tag category required/optional state**: `JsonFormsTaggedControl` shows an asterisk per tag category based on the `isRequired` flag from collection settings (cosmetic only — does not block saves)
- **CategoryId dropdown**: non-clearable and visually marked as required
- **MultiSelect pill styling**: selected tag options rendered as pills (`border-radius: 50px`)
- **Cleanup**: removed stale "contact your site owner(s)" description from the `tagged` field schema
- **Published-only tag categories**: `getCollectionTags` now accepts an `isPublishedCategories` boolean (defaults to `false`); the collection item editor passes `true` so only published collection settings are shown, consistent with `getCategoryOptions`. The original draft-fallback behaviour is preserved for other callers (e.g. `GazetteSubcategoriesContext`)

Closes https://linear.app/ogp/issue/ISOM-2306/fix-surface-tag-category-and-categoryid-ux-for-collection-item-editors

## Test plan

- [ ] Edit a CollectionLink — confirm `categoryId` is preserved after saving
- [ ] Open a collection item in a collection with required tag categories — confirm asterisk appears on required categories and not on optional ones
- [ ] Open a collection item with a `categoryId` dropdown — confirm it is non-clearable and shows the required asterisk
- [ ] Verify MultiSelect selected tag options render as pills
- [ ] Before publishing collection settings: confirm tag categories/categoryId dropdown shows no options in the item editor
- [ ] After publishing collection settings: confirm tag categories and categoryId options appear correctly in the item editor

### DIFF
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
@@ -54,7 +54,7 @@ function SuspendableJsonFormsCategoryIdSelect({
 
 export function JsonFormsCategoryIdControl({
   description,
-  required,
+  required: _required,
   label,
   ...props
 }: ControlProps) {
--- a/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
+++ b/apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
@@ -44,7 +44,7 @@ function SuspendableJsonFormsCategoryIdSelect({
         label: optionLabel,
         value: id,
       }))}
-      isClearable={true} // TODO: change to false after migration where it's no longer optional
+      isClearable={false}
       onChange={(value) => {
         handleChange(path, value)
       }}
@@ -77,7 +77,7 @@ export function JsonFormsCategoryIdControl({
   if (resourceId === undefined) return null
 
   return enabledSites.includes(parsedQuery.data.siteId.toString()) ? (
-    <FormControl isRequired={required} gap="0.5rem">
+    <FormControl isRequired gap="0.5rem">
       <FormLabel description={description}>{label}</FormLabel>
       <ErrorBoundary fallbackRender={() => null}>
         <Suspense fallback={<Skeleton />}>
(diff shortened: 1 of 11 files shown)


apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
```diff
@@ -44,7 +44,21 @@ function SuspendableJsonFormsCategoryIdSelect({
...
+      //   State                                 | isClearable=false
+      //   --------------------------------------|------------------------------------------
+      //   Pre-migration, old item (no categoryId) | nothing to clear — harmless
+      //   Pre-migration, new item (has categoryId) | can't clear — correct
+      //   Migration ran, schema not yet updated  | can't clear — protects migrated data
+      //   Migration ran + schema updated         | can't clear — correct
+      //
+      // Once migration is complete and categoryId is made required in the schema,
+      // replace with isClearable={!required}.
+      isClearable={false}
```
seaerchin, line 61: comment: if this is important to ensure correctness, we might want to also block on backend.

apps/studio/src/schemas/collection.ts
```diff
@@ -87,6 +88,7 @@ export const getCollectionTagsSchema = z
     resourceId: z.number().min(1).optional(),
     collectionId: z.number().min(1).optional(),
     siteId: z.number().min(1),
+    isPublishedCategories: z.boolean().default(false),
```
seaerchin: comment: i saw that your PR added this to account for GazetteSubcategoriesContext, but can just remove and set it to `true`; i've already published the index page for egazette so it should fetch the published subcategories regardless

seaerchin (approved):
lgtm; i think we should collapse the isPublishedCategories to be `true`, but happy to do that as a follow-up. can we file a ticket on linear so we don't lose track of it?

apps/studio/src/schemas/collection.ts
adriangohjw: removed, thanks! https://github.com/opengovsg/isomer/pull/2514/changes/ddf33b8b0952a75cbcc23ed9ea053c40e70ea32c

apps/studio/src/features/editing-experience/components/form-builder/renderers/controls/JsonFormsCategoryIdControl.tsx
adriangohjw: thought about it but decided not to proceed - this is only an issue if someone calls the TRPC procedure directly and not possible via the UI. Im okay to risk accept it and keep the code simpler for now - esp. since this will be fixed once post-migration

LMK if this is an issue for you!

PR #2515, opened 2026-06-18
🔖 chore: migration script — populate categoryOptions and categoryId from legacy category strings (DO NOT MERGE)

adriangohjw:
## Summary

- Adds a one-shot off-hours migration script \`apps/studio/prisma/scripts/migrate-category-to-categoryid.ts\` that populates \`categoryOptions\` on each Collection Index and backfills \`categoryId\` on each Collection Item (CollectionPage + CollectionLink) from the existing legacy free-text \`category\` string field
- Strategy: auto-migrate as-is — deduplicate category strings (case-sensitive, after trim), generate a UUID per distinct label, assign it as \`categoryId\` on each child item; legacy \`category\` strings are preserved (not removed)
- Idempotent: a collection is skipped when its Collection Index draft blob already contains a non-empty \`categoryOptions\` array. This is safe because the index \`categoryOptions\` write and all child \`categoryId\` writes happen in a single transaction — non-empty \`categoryOptions\` guarantees the children were also stamped. ⚠️ Do NOT manually clear \`categoryOptions\` and re-run: the script generates fresh UUIDs each time, which would break existing \`categoryId\` references.
- Operates only on draft blobs; published content is used as a fallback for items that have no draft
- Adds 25 unit tests covering pure data-transformation helpers (\`hasCategoryOptions\`, \`buildLabelMap\`, \`buildCategoryOptions\`) with no DB or Docker dependency. Tests live in \`prisma/scripts/__tests__/\` and are picked up by the standard \`vitest.config.ts\`.

> ⚠️ Run this script only after ISOM-2301, ISOM-2302, ISOM-2303, and ISOM-2304 are merged and deployed.

Closes https://linear.app/ogp/issue/ISOM-2305/chore-migration-script-populate-categoryoptions-and-categoryid-from

## Test plan

- [ ] Unit tests pass: \`cd apps/studio && pnpm exec vitest run --config vitest.config.ts prisma/scripts/__tests__/migrate-category-to-categoryid.test.ts\`
- [ ] Run against a staging DB and verify:
  - [ ] Collections with existing \`categoryOptions\` are skipped (idempotency)
  - [ ] Collections without \`categoryOptions\` get the correct options generated from child \`category\` strings
  - [ ] Each Collection Item's draft blob gains a \`categoryId\` matching the corresponding option UUID
  - [ ] Re-running the script a second time produces no additional changes (idempotency confirmed)
- [ ] Run against production DB during off-hours after ISOM-2301–2304 are deployed


adriangohjw (commented):
## Summary

Adds a one-shot data migration script (`prisma/scripts/migrate-category-to-categoryid.ts`) that reads legacy free-text `category` strings from CollectionPage/CollectionLink draft blobs, synthesises a deduplicated `categoryOptions` array on each Collection Index, and stamps a matching `categoryId` UUID on each child item. Pure data-transformation helpers are extracted and covered by 25 unit tests with no DB dependency.

**Code flow**: `main()` iterates all sites → all Collections per site → per-collection Kysely transaction (steps 3–11 in the file). The single-transaction-per-collection design is the key correctness invariant: atomicity means the idempotency check (non-empty `categoryOptions` on the index draft blob) is sufficient to detect a completed migration. If a transaction throws, the whole script aborts; a re-run skips already-migrated collections.

**Overall**: Solid work for a migration script. The helper extraction + unit tests is the right call for a script touching production data. A few things to address before running against prod.

---

**issue (non-blocking):** `prisma/scripts/migrate-category-to-categoryid.ts:L330` — `existingDraftBlobId = newBlob.id` is dead code. The variable is read at lines 210 and 309, but nothing reads it after line 330 within the transaction. The reassignment is leftover from an earlier iteration and can be removed without any behavioural change.

Confidence: high

---

**issue:** The PR body and test plan are stale and will mislead whoever runs the pre-deploy checklist:
- **PR body** still says the script lives at `apps/studio/scripts/...` (it was moved to `prisma/scripts/`).
- **PR body** lists `vitest.scripts.config.ts` as a deliverable — that file was deleted.
- **PR body** says "Minor fix: preserve `categoryId` when saving a collection link" — that change was reverted and is in PR #2516.
- **Test plan item 1** references `pnpm exec vitest run --config vitest.scripts.config.ts`, which will fail with "config file not found". The correct command is now `pnpm test:unit` or `pnpm exec vitest run --config vitest.config.ts prisma/scripts/__tests__/migrate-category-to-categoryid.test.ts`.
- **Test plan item 2** says "Confirm the script prints a dry-run summary before writing to DB" — there is no dry-run mode; the script writes immediately.

Confidence: high

---

**note:** `config({ path: resolve(__dirname, "../../.env") })` runs at module level (line 37), outside the `process.argv[1]` guard. Any import of this file (e.g. from the test suite) will load `apps/studio/.env` into `process.env`. This is benign for the current unit tests since none of them make DB calls, but it means running the test suite on a machine with a real `.env` will inject `DATABASE_URL` (and any other local secrets) into the test process environment. No action required for this PR specifically, but worth knowing if additional tests are added that assert on env state.

Confidence: high

---

**thought:** The idempotency check is: _skip if `categoryOptions` is a non-empty array on the collection index draft blob_. The single-transaction design makes this sufficient for the happy path. But if an operator manually clears `categoryOptions` from a collection's draft and re-runs the script, the script will generate **new** UUIDs for the same category labels. Any `categoryId` values the app already stored (from the first run) would no longer match the new UUIDs, silently breaking category filtering for items in that collection. This edge case probably doesn't matter for a one-shot migration, but it's worth a line in the runbook: "do not re-run after partial manual edits to `categoryOptions`."

Confidence: medium

PR #2526, opened 2026-06-18
🔖 fix: show correct category and tags in collection link preview

adriangohjw:
## Problem

The collection link preview (`EditLinkPreview`) was rendering items with:
1. **Wrong category label** — always fell back to the legacy `category` string instead of deriving the label from `categoryId` + `categoryOptions`
2. **No tag pills** — `tagged` UUIDs were never resolved to tag labels because `tagCategories` wasn't passed to the layout
3. **Wrong filters** — filter options were empty since `tagCategories` and `categoryOptions` weren't present

All three stem from `PreviewWithCustomSitemap` being called with `page={{ title: parentTitle }}`. `CollectionLayout` reads `tagCategories` and `categoryOptions` from `page` to process items and build filters, but neither field was present.

Closes https://linear.app/ogp/issue/ISOM-2302/feat-render-categoryid-correctly-on-published-collection-pages

## Fix

- Fetches `categoryOptions` via `trpc.page.getCategoryOptions` (same link ID pattern as the existing `getCollectionTags` call)
- Passes both `tagCategories` and `categoryOptions` in the `page` prop so `CollectionLayout` can:
  - Map `categoryId` → display label
  - Convert `tagged` UUID arrays → tag objects
  - Build correct filter options

The `sitemap.ts` type (`categoryOptions` in `collectionPagePageProps`) is already added by the parent branch — no new type changes needed here.

## Not covered by existing branches

Verified that neither `fix/required-tag-ux` nor `feat/categoryid-published-site-rendering` touches `EditLinkPreview.tsx`.

## Test plan

- [ ] Open a collection link with `categoryId` set — preview shows the correct category label (not "Others" or the raw string)
- [ ] Open a collection link with `tagged` values — preview shows tag pills
- [ ] Filter dropdown in the collection preview shows correct category options and tag categories

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2524, opened 2026-06-18
chore(deps): bump lodash-es in the lodash-es group (#2517)

adriangohjw:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [ ] No - this PR is backwards compatible

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

**Manual Verification Steps**:

- [ ] Step 1
- [ ] Step 2
- [ ] Step 3

**New scripts**:

- `script` : script details

**New dependencies**:

- `dependency` : dependency details

**New dev dependencies**:

- `dependency` : dependency details

PR #2527, opened 2026-06-18
🔖 fix: include categoryId and tagged in localised sitemap for collection preview

adriangohjw:
## Problem

The collection index page preview (when editing a collection's display settings) showed items with:
1. **Wrong category label** — `getLocalisedSitemap` only extracted the legacy `category` string from the published blob; `categoryId` was never included, so `getCollectionItems` couldn't map it to the correct label
2. **No tag pills** — `tagged` UUIDs were missing from sitemap items, so `getCollectionItems` couldn't resolve them to tag objects
3. **Wrong filter options** — filters derive from processed items, so they also showed incorrect data

`page.tagCategories` and `page.categoryOptions` were already correct (from `previewPageState`). The missing piece was `categoryId` and `tagged` on each child item in the sitemap.

Closes https://linear.app/ogp/issue/ISOM-2302/feat-render-categoryid-correctly-on-published-collection-pages

## Fix

**`resource.service.ts`** — adds two new SQL expressions to `getLocalisedSitemap`:
- `categoryIdSql`: extracts `categoryId` from the published blob for article/link items; `NULL` for others
- `taggedSql`: extracts `tagged` (as JSON text) from the published blob for article/link items; `NULL` for others

Threads both columns through all CTEs (null literals for ancestor/nestedResources rows, which are folders/collections/index pages).

**`sitemap.ts`** — updates `ResourceDto` to include the new fields, and passes `categoryId` and parsed `tagged` into the `CollectionPage` and `CollectionLink` sitemap nodes in `getSitemapTreeFromArray`.

## Scope

Only affects `getLocalisedSitemap` (the Studio preview path). Published site rendering is unchanged.

**Note**: The sitemap only includes *published* collection items (pre-existing behaviour — draft items don't appear in the preview).

## Test plan

- [ ] Edit a collection's display settings — preview shows items with correct `categoryId`-derived category labels
- [ ] Preview shows tag pills on items that have `tagged` values
- [ ] Filter dropdown shows correct category options and tag categories

### DIFF
--- a/apps/studio/src/server/modules/resource/resource.service.ts
+++ b/apps/studio/src/server/modules/resource/resource.service.ts
@@ -409,6 +409,20 @@ export const getLocalisedSitemap = async (
       ELSE ''
     END
 `.as("content")
+  const categoryIdSql = sql<string | null>`
+    CASE
+      WHEN (published.content ->> 'layout') IN ('article','link')
+      THEN (published.content -> 'page' ->> 'categoryId')
+      ELSE NULL
+    END
+`.as("categoryId")
+  const taggedSql = sql<string | null>`
+    CASE
+      WHEN (published.content ->> 'layout') IN ('article','link')
+      THEN (published.content -> 'page' ->> 'tagged')
+      ELSE NULL
+    END
+`.as("tagged")
 
   // Get the actual resource first
   const resource = await getById(db, { resourceId, siteId })
@@ -431,6 +445,8 @@ export const getLocalisedSitemap = async (
           categorySql,
           dateSql,
           contentSql,
+          categoryIdSql,
+          taggedSql,
           ...defaultResourceSelect,
         ])
         .unionAll((fb) =>
@@ -449,6 +465,8 @@ export const getLocalisedSitemap = async (
               eb.cast<string>(eb.val(""), "text").as("category"),
               eb.cast<string>(eb.val(""), "text").as("date"),
               eb.cast<string>(eb.val(""), "text").as("content"),
+              eb.cast<string | null>(eb.val(null), "text").as("categoryId"),
+              eb.cast<string | null>(eb.val(null), "text").as("tagged"),
               ...defaultResourceSelect,
             ]),
         ),
@@ -476,6 +494,8 @@ export const getLocalisedSitemap = async (
           categorySql,
           dateSql,
           contentSql,
+          categoryIdSql,
+          taggedSql,
           ...defaultResourceSelect,
         ]),
     )
@@ -492,12 +512,14 @@ export const getLocalisedSitemap = async (
         .where("Resource.state", "=", ResourceState.Published)
         .leftJoin("Version", "Version.id", "Resource.publishedVersionId")
         .leftJoin("Blob as published", "Version.blobId", "published.id")
-        .select(() => [
+        .select(({ eb }) => [
           headerSql,
           thumbnailSql,
           categorySql,
           dateSql,
           contentSql,
+          eb.cast<string | null>(eb.val(null), "text").as("categoryId"),
+          eb.cast<string | null>(eb.val(null), "text").as("tagged"),
           ...defaultResourceSelect,
         ])
         .unionAll((fb) =>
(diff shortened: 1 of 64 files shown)


adriangohjw:
* **#2527** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2527?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2527?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2526** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2526?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2512** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2512?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2524](https://github.com/opengovsg/isomer/pull/2524) <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2524?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

adriangohjw (commented):
## Summary

This PR fixes the collection index page preview by extracting `categoryId` and `tagged` fields from published blobs in `getLocalisedSitemap`, then threading them into the sitemap tree so `getCollectionItems` can resolve category labels and tag pills correctly.

**Code flow:**
- `resource.service.ts:412-425` — two new SQL expressions (`categoryIdSql`, `taggedSql`) using PostgreSQL's `->>` operator to extract values from `published.content -> 'page'`. NULL literals are used for non-article/link rows (ancestors, nestedResources, index pages) so all CTEs have the same column shape.
- `resource.service.ts:448-590` — both columns threaded through all four CTEs and the final UNION select.
- `sitemap.ts:119-122, 145-148` — `categoryId` passed as `undefined` when null; `tagged` JSON-parsed from the raw string returned by `->>` into `string[]`.

**Overall assessment:** Straightforward read-only change with correct null-handling throughout. The rest of this review focuses on the specific safety question: is this safe to merge before the PR #2515 data migration has run?

---

**answer: yes, safe to merge before the #2515 migration runs**

The concern is: before the migration backfills `categoryId` / `tagged` into existing published blobs, those JSON keys don't exist. Here's why that's safe:

1. PostgreSQL's `->>` operator returns `NULL` (not an error) for a missing JSON key. So `(published.content -> 'page' ->> 'categoryId')` silently returns `NULL` for every pre-migration item.
2. `NULL` published blobs are also handled — the `LEFT JOIN` to `Version` / `Blob` means `published.content` can be NULL. `NULL ->> 'layout'` is NULL, which is not `IN ('article','link')`, so the CASE falls to `ELSE NULL`. No crash.
3. `categoryId: resource.categoryId ?? undefined` (`sitemap.ts:119`, `145`) — NULL → `undefined`, fine.
4. `tagged: resource.tagged ? JSON.parse(...) : undefined` (`sitemap.ts:120-122`, `146-148`) — NULL is falsy → `undefined`, fine.

Pre-migration behaviour: both fields are `undefined` on every item → preview is identical to before this PR (no category/tag info shown). Post-migration behaviour: fields are populated, preview improves. No regression either way.

---

**issue (non-blocking):** `sitemap.ts:121`, `sitemap.ts:147` — `JSON.parse(resource.tagged)` is called without a try/catch. If a published blob somehow contains a malformed JSON string for `tagged` (e.g. a partially-written record, a manual DB edit), this throws an unhandled exception that would bubble up as a 500 on the preview endpoint. The mitigation is a try/catch with a fallback to `undefined`. Low probability given that application writes are schema-validated, but worth noting for robustness.

Confidence: medium

---

**note:** The Vercel deployment shows an error status on this PR, but that's a pre-existing CI issue unrelated to this change (no code was added that would cause a build failure — the diff is purely additive SQL expressions and two type fields).

Confidence: high (based on diff scope)

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -449,6 +465,8 @@ export const getLocalisedSitemap = async (
               eb.cast<string>(eb.val(""), "text").as("category"),
               eb.cast<string>(eb.val(""), "text").as("date"),
               eb.cast<string>(eb.val(""), "text").as("content"),
+              eb.cast<string | null>(eb.val(null), "text").as("categoryId"),
+              eb.cast<string | null>(eb.val(null), "text").as("tagged"),
```
dcshzj, line 469: question(non-blocking): unrelated to this PR but doesn't this mean that actually the other fields that we are extracting (e.g. `category` and `date`) should also have their types updated?

Seems like the NULL can also appear for them

dcshzj (approved):

apps/studio/src/server/modules/resource/resource.service.ts
adriangohjw: yeah ideally but given that they are likely created via studio, those values should be there. in this case, these values like `categoryId` is here because its a field to be patched, so there's a much likelier chance of it being not present if the deployment was made prior to the migration

dcshzj: > yeah ideally but given that they are likely created via studio, those values should be there

I think not quite true especially for `date`, as it is an optional field, unless we force it to be an empty string but from what I recall we completely remove the key.

PR #2584, opened 2026-06-28
fix(studio): guard site admin config page

### LINKED ISSUE #2377: (warden) Missing early return after auth check renders admin form to non-admin users and fires side effects during render
# Missing early return after auth check renders admin form to non-admin users and fires side effects during render

**Skill:** code-review

**Severity:** 🟡 MEDIUM

**ID:** `CFZ-WE6`

**Confidence:** high

**Location:** `apps/studio/src/pages/sites/[siteId]/admin.tsx:54-59`

## Description

The authorization check at lines 54–59 calls `toast()` and `void router.push()` but never returns early, so the component continues past the guard and renders the full admin form—including all four site config text areas populated with real config data—to any authenticated user who is not an Isomer admin, for as long as the async `router.push` takes to complete. Additionally, both `toast()` and `router.push()` are called directly in the render body rather than inside `useEffect`; with `reactStrictMode: true` confirmed in `next.config.mjs`, React Strict Mode double-invokes the render function in development, so non-admin users see two error toasts and two concurrent navigation calls are fired. The fix is to add `return null` (or a loading spinner) immediately after the redirect block so rendering stops for unauthorized users.

## Verification

Confirmed `reactStrictMode: true` in apps/studio/next.config.mjs. Checked apps/studio/src/hooks/useIsUserIsomerAdmin.ts: returns `{ isAdmin: isAdmin ?? false, isLoading }` where `isAdmin` comes from a regular `useQuery` (not suspense), so `isLoading` can be false before the suspense queries settle. When `!isLoading && !isUserIsomerAdmin` is true, no `return null` exists—execution falls through to all four `useSuspenseQuery` calls and the JSX return that renders the full form with site config data.

Benjam11n:
## Problem

The site admin configuration page redirects non-Isomer-admin users, but the redirect and toast were previously triggered during render and the component continued rendering before navigation completed.

This meant unauthorized authenticated users could briefly mount the site config queries and see a flash of the admin form populated with config data.

Closes #2377

## Solution

Moved the sensitive admin form into a separate `SiteAdminForm` component that is only rendered after the Isomer admin check passes.

The page now:
- waits for the Isomer admin check to finish
- returns `null` for loading or unauthorized users
- triggers the unauthorized toast and redirect from `useEffect`
- guards the redirect effect so it only fires once

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- N/A

**Bug Fixes**:

- Prevent unauthorized users from seeing even a transient flash of the site admin configuration form.
- Prevent admin-only config, theme, navbar, and footer queries from mounting before the Isomer admin check passes.
- Move unauthorized redirect and toast side effects out of render.

## Before & After Screenshots

**BEFORE**:

N/A

**AFTER**:

N/A

## Tests

- [ ] Log in as a user who is not an Isomer Core or Migrator admin.
- [ ] Open browser DevTools and open the Network tab.
- [ ] Navigate directly to `/sites/1/admin`.
- [ ] Verify that the admin configuration form does not flash on screen at any point before redirect.
- [ ] Verify that the user sees the unauthorized error toast and is redirected to `/sites/1`.
- [ ] In the Network tab, verify that the page does not fetch the admin config data endpoints for site config, theme, navbar, or footer.
- [ ] Log in as an Isomer Core or Migrator admin.
- [ ] Navigate to `/sites/1/admin`.
- [ ] Verify that the site configuration form loads normally and the config, theme, navbar, and footer fields are populated.

**New scripts**:

- N/A

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

This PR fixes a security gap where unauthorized (but authenticated) users could briefly see the admin configuration form and trigger its data-fetching queries before the client-side redirect completed. The fix moves the Isomer admin check from a `useIsUserIsomerAdmin` hook inside the component to a `getServerSideProps` function that runs before the page is sent to the browser.

- **`admin.tsx`**: Adds `getServerSideProps` that reads the iron-session cookie server-side, calls `isActiveIsomerAdmin`, and returns a 302 redirect for non-admin users — the page component and its tRPC queries never mount for unauthorized visitors. The component now receives `siteId` as a typed prop instead of parsing it from the URL at runtime.
- **`session.ts`**: Makes the parameter to `generateSessionOptions` optional (adds `= {}` default destructuring) so the function can be called without arguments from `getServerSideProps`; all existing call-sites remain unaffected.
</details>

Safe to merge — the server-side guard is correctly implemented, the session and admin checks are well-tested utilities already used throughout the codebase, and the component-level change is a straightforward prop refactor.

The auth check is moved to getServerSideProps using the same isActiveIsomerAdmin and getIronSession utilities already trusted in the server layer. The generateSessionOptions change is minimal and backward-compatible. No new logic paths or error cases are introduced.

</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/pages/sites/[siteId]/admin.tsx | Replaces the client-side Isomer admin check (useIsUserIsomerAdmin hook) with a server-side guard via getServerSideProps using iron-session and isActiveIsomerAdmin; eliminates the transient flash of admin config data for unauthorized users and prevents the tRPC admin queries from ever mounting. |
| apps/studio/src/server/modules/auth/session.ts | Adds a default argument (= {}) to generateSessionOptions so it can be called without arguments from getServerSideProps; backward-compatible change with no functional impact on existing callers. |

</details>

```mermaid
sequenceDiagram
    participant Browser
    participant Next as Next.js Server (getServerSideProps)
    participant IronSession as iron-session
    participant DB as Database (isActiveIsomerAdmin)

    Browser->>Next: GET /sites/:siteId/admin
    Next->>Next: siteAdminSchema.safeParse(params)
    alt Invalid params
        Next-->>Browser: 404 Not Found
    else Valid params
        Next->>IronSession: getIronSession(req, res, options)
        IronSession-->>Next: session (userId or empty)
        alt No userId in session
            Next-->>Browser: 302 → /sites/:siteId
        else userId present
            Next->>DB: isActiveIsomerAdmin(userId, [Core, Migrator])
            DB-->>Next: boolean
            alt Not admin
                Next-->>Browser: 302 → /sites/:siteId
            else Is admin
                Next-->>Browser: "200 + { siteId } props"
                Browser->>Browser: Render SiteAdminPage with siteId prop
                Browser->>Browser: Mount tRPC queries (config, theme, navbar, footer)
            end
        end
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant Browser
    participant Next as Next.js Server (getServerSideProps)
    participant IronSession as iron-session
    participant DB as Database (isActiveIsomerAdmin)

    Browser->>Next: GET /sites/:siteId/admin
    Next->>Next: siteAdminSchema.safeParse(params)
    alt Invalid params
        Next-->>Browser: 404 Not Found
    else Valid params
        Next->>IronSession: getIronSession(req, res, options)
        IronSession-->>Next: session (userId or empty)
        alt No userId in session
            Next-->>Browser: 302 → /sites/:siteId
        else userId present
            Next->>DB: isActiveIsomerAdmin(userId, [Core, Migrator])
            DB-->>Next: boolean
            alt Not admin
                Next-->>Browser: 302 → /sites/:siteId
            else Is admin
                Next-->>Browser: "200 + { siteId } props"
                Browser->>Browser: Render SiteAdminPage with siteId prop
                Browser->>Browser: Mount tRPC queries (config, theme, navbar, footer)
            end
        end
    end
```

</a>
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(studio): simplify admin page server ..."](https://github.com/opengovsg/isomer/commit/d0611ff9b773966ecba38828915ee7481f95c574) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40361512)</sub>


apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -205,6 +195,36 @@ const SiteAdminPage: NextPageWithLayout = () => {
...
+      ...BRIEF_TOAST_SETTINGS,
+    })
+    void router.push(`/sites/${siteId}`)
+  }, [isLoading, isUserIsomerAdmin, router, siteId, toast])
+
+  if (isLoading || !isUserIsomerAdmin) {
+    return null
+  }
+
+  return <SiteAdminForm siteId={siteId} />
```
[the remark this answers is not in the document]
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: This is the current behaviour used throughout the application and should not be updated in this pr solving #2377

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -205,6 +195,36 @@ const SiteAdminPage: NextPageWithLayout = () => {
...
+      title: "You do not have permission to access this page.",
+      status: "error",
+      ...BRIEF_TOAST_SETTINGS,
+    })
+    void router.push(`/sites/${siteId}`)
+  }, [isLoading, isUserIsomerAdmin, router, siteId, toast])
+
+  if (isLoading || !isUserIsomerAdmin) {
+    return null
+  }
```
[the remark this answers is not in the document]
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: For TanStack Query, a background refetch with cached data should generally be represented by `isFetching`, not `isLoading`, so I don’t think the stale-while-revalidate blank-screen case applies here.

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -205,6 +195,36 @@ const SiteAdminPage: NextPageWithLayout = () => {
...
+    }
+
+    hasRedirectedRef.current = true
+    toast({
+      title: "You do not have permission to access this page.",
+      status: "error",
+      ...BRIEF_TOAST_SETTINGS,
+    })
+    void router.push(`/sites/${siteId}`)
+  }, [isLoading, isUserIsomerAdmin, router, siteId, toast])
```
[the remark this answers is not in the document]
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: fixed in latest commit

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -205,6 +197,39 @@ const SiteAdminPage: NextPageWithLayout = () => {
   )
 }
 
+const SiteAdminPage: NextPageWithLayout = () => {
+  const toast = useToast()
+  const router = useRouter()
+  const hasRedirectedRef = useRef(false)
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx]: question: why do we need a ref?

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -205,6 +197,39 @@ const SiteAdminPage: NextPageWithLayout = () => {
...
+    roles: [IsomerAdminRole.Core, IsomerAdminRole.Migrator],
+  })
+
+  useEffect(() => {
+    if (isLoading || isUserIsomerAdmin || hasRedirectedRef.current) {
+      return
+    }
+
+    hasRedirectedRef.current = true
+    if (!toast.isActive(UNAUTHORIZED_TOAST_ID)) {
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx]: question: why do we need this?

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -205,6 +197,39 @@ const SiteAdminPage: NextPageWithLayout = () => {
...
+        status: "error",
+        ...BRIEF_TOAST_SETTINGS,
+      })
+    }
+    void router.push(`/sites/${siteId}`)
+  }, [isLoading, isUserIsomerAdmin, router, siteId, toast])
+
+  if (isLoading || !isUserIsomerAdmin) {
+    return null
+  }
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx]: im thinking we can actually consider either of these approaches instead. WDYT? otherwise the repeated check for `isLoading || isUserIsomerAdmin` in both useeffect and the rendering feels sus

1. Extract a hook — move the useEffect + useIsUserIsomerAdmin into a useRequireIsomerAdmin(siteId) hook

OR

2. Server-side redirect via getServerSideProps — check admin status on the server and redirect before the page renders. Eliminates useEffect, isLoading, and the null guard entirely. The page component would just render <SiteAdminForm> unconditionally. This is the most meaningful simplification.

Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: Went with the `getServerSideProps` as it's simpler and suits this use case well. However, this duplicates the auth checks that previously happened via the `protectedProcedure` path in `user.isIsomerAdmin` and does not show the error toast.

Does this feel okay, or would you prefer the extracted hook approach? I think that would better match existing project conventions, but server-side redirect suits this use case well.

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: good catch, I don’t think we need this anymore. It was added to prevent the effects(toast, redirect) from running twice as we redirect from the page. However, the duplicate toast issue was solved by using toastIds.

(I went with the server side redirection so this change is not relevant anymore)

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: this was in response to a copilot comment above and is to prevent duplicate toasts in 1. react strict mode and 2. during dependency changes in the useEffect before the redirect completes.

(I went with the server side redirection so this change is not relevant anymore)

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -41,24 +46,62 @@ const SUPPORTED_SITE_CONFIG_TYPES = [
...
+
+  if (!parsedParams.success) {
+    return { notFound: true }
   }
 
+  const { siteId } = parsedParams.data
+  const session = await getIronSession<SessionData>(
+    req,
+    res,
+    generateSessionOptions({ ttlInHours: 1 }),
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx]: i thinik we dont need to pass in the `ttlInHours` - `generateSessionOptions()` should be sufficient

we can update `generateSessionOptions` itself to default to empty {} if no props is passed in to fix the type error

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -41,24 +46,62 @@ const SUPPORTED_SITE_CONFIG_TYPES = [
...
+      .where("deletedAt", "is", null)
+      .executeTakeFirst())
+
+  const isUserIsomerAdmin =
+    !!user &&
+    (await isEmailWhitelisted(user.email)) &&
+    (await isActiveIsomerAdmin(user.id, [
+      IsomerAdminRole.Core,
+      IsomerAdminRole.Migrator,
+    ]))
```
Comment (adriangohjw) [inline apps/studio/src/pages/sites/[siteId]/admin.tsx]: we can simplify and skip the check for `user` too - since if their iron session cookie is still valid, we can assume they still have permissions to access the platform. same for `isEmailWhitelisted` too

so a minimal version will be
```ts
const isUserIsomerAdmin =
    !!session.userId
    (await isActiveIsomerAdmin(session.userId as string, [
      IsomerAdminRole.Core,
      IsomerAdminRole.Migrator,
    ]))
```

adriangohjw (commented):
good attempt! i think we can tighten things up so its less verbose

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (adriangohjw) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: answered you in the review comments :D

apps/studio/src/pages/sites/[siteId]/admin.tsx
```diff
@@ -41,24 +46,62 @@ const SUPPORTED_SITE_CONFIG_TYPES = [
...
+
+  if (!isUserIsomerAdmin) {
+    return {
+      redirect: {
+        destination: `/sites/${siteId}`,
+        permanent: false,
+      },
+    }
+  }
```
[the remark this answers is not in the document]
Comment (adriangohjw) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx:80]: thanks bro, we accept this

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: addressed this in the latest commit!

apps/studio/src/pages/sites/[siteId]/admin.tsx
Comment (Benjam11n) [reply apps/studio/src/pages/sites/[siteId]/admin.tsx]: addressed this in the latest commit!

adriangohjw (approved):

adriangohjw:
@Benjam11n i will be merging this! can i have your help to do similar changes to routes in the `godmode` routes? they are also reserved by isomer admins. very similar to this PR. thank you!

Benjam11n:
sure!

PR #2603, opened 2026-06-30
feat(db): add AuditLogExportRequest model and migration

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 3 | +183 | -15 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 3 | +49 | -0 |
| 🚫 Ignore | 3 | +32 | -0 |

## Problem

The async Audit Log Export feature lets a Site Admin request a downloadable report of a site's audit/access data for a selected month and receive it by email. That workflow needs durable, queryable per-request state — who requested what, the lifecycle status, retry count, and the resulting object key — but no such table exists.

No linked issue (stacked feature work). This is the base of a 6-PR stack:

1. **#2603 (this PR)** — schema: `AuditLogExportRequest` model + migration
2. #2604 — Zod input schema
3. #2605 — report queries + CSV
4. #2606 — S3 + email delivery plumbing
5. #2607 — request endpoint + cron orchestrator
6. #2608 — Settings UI

## Solution

Add an **additive, backward-compatible** Prisma model and two enums for the export workflow, plus the architectural decision records and glossary entries that explain the non-obvious choices.

**Breaking Changes**

- [ ] Yes
- [x] No — this PR is backwards compatible. The migration only adds new types, a new table, indexes, and foreign keys; no existing object is altered or dropped.

**Features**:

- `AuditLogExportRequest` table — `id`, `siteId`, `userId`, `auditLogDateRange` (native Postgres `daterange`: SGT calendar dates, half-open, canonical `[YYYY-MM-DD,YYYY-MM-DD)` — e.g. `[2026-04-01,2026-05-01)` for April; current-month requests are clamped to SGT-today + 1 at insert time), `reportType`, `status` (default `Pending`), `attempts`, `errorMessage`, `objectKey` (nullable — each job produces exactly one CSV), timestamps. A named CHECK constraint rejects empty or unbounded ranges. Indexed on `status` (cron sweep), plus a partial unique index on `(siteId, userId, auditLogDateRange, reportType) WHERE status IN ('Pending','Processing')` (in-flight dedupe by range **equality** — Postgres canonicalizes `daterange`, so equality is representation-proof).
- `AuditLogExportReportType` enum — `Access` | `Activity`. `Both` is deliberately **not** a DB value: it is input vocabulary only, which the service (#2607) fans out into two rows, one per report.
- `AuditLogExportStatus` enum — `Pending` | `Processing` | `Done` | `Failed`.

**Improvements**:

- `docs/adr/0003-point-in-time-access-report.md` — records why the Access report reconstructs *historical* access (as-of month end) instead of reading current permissions.
- `docs/adr/0004-emailed-signed-url-delivery-for-audit-exports.md` — records the emailed long-lived signed-URL delivery trade-off and its mitigations.
- `CONTEXT.md` — glossary entries: **Site Admin**, **Audit Log**, **Audit Log Export**, **Access report**, **Activity report**.

**Bug Fixes**:

- None.

## Tests

Schema-only PR; no runtime behaviour yet. Verified that:

- `pnpm --filter @isomer/db migrate` applies `20260630031102_add_audit_log_export_request` cleanly (additive SQL only).
- `pnpm generate` refreshes the Prisma + Kysely types with no errors.
- `pnpm typecheck` is green.

**New dependencies**: none.

Adds the `AuditLogExportRequest` Prisma model and its migration as the schema foundation for the async Audit Log Export feature. The change is purely additive — two new enums, one new table with appropriate indexes, FK constraints, and a bounded-range CHECK constraint.

- **Schema**: `AuditLogExportRequest` table with a native Postgres `daterange` column (`auditLogDateRange`), a `status` sweep index, and a hand-written partial UNIQUE index `WHERE status IN ('Pending','Processing')` for in-flight deduplication; the plain `@@index` / partial-UNIQUE drift is explicitly acknowledged in both the schema and migration comments.
- **Code generation**: `patchUnsupportedColumns` in `generate.cts` re-injects `Unsupported(\"daterange\")` fields (annotated with `/// @kyselyType(...)`) into the prisma-kysely output, correctly producing `auditLogDateRange: string` in the Kysely types.
- **Docs**: Two new ADRs (point-in-time access reconstruction, emailed signed-URL delivery trade-off) and glossary entries in `CONTEXT.md`.
</details>

Safe to merge — purely additive schema change with no modifications to existing tables, indexes, or types.

Every change is additive: new enums, a new table, new indexes, and new generated type files. Existing tables are untouched. The migration is backward-compatible and the Kysely type injection is correctly guarded for idempotency. The only non-trivial concern is that `generate().catch` logs but does not exit non-zero on failure, which would hide a broken generation step in CI — but this does not affect the correctness of the committed generated files.

**Files Needing Attention:** packages/db/prisma/generate.cts — the new `patchUnsupportedColumns` code path throws on mismatch but the top-level error handler exits 0.
</details>

| Filename | Overview |
|----------|----------|
| packages/db/prisma/generate.cts | Adds `patchUnsupportedColumns` to inject `Unsupported("...")`-typed fields (annotated with `/// @kyselyType(...)`) into the prisma-kysely output; logic is sound but the top-level `.catch` silently swallows errors (exits 0), which can hide injection failures. |
| packages/db/prisma/migrations/20260630031102_add_audit_log_export_request/migration.sql | Clean additive migration: two enums, one table, a bounded-range CHECK constraint, a status sweep index, and a partial UNIQUE index for in-flight deduplication; index names are within Postgres's 63-char limit and both FKs carry ON DELETE RESTRICT. |
| packages/db/prisma/schema.prisma | Adds `AuditLogExportRequest` model with `daterange` column typed as `Unsupported` + `/// @kyselyType(string)` annotation; both FK relations carry explicit `onDelete: Restrict`; the plain `@@index` vs. migration's partial UNIQUE index drift is clearly documented inline. |
| packages/db/src/generated/generatedTypes.ts | Correctly adds the `AuditLogExportRequest` interface and registers it in the `DB` map; `auditLogDateRange` appears first (injected by `patchUnsupportedColumns`) while the DB column is 4th — cosmetic inconsistency, not a runtime issue. |
| packages/db/src/generated/generatedEnums.ts | Generates `AuditLogExportReportType` and `AuditLogExportStatus` const-enums matching the Prisma schema values exactly. |
| packages/db/src/generated/selectableTypes.ts | Adds `AuditLogExportRequest` selectable type, consistent with the rest of the file. |
| docs/adr/0003-point-in-time-access-report.md | New ADR documenting the point-in-time access reconstruction approach; clearly explains why the query uses `createdAt`/`deletedAt` instead of reading current permissions. |
| docs/adr/0004-emailed-signed-url-delivery-for-audit-exports.md | New ADR documenting the emailed signed-URL delivery trade-off and its mitigations; the security implications are candidly acknowledged. |
| CONTEXT.md | Adds glossary entries for Site Admin, Audit Log, Audit Log Export, Access report, and Activity report; consistent with the new schema concepts. |

</details>

<details open><summary><h3>Entity Relationship Diagram</h3></summary>

```mermaid
erDiagram
    Site {
        int id PK
        string name
    }
    User {
        string id PK
        string email
    }
    AuditLogExportRequest {
        bigint id PK
        int siteId FK
        string userId FK
        daterange auditLogDateRange
        AuditLogExportReportType reportType
        AuditLogExportStatus status
        int attempts
        string errorMessage
        string objectKey
        timestamp completedAt
        timestamp createdAt
        timestamp updatedAt
    }
    AuditLogExportReportType {
        string Access
        string Activity
    }
    AuditLogExportStatus {
        string Pending
        string Processing
        string Done
        string Failed
    }
    Site ||--o{ AuditLogExportRequest : "siteId (RESTRICT)"
    User ||--o{ AuditLogExportRequest : "userId (RESTRICT)"
    AuditLogExportRequest }o--|| AuditLogExportReportType : "reportType"
    AuditLogExportRequest }o--|| AuditLogExportStatus : "status"
```
</details>

<sub>Reviews (18): Last reviewed commit: ["feat(db): add AuditLogExportRequest mode..."](https://github.com/opengovsg/isomer/commit/f5b3593765cba011cfe260c9c7ee50fe066a1183) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40759279)</sub>

### DIFF
--- a/docs/adr/0004-emailed-signed-url-delivery-for-audit-exports.md
+++ b/docs/adr/0004-emailed-signed-url-delivery-for-audit-exports.md
@@ -1,0 +1,16 @@
+---
+status: accepted
+---
+
+# Audit Log Exports are delivered as long-lived signed S3 URLs in email
+
+A completed Audit Log Export is delivered by emailing the requesting Site Admin a **signed S3 GET URL valid for 3 days**, after which the download link expires (the presigned URL is no longer usable). The link is a standalone bearer credential: anyone who receives the email can download the CSV — which contains sensitive data (user emails, roles, IP addresses, login/logout history) — with no re-authentication. We accepted this for simplicity over building an authenticated in-Studio download flow.
+
+## Considered options
+
+- **Link to an authenticated Studio route that mints a short-lived signed URL on click** — recipient must be a logged-in Admin of the site; access is re-checked at download time, so no bearer credential leaves Studio. Rejected for now in favour of a simpler email-only flow, despite being the stronger choice for sensitive data.
+- **Studio "Exports" page with history + auth-gated re-download** — most robust, but more UI than this iteration warrants.
+
+## Consequences
+
+This is a deliberate convenience-over-security trade-off and a reviewer will (rightly) question it. Mitigations: the presigned download URL expires after 3 days, bounding how long the emailed bearer credential works; exports are Site-Admin-only; and the default 5-minute signed-URL expiry in `src/lib/s3.ts` must be raised specifically for this path. Note that URL expiry does not delete the underlying S3 object — a separate object-lifecycle/deletion policy would be needed to bound data-at-rest exposure, which is not yet implemented. If the threat model tightens, switch to the authenticated-route option above without changing the rest of the workflow.
--- a/packages/db/prisma/generate.cts
+++ b/packages/db/prisma/generate.cts
@@ -6,6 +6,63 @@ const TYPE_FOLDER = "../src/generated"
 // THIS MUST BE PREFIXED WITH A DOT
 const TYPE_FILE = "./generatedTypes.ts"
 const GENERATED_FILE = "./selectableTypes.ts"
+const SCHEMA_FILE = "./schema.prisma"
+
+// prisma-kysely drops `Unsupported(...)` columns entirely (it skips fields of
+// kind "unsupported" before applying `@kyselyType` doc-comment overrides), so
+// we re-inject those fields into the generated Kysely types here, using each
+// field's `/// @kyselyType(X)` doc comment in schema.prisma as the emitted
+// type. Fields of type `Unsupported(...)` without a `/// @kyselyType(...)`
+// doc comment stay omitted, matching prisma-kysely's default behaviour.
+const patchUnsupportedColumns = async () => {
+  const schema = await fs.readFile(path.join(__dirname, SCHEMA_FILE), "utf8")
+  const typesPath = path.join(__dirname, TYPE_FOLDER, TYPE_FILE)
+  let types = await fs.readFile(typesPath, "utf8")
+
+  const modelPattern = /model\s+(\w+)\s+\{([\s\S]*?)\n\}/g
+  const fieldPattern =
+    /\/\/\/\s*@kyselyType\(([^)]+)\)\s*\n\s*(\w+)\s+Unsupported\("[^"]+"\)(\?)?/g
+
+  for (const [, modelName, modelBody] of schema.matchAll(modelPattern)) {
+    if (!modelName || !modelBody) {
+      continue
+    }
+    for (const [, kyselyType, fieldName, nullable] of modelBody.matchAll(
+      fieldPattern,
+    )) {
+      if (!kyselyType || !fieldName) {
+        continue
+      }
+      const fieldType = `${kyselyType}${nullable ? " | null" : ""}`
+      // prisma-kysely emits `export type X = {...};` (later rewritten to
+      // `export interface X {...}` by oxlint's --fix), so handle both forms.
+      const headers = [
+        {
+          header: `export type ${modelName} = {\n`,
+          property: `    ${fieldName}: ${fieldType};\n`,
+        },
+        {
+          header: `export interface ${modelName} {\n`,
+          property: `  ${fieldName}: ${fieldType}\n`,
+        },
+      ]
+      const match = headers.find(({ header }) => types.includes(header))
+      if (!match) {
+        throw new Error(
+          `Could not find generated type for model ${modelName} while injecting Unsupported column ${fieldName}`,
+        )
+      }
+      // Idempotency guard: `prisma generate` rewrites the file before this
+      // script runs, but skip re-injection if the script is run standalone.
+      if (new RegExp(`\\b${fieldName}\\s*[?:]`).test(types)) {
+        continue
+      }
+      types = types.replace(match.header, match.header + match.property)
+    }
+  }
+
+  await fs.writeFile(typesPath, types)
+}
 
 const KYSELY_IMPORT = ts.factory.createImportDeclaration(
   undefined,
@@ -87,6 +144,8 @@ const extractTableTypes = (source: ts.SourceFile) => {
 }
 
 export const generate = async () => {
+  await patchUnsupportedColumns()
+
   const file = await fs.readFile(path.join(__dirname, TYPE_FOLDER, TYPE_FILE))
   const source = ts.createSourceFile(
     TYPE_FILE,
--- a/packages/db/prisma/migrations/20260630031102_add_audit_log_export_request/migration.sql
+++ b/packages/db/prisma/migrations/20260630031102_add_audit_log_export_request/migration.sql
@@ -1,0 +1,62 @@
+-- CreateEnum
+CREATE TYPE "AuditLogExportReportType" AS ENUM ('Access', 'Activity');
+
+-- CreateEnum
+CREATE TYPE "AuditLogExportStatus" AS ENUM ('Pending', 'Processing', 'Done', 'Failed');
+
+-- CreateTable
+-- "auditLogDateRange" bounds are SGT calendar dates, half-open [) (Postgres
+-- canonicalizes daterange to this form); a full month is e.g.
+-- [2026-04-01,2026-05-01). Current-month requests are clamped to today+1 at
+-- insert time.
+CREATE TABLE "AuditLogExportRequest" (
+    "id" BIGSERIAL NOT NULL,
+    "siteId" INTEGER NOT NULL,
+    "userId" TEXT NOT NULL,
+    "auditLogDateRange" daterange NOT NULL,
+    "reportType" "AuditLogExportReportType" NOT NULL,
+    "status" "AuditLogExportStatus" NOT NULL DEFAULT 'Pending',
+    "attempts" INTEGER NOT NULL DEFAULT 0,
+    "errorMessage" TEXT,
+    "objectKey" TEXT,
+    -- set when the request reaches Done; a completed row whose "completedAt" >=
+    -- the end of its "auditLogDateRange" holds a Complete Artifact eligible for reuse
+    "completedAt" TIMESTAMP(3),
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "AuditLogExportRequest_pkey" PRIMARY KEY ("id")
+);
+
+-- AddCheckConstraint
+-- The export range must be non-empty and bounded on both sides: an unbounded
+-- or empty range is never a valid audit log export request.
+ALTER TABLE "AuditLogExportRequest"
+    ADD CONSTRAINT "AuditLogExportRequest_auditLogDateRange_bounded_check"
+    CHECK (
+        NOT isempty("auditLogDateRange")
+        AND lower("auditLogDateRange") IS NOT NULL
+        AND upper("auditLogDateRange") IS NOT NULL
+    );
+
+-- CreateIndex
+CREATE INDEX "AuditLogExportRequest_status_idx" ON "AuditLogExportRequest"("status");
+
+-- CreateIndex
+-- Partial UNIQUE index: prevents duplicate in-flight export requests for the same
+-- (site, user, range, reportType) while one is still Pending/Processing, closing the
+-- SELECT-then-INSERT race in the service layer. Once a request reaches Done/Failed the
+-- row drops out of the index, so the same range can be re-requested later.
+-- Dedupe is by range EQUALITY (not overlap); Postgres canonicalizes daterange to
+-- half-open [) form, so equality is representation-proof (e.g. [a,b] and [a,b+1)
+-- store identically).
+-- Prisma cannot express a partial unique index, so this is hand-written SQL and the
+-- Prisma schema keeps a plain @@index (expect migrate-dev drift on this index).
+-- (Index name kept under Postgres's 63-char identifier limit.)
+CREATE UNIQUE INDEX "AuditLogExportRequest_siteId_userId_dateRange_reportType_idx" ON "AuditLogExportRequest"("siteId", "userId", "auditLogDateRange", "reportType") WHERE "status" IN ('Pending', 'Processing');
+
+-- AddForeignKey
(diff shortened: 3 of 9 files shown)


seaerchin:
* **#2607** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2605** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2605?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2603** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `feat/audit-logs`


<h2></h2>

packages/db/prisma/schema.prisma
```diff
@@ -366,3 +368,35 @@ model PushDocumentJob {
...
+  Access // who had access (point-in-time)
+  Activity // what happened (events)
+  Both
+}
+
+enum AuditLogExportStatus {
+  Pending
+  Processing
+  Done
+  Failed
```
seaerchin, line 383: comment: will check for clean-up

packages/db/prisma/schema.prisma
```diff
@@ -366,3 +368,35 @@ model PushDocumentJob {
...
+  siteId       Int
+  user         User                     @relation(fields: [userId], references: [id])
+  userId       String
+  month        String // selected month as "yyyy-MM" (Asia/Singapore)
+  reportType   AuditLogExportReportType
+  status       AuditLogExportStatus     @default(Pending)
+  attempts     Int                      @default(0)
+  errorMessage String?
+  objectKeys   String[]                 @default([]) // S3 keys of generated CSV(s)
+  createdAt    DateTime                 @default(now())
```
seaerchin: comment: note that because users can request for the month in flight, if `createdAt` has the same month/year as `month`, this means that we are **not able** to generate the full audit log for that requested month

docs/adr/0004-emailed-signed-url-delivery-for-audit-exports.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Considered options
+
+- **Link to an authenticated Studio route that mints a short-lived signed URL on click** — recipient must be a logged-in Admin of the site; access is re-checked at download time, so no bearer credential leaves Studio. Rejected for now in favour of a simpler email-only flow, despite being the stronger choice for sensitive data.
+- **Studio "Exports" page with history + auth-gated re-download** — most robust, but more UI than this iteration warrants.
+
+## Consequences
+
+This is a deliberate convenience-over-security trade-off and a reviewer will (rightly) question it. Mitigations: the URL expires after 3 days and the object is then deleted, bounding the exposure window; exports are Site-Admin-only; and the default 5-minute signed-URL expiry in `src/lib/s3.ts` must be raised specifically for this path. If the threat model tightens, switch to the authenticated-route option above without changing the rest of the workflow.
```
adriangohjw: `object is then deleted` -> oh i tot its just the URL expires but the object still there?

packages/db/prisma/schema.prisma
```diff
@@ -366,3 +368,35 @@ model PushDocumentJob {
...
+  Failed
+}
+
+model AuditLogExportRequest {
+  id           BigInt                   @id @default(autoincrement())
+  site         Site                     @relation(fields: [siteId], references: [id], onDelete: Restrict)
+  siteId       Int
+  user         User                     @relation(fields: [userId], references: [id])
+  userId       String
+  month        String // selected month as "yyyy-MM" (Asia/Singapore)
```
adriangohjw: issue(?): should be `yearmonth` instead of `month`? `month` might be a bit misleading since its actually yyyy-MM

packages/db/prisma/schema.prisma
```diff
@@ -366,3 +368,35 @@ model PushDocumentJob {
...
+  reportType   AuditLogExportReportType
+  status       AuditLogExportStatus     @default(Pending)
+  attempts     Int                      @default(0)
+  errorMessage String?
+  objectKeys   String[]                 @default([]) // S3 keys of generated CSV(s)
```
adriangohjw: note to self: im not sure how these will be used for now but it seems a bit sus. But okay to approve since this is pre-launch feature and we can always make another contracting migrations

adriangohjw (approved):
some potential issue - but i don;t think is blocker-worthy

packages/db/prisma/schema.prisma
```diff
@@ -366,3 +368,35 @@ model PushDocumentJob {
   @@index([scheduledAt])
   @@index([resourceId])
 }
+
+enum AuditLogExportReportType {
+  Access // who had access (point-in-time)
+  Activity // what happened (events)
+  Both
```
dcshzj: question: this seems like we are locking ourselves into just these two audit log export types. Why not make two audit log export requests instead, since anyway there are two files that the user will need to download

packages/db/prisma/schema.prisma
dcshzj: issue: I think we shouldn't lock ourselves to a specific month? We can consider like a time range field instead so we have the flexibility for month, quarter, half yearly and yearly reports

docs/adr/0004-emailed-signed-url-delivery-for-audit-exports.md
seaerchin: the plan mode not good... the presigned url expires only

packages/db/prisma/schema.prisma
seaerchin: agreed - i think we should just use `DateTime` here; will tell claude to refactor >:(

packages/db/prisma/schema.prisma
seaerchin: the issue is that each job is async but we need to send a single email out, so we need a type denominating "all the items".

i can check w/ design if this is required or if we can relax the constraint

packages/db/prisma/schema.prisma
seaerchin: there are 2 ways we could do this:

1. naive month 
2. start + interval 

i went w/ naive month because i'm not sure if the added complexity of implementing a start + interval will pay off. this is more of a product question that i'll discuss tomorrow ba

packages/db/prisma/schema.prisma
adriangohjw: question: why not remove Both  and if users want both, we just trigger 2 separate job?

seaerchin:
@greptileai

packages/db/prisma/schema.prisma
```diff
@@ -366,3 +368,47 @@ model PushDocumentJob {
...
+  updatedAt         DateTime                 @default(now()) @updatedAt
+
+  @@index([status])
+  // NOTE: the migration promotes this to a PARTIAL UNIQUE index
+  // (`WHERE status IN ('Pending', 'Processing')`) to prevent duplicate
+  // in-flight requests for the same range. Prisma's schema DSL cannot express
+  // a partial unique index, so it is defined as raw SQL in the migration and
+  // this stays a plain `@@index`. As a result `prisma migrate dev` against an
+  // existing DB will report drift here; that divergence is expected.
+  @@index([siteId, userId, auditLogDateRange, reportType])
```
adriangohjw, line 414: question: this will mean that if the attempt Failed, there's no way to raise another request? im assuming down the line we will have mechanism to retry it (and update it back to Pending)?

packages/db/prisma/generate.cts
```diff
@@ -6,6 +6,63 @@ const TYPE_FOLDER = "../src/generated"
...
 const GENERATED_FILE = "./selectableTypes.ts"
+const SCHEMA_FILE = "./schema.prisma"
+
+// prisma-kysely drops `Unsupported(...)` columns entirely (it skips fields of
+// kind "unsupported" before applying `@kyselyType` doc-comment overrides), so
+// we re-inject those fields into the generated Kysely types here, using each
+// field's `/// @kyselyType(X)` doc comment in schema.prisma as the emitted
+// type. Fields of type `Unsupported(...)` without a `/// @kyselyType(...)`
+// doc comment stay omitted, matching prisma-kysely's default behaviour.
+const patchUnsupportedColumns = async () => {
```
adriangohjw, line 17: i assume this is clauded - im okay to accept this part since its script we run locally

adriangohjw (commented):
largely okay, just one question on the index we have

packages/db/prisma/generate.cts
seaerchin: yeap this is clauded

packages/db/prisma/schema.prisma
seaerchin: we can ah - the actual index (from the generated sql) is as follows: 
```sql
CREATE UNIQUE INDEX "AuditLogExportRequest_siteId_userId_dateRange_reportType_idx" ON "AuditLogExportRequest"("siteId", "userId", "auditLogDateRange", "reportType") WHERE "status" IN ('Pending', 'Processing');
```

the unique index is only where `status in PENDING, PROCESSING`, so if the report type is `FAILED`, we can retry (status below)

```ts
enum AuditLogExportStatus {
  Pending
  Processing
  Done
  Failed
}
```

packages/db/prisma/schema.prisma
seaerchin: Still open — needs a decision on cleanup of export artifacts/rows (S3 lifecycle policy vs a sweep); not blocking this stack.

packages/db/prisma/schema.prisma
seaerchin: Superseded by the daterange design: a current-month request is clamped to SGT-today+1 at insert time, so the export covers everything up to the request day, and such in-flight snapshots are excluded from Complete-Artifact reuse in #2832.

packages/db/prisma/schema.prisma
seaerchin: `objectKey` is the S3 key of the generated CSV — the ready email signs a download URL against it. `completedAt` marks when the row reached `Done`, and is what #2832 compares against the range end to decide whether the artifact can be reused.

packages/db/prisma/schema.prisma
seaerchin: Addressed in the stack: the DB enum has no `Both` — a "Both" ask fans out into two rows (one Access, one Activity) in `createAuditLogExportRequest`, each fulfilled independently with its own email. `Both` is UX vocabulary only (see `schemas/audit.ts`).

packages/db/prisma/schema.prisma
seaerchin: Addressed: the column is now `auditLogDateRange` (Postgres `daterange`, half-open SGT calendar dates), so quarters/half-years are representable without a migration — the month picker is just the current UI input.

packages/db/prisma/migrations/20260630031102_add_audit_log_export_request/migration.sql
```diff
@@ -0,0 +1,59 @@
...
+    "reportType" "AuditLogExportReportType" NOT NULL,
+    "status" "AuditLogExportStatus" NOT NULL DEFAULT 'Pending',
+    "attempts" INTEGER NOT NULL DEFAULT 0,
+    "errorMessage" TEXT,
+    "objectKey" TEXT,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "AuditLogExportRequest_pkey" PRIMARY KEY ("id")
+);
```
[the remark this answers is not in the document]
seaerchin: Deferred: `status = Done` and `objectKey` are written in the same UPDATE at the single call site, so the invariant holds at the write. Can add the CHECK as follow-up hardening if we want it enforced at the DB level too.

PR #2604, opened 2026-06-30
feat(studio): add audit log export request input schema

seaerchin:
## Problem

The Audit Log Export endpoint needs validated, type-safe input — a site, a calendar month, and which report(s) to produce — with user-facing error messages, before any server logic runs.

No linked issue (stacked feature work). Part of the stack starting at #2603 (depends on the schema PR for the `AuditLogExportReportType` enum).

## Solution

Add a Zod schema for the create-export-request input under `~/schemas/`, following the area conventions (one file per domain, `<action><Entity>Schema` naming, user-facing messages on every constraint, inferred type exported alongside).

**Breaking Changes**

- [ ] Yes
- [x] No — additive, no existing schema changed.

**Features**:

- `createAuditLogExportRequestSchema` in `apps/studio/src/schemas/audit.ts`:
  - `siteId` — positive integer.
  - `month` — string matching `^\d{4}-(0[1-9]|1[0-2])$` (e.g. `2026-03`), **plus** two refinements enforcing the export window: not in the future, and not older than the 12-month window — both relative to the current Singapore month. Because the client form derives from this schema, the window is now validated client-side (zodResolver) and server-side (router input) as well as in the service guard.
  - `reportType` — `z.enum(AuditLogExportRequestedReportType)` (`Access` | `Activity` | `Both`). `AuditLogExportRequestedReportType` is **input vocabulary**, defined in this file by spreading the Prisma `AuditLogExportReportType` enum (`Access` | `Activity`) plus `Both` — so it still fails type-check if the DB enum changes. `Both` never reaches the DB: the service (#2607) fans a `Both` request out into two rows (one `Access`, one `Activity`).
- Exported inferred type `CreateAuditLogExportRequestInput`.
- `AUDIT_LOG_EXPORT_MAX_MONTHS` (= 12), `getEarliestExportableMonth(currentMonth)`, and `getCurrentSingaporeMonth()` — the shared export-window policy, consumed by the schema refinements, the server guard (#2607), and the client month picker (#2608) so they can't drift. Note: adding the refinements makes this schema time-dependent by design (its tests use dynamic months).

**Improvements**:

- Uses the zod 4 `z.enum(EnumObject)` form (not the deprecated `z.nativeEnum`).

**Bug Fixes**:

- None.

## Tests

[...]

PR #2605, opened 2026-06-30
feat(studio): add audit log export report queries and CSV

seaerchin:
## Problem

An audit log export must answer two questions for a site and a selected month: *who had access* (Access report) and *what happened* (Activity report) — and serialize each to CSV. A one-off script (`prisma/scripts/getAuditLogs.ts`) did this manually for a hardcoded site allowlist, with one important wrong-for-audit behaviour: its access query returned who has access **right now**, ignoring the month.

No linked issue (stacked feature work). Part of the stack starting at #2603 (depends on the schema PR).

## Solution

Port the script's logic into a reusable, parameterized query module, with one deliberate semantic change to the access report and correct Singapore-time month boundaries.

**Breaking Changes**

- [ ] Yes
- [x] No — new module; the original script is untouched.

**Features**:

- `getAccessReportRows({ siteId, auditLogDateRange })` — **point-in-time** reconstruction of who had access *as of the end of the export range* (`createdAt <= rangeEnd AND (deletedAt IS NULL OR deletedAt > rangeEnd)`), per ADR-0003. A current-month range is clamped to SGT-today + 1, so it naturally collapses to "now".
- `getActivityReportRows({ siteId, auditLogDateRange })` — range-scoped `AuditLog` events, a faithful port of the script's event/description logic (the hardcoded site allowlist is dropped; `siteId` is a parameter).
- `toCsv(rows)` — Papa Parse serialization matching the script's stringification.
- `getMonthDateRange(month, now)` — the canonical `daterange` string for a `yyyy-MM` month (full calendar month, e.g. `[2026-04-01,2026-05-01)`; the current month is clamped to SGT-today + 1 so today's partial day is included and the range is never empty). `getExportRange(auditLogDateRange)` — converts a stored range to half-open UTC instants (`rangeStart`/`rangeEnd`) for `Asia/Singapore` (via `date-fns-tz`).

**Improvements**:

- Both reports preserve the existing `@open.gov.sg` (Isomer team) exclusion.

**Bug Fixes**:

- Access report is now time-correct for audits (was effectively "current access" in the script).

## Tests

- `apps/studio/src/server/modules/audit/__tests__/auditLogExport.query.test.ts` — 12 DB-backed cases: all four point-in-time boundary cases for access; `@open.gov.sg` exclusion; an SGT-vs-UTC boundary event (`2024-03-31T23:30:00Z` → April SGT, excluded from March); activity in/out of month with non-empty descriptions; `toCsv` header + row count.
- `pnpm test:unit -- src/server/modules/audit/__tests__/auditLogExport.query.test.ts` → 12/12 passing. Typecheck green.

**New dependencies**: none (`date-fns-tz` and `papaparse` already present).

This PR moves audit log export logic into a reusable server module. The main changes are:

- Added shared access and activity report queries for audit log exports.
- Added Singapore-time month range handling for stored audit export ranges.
- Refactored the one-off Prisma script to call the shared query and CSV helpers.
- Added DB-backed tests for range boundaries, access reconstruction, activity filtering, descriptions, and CSV output.
</details>

Safe to merge based on the reviewed changed code and tests.

The core paths are covered for timezone boundaries, access lifecycle boundaries, activity filtering, CSV formatting, and previously reported edge cases.

**Files Needing Attention:** No files require special attention beyond normal human review for the high-risk audit module path.
</details>

**What T-Rex did**
- Validated the base access semantics for non-admin users, confirming that only rows with deletedAt IS NULL are included and that the row created at rangeEnd is present while the row deleted at rangeEnd is not.
- Validated head range behavior and the current-month clamp by checking the head predicate results, showing active@agency.gov.sg and deleted-at-range-end@agency.gov.sg included, while deleted-before-end@agency.gov.sg, created-at-range-end@agency.gov.sg, and internal@open.gov.sg are excluded.
- Captured the before-run and after-run sequences for the head-range tests, including the shared testcontainers/container-runtime blocker; captured the head fixture source alongside notes about the base fixture origin and the untracked file being removed in head checkout.
- Validated serialization and range helpers: dates serialize to ISO strings, nulls become empty strings, and objects use JSON.stringify; getMonthDateRange and getExportRange produce the expected ranges and UTC instants for export.
- Compiled and organized artifacts for review, linking each artifact to its corresponding proof to support verification of base, head, and export-related tests.

<a href="https://app.greptile.com/trex/runs/17201519/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/audit/auditLogExport.query.ts | Adds reusable access and activity export queries, Singapore-time date range conversion, CSV serialization, and fixes previously identified audit boundary and description issues. |
| apps/studio/src/server/modules/audit/__tests__/auditLogExport.query.test.ts | Adds DB-backed coverage for SGT month ranges, point-in-time access reconstruction, activity filtering, descriptions, and CSV serialization. |
| apps/studio/prisma/scripts/getAuditLogs.ts | Refactors the one-off audit log script into a wrapper over the shared export query module while preserving per-site CSV output. |
| apps/studio/src/server/modules/audit/audit.types.ts | Introduces shared audit export query parameter and access report row types matching Kysely's quoted alias behavior. |
| apps/studio/prisma/scripts/__tests__/getAuditLogs.test.ts | Removes the old script-specific audit log tests after moving query coverage to the server module tests. |

</details>

```mermaid
sequenceDiagram
participant Script as prisma/scripts/getAuditLogs.ts
participant Export as auditLogExport.query.ts
participant DB as PostgreSQL
participant CSV as Papa Parse CSV

Script->>Export: getMonthDateRange(monthYear, now)
Export-->>Script: canonical daterange [SGT start, SGT end)
Script->>Export: getAccessReportRows(siteId, auditLogDateRange)
Export->>DB: Query ResourcePermission/User at rangeEnd
DB-->>Export: AccessReportRow[]
Script->>Export: getActivityReportRows(siteId, auditLogDateRange)
Export->>DB: Query AuditLog within [rangeStart, rangeEnd)
DB-->>Export: ActivityReportRow[]
Script->>Export: toCsv(rows)
Export->>CSV: stringify values and headers
CSV-->>Script: CSV text
Script->>Script: "write useraccess_*.csv and auditlogs_*.csv"
```
</details>

<sub>Reviews (25): Last reviewed commit: ["refactor(studio): simplify audit export ..."](https://github.com/opengovsg/isomer/commit/f385ae7f7dcf0f83425b76d2ab57aed7cf7e6305) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41055317)</sub>


seaerchin:
* **#2607** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2605** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2605?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2603** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/audit-logs`


<h2></h2>

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,375 @@
...
+    year === undefined ||
+    monthIndex === undefined ||
+    Number.isNaN(year) ||
+    Number.isNaN(monthIndex) ||
+    monthIndex < 1 ||
+    monthIndex > 12
+  ) {
+    // oxlint-disable-next-line @typescript-eslint/restrict-template-expressions
+    throw new Error(`Invalid month, expected "yyyy-MM" but got: ${month}`)
+  }
```
seaerchin: comment: ~~no need to account for FE/BE time boundary~~; for month in flight, we will just use the current time cos the system time on backend is UTC and the DB time is also UTC.

we will take **all time from frontend** as sgt and convert it into the appropriate UTC time range using date-fns.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,370 @@
...
+  // Treat the month as Singapore time: express an instant within it as SGT
+  // wall-clock, let date-fns find the month's start/end in that wall clock,
+  // then convert both boundaries back to UTC instants.
+  const zoned = toZonedTime(
+    new Date(Date.UTC(year, monthIndex - 1, 1)),
+    SINGAPORE_TIME_ZONE,
+  )
+  const monthStart = fromZonedTime(startOfMonth(zoned), SINGAPORE_TIME_ZONE)
+  const monthEnd = fromZonedTime(endOfMonth(zoned), SINGAPORE_TIME_ZONE)
+
```
seaerchin: comment: this first constructs a **utc instant** of 15 <month> <year> and then converts it into an sg aware TZ

then, it converts that sg tz to first of month/end of month then returns that

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin, line 1: taken from existing file, left unchanged except for shifting siteId out

seaerchin:
@greptileai review

seaerchin:
@greptileai review

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,486 @@
...
+  if (!match?.[1] || !match[2]) {
+    throw new Error(
+      `Invalid audit log date range, expected "[YYYY-MM-DD,YYYY-MM-DD)" but got: ${auditLogDateRange}`,
+    )
+  }
+  return { lowerInclusive: match[1], upperExclusive: match[2] }
+}
+
+/**
+ * Convert a `yyyy-MM` month (interpreted in Singapore time) into the stored
```
seaerchin, line 54: comment: this first constructs a **utc instant** of 15 <month> <year> and then converts it into an sg aware TZ

then, it converts that sg tz to first of month/end of month then returns that

seaerchin:
@greptileai

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,487 @@
...
+        .then(AUDIT_LOGS_EVENTS_QUERIES[AuditLogEvent.ResourceDelete])
+        .when("al.eventType", "=", AuditLogEvent.Publish)
+        .then(AUDIT_LOGS_EVENTS_QUERIES[AuditLogEvent.Publish])
+        .when("al.eventType", "=", AuditLogEvent.NavbarUpdate)
+        .then(AUDIT_LOGS_EVENTS_QUERIES[AuditLogEvent.NavbarUpdate])
+        .when("al.eventType", "=", AuditLogEvent.FooterUpdate)
+        .then(AUDIT_LOGS_EVENTS_QUERIES[AuditLogEvent.FooterUpdate])
+        .when("al.eventType", "=", AuditLogEvent.SiteConfigUpdate)
+        .then(AUDIT_LOGS_EVENTS_QUERIES[AuditLogEvent.SiteConfigUpdate])
+        .when("al.eventType", "=", AuditLogEvent.PermissionCreate)
```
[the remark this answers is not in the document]
seaerchin: Why is this a concern? RedirectsCreate and RedirectsDelete are self-evident so showing Description = `-` should be ok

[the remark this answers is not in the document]
dcshzj: RedirectsCreate and RedirectsDelete doesn't even exist in our audit logs as a type, why are you proposing to add them?

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,485 @@
...
+// (A type alias — not an interface — because only type aliases get an
+// implicit index signature, making rows assignable to `Record<string,
+// unknown>` so they can be passed straight to `toCsv`.)
+// oxlint-disable-next-line typescript/consistent-type-definitions
+export type AccessReportRow = {
+  Email: string | null
+  '"Last login"': Date | null
+  Role: string
+  '"Date added"': Date
+}
```
dcshzj: nit: for better organisation can these types be placed in a separate file, e.g. `audit.types.ts`

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,485 @@
...
+        'rp.createdAt as "Date added"',
+      ])
+      .where("rp.siteId", "=", siteId)
+      // Exclude internal Isomer admins from agency-facing reports
+      .where("u.id", "not in", (eb) =>
+        eb.selectFrom("IsomerAdmin").select("IsomerAdmin.userId"),
+      )
+      // Point-in-time predicate: permission must already exist as of the last
+      // covered instant, and must not have been revoked at or before it.
+      .where("rp.createdAt", "<=", rangeEndInclusive)
```
dcshzj: question: why do you need to convert the end to inclusive? Why not keep it exclusive, then using `<` for this comparator (`>=` in line 197)

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,485 @@
...
+          "permission.createdAt as grantedAt",
+          "permission.deletedAt as revokedAt",
+        ])
+        .where("permission.siteId", "=", siteId)
+        // Exclude internal Isomer admins from agency-facing reports
+        .where("collaboratorUser.id", "not in", (ieb) =>
+          ieb.selectFrom("IsomerAdmin").select("IsomerAdmin.userId"),
+        )
+        // Restrict to windows that could overlap the export range
+        .where("permission.createdAt", "<=", rangeEndInclusive)
```
dcshzj: question: same for this also, so we skip the calculation for the end timestamp

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,485 @@
...
+ * - Past month: the full calendar month, e.g. "2024-03" →
+ *   `[2024-03-01,2024-04-01)`.
+ * - Current SGT month (the month of `now` in Asia/Singapore): the upper bound
+ *   is clamped to SGT-today + 1 day so today's partial day is included. On the
+ *   1st this yields `[yyyy-MM-01,yyyy-MM-02)` — never empty.
+ *
+ * Future months are rejected upstream by zod; `now` is an explicit parameter
+ * (no internal clock reads) for testability.
+ */
+export const getMonthDateRange = (month: IsoMonth, now: Date): string => {
```
dcshzj, line 66: question: why do we need to pass in the `now` param? Is it for the unit tests?

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Changed in 1a9db51ee — dropped the -1ms arithmetic and now compare directly against the exclusive boundary: `createdAt < rangeEnd` and `deletedAt >= rangeEnd`. Besides being simpler, it's more correct: Postgres timestamps have microsecond precision, so the -1ms inclusive bound misclassified rows landing in the final millisecond of the range.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Changed in 1a9db51ee — same as above, the CTE now uses `permission.createdAt < rangeEnd` and `rangeEndInclusive` is gone entirely.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Moved in 1a9db51ee — `AuditReportQueryParams` and `AccessReportRow` now live in `audit.types.ts` (re-exported from the query module so the upcoming router PR can import from either). `ActivityReportRow` stays in the query file since it's inferred from the query's select list.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Yes — `now` is injected so the clamp for an in-progress month is deterministic in tests (no internal clock reads). Callers pass `new Date()`; the future-month guard itself lives upstream in the zod schema.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,485 @@
...
+ */
+export const getAccessReportRows = async ({
+  siteId,
+  auditLogDateRange,
+}: AuditReportQueryParams): Promise<AccessReportRow[]> => {
+  const { rangeEnd } = getExportRange(auditLogDateRange)
+  // The last covered instant of the half-open range (the exclusive boundary
+  // minus one millisecond). Using this — rather than the exclusive `rangeEnd`
+  // — is what excludes a revocation landing exactly on the boundary.
+  const rangeEndInclusive = new Date(rangeEnd.getTime() - 1)
```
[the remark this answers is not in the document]
seaerchin: Fixed in 1a9db51ee — the predicates now compare against the exclusive `rangeEnd` directly (`createdAt < rangeEnd`, `deletedAt >= rangeEnd`), so there is no -1ms window left. Same change applied to the `collaboratorWindows` CTE.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,461 @@
...
+): { rangeStart: Date; rangeEnd: Date } => {
+  const { lowerInclusive, upperExclusive } =
+    parseAuditLogDateRange(auditLogDateRange)
+  return {
+    rangeStart: fromZonedTime(lowerInclusive, SINGAPORE_TIME_ZONE),
+    rangeEnd: fromZonedTime(upperExclusive, SINGAPORE_TIME_ZONE),
+  }
+}
+
+export type { AccessReportRow, AuditReportQueryParams } from "./audit.types"
```
adriangohjw: nit/ask: don't think this export is being used - can remove

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,461 @@
...
+      // Point-in-time predicate against the exclusive trailing edge: the
+      // permission must have been created inside the range and must not have
+      // been revoked before its end.
+      .where("rp.createdAt", "<", rangeEnd)
+      .where((eb) =>
+        eb.or([
+          eb("rp.deletedAt", "is", null),
+          eb("rp.deletedAt", ">=", rangeEnd),
+        ]),
+      )
```
adriangohjw, line 174: suggestion (non-blocking) - duplicated logic found in other code. maybe can have a separate PR to deduplicate it with a shared helper

| Occurrence | Lines | Predicate |
|------------|:-----:|-----------|
| **collaboratorWindows CTE** range-overlap check | 273–278 | `permission.createdAt < rangeEnd AND (permission.deletedAt IS NULL OR permission.deletedAt > rangeStart)` — different operator (`>` not `>=`) and different variable (`rangeStart` not `rangeEnd`) |
| **Login gating** subquery | 389–395 | `cw.grantedAt <= al."createdAt" AND (cw.revokedAt IS NULL OR cw.revokedAt > al."createdAt")` |
| **Logout gating** subquery | 405–411 | Identical shape to Login, verbatim duplicate of it |

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,461 @@
...
+// handled. This module is the single source of truth for the audit report
+// queries — `prisma/scripts/getAuditLogs.ts` is a thin wrapper over them.
+type DisplayableAuditLogEvent = Exclude<
+  AuditLogEvent,
+  | "UserCreate"
+  | "UserUpdate"
+  | "UserDelete"
+  | "PermissionUpdate"
+  | "SchedulePublish"
+  | "CancelSchedulePublish"
```
adriangohjw, line 190: question - why don't we want to show this action?
_(non-blocking since it was ported over, just wanna understand)_

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,461 @@
...
+    year === undefined ||
+    monthIndex === undefined ||
+    Number.isNaN(year) ||
+    Number.isNaN(monthIndex) ||
+    monthIndex < 1 ||
+    monthIndex > 12
+  ) {
+    // oxlint-disable-next-line @typescript-eslint/restrict-template-expressions
+    throw new Error(`Invalid month, expected "yyyy-MM" but got: ${month}`)
+  }
```
adriangohjw: suggestion(non-blocking): i think we can simplify to
```ts
const ISO_MONTH_REGEX = /^\d{4}-(0[1-9]|1[0-2])$/ // yyyy-MM pattern
if (!ISO_MONTH_REGEX.test(month)) {
  throw new Error(`Invalid month, expected "yyyy-MM" but got: ${month}`)
}
```

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,461 @@
...
+
+  // A UTC instant mid-month falls inside the target month in every timezone,
+  // so we can derive the SGT month start from it without boundary surprises.
+  // `toZonedTime` re-labels the instant with SGT wall-clock fields, so the
+  // plain date-fns `format`/`startOfMonth`/`addMonths` below all operate on
+  // SGT calendar dates regardless of the server timezone.
+  const midMonth = new Date(Date.UTC(year, monthIndex - 1, MID_OF_MONTH))
+  const monthStart = startOfMonth(toZonedTime(midMonth, SINGAPORE_TIME_ZONE))
+  const lowerInclusive = format(monthStart, "yyyy-MM-dd")
+  const nextMonthStart = format(addMonths(monthStart, 1), "yyyy-MM-dd")
```
adriangohjw, line 82: suggestion(non-blocking): i don't think we need time date and time conversion here
```ts
const [yearNumber, monthNumber] = month.split("-").map(Number)

// Lower bound is always the first of the month
const lowerInclusive = `${yearNumber}-${monthNumber}-01`

// Calculate the start of the next month
const nextMonthStart =
  monthNumber === 12
    ? `${yearNumber + 1}-01-01`
    : `${yearNumber}-${monthNumber + 1}-01`
```

apps/studio/src/server/modules/audit/auditLogExport.query.ts
```diff
@@ -0,0 +1,461 @@
...
+ * day, whose trailing edge is still in the future) the predicate naturally
+ * collapses to "who has access now" — this is intended.
+ *
+ * Internal Isomer admins (rows in the `IsomerAdmin` table) are excluded,
+ * matching the script (PR #2612).
+ */
+export const getAccessReportRows = async ({
+  siteId,
+  auditLogDateRange,
+}: AuditReportQueryParams): Promise<AccessReportRow[]> => {
```
adriangohjw, line 144: nit/ask: i don't think we need to add the return type here since it doesn't quite serve any purpose? it will also save us the need to declare the type `AccessReportRow` in `apps/studio/src/server/modules/audit/audit.types.ts` - which isn't referenced anywhere too

adriangohjw (approved):
no blockers, just some nits/asks which aren't hard blockers for me

apps/studio/src/server/modules/audit/auditLogExport.query.ts
adriangohjw: can create this in a new github issue if you want haha

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Superseded by the daterange design — months convert to SGT calendar-date ranges via `getMonthDateRange`, and `getExportRange` maps the bounds to UTC instants.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Removed in 352784fff.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Will split into a follow-up issue as suggested — the three predicates (collaboratorWindows overlap, Login/Logout gating) can share a half-open interval helper.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Ported unchanged from the existing script's displayable set: the activity report covers site-facing resource activity, so user-lifecycle/permission bookkeeping events are excluded (access surfaces in the Access report instead), and schedule bookkeeping shows up as the Publish events it produces. Happy to revisit the list separately.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Equivalent behavior either way — kept the manual guard since zod + the `IsoMonth` type already enforce the shape upstream and this is defense-in-depth for the script path. No objection to the regex form if you prefer it.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: The string-math version loses two things: zero-padding (month 3 → `2024-3-01`, which breaks the `[YYYY-MM-DD,YYYY-MM-DD)` canonical form the parser and DB expect) and the current-month clamp to SGT-today+1 — the `toZonedTime` step is what makes "today" an SGT-calendar answer on a UTC host. Keeping the conversion.

apps/studio/src/server/modules/audit/auditLogExport.query.ts
seaerchin: Kept the annotation — it's the declared contract for the hand-rolled query, so column drift fails at the definition site rather than at callers. The genuinely-unused re-export is removed in 352784fff.

PR #2606, opened 2026-06-30
feat(studio): add audit log export S3 and email delivery

seaerchin:
**TODO: create bucket on infra and update env vars**

## Problem

A completed export must be stored privately and delivered to the requester by email with a download link. Audit/access CSVs contain sensitive data (emails, roles, IP addresses, login history), so they must not live in the public assets bucket, and the existing signed-URL helper only supported a 5-minute expiry.

No linked issue (stacked feature work). Part of the stack starting at #2603.

## Solution

Add the storage + email plumbing the orchestrator (#2607) will call.

**Breaking Changes**

- [ ] Yes
- [x] No — `generateSignedGetUrl`'s default expiry is unchanged (5 min); all other additions are new.

**Features**:

- `uploadAuditLogExport({ key, body })` (`lib/s3.ts`) — uploads a CSV to the private audit-export bucket with `text/csv` + `attachment` disposition; throws a clear error if the bucket env var is unconfigured.
- `getAuditLogExportBucketName()` + `AUDIT_LOG_EXPORT_URL_EXPIRY_SECONDS` (3 days) constant.
- `generateSignedGetUrl(..., expiresIn?)` — optional expiry param (default still `60 * 5`).
- Email templates `auditLogExportReady` (a **single** download link — the template takes `link: { label, url }`; each export job produces one CSV and sends its own email, so a "Both" request results in two independent emails — "valid for 3 days") and `auditLogExportFailed`, with `sendAuditLogExportReadyEmail` / `sendAuditLogExportFailedEmail` senders.
- New **optional** env var `S3_STUDIO_ASSETS_BUCKET_NAME` (server-only) with a use-time guard.

**Improvements**:

- Sensitive exports are isolated to a dedicated private bucket; expiry is explicit (ADR-0004).

**Bug Fixes**:

- None.

## Tests

- `lib/__tests__/uploadAuditLogExport.test.ts` — asserts the PutObject call uses the configured bucket, `text/csv`, attachment disposition, and throws when the bucket env var is unset (S3 mocked).
- `features/mail/templates/__tests__/auditLogExport.test.ts` — ready template renders exactly one download link with the right href + label per report kind, per-kind subjects, "3 days" validity note present, HTML-escaping of unsafe site names and link hrefs; failed template sanity.
- `pnpm typecheck` green; both test files passing.

**New dependencies**: none.

[...]

PR #2607, opened 2026-06-30
feat(studio): add audit log export request endpoint and cron

seaerchin:
## Problem

Site Admins need an endpoint to request an export, and the system needs to fulfil those requests asynchronously (a month of activity for a large site can be slow to generate), deliver the result by email, and survive transient failures — without blocking the request.

No linked issue (stacked feature work). Part of the stack starting at #2603 (depends on the schema, Zod, queries, and delivery PRs).

## Solution

Add the request-creation endpoint and the pg-boss cron that drains pending requests, following the repo's existing `PushDocumentJob` poll pattern (the wrapper is schedule-only; no ad-hoc queue).

**Breaking Changes**

- [ ] Yes
- [x] No — new module + new cron job registration; existing routers untouched.

**Features**:

- `audit.createExportRequest` mutation (registered in `_app.ts`):
  - **Admin-only** — reuses `validatePermissionsForManagingUsers({ action: "manage" })`, the same gate as `/sites/[siteId]/users`; Editors/Publishers are rejected.
  - **Month-window guard** (Asia/Singapore) → `BAD_REQUEST` for a future month, and for any month older than the 12-month window (current month + previous 11), using the shared `getEarliestExportableMonth` helper.
  - **In-flight dedupe** — rejects with `CONFLICT` if an identical `(siteId, userId, auditLogDateRange, reportType)` request is already `Pending`/`Processing` (backed by the partial unique index; `daterange` canonicalization makes the equality check representation-proof).
  - **`Both` fan-out** — a `Both` request is input vocabulary only: the service inserts **two** rows (one `Access`, one `Activity`) inside a single transaction, all-or-nothing — if either insert loses to the unique index, the whole transaction rolls back and the request fails with `CONFLICT`. The endpoint returns the array of inserted rows (1 or 2).
  - Rate-limited (`.meta({ rateLimitOptions: { max: 5, windowMs: 60_000 } })`) since it triggers downstream email/S3.
  - Inserts `Pending` row(s) with the month's `daterange` (a current-month request is clamped to SGT-today + 1 at insert time, so today's partial day is included); errors are logged via `ctx.logger` and never leaked.
- `audit-log-export` cron job (every minute) → `processPendingAuditLogExports()`:
  - Atomically claims a row (`UPDATE ... WHERE id = ? AND status = 'Pending' RETURNING`).
  - Each row is exactly one report: generates it, uploads **one** CSV (inclusive-end key slug, e.g. `access-2026-04-01-to-2026-04-30.csv`), signs a 3-day URL, sends a ready email with a **single** download link, marks `Done`. A `Both` request (two rows) therefore yields two independent emails.
  - **Bounded retry**: on error, `attempts++`; `< 3` → back to `Pending` (retried next sweep); `>= 3` → `Failed` + failure email.
  - Empty results still produce a (header-only) CSV and a link.

**Improvements**:

- Per-row error isolation so one bad request never aborts the sweep.

**Bug Fixes**:

- None.

## Tests

- `__tests__/audit.router.test.ts` — Admin happy path (one `Pending` row), Editor/no-permission → `FORBIDDEN`, duplicate → `CONFLICT`, future month → `BAD_REQUEST`, unauthed → 401.
- `__tests__/auditLogExport.processor.test.ts` — Access happy path (one upload with an inclusive-end key, one link, `Done`), fanned-out `Both` (two rows → two uploads + two single-link emails, both `Done`), empty results (header-only, still `Done`), retry-then-fail (attempts 1→2 stay `Pending`, attempt 3 → `Failed` + failure email, ready email never sent), and claiming (a `Done` row not reprocessed). DB is real; S3/mail mocked.
- `pnpm test:unit -- src/server/modules/audit/` → all passing. Typecheck + lint green.

**New dependencies**: none.

This PR adds an audit log export request flow for Studio. The main changes are:

- A new admin-only `audit.createExportRequest` tRPC mutation with month validation, rate limiting, and in-flight dedupe.
- Fan-out from `Both` requests into separate `Access` and `Activity` export rows.
- A scheduled `audit-log-export` cron job that claims pending or stale requests and processes them asynchronously.
- CSV generation, S3 upload, signed download links, and ready/failure email delivery for each export row.
- Tests covering authorization, duplicate handling, fan-out, retries, stale claims, empty exports, and mail templates.
</details>

This PR appears safe to merge based on the reviewed changed paths.

No new non-duplicative blocking issues were identified. The endpoint follows the reviewed authorization, validation, and rate-limit conventions. The processor has focused tests for retries, stale claims, fan-out, empty exports, and terminal states.

**Files Needing Attention:** `apps/studio/src/server/modules/audit/auditLogExport.service.ts` remains the highest-risk file because it owns claim/retry state transitions plus S3 and email side effects.
</details>

**What T-Rex did**
- Attempted to run the audit create export request, but environment and setup failures blocked the execution.
- Tried the audit log export processor tests; the base invocation failed due to a missing container runtime, and the after state shows a blocked head invocation plus a passing cron-registration test with exit code 0.
- Ran the audit export persistence dedupe tests; the after state shows a partial unique index and service implementation with four tests passing, while the fan-out test failed due to an unexpected report type order.

<a href="https://app.greptile.com/trex/runs/17211980/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/audit/auditLogExport.service.ts | Implements request creation, in-flight dedupe, report fan-out, export processing, retry handling, S3 upload, and email delivery. |
| apps/studio/src/server/modules/audit/audit.router.ts | Adds the admin-only, rate-limited tRPC mutation for creating audit log export requests. |
| apps/studio/src/server/modules/permissions/permissions.service.ts | Adds an active Site Admin permission validator for the new audit export endpoint. |
| apps/studio/src/schemas/audit.ts | Adds audit export request validation and month-window helpers for the new endpoint. |
| apps/studio/src/server/cron/jobs/auditLogExportJob.ts | Adds a scheduled pg-boss wrapper that runs pending audit log export processing every minute. |
| apps/studio/src/features/mail/templates/templates.ts | Adds audit export ready/failed mail templates with escaped HTML bodies and raw text subjects. |

</details>

```mermaid
sequenceDiagram
  actor Admin as Site Admin
  participant Router as audit.createExportRequest
  participant Perm as validateUserIsSiteAdmin
  participant DB as AuditLogExportRequest
  participant Cron as audit-log-export cron
  participant Query as Report queries
  participant S3 as S3
  participant Mail as Email service

  Admin->>Router: request export(siteId, month, reportType)
  Router->>Perm: verify active Site Admin
  Perm-->>Router: allowed
  Router->>DB: transaction: dedupe + insert Pending row(s)
  DB-->>Router: inserted Access/Activity request(s)
  Router-->>Admin: accepted rows

  Cron->>DB: select Pending or stale Processing batch
  loop each request
    Cron->>DB: atomically claim Processing + attempts + 1
    Cron->>Query: build Access or Activity rows
    Query-->>Cron: report rows
    Cron->>S3: upload CSV
    S3-->>Cron: object stored
    Cron->>S3: generate signed URL
    S3-->>Cron: 3-day URL
    Cron->>Mail: send ready email
    Cron->>DB: mark Done
  end

  alt processing error
    Cron->>DB: "requeue Pending if attempts < 3"
    Cron->>DB: "mark Failed if attempts >= 3"
    Cron->>Mail: send failure email after final attempt
  end
```
</details>

<sub>Reviews (20): Last reviewed commit: ["chore: unescape the title of the emali"](https://github.com/opengovsg/isomer/commit/fd5166c44f4beda49ecbbe4d712cc7205fa96028) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41594814)</sub>

### DIFF
--- a/apps/studio/src/server/CLAUDE.md
+++ b/apps/studio/src/server/CLAUDE.md
@@ -40,7 +40,7 @@ Never write your own auth middleware in a router. If you need a new auth shape, 
 - Define the Zod schema in `apps/studio/src/schemas/<area>.ts` and import it in the router.
 - Routers should not declare ad-hoc inline schemas. If a schema is router-internal and tiny, keep it inline; otherwise lift it.
 
-### Permission checks belong in the router 
+### Permission checks belong in the router
 
 - Routers wire input → service. The router performs the permission check (`bulkValidateUserPermissionsForResources` and friends) **before** mutating and calling services.
 - Do not gate purely on the existence of a session — that only proves the caller is logged in, not authorised for the resource.
--- a/apps/studio/src/server/CLAUDE.md
+++ b/apps/studio/src/server/CLAUDE.md
@@ -42,7 +42,7 @@ Never write your own auth middleware in a router. If you need a new auth shape, 
 
 ### Permission checks belong in the service
 
-- Routers wire input → service. The service performs the permission check (`bulkValidateUserPermissionsForResources` and friends) **before** mutating.
+- Routers wire input → service. The router performs the permission check (`bulkValidateUserPermissionsForResources` and friends) **before** mutating and calling services.
 - Do not gate purely on the existence of a session — that only proves the caller is logged in, not authorised for the resource.
 
 ### Audit-log every state change
--- a/apps/studio/src/server/CLAUDE.md
+++ b/apps/studio/src/server/CLAUDE.md
@@ -40,7 +40,7 @@ Never write your own auth middleware in a router. If you need a new auth shape, 
 - Define the Zod schema in `apps/studio/src/schemas/<area>.ts` and import it in the router.
 - Routers should not declare ad-hoc inline schemas. If a schema is router-internal and tiny, keep it inline; otherwise lift it.
 
-### Permission checks belong in the service
+### Permission checks belong in the router 
 
 - Routers wire input → service. The router performs the permission check (`bulkValidateUserPermissionsForResources` and friends) **before** mutating and calling services.
 - Do not gate purely on the existence of a session — that only proves the caller is logged in, not authorised for the resource.
--- a/apps/studio/src/server/modules/audit/audit.router.ts
+++ b/apps/studio/src/server/modules/audit/audit.router.ts
@@ -1,0 +1,44 @@
+import { TRPCError } from "@trpc/server"
+import { createAuditLogExportRequestSchema } from "~/schemas/audit"
+
+import { protectedProcedure, router } from "../../trpc"
+import { createAuditLogExportRequest } from "./auditLogExport.service"
+
+export const auditRouter = router({
+  createExportRequest: protectedProcedure
+    .input(createAuditLogExportRequestSchema)
+    // Rate-limited because each accepted request eventually triggers downstream
+    // work that hits external services (CSV generation, S3 upload, email).
+    // Arbitrary low limit to prevent abuse; tune if legitimate usage is blocked.
+    .meta({ rateLimitOptions: { max: 5, windowMs: 60_000 } })
+    .mutation(async ({ ctx, input: { siteId, month, reportType } }) => {
+      try {
+        return await createAuditLogExportRequest({
+          siteId,
+          userId: ctx.user.id,
+          month,
+          reportType,
+        })
+      } catch (error) {
+        // Permission / validation / dedupe failures are already typed
+        // TRPCErrors with safe, user-facing messages — let them through.
+        if (error instanceof TRPCError) {
+          throw error
+        }
+
+        // Anything else (e.g. a DB error) may leak request internals; log it
+        // with the request context and surface a generic error to the client.
+        ctx.logger.error({
+          error,
+          message: "Failed to create audit log export request",
+          siteId,
+          month,
+          reportType,
+        })
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Failed to create audit log export request",
+        })
+      }
+    }),
+})
--- a/apps/studio/src/server/modules/audit/audit.router.ts
+++ b/apps/studio/src/server/modules/audit/audit.router.ts
@@ -2,6 +2,7 @@ import { TRPCError } from "@trpc/server"
 import { createAuditLogExportRequestSchema } from "~/schemas/audit"
 
 import { protectedProcedure, router } from "../../trpc"
+import { validateUserIsSiteAdmin } from "../permissions/permissions.service"
 import { createAuditLogExportRequest } from "./auditLogExport.service"
 
 export const auditRouter = router({
@@ -12,6 +13,13 @@ export const auditRouter = router({
     // Arbitrary low limit to prevent abuse; tune if legitimate usage is blocked.
     .meta({ rateLimitOptions: { max: 5, windowMs: 60_000 } })
     .mutation(async ({ ctx, input: { siteId, month, reportType } }) => {
+      // Permission check FIRST, before any mutation. Audit log export is a
+      // Site Admin-only capability so we reject any attempts if they are not an admin
+      await validateUserIsSiteAdmin({
+        siteId,
+        userId: ctx.user.id,
+      })
+
       try {
         return await createAuditLogExportRequest({
           siteId,
--- a/apps/studio/src/server/modules/audit/auditLogExport.service.ts
+++ b/apps/studio/src/server/modules/audit/auditLogExport.service.ts
@@ -31,7 +31,6 @@ import type { BaseLogger } from "@isomer/logging"
 import type { AuditLogExportReportType } from "../database"
 import { db } from "../database"
 import { PG_ERROR_CODES } from "../database/constants"
-import { validatePermissionsForManagingUsers } from "../permissions/permissions.service"
 import {
   getAccessReportRows,
   getActivityReportRows,
@@ -68,17 +67,6 @@ export const createAuditLogExportRequest = async ({
   month,
   reportType,
 }: CreateAuditLogExportRequestProps) => {
-  // Permission check FIRST, before any mutation. Audit log export is a
-  // Site Admin-only capability — we reuse the same admin-only gate as the
-  // user-management surface (`manage` on `UserManagement`), which only grants
-  // the `manage` action to the `Admin` role. Editors/Publishers can `read`
-  // but not `manage`, so they are rejected here with FORBIDDEN.
-  await validatePermissionsForManagingUsers({
-    siteId,
-    userId,
-    action: "manage",
-  })
-
   // Reject months outside the allowed window (in Singapore time). This mirrors
   // the schema's window refinements as defense-in-depth, and uses the same
   // date-fns comparators (parseISO + isBefore/isAfter/isSameMonth) so the
--- a/apps/studio/src/server/modules/audit/auditLogExport.service.ts
+++ b/apps/studio/src/server/modules/audit/auditLogExport.service.ts
@@ -20,7 +20,7 @@ import {
   uploadAuditLogExport,
 } from "~/lib/s3"
 import {
-  type AuditLogExportRequestedReportType,
+  AuditLogExportRequestedReportType,
   type CreateAuditLogExportRequestInput,
   getCurrentSingaporeMonth,
   getEarliestExportableMonth,
@@ -28,7 +28,7 @@ import {
 
 import type { BaseLogger } from "@isomer/logging"
 
-import type { AuditLogExportReportType } from "../database"
+import { AuditLogExportReportType } from "../database"
 import { AuditLogExportStatus, db } from "../database"
 import { PG_ERROR_CODES } from "../database/constants"
 import {
@@ -59,9 +59,14 @@ const REPORT_TYPES_BY_REQUESTED_TYPE: Record<
   AuditLogExportRequestedReportType,
   readonly AuditLogExportReportType[]
 > = {
-  Access: ["Access"],
-  Activity: ["Activity"],
-  Both: ["Access", "Activity"],
+  [AuditLogExportRequestedReportType.Access]: [AuditLogExportReportType.Access],
+  [AuditLogExportRequestedReportType.Activity]: [
+    AuditLogExportReportType.Activity,
+  ],
+  [AuditLogExportRequestedReportType.Both]: [
+    AuditLogExportReportType.Activity,
+    AuditLogExportReportType.Access,
+  ],
 }
 
 export const createAuditLogExportRequest = async ({
@@ -219,8 +224,14 @@ const PROCESSING_LEASE_MS = 15 * 60 * 1000
 // request time (see REPORT_TYPES_BY_REQUESTED_TYPE), so every row here maps to
 // exactly one report — one CSV, one download link, one email.
 const REPORT_BY_TYPE = {
-  Access: { kind: "Access", label: "access" },
-  Activity: { kind: "Activity", label: "audit" },
+  [AuditLogExportReportType.Access]: {
+    kind: AuditLogExportReportType.Access,
+    label: "access",
+  },
+  [AuditLogExportReportType.Activity]: {
+    kind: AuditLogExportReportType.Activity,
+    label: "audit",
+  },
 } as const
 
 /**
--- a/apps/studio/src/server/modules/audit/auditLogExport.service.ts
+++ b/apps/studio/src/server/modules/audit/auditLogExport.service.ts
@@ -359,9 +359,7 @@ export const processAuditLogExportRequest = async (
             auditLogDateRange: request.auditLogDateRange,
           })
 
-    // The report row types are precise interfaces without an index
-    // signature; `toCsv` only needs string-keyed records, so widen here.
-    const csv = toCsv(rows as unknown as Record<string, unknown>[])
+    const csv = toCsv(rows)
     const rangeSlug = getRangeSlug(request.auditLogDateRange)
     const objectKey = `audit-log-exports/${request.siteId}/${requestId}/${report.kind.toLowerCase()}-${rangeSlug}.csv`
 
--- a/apps/studio/src/server/modules/audit/auditLogExport.service.ts
+++ b/apps/studio/src/server/modules/audit/auditLogExport.service.ts
@@ -29,7 +29,7 @@ import {
 import type { BaseLogger } from "@isomer/logging"
 
 import type { AuditLogExportReportType } from "../database"
-import { db } from "../database"
+import { AuditLogExportStatus, db } from "../database"
 import { PG_ERROR_CODES } from "../database/constants"
 import {
   getAccessReportRows,
@@ -46,7 +46,10 @@ type CreateAuditLogExportRequestProps = CreateAuditLogExportRequestInput & {
 // Statuses that represent an export that is still in-flight; a duplicate
 // request for the same (site, user, range, report type) should be rejected
 // while one of these exists.
-const IN_FLIGHT_STATUSES = ["Pending", "Processing"] as const
+const IN_FLIGHT_STATUSES = [
+  AuditLogExportStatus.Pending,
+  AuditLogExportStatus.Processing,
+]
 
 // Fan-out from the requested (input) report type to the DB rows to insert.
 // `Both` is UX vocabulary only (see schemas/audit.ts): it becomes TWO
--- a/apps/studio/src/server/modules/audit/auditLogExport.service.ts
+++ b/apps/studio/src/server/modules/audit/auditLogExport.service.ts
@@ -16,7 +16,7 @@ import { createBaseLogger } from "~/lib/logger"
 import {
   AUDIT_LOG_EXPORT_URL_EXPIRY_SECONDS,
   generateSignedGetUrl,
-  getAuditLogExportBucketName,
+  getStudioAssetsBucketName,
   uploadAuditLogExport,
 } from "~/lib/s3"
 import {
@@ -346,7 +346,7 @@ export const processAuditLogExportRequest = async (
     // URL. `Both` requests were fanned out into two rows at request time, so
     // one row is always exactly one report.
     const report = REPORT_BY_TYPE[request.reportType]
-    const bucket = getAuditLogExportBucketName()
+    const bucket = getStudioAssetsBucketName()
 
     const rows =
       report.kind === "Access"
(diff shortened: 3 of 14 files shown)


seaerchin:
* **#2607** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2605** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2605?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2603** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/audit-logs`


<h2></h2>

seaerchin:
@greptileai review

seaerchin:
@greptileai review

seaerchin:
@greptileai review

seaerchin:
@greptileai review

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,504 @@
...
+        status: "Done",
+        objectKey,
+        errorMessage: null,
+        updatedAt: new Date(),
+      })
+      .where("id", "=", requestId)
+      .execute()
+  } catch (error) {
+    // Step 7: failure handling. Increment attempts; re-queue or fail.
+    const attempts = request.attempts + 1
```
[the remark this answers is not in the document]
seaerchin: Confirmed — the claim UPDATE incremented `attempts` for stale re-claims and the catch added 1 to the post-UPDATE value (`returningAll` returns the post-increment row), so a stale re-claim that then failed was charged twice: a row could jump 1 → 3 and skip its middle retry, going straight to `Failed` + failure email.

Fixed in 22ebdd112: `attempts` now counts processing attempts **started** — every claim (fresh or stale) charges `attempts + 1` atomically, and the catch only reads the already-charged value to decide re-queue vs `Failed`. Each attempt is counted exactly once whether it ends in a caught error, a dead worker, or success. Also added the missing stale-reclaim-then-failure test: a stale `Processing` row with `attempts: 1` that fails again ends up re-queued as `Pending` with `attempts: 2`, not `Failed`.

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+  )
+  if (
+    !isSameMonth(requestedMonthDate, earliestMonthDate) &&
+    isBefore(requestedMonthDate, earliestMonthDate)
+  ) {
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: "You can only export audit logs from the past 12 months",
+    })
+  }
```
adriangohjw: issue(?): we should let the input schema handle this?

apps/studio/src/server/modules/audit/audit.router.ts
```diff
@@ -0,0 +1,44 @@
...
+          message: "Failed to create audit log export request",
+          siteId,
+          month,
+          reportType,
+        })
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Failed to create audit log export request",
+        })
+      }
```
adriangohjw, line 50: issue:
- currently the permission check `validatePermissionsForManagingUsers` is in `createAuditLogExportRequest` -> but we should stick with existing convention of having it the router instead of the service
- i believe we should move this try-catch and error throwing stuff into the service itself?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+  // Permission check FIRST, before any mutation. Audit log export is a
+  // Site Admin-only capability — we reuse the same admin-only gate as the
+  // user-management surface (`manage` on `UserManagement`), which only grants
+  // the `manage` action to the `Admin` role. Editors/Publishers can `read`
+  // but not `manage`, so they are rejected here with FORBIDDEN.
+  await validatePermissionsForManagingUsers({
+    siteId,
+    userId,
+    action: "manage",
+  })
```
adriangohjw: nit: logically correct but feels iffy reusing this validation since its for managing users - can be confusing from a naming POV and unnecessarily tying them together. would prefer if we have a simple `validateUserIsSiteAdmin` (maybe even extracted as shared method)

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+} from "./auditLogExport.query"
+
+type CreateAuditLogExportRequestProps = CreateAuditLogExportRequestInput & {
+  userId: string
+}
+
+// Statuses that represent an export that is still in-flight; a duplicate
+// request for the same (site, user, range, report type) should be rejected
+// while one of these exists.
+const IN_FLIGHT_STATUSES = ["Pending", "Processing"] as const
```
adriangohjw: nit: can we use the value from `AuditLogExportStatus` from the generated enums instead?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+              attempts: 0,
+            })
+            .returningAll()
+            .executeTakeFirstOrThrow(),
+        )
+      }
+      // Return every inserted row (one for Access/Activity, two for Both).
+      // The UI ignores the payload, so the array shape is chosen purely to
+      // reflect the fan-out honestly.
+      return inserted
```
adriangohjw: suggestion: should we combine to reduce number of API calls + keep it cleaner?

```suggestion
      return await tx
        .insertInto("AuditLogExportRequest")
        .values(
          reportTypes.map((dbReportType) => ({
            siteId,
            userId,
            auditLogDateRange,
            reportType: dbReportType,
            status: "Pending",
            attempts: 0,
          })),
        )
        .returningAll()
        .execute()
```

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+        inserted.push(
+          await tx
+            .insertInto("AuditLogExportRequest")
+            .values({
+              siteId,
+              userId,
+              auditLogDateRange,
+              reportType: dbReportType,
+              status: "Pending",
+              attempts: 0,
```
adriangohjw, line 156: nit: should not need this since its default to 0 in the DB but okay to keep for explicitness

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+// `Both` is UX vocabulary only (see schemas/audit.ts): it becomes TWO
+// AuditLogExportRequest rows — one Access, one Activity — each fulfilled as an
+// independent job. The DB enum has no `Both` member.
+const REPORT_TYPES_BY_REQUESTED_TYPE: Record<
+  AuditLogExportRequestedReportType,
+  readonly AuditLogExportReportType[]
+> = {
+  Access: ["Access"],
+  Activity: ["Activity"],
+  Both: ["Access", "Activity"],
```
adriangohjw: nit - to use DB enum `AuditLogExportReportType` instead?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
adriangohjw, line 1: nit: in general i think we can use the generated enum whenever possible instead of hardcoing as string. couple of instances so i only named some to not cloud the review

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+    // Step 5: one ready email with the single download link. A "Both" user
+    // request is two independent rows, so it yields two independent emails —
+    // no cross-job coordination.
+    await sendAuditLogExportReadyEmail({
+      recipientEmail,
+      siteName,
+      month: getExportPeriodLabel(request.auditLogDateRange),
+      link: { label: report.label, url },
+    })
+
```
adriangohjw, line 393: question: this is outdated and no longer relevant since we are switching to generating signed URL on demand due to the TTL?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+          "AuditLogExportRequest.auditLogDateRange as auditLogDateRange",
+        ])
+        .executeTakeFirstOrThrow()
+
+      const failedConfig = failed.config as { siteName?: string } | null
+      await sendAuditLogExportFailedEmail({
+        recipientEmail: failed.recipientEmail,
+        siteName: failedConfig?.siteName || failed.siteName,
+        month: getExportPeriodLabel(failed.auditLogDateRange),
+      })
```
adriangohjw: question: i think we don't need to query this again since we already have the site and user from the above code? just need to pull those out of the try-catch loop and we can reuse it here to remove this DB query

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+    await db
+      .updateTable("AuditLogExportRequest")
+      .set({
+        status: "Done",
+        objectKey,
+        errorMessage: null,
+        updatedAt: new Date(),
+      })
+      .where("id", "=", requestId)
+      .execute()
```
adriangohjw, line 405: issue: we need to set `completedAt` too

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: the input schema already does this; this is just a dup for sanity so that the service also enforces the guarantee:

```ts
  month: z
    .string()
    .regex(MONTH_REGEX, {
      message: "Enter a month in the format YYYY-MM, e.g. 2026-03",
    })
    .refine(
      (month) => {
        const parsed = parseISO(`${month}-01`)
        const current = parseISO(`${getCurrentSingaporeMonth()}-01`)
        return isSameMonth(parsed, current) || isBefore(parsed, current)
      },
      {
        message:
          "You cannot export audit logs for a month that is in the future",
      },
    )
    .refine(
      (month) => {
        const parsed = parseISO(`${month}-01`)
        const earliest = parseISO(
          `${getEarliestExportableMonth(getCurrentSingaporeMonth())}-01`,
        )
        return isSameMonth(parsed, earliest) || isAfter(parsed, earliest)
      },
      { message: "You can only export audit logs from the past 12 months" },
    )
    // The regex above guarantees the shape at runtime; narrow the inferred
    // output from `string` to `IsoMonth` so consumers get the real type.
    .transform((month) => month as IsoMonth),
```

seaerchin: i'm happy to delete if you want!

apps/studio/src/server/modules/audit/audit.router.ts
seaerchin: sounds good - lemme shift it

seaerchin: The service itself does a `try/catch`; the thing is that it does rethrow errors which the service itself doesn't know how to handle (eg: db insertions perhaps due to pool being all used)

The pattern is for the service to handle/throw the appropriate error if it knows how to do so and rethrow errors it doesn't know how to handle. 

The added `try/catch` on the router level is just adding hte logger + a nice(r) error to end user - happy to remove it you deem useless, but i don't think we should shift it into the service (or at least we should iron out where, for example, db errors live - do they get caught by service and rethrown as opaque errors are you're suggesting or the router just handles the raw error itself)

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: fixed - added your method as suggested

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: hmm, wouldn't this pattern fail a whole insert if, for example, only 1 row was a duplicate?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: tbh, the type system enforces this for us so i didn't think of this as an issue - if you were to use "access" insetad of "Access", it'd be a type error but i'm happy to change!

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: done downstream

apps/studio/src/server/modules/audit/audit.router.ts
dcshzj: Btw I also noticed that it was inside our Claude instructions that the permissions validation is to be placed inside the service methods. I think we should stick to having the permissions validation at the router level, so the instructions should be updated as well.

apps/studio/src/server/modules/audit/auditLogExport.service.ts
dcshzj: We might be able to abstract out as a common utility function but regardless, agree we should also have this check here as well

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,501 @@
...
+      for (const dbReportType of reportTypes) {
+        inserted.push(
+          await tx
+            .insertInto("AuditLogExportRequest")
+            .values({
+              siteId,
+              userId,
+              auditLogDateRange,
+              reportType: dbReportType,
+              status: "Pending",
```
dcshzj: nit: `AuditLogExportStatus.Pending`

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,501 @@
...
+      if (
+        error instanceof Error &&
+        "code" in error &&
+        error.code === PG_ERROR_CODES.uniqueViolation
+      ) {
+        throw new TRPCError({
+          code: "CONFLICT",
+          message:
+            "An export for this period and report type is already being generated",
+        })
```
dcshzj: question: if only one row fails but the other passes, wouldn't the error message be confusing in the sense that the user would not know which report type is the one that failed?

Also saw the unit test case `should throw CONFLICT for a Both request when one of its report types is already in flight, committing nothing`, seems like the intent is that we do not commit anything even if only one is in conflict, so I believe we should just fail all?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,501 @@
...
+      .select(["id", "name", "config"])
+      .executeTakeFirstOrThrow()
+
+    const user = await db
+      .selectFrom("User")
+      .where("id", "=", request.userId)
+      .select(["id", "email"])
+      .executeTakeFirstOrThrow()
+
+    const siteConfig = site.config as { siteName?: string } | null
```
dcshzj: question(non-blocking): why do we need this cast ah, is it to force `siteName` to be optional?

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,498 @@
...
+    })
+    .where("id", "=", requestId)
+    .where((eb) =>
+      eb.or([
+        eb("status", "=", "Pending"),
+        eb.and([
+          eb("status", "=", "Processing"),
+          eb("updatedAt", "<", staleCutoff),
+        ]),
+      ]),
```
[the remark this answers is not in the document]
dcshzj: @seaerchin I think this is good to have even though we would have change the state to Failed, in case there is some race condition in the way the attempts are being read and incremented

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,501 @@
...
+
+    // Step 5: one ready email with the single download link. A "Both" user
+    // request is two independent rows, so it yields two independent emails —
+    // no cross-job coordination.
+    await sendAuditLogExportReadyEmail({
+      recipientEmail,
+      siteName,
+      month: getExportPeriodLabel(request.auditLogDateRange),
+      link: { label: report.label, url },
+    })
```
dcshzj, line 392: suggestion: btw might want to re-check that the recipient is still a valid site admin before proceeding to send the report (but minor uh since the window is very small anyway)

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: yep updated

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: good call, updated

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: removed the cast

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -0,0 +1,506 @@
...
+
+  const user = await db
+    .selectFrom("User")
+    .innerJoin("ResourcePermission", "ResourcePermission.userId", "User.id")
+    .where("User.id", "=", request.userId)
+    .where("ResourcePermission.userId", "=", request.userId)
+    .where("ResourcePermission.role", "=", "Admin")
+    .where("ResourcePermission.siteId", "=", site.id)
+    .select(["User.id", "User.email"])
+    .executeTakeFirst()
```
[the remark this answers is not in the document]
seaerchin: !fixed

apps/studio/src/server/CLAUDE.md
```diff
@@ -42,7 +42,7 @@ Never write your own auth middleware in a router. If you need a new auth shape,
 
 ### Permission checks belong in the service
```
dcshzj: nit: the heading here is not quite right yeah, it should say router instead of service

dcshzj (approved):

PR #2713, opened 2026-07-08
ISOM-2313 fix(resource): guard getSearchResults against empty resourceTypes

Ssatu:
Linear
https://linear.app/ogp/issue/ISOM-2313/getsearchresults-crashes-with-postgresql-syntax-error-when

## Problem

Closes ISOM-2313

`getSearchResults` throws a 500 error when called with an empty `resourceTypes` array. Kysely compiles `.where("Resource.type", "in", [])` to an invalid `IN ()` SQL clause, which Postgres rejects.

## Solution

Added an early-return guard in `getSearchResults` that short-circuits when `resourceTypes` is empty, returning an empty result set instead of hitting the database with invalid SQL. This mirrors the existing empty-array guard already used in `getWithFullPermalink` in the same file.

**Breaking Changes**

**Bug Fixes**:

- `getSearchResults` (`resource.service.ts`) now returns `{ resources: [], totalCount: 0 }` when `resourceTypes` is an empty array, instead of throwing a 500 from an invalid `IN ()` SQL clause.

## Tests

Added unit tests for `getSearchResults` in `resource.service.test.ts`:
- Returns an empty result (`resources: []`, `totalCount: 0`) when `resourceTypes` is empty.
- Returns matching resources when `resourceTypes` includes the resource's type, confirming the guard doesn't affect the normal search path.

- [ ] In Studio, open the resource search/picker for a site and search with at least one resource type filter selected — confirm results still return normally.
- [ ] Trigger a search request with `resourceTypes` set to an empty array (e.g. via the tRPC panel or by adjusting the client filter state) and confirm it returns an empty result set with `totalCount: 0` instead of a server error.

### DIFF
--- a/apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts
+++ b/apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts
@@ -22,6 +22,7 @@ import {
   getLocalisedSitemap,
   getNavBar,
   getPageById,
+  getSearchResults,
   getSiteResourceById,
   getWithFullPermalink,
   updateBlobById,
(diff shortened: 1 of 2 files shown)


adriangohjw (approved):
nice!

adriangohjw:
@Ssatu tip: you can link the linear issue in the PR description and it should auto link, so when this PR merges, the linear ticket will be closed automatically too

PR #2730, opened 2026-07-09
chore(deps): migrate monorepo from TypeScript 6 to TypeScript 7

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 13 | +62 | -144 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 1 | +425 | -229 |

## Problem

The monorepo was using TypeScript 6 for compilation while relying on `@typescript/native-preview` (`tsgo`) as a preview path toward TypeScript 7 typechecking. TypeScript 7 is now stable, so we should adopt the standard `tsc` compiler across the monorepo and remove the native-preview dependency.

## Solution

Upgraded the monorepo to TypeScript 7 and simplified the typecheck toolchain by replacing `tsgo` with `tsc --noEmit`. `@isomer/db` intentionally remains on TypeScript 6 because `prisma-json-types-generator` and `generate.cts` depend on the TypeScript compiler API, which is not yet compatible with TS 7 in that package.

**Breaking Changes**

  - Details ...

This is a developer-tooling change only. There are no runtime or API breaking changes for production code. TypeScript 7 may surface new type errors during `pnpm typecheck`, but emitted JavaScript behaviour is unchanged.

**Features**:

- None

**Improvements**:

- Upgrade workspace `typescript` catalog from `^6.0.3` to `^7.0.2`
- Remove `@typescript/native-preview` from all packages and the pnpm catalog
- Replace `tsgo` with `tsc --noEmit` in `typecheck` scripts across `apps/studio`, `packages/components`, `packages/db`, `packages/logging`, `packages/pgboss`, and `packages/validators`
- Add a `typescript6` pnpm catalog and pin `@isomer/db` to it so Prisma code generation continues to use the TS 6 compiler API
- Configure Sherif to ignore intentional `typescript` version mismatches between `@isomer/db` and the rest of the monorepo
- Bump `prisma-json-types-generator` from `5.0.0` to `5.1.0`
- Remove deprecated `baseUrl` from `tooling/migrate-tags` and `tooling/site-launch` tsconfigs for TS 7 compatibility

**Bug Fixes**:

- None

## Before & After Screenshots

**BEFORE**:

N/A — tooling/dependency change only.

**AFTER**:

N/A — tooling/dependency change only.

## Benchmarking

Turbo `typecheck` task duration in CI:

| | Duration |
|---|---|
| **Before** (TS 6 + `@typescript/native-preview` / `tsgo`) | 36s |
| **After** (TS 7 + `tsc --noEmit`) | 22s |

~39% faster (~14s saved) on the CI typecheck pipeline.

## Tests

- [ ] `pnpm install`
- [ ] `pnpm typecheck` passes across the monorepo
- [ ] `pnpm lint` passes
- [ ] `pnpm build` succeeds
- [ ] From `packages/db`, run `pnpm generate` to confirm Prisma/Kysely type generation still works on the pinned TS 6 toolchain

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- `typescript@^7.0.2` : stable TypeScript 7 compiler, replacing `@typescript/native-preview` for monorepo typechecking (except `@isomer/db`, which remains on `typescript@^6.0.3` via the `typescript6` catalog)


adriangohjw:
## Blocked: TypeScript 7 is incompatible with current Next.js and dd-trace

I dug into the failing checks (`build`, `End-to-end tests`, `Publishing script tests`, `Components bundle size`, `Vercel`). All five trace back to the same root cause and are not fixable with a small patch in this repo — converting this PR to draft until upstream support lands.

### Root cause

TypeScript 7.0.2 is a ground-up rewrite that removed the classic CommonJS Compiler API (`typescript/lib/typescript.js`, `ts.createProgram`, `ts.transpileModule`, etc.), replacing it with a new `./unstable/*` async API surface. Two tools in our stack still hard-depend on the old API:

1. **Next.js** (16.2.6, and I confirmed the same on the latest stable 16.2.10) hardcodes a file-existence check for `typescript/lib/typescript.js` to detect whether TypeScript is installed. That file no longer exists in TS 7, so Next.js concludes TypeScript isn't installed and aborts the production build immediately in CI (`isCI` skips its normal auto-install fallback). This is why `build`, `End-to-end tests` (never gets past building the app), and the `Vercel` deploy all fail.
2. **`Components bundle size`** hits the same Next.js limitation from another angle: Next parses `tsconfig.json` path aliases (`@/*`) via the same legacy TypeScript API for webpack module resolution, so it silently fails to resolve aliased imports (e.g. `@/data/config.json`) instead of throwing the loud "missing deps" error.
3. **`Publishing script tests`** fails differently: Datadog's `dd-trace` CI wrapper instruments module loading and throws `ERR_REQUIRE_CYCLE_MODULE` when it tries to hook TypeScript 7's new ESM-based `bin/tsc`.

### Why this isn't a quick fix

Patching around the Next.js file-existence check alone isn't sufficient — Next's internal TypeScript integration (tsconfig parsing, type-checking during build, `next-env.d.ts` generation) calls into the classic synchronous Compiler API, which TypeScript 7 no longer exposes via `require('typescript')` (its `.` export now only resolves to a version file). Making that work would require Next.js to adopt the new `unstable` API, which hasn't shipped yet.

### Recommendation

Hold off on this migration until Next.js and dd-trace both ship official TypeScript 7 support, then revisit.

packages/db/prisma/generate.cts
adriangohjw, line 1: - **Purpose:** Prisma-kysely writes `generatedTypes.ts` with a `DB` schema and table types oriented around writes; the app needs **select** row types (`Selectable<T.Table>`) in `selectableTypes.ts` for Kysely queries, tRPC, and publishing code.
- **Old approach:** `generate.cts` used the TypeScript compiler API to read `DB` and emit those `Selectable` aliases.
- **What broke:** The script expected an older Kysely shape; current output is `export type DB = { AuditLog: AuditLog; … };`, so `extractTableTypes` threw on a normal generate.
- **Why it went unnoticed:** Errors were only logged (`console.log`) and the process still exited 0, so `pnpm generate` continued and **stale** `selectableTypes.ts` stayed in the repo.
- **Why we rewrote it:** String parsing matches today’s generator output, fails loudly on errors, and drops the compiler API — which TS 7 no longer provides — so `@isomer/db` can use TypeScript 7 without a separate `typescript6` catalog.

adriangohjw:
@cursor pull from main and resolve conflict (do pnpm i)

PR #2778, opened 2026-07-14
feat(redirect): bulk-upload redirects modal and feature flag

dcshzj:
### 📊 PR diff breakdown

| Bucket | Files | +Lines | -Lines |
|---|---|---|---|
| ⚙️ App | 8 | +583 | -48 |
| 🧪 Test | 3 | +170 | -1 |
| 📄 Doc | 1 | +20 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

> **📚 Top of a 3-PR stack** — bulk upload redirects, split from this PR's original single-PR form. Review and merge bottom-up:
> 1. #2805 — CSV parsing + validation schema
> 2. #2806 — bulk validate/create endpoints
> 3. **#2778 — modal UI + feature flag ← _this PR_**
>
> This PR now contains only the **frontend layer**; the CSV parser/schema is in #2805 and the server endpoints are in #2806.

## Problem

Closes ISOM-2271

Editors could only add redirects one at a time through the inline form on the redirects settings page. Setting up or migrating many redirects at once was slow and error-prone, with no way to validate a whole set before publishing.

## Solution

Adds the **Bulk upload redirects** UI: an editor downloads a CSV template, fills it in, and uploads it. The file is parsed in the browser for instant feedback, validated authoritatively on the server (#2806), and — only if every row is clean — published in one step. The entry point is gated behind the `is-advanced-redirects-enabled` GrowthBook flag for staged rollout.

Design reference: [Figma — Bulk upload redirects](https://www.figma.com/design/g7LnLn4IUcUC9dYxgsHEJM/Isomer-Studio-working-file?node-id=22241-11950&m=dev).

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible

**Features**:

- **Inline "bulk upload with a .csv instead." CTA** in the add-redirect card on the redirects settings page, gated behind `is-advanced-redirects-enabled` via `useIsAdvancedRedirectsEnabled`.
- **Multi-step modal**: download template → upload CSV → client-parse for instant file-level feedback → server `bulkValidate` → fix errors / review preview → `bulkCreate` → publish.
- **Errors flow**: when any row fails, the modal shows the count and offers a downloadable errors file (failures listed first) with plain-language reasons, in the same column shape as the template so it re-uploads directly after fixes.
- **Ready-to-publish preview**: an "All N redirects are good to go" screen with an expandable list of the rows to be added.
[...]

PR #2806, opened 2026-07-15
feat(redirect): add bulk validate/create redirect endpoints

dcshzj:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 5 | +889 | -113 |
| 🧪 Test | 1 | +496 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

> **📚 Part of a 3-PR stack** — bulk upload redirects, split from the original #2778. Review and merge bottom-up:
> 1. #2805 — CSV parsing + validation schema
> 2. **#2806 — bulk validate/create endpoints ← _this PR_**
> 3. #2778 — modal UI + feature flag

## Problem

Bulk-uploading redirects must be validated **authoritatively on the server** (the trust boundary — the browser check is only for fast feedback) and, only if every row is clean, created as one all-or-nothing batch that publishes the site exactly once. A partially-applied batch, or one that silently skips bad rows, would leave the site's redirects in a confusing state.

This layer adds the two tRPC endpoints on top of the parsing/validation foundation in #2805. There is **no UI yet** — the modal that calls these lands in #2778.

## Solution

Adds `redirect.bulkValidate` and `redirect.bulkCreate`, both sent over POST (the CSV body is too large for a GET query string).

**Breaking Changes**

**Features**:

- **`redirect.bulkValidate`** (read-gated): re-parses the CSV server-side and returns a file-level error plus per-row verdicts, covering:
  - format rules (shared `redirectRowSchema` from #2805) and malformed/stray-column rows;
  - a source listed more than once **within the file**;
  - a source already redirected on the live table;
  - a source that **shadows a live published page/folder**;
  - **loops of any length** across the combined graph of the uploaded batch + the live redirects table.
- **`redirect.bulkCreate`** (create / site-admin gated): revalidates server-side, resolves internal destinations to `[resource:…]` references, then in a single transaction batch-upserts every row (reviving soft-deleted rows for reused sources), writes one `RedirectCreate` audit per row plus a single `Publish`, and rebuilds the site **once**. All-or-nothing: any error blocks the whole batch. If a concurrent change invalidates the batch between validate and insert, it returns fresh row verdicts instead of publishing.

**Improvements**:

- Loop detection follows each chain to its end or to a repeated node (visited-set), rather than capping at a fixed number of hops — so a long batch loop is caught rather than silently published.

**Bug Fixes**:

- None (new endpoints).

## Before & After Screenshots

N/A — backend endpoints, no UI in this layer. The rendered flow lives in #2778.

## Tests

Integration tests (`__tests__/bulkRedirects.router.test.ts`) run against a real Postgres:

- `bulkValidate`: 403 without read access, missing header column, malformed row, duplicate-within-file, source already on the table, shadowing a published page, loop within the file, loop against the live table, a >10-hop (any-length) cycle, and a fully-valid file.
- `bulkCreate`: 403 without create access, error → no publish, happy-path insert + exactly one publish, soft-deleted revive, and one `RedirectCreate` per row plus one `Publish`.

These endpoints are only reachable through the modal in #2778 (behind the `is-advanced-redirects-enabled` flag). Once the stack is deployed, verify this layer's behaviour via that flow:

- [ ] Upload a CSV whose rows form a loop (e.g. `/a → /b` and `/b → /a`) and confirm **both** rows are rejected and nothing is published.
- [ ] Upload a CSV with a source that already exists in the redirects table, and one that matches a live published page's URL — confirm both are flagged.
- [ ] Upload a valid CSV, publish, and confirm every row appears in the redirects table and the site republishes **once** (check the audit log shows one `RedirectCreate` per row and a single `Publish`).
- [ ] Re-upload a CSV reusing a source that was previously deleted, and confirm the redirect is revived with the new destination rather than erroring.

**New scripts**:

- None.

**New dependencies**:

- None.

**New dev dependencies**:

- None.

This PR adds two tRPC endpoints — `redirect.bulkValidate` and `redirect.bulkCreate` — that let editors upload a CSV of redirects and publish them as a single all-or-nothing batch. It addresses multiple race conditions raised in previous review rounds: the per-site `pg_advisory_xact_lock` serialises all redirect writes; an in-transaction shadow check catches pages published between validation and insert; and the combined-graph loop check runs post-insert inside the transaction so a concurrently committed redirect cannot close a loop.

- **`bulkValidate`** parses the CSV server-side, then checks format, in-file duplicate sources, already-live redirects, published-page shadows, and loops (including `[resource:...]` reference destinations) in a single read pass.
- **`bulkCreate`** re-validates, resolves destinations to `[resource:...]` references, acquires the per-site advisory lock, upserts rows (reviving soft-deleted ones), re-enforces the shadow and loop guards inside the transaction, writes one `RedirectCreate` audit per row and a single `Publish`, then rebuilds the site once.
- **`resource.service.ts`** receives a performance fix: `getResourceIdsByPermalinks` now builds a `Map<string,string>` index for O(1) segment lookup and chunks the `WHERE permalink IN (...)` query to stay under Postgres' 65 535 bind-parameter cap.
</details>

Safe to merge. The batch write path is all-or-nothing, correctly serialised with a per-site advisory lock, and all previously raised race windows (page shadow, loop after concurrent insert, shortfall on conflicting live redirect) are re-checked inside the transaction.

Every race condition surfaced in the previous review rounds has been addressed: the per-site advisory lock is acquired by both createRedirect and bulkCreateRedirects; the in-transaction shadow guard catches pages published between validation and insert; the combined-graph loop check runs after the insert so a concurrently committed redirect that would close a cycle is detected before commit. The resource.service.ts optimisation is correct and preserves the original find semantics. Integration tests cover the major failure modes including the page-shadow race. No functional regressions were found.

</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/redirect/redirect.service.ts | Core bulk logic: adds runBulkValidation, bulkValidateRedirects, and bulkCreateRedirects with per-site advisory lock, in-transaction shadow/loop rechecks, chunked upsert, and audit logging. Also adds the lock to createRedirect. All previously flagged race conditions are addressed. |
| apps/studio/src/server/modules/redirect/redirect.router.ts | Adds bulkValidate (read-gated, POST mutation) and bulkCreate (create-gated, POST mutation) endpoints; permission checks and input schemas look correct. |
| apps/studio/src/server/modules/redirect/__tests__/bulkRedirects.router.test.ts | Integration tests cover 403 guards, file errors, per-row validation (malformed, duplicate, on-table, shadow, loops of all lengths, reference destinations), happy-path insert + single publish, soft-delete revive, audit log counts, and the page-shadow race condition. |
| apps/studio/src/server/modules/resource/resource.service.ts | getResourceIdsByPermalinks optimised: candidate lookup chunked to avoid the 65 535 bind-parameter cap, and Array.find replaced with a Map index for O(1) per-segment lookups. Map key construction is injective for numeric parentIds. |
| apps/studio/src/lib/redirectCsv.ts | New isomorphic CSV parser: BOM stripping, blank-line-aware row numbering, header matching by name, malformed-row detection, and errors-file builder. Correctly handles re-uploaded error files with an extra Error column. |
| apps/studio/src/schemas/redirect.ts | Adds bulkRedirectsCsvSchema with a dual code-unit / UTF-8-byte cap and redirectRowSchema (shared by form and CSV), plus normalizeDestinationScheme and refineSourceDestinationDiffer helpers. |
| apps/studio/src/lib/redirectCsv.test.ts | Unit tests for parseRedirectCsv and buildRedirectErrorsCsv: covers BOM, header variants, blank lines, malformed rows, empty file, template round-trip. |
| apps/studio/src/server/modules/database/constants.ts | Adds lockTimeout (55P03) to the PG_ERROR_CODES map for the advisory-lock timeout handling. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant Router
    participant Service
    participant DB

    Note over Client,DB: bulkValidate (read-only)
    Client->>Router: "bulkValidate({ siteId, csv })"
    Router->>Service: runBulkValidation(siteId, csv)
    Service->>Service: parseRedirectCsv → per-row Zod validation
    Service->>DB: SELECT live redirects (source check)
    Service->>DB: getResourceIdsByPermalinks (shadow check)
    Service->>DB: resolveStoredDestinationsToSources (loop graph)
    Service->>Service: findCycleNodes → flag loop rows
    Service-->>Client: BulkRedirectValidationResult

    Note over Client,DB: bulkCreate (write, all-or-nothing)
    Client->>Router: "bulkCreate({ siteId, csv })"
    Router->>Service: bulkCreateRedirects(siteId, csv, userId)
    Service->>Service: runBulkValidation (pre-flight)
    alt any error
        Service-->>Client: ok false + validation
    end
    Service->>DB: resolveDestinationsForStorage (batch)
    Service->>DB: BEGIN TRANSACTION
    Service->>DB: "SET LOCAL lock_timeout = 15s"
    Service->>DB: pg_advisory_xact_lock(RD, siteId)
    Service->>DB: SELECT existing rows (audit before-state)
    Service->>DB: shadow recheck (getResourceIdsByPermalinks)
    alt shadow race
        Service->>DB: ROLLBACK
        Service-->>Client: ok false + fresh validation
    end
    Service->>DB: INSERT ON CONFLICT upsert (chunked)
    alt shortfall race
        Service->>DB: ROLLBACK
        Service-->>Client: ok false + fresh validation
    end
    Service->>DB: SELECT liveAfterInsert via tx
    Service->>Service: findCycleNodes (loop recheck)
    alt loop race
        Service->>DB: ROLLBACK
        Service-->>Client: ok false + fresh validation
    end
    Service->>DB: INSERT AuditLog RedirectCreate x N
    Service->>DB: INSERT AuditLog Publish x 1
    Service->>DB: COMMIT (releases advisory lock)
    Service->>DB: publishSite (CodeBuild, outside txn)
    Service-->>Client: ok true + publishedCount
```
</details>

<sub>Reviews (10): Last reviewed commit: ["fix(redirect): address review — chunk IN..."](https://github.com/opengovsg/isomer/commit/a0eb60d144d2b8d64a652a785b454c2b2b41fa59) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44385295)</sub>

### DIFF
--- a/apps/studio/src/server/modules/redirect/redirect.router.ts
+++ b/apps/studio/src/server/modules/redirect/redirect.router.ts
@@ -1,4 +1,5 @@
 import {
+  bulkRedirectsCsvSchema,
   countRedirectsByDestinationSchema,
   countRedirectsSchema,
   createRedirectSchema,
@@ -11,6 +12,8 @@ import { protectedProcedure, router } from "~/server/trpc"
 
 import { validateUserPermissionsForSite } from "../site/site.service"
 import {
+  bulkCreateRedirects,
+  bulkValidateRedirects,
   countRedirects,
   countRedirectsPointingToResource,
   createRedirect,
@@ -103,6 +106,42 @@ export const redirectRouter = router({
       })
 
       return countRedirectsPointingToResource(input)
+    }),
+
+  // Validates a whole uploaded CSV without writing anything, so the bulk-upload
+  // modal can show row errors or a ready-to-publish preview. A mutation (not a
+  // query) purely so the CSV rides in the POST body — a query serialises its
+  // input into the request URL, which the server/proxy rejects once the file
+  // nears the size cap (a connection reset). Still read-only and gated on "read"
+  // like the single-redirect `validate` preflight above.
+  bulkValidate: protectedProcedure
+    .input(bulkRedirectsCsvSchema)
+    .mutation(async ({ ctx, input }) => {
+      await validateUserPermissionsForSite({
+        siteId: input.siteId,
+        userId: ctx.user.id,
+        action: "read",
+      })
+
+      return bulkValidateRedirects(input)
+    }),
+
+  // Commits a validated batch: re-validates server-side, inserts every row in
+  // one transaction, and publishes once. Site-admin only like `create`.
+  bulkCreate: protectedProcedure
+    .input(bulkRedirectsCsvSchema)
+    .mutation(async ({ ctx, input }) => {
+      await validateUserPermissionsForSite({
+        siteId: input.siteId,
+        userId: ctx.user.id,
+        action: "create",
+      })
+
+      return bulkCreateRedirects({
+        ...input,
+        byUserId: ctx.user.id,
+        logger: ctx.logger,
+      })
     }),
 
   // create/delete publish immediately (no separate publish step). Site-wide
--- a/apps/studio/src/server/modules/redirect/redirect.service.ts
+++ b/apps/studio/src/server/modules/redirect/redirect.service.ts
@@ -20,7 +20,13 @@ import {
 } from "@opengovsg/isomer-components"
 import { TRPCError } from "@trpc/server"
 import { sql } from "kysely"
-import { REDIRECT_MESSAGES, RedirectValidationCode } from "~/constants/redirect"
+import { chunk } from "lodash-es"
+import {
+  BULK_DUPLICATE_SOURCE_MESSAGE,
+  BULK_MALFORMED_ROW_MESSAGE,
+  REDIRECT_MESSAGES,
+  RedirectValidationCode,
+} from "~/constants/redirect"
 import { parseRedirectCsv } from "~/lib/redirectCsv"
 import {
   isValidExternalDestination,
@@ -596,22 +602,12 @@ const BULK_REDIRECT_INSERT_CHUNK_SIZE = 500
 // automatically when the transaction ends.
 const REDIRECT_WRITE_LOCK_NAMESPACE = 0x5244 // "RD"
 
-// Shown when a row is split by an unquoted comma into extra columns.
-const BULK_MALFORMED_ROW_MESSAGE =
-  "This row has extra columns. Put double quotes around any value that contains a comma."
-
 // Thrown inside the bulk insert when a source stopped being publishable between
 // validation and the insert — either it gained a live redirect (the upsert
 // skipped it, a shortfall) or a page was published at it (the in-txn recheck).
 // bulkCreate catches it to re-validate and return fresh row verdicts rather than
 // surface a generic failure.
 class BulkRedirectRaceError extends Error {}
-
-// Cross-file duplicate copy. The other row errors reuse the shared Studio
-// messages (REDIRECT_MESSAGES) so the errors file reads the same as the inline
-// form errors, per the design.
-const BULK_DUPLICATE_SOURCE_MESSAGE =
-  "This source is listed more than once in your file."
 
 // One row after evaluation: the values as the editor typed them (for the errors
 // file + preview), the normalized values (for the DB checks + insert), and the
@@ -693,6 +689,10 @@ const resolveDestinationsForStorage = async (
   destinations: string[],
 ): Promise<Map<string, string>> => {
   const result = new Map<string, string>()
+  // First pass resolves everything storable as-is (query/fragment paths,
+  // external URLs, existing references) and sets those directly, collecting the
+  // internal paths that still need a resource lookup. parseDestination returns
+  // value === destination for an internal path, so the path doubles as the key.
   const internalPaths: string[] = []
   for (const destination of new Set(destinations)) {
     const parsed = parseDestination(destination)
@@ -707,18 +707,18 @@ const resolveDestinationsForStorage = async (
     internalPaths.push(parsed.value)
   }
 
+  // Second pass resolves only those internal paths: each becomes a [resource:...]
+  // reference when a resource lives at that permalink, or stays literal when none
+  // does (a typo or an as-yet-uncreated page — it just won't follow moves).
   const idByPath = internalPaths.length
     ? await getResourceIdsByPermalinks(siteId, internalPaths)
     : new Map<string, number | null>()
-  for (const destination of new Set(destinations)) {
-    if (result.has(destination)) {
-      continue
-    }
-    const resourceId = idByPath.get(destination) ?? null
+  for (const path of internalPaths) {
+    const resourceId = idByPath.get(path) ?? null
     result.set(
-      destination,
+      path,
       resourceId === null
-        ? destination
+        ? path
         : getReferenceLink({
             siteId: String(siteId),
             resourceId: String(resourceId),
@@ -738,8 +738,7 @@ const resolveDestinationsForStorage = async (
 // leading INTO a cycle is (correctly) not included.
 const findCycleNodes = (edges: Map<string, string | null>): Set<string> => {
   const cycleNodes = new Set<string>()
-  // 0 = on the path currently being walked, 1 = fully explored.
-  const state = new Map<string, 0 | 1>()
+  const state = new Map<string, "walking" | "explored">()
   for (const start of edges.keys()) {
     if (state.has(start)) {
       continue
@@ -749,19 +748,26 @@ const findCycleNodes = (edges: Map<string, string | null>): Set<string> => {
     const path: string[] = []
     let current: string | null = start
     while (current !== null && !state.has(current)) {
-      state.set(current, 0)
+      state.set(current, "walking")
       path.push(current)
       current = edges.get(current) ?? null
     }
-    // Stopping on a node still marked "on the current path" closed a cycle; every
-    // node from it to the end of the path lies on that cycle.
-    if (current !== null && state.get(current) === 0) {
+    // The walk stops in one of three ways, and only the last is a cycle:
+    //   - current === null       → the chain terminates; no cycle.
+    //   - state is "explored"    → it merged into an already-resolved chain
+    //                              (e.g. a tail leading into someone else's
+    //                              path); no new cycle on this walk.
+    //   - state is "walking"     → we looped back onto the current path; every
+    //                              node from there to the end lies on the cycle.
+    // (So we can't assert current === null for the no-cycle case — the explored
+    // exit is equally valid.)
+    if (current !== null && state.get(current) === "walking") {
       for (const node of path.slice(path.indexOf(current))) {
         cycleNodes.add(node)
       }
     }
     for (const node of path) {
-      state.set(node, 1)
+      state.set(node, "explored")
     }
   }
   return cycleNodes
@@ -1076,18 +1082,13 @@ export const bulkCreateRedirects = async ({
       }
 
       const insertedRows: typeof existingRows = []
-      for (
-        let i = 0;
-        i < valuesToInsert.length;
-        i += BULK_REDIRECT_INSERT_CHUNK_SIZE
-      ) {
-        const chunk = valuesToInsert.slice(
-          i,
-          i + BULK_REDIRECT_INSERT_CHUNK_SIZE,
-        )
+      for (const batch of chunk(
+        valuesToInsert,
+        BULK_REDIRECT_INSERT_CHUNK_SIZE,
+      )) {
         const inserted = await tx
           .insertInto("Redirect")
-          .values(chunk)
+          .values(batch)
           .onConflict((oc) =>
             oc
               .columns(["siteId", "source"])
@@ -1160,15 +1161,8 @@ export const bulkCreateRedirects = async ({
         userId: byUser.id,
         metadata: {},
       }))
-      for (
-        let i = 0;
-        i < auditValues.length;
-        i += BULK_REDIRECT_INSERT_CHUNK_SIZE
-      ) {
-        await tx
-          .insertInto("AuditLog")
-          .values(auditValues.slice(i, i + BULK_REDIRECT_INSERT_CHUNK_SIZE))
-          .execute()
+      for (const batch of chunk(auditValues, BULK_REDIRECT_INSERT_CHUNK_SIZE)) {
+        await tx.insertInto("AuditLog").values(batch).execute()
       }
 
       // The whole batch republishes once, audited as a single Publish event.
@@ -1194,7 +1188,18 @@ export const bulkCreateRedirects = async ({
       // shows the offending row(s) rather than a generic failure — this keeps
       // the ok:false contract.
       const { fileError, rows } = await runBulkValidation(siteId, csv)
-      return { ok: false, validation: toValidationResult(fileError, rows) }
+      const validation = toValidationResult(fileError, rows)
+      // The race may have cleared by the time we re-validate (e.g. the
+      // conflicting redirect was deleted). With nothing left to flag, ok:false
+      // would strand the modal on an errors screen listing no rows, so surface a
+      // transient conflict instead — the client prompts a retry, which succeeds.
+      if (validation.fileError === null && validation.errorCount === 0) {
+        throw new TRPCError({
+          code: "CONFLICT",
+          message: "The site changed while publishing. Please try again.",
+        })
+      }
+      return { ok: false, validation }
     }
     throw error
   }
--- a/apps/studio/src/server/modules/redirect/redirect.service.ts
+++ b/apps/studio/src/server/modules/redirect/redirect.service.ts
@@ -3,6 +3,7 @@ import type {
   RedirectValidationResult,
 } from "~/constants/redirect"
 import type {
+  BulkRedirectsCsvInput,
   CountRedirectsByDestinationInput,
   CountRedirectsInput,
   CreateRedirectInput,
@@ -20,10 +21,12 @@ import {
 import { TRPCError } from "@trpc/server"
 import { sql } from "kysely"
 import { REDIRECT_MESSAGES, RedirectValidationCode } from "~/constants/redirect"
+import { parseRedirectCsv } from "~/lib/redirectCsv"
 import {
   isValidExternalDestination,
   normalizeRedirectPath,
   normalizeRedirectSource,
+  redirectRowSchema,
 } from "~/schemas/redirect"
 import { getReferenceLink } from "~/utils/link"
 import { ResourceType } from "~prisma/generated/generatedEnums"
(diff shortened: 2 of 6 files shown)


dcshzj:
* **#2778** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2806** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2806?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2805** <a href="https://app.graphite.com/github/pr/opengovsg/isomer/2805?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,510 @@ export const createRedirect = async ({
...
+  logger,
+}: BulkRedirectsCsvInput & {
+  byUserId: string
+  logger: Logger<string>
+}): Promise<BulkCreateRedirectsResult> => {
+  const byUser = await getByUser(byUserId)
+
+  const { fileError, rows } = await runBulkValidation(siteId, csv)
+  if (fileError !== null || rows.some((row) => row.error !== null)) {
+    return { ok: false, validation: toValidationResult(fileError, rows) }
```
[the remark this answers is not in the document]
dcshzj: Fixed on this branch. The source-vs-published-page guard is now re-checked inside the write transaction, right before the insert (mirroring `createRedirect`): it resolves the batch's sources to resource ids and, if any now maps to a live published page/container, throws `BulkRedirectRaceError`. `bulkCreate` catches that on the existing race path, re-validates against the current table, and returns `ok: false` with the offending row(s) flagged `sourceIsExistingPage` — so a redirect that would shadow a page published between validation and the insert is never committed or published.

Added an integration test that simulates the race (validation is made to miss the page; the in-transaction recheck catches it) and asserts nothing publishes and the row comes back flagged.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,538 @@ export const createRedirect = async ({
...
+      return true
+    }
+    if (visited.has(current)) {
+      return false
+    }
+    visited.add(current)
+    current = edges.get(current) ?? null
+  }
+  return false
+}
```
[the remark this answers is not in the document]
dcshzj: !fixed
Replaced the per-row `formsLoop` (a fresh chain walk from every source — O(N^2) for a long chain) with a single pass that finds every node on a cycle in the functional redirect graph and flags rows whose source is one, i.e. O(total edges). Behaviour is identical (a source is flagged iff its chain returns to itself; a tail leading into a loop stays unflagged), so a long chained upload no longer blocks the event loop. Added a tail-into-cycle test guarding that equivalence.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,538 @@ export const createRedirect = async ({
...
+      // ok:false contract) rather than committing a shadowing redirect.
+      if (sources.length > 0) {
+        const idByPermalink = await getResourceIdsByPermalinks(siteId, sources)
+        const pageResourceIds = [...idByPermalink.values()].filter(
+          (id): id is number => id !== null,
+        )
+        const publishedState = await getPublishedStateByResourceIds(
+          siteId,
+          pageResourceIds,
+        )
```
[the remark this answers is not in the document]
dcshzj: Assessed — leaving this as a known, recoverable race rather than changing it:

1. It matches the single-create path. `createRedirect` runs the same source-vs-published-page recheck inside its transaction via `db` too (its comment calls it "a plain read, so [an] accepted race"). Bulk now mirrors that exactly.

2. Switching these reads to `tx` would not actually close the race. `db.transaction()` runs at Postgres' default READ COMMITTED, where each statement sees data committed as of statement start — so a page publish that commits after the recheck read but before the insert commit is invisible whether the read used `tx` or `db`. Passing `tx` would look like a fix without changing behaviour. Genuinely closing it needs SERIALIZABLE + retry, or explicit `SELECT ... FOR UPDATE` locks on the resources — which the codebase deliberately doesn't do for redirects (a shadowing redirect is recoverable by deleting it), and locking still wouldn't cover a page newly created at the path concurrently.

The prior commit already narrowed the window by rechecking immediately before the insert (vs only at validation). Happy to implement serializable+retry if the team wants to fully close it, but that's a deliberate design change beyond parity with the existing create path.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,554 @@ export const createRedirect = async ({
...
+          ? [row.normalizedSource]
+          : [],
+      ),
+    ),
+  ]
+  if (sourcesToCheck.length > 0) {
+    const idByPermalink = await getResourceIdsByPermalinks(
+      siteId,
+      sourcesToCheck,
+    )
```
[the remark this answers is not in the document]
dcshzj: !fixed
`getResourceIdsByPermalinks` still reads in two round-trips, but its in-memory resolution walked each path segment with a linear `candidates.find` — O(segments * candidates). Indexed the candidates into a Map keyed by (parentId, permalink) so each segment step is an O(1) lookup, making resolution O(total segments). Behaviour is unchanged (first-match semantics preserved) and it benefits every caller of the helper, not just the bulk path. Verified against the full redirect suite (103 tests).

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,538 @@ export const createRedirect = async ({
...
+                // Revived rows republish now, so refresh createdAt (the publish
+                // time shown to users).
+                createdAt: dbNow,
+              }))
+              // Only soft-deleted rows may be revived; a live one must not be
+              // silently overwritten.
+              .where("Redirect.deletedAt", "is not", null),
+          )
+          .returningAll()
+          .execute()
```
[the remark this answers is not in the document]
dcshzj: Fixed. `bulkCreate` now rechecks the combined-graph loop guard inside the transaction, right after the insert: it reads the table via `tx` (so it sees the just-inserted batch plus any redirect committed concurrently), rebuilds the source -> next-source graph, and if a batch source now lies on a cycle it aborts via `BulkRedirectRaceError` — the same path that re-validates and returns fresh row verdicts. So a concurrent `/b -> /a` landing after a batch `/a -> /b` was validated no longer publishes a loop. Mirrors `createRedirect`'s in-transaction loop guard, and reuses the single-pass cycle detector from the earlier O(N) loop fix.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,584 @@ export const createRedirect = async ({
...
+  }
+  for (const row of rows) {
+    if (row.error !== null || row.normalizedSource === null) {
+      continue
+    }
+    const destination = row.normalizedDestination
+    const nextSource =
+      destination !== null && destination.startsWith("/")
+        ? normalizeRedirectSource(destination)
+        : null
```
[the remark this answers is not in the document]
dcshzj: Fixed. Bulk validation now resolves the batch's destinations via `resolveStoredDestinationsToSources` — the same path existing stored redirects go through — instead of the old path-only check, so a `[resource:...]` destination contributes the correct edge to the loop graph. That means a reference-formed loop is caught at validation (flagged as a row error) and validation now agrees with the in-transaction recheck, so the `ok: false` with zero row errors you spotted can't happen. Added a test that a `/a -> [resource:...page at /b]` + `/b -> /a` batch is flagged as a loop on both rows.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,596 @@ export const createRedirect = async ({
...
+      // published-page rechecks don't catch that. Read the table via tx (so it
+      // sees this batch and any committed concurrent redirect), rebuild the
+      // combined graph, and abort if a batch source now lies on a cycle (mirrors
+      // createRedirect's in-transaction loop guard).
+      const liveAfterInsert = await tx
+        .selectFrom("Redirect")
+        .select(["source", "destination"])
+        .where("siteId", "=", siteId)
+        .where("deletedAt", "is", null)
+        .execute()
```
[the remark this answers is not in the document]
dcshzj: Fixed. `bulkCreate` now takes a per-site `pg_advisory_xact_lock` as the first statement in its transaction, so concurrent bulk creates on the same site serialize: the second waits until the first commits, then its in-transaction loop recheck sees the committed rows and aborts via `BulkRedirectRaceError`. The `/a -> /b` + `/b -> /a` interleaving you described can't both commit anymore.

Scope note: this serializes bulk-vs-bulk (the raised case). A single `createRedirect` racing a bulk isn't in the same lock yet — a separate, recoverable case I can extend `createRedirect` to cover if you'd like.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,596 @@ export const createRedirect = async ({
...
+        await logRedirectEvent(tx, {
+          siteId,
+          by: byUser,
+          eventType: AuditLogEvent.RedirectCreate,
+          delta: {
+            before: existingBySource.get(row.source) ?? null,
+            after: row,
+          },
+        })
+      }
```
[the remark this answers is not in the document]
dcshzj: !fixed
The per-row RedirectCreate audit inserts are now batched in the same 500-row chunks as the redirect insert, instead of one sequential INSERT per row. A ~15k-row upload now runs ~30 audit inserts instead of 15k, cutting the round-trips (and the time locks and the connection are held) inside the transaction. The chunk size also keeps each statement under Postgres' bind-parameter limit.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +571,626 @@ export const createRedirect = async ({
...
+    const created = await db.transaction().execute(async (tx) => {
+      // Serialise redirect writes for this site: two concurrent bulk creates
+      // could otherwise each miss the other's uncommitted rows in the rechecks
+      // below (READ COMMITTED doesn't see them) and both publish — e.g. one
+      // inserting /a -> /b while the other inserts /b -> /a, forming a loop. The
+      // xact lock makes the later transaction wait, so its recheck sees the
+      // committed rows and aborts. Released automatically at commit/rollback.
+      await sql`SELECT pg_advisory_xact_lock(${REDIRECT_WRITE_LOCK_NAMESPACE}, ${siteId})`.execute(
+        tx,
+      )
```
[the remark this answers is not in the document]
dcshzj: Fixed. `createRedirect` now acquires the same per-site `pg_advisory_xact_lock` as `bulkCreateRedirects`, as the first statement in its transaction (before the loop/shadow guards and the insert). So a single create and a bulk create on the same site serialise: whichever runs second sees the other's committed row in its loop recheck and is blocked/rejected, closing the `/a -> /b` + `/b -> /a` interleaving across the two paths. (The server-derived page-move redirect path runs inside the publish transaction and isn't a user-controlled writer, so it's out of scope here.)

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1083,6 +1083,21 @@ export const getResourceIdsByPermalinks = async (
...
 
+  // Index candidates by (parentId, permalink) so each segment step is an O(1)
+  // lookup. A linear `candidates.find` per segment is O(segments * candidates),
+  // which a large bulk upload against a resource-heavy site pushes into hundreds
+  // of millions of comparisons — enough to block the event loop. The " "
+  // separator is safe: a parentId is only digits, so the concatenation is
+  // injective (the first space always delimits parentId from permalink).
+  const idByParentAndPermalink = new Map<string, string>()
+  for (const candidate of candidates) {
+    const key = `${candidate.parentId ?? ""} ${candidate.permalink}`
```
seaerchin: comment: in my editor i see a null byte between these 2? not sure if intended

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+// statement at 65535 bind parameters, so the chunk keeps every batched insert
+// (redirects and their audit rows) well under that.
+const BULK_REDIRECT_INSERT_CHUNK_SIZE = 500
+
+// Arbitrary namespace for the per-site advisory lock that serialises redirect
+// writes (single create + bulk create) so their loop/shadow rechecks can't race
+// each other. There is no other advisory-lock user, but the two-key form keeps
+// this from ever colliding with a future one; pg_advisory_xact_lock releases
+// automatically when the transaction ends.
+const REDIRECT_WRITE_LOCK_NAMESPACE = 0x5244 // "RD"
```
seaerchin, line 615: question: do you mind elaborating on how this works? it seems quite clean so was thinking we could reuse this

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+// file + preview), the normalized values (for the DB checks + insert), and the
+// single chosen error (null when the row passed).
+interface EvaluatedRedirectRow {
+  rowNumber: number
+  source: string
+  destination: string
+  normalizedSource: string | null
+  normalizedDestination: string | null
+  error: string | null
+}
```
seaerchin, line 663: comment: should we shift these out to `redirect.types.ts` and `constants.ts`? ok w/ not doing if only used here

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+  destinations: string[],
+): Promise<Map<string, string>> => {
+  const result = new Map<string, string>()
+  const internalPaths: string[] = []
+  for (const destination of new Set(destinations)) {
+    const parsed = parseDestination(destination)
+    if (
+      parsed.type !== "internalPath" ||
+      parsed.value.includes("?") ||
+      parsed.value.includes("#")
```
seaerchin, line 749: question: why not just check for presence of a protocol? (eg: https/mailto etc)

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+      continue
+    }
+    const resourceId = idByPath.get(destination) ?? null
+    result.set(
+      destination,
+      resourceId === null
+        ? destination
+        : getReferenceLink({
+            siteId: String(siteId),
+            resourceId: String(resourceId),
```
seaerchin, line 771: question: how come we set the value as `destination` when `resourceId === null`? this should resolve nicely since we've already filtered the external links above, no? ie, this only happens when the resource has been deleted?

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+    internalPaths.push(parsed.value)
+  }
+
+  const idByPath = internalPaths.length
+    ? await getResourceIdsByPermalinks(siteId, internalPaths)
+    : new Map<string, number | null>()
+  for (const destination of new Set(destinations)) {
+    if (result.has(destination)) {
+      continue
+    }
```
seaerchin: question: this is just defence in depth right since we're iterating over a `Set`, this should be impossible?

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+// graph — O(total edges) — instead of re-walking the chain from each row, which
+// is O(N^2) and lets a long uploaded chain (up to ~100k rows within the 1MB cap)
+// block the single-threaded event loop. A node is on a cycle iff following its
+// chain returns to it, so membership in the returned set is exactly the old
+// per-row "chain from `start` returns to `start`" check — a source on the tail
+// leading INTO a cycle is (correctly) not included.
+const findCycleNodes = (edges: Map<string, string | null>): Set<string> => {
+  const cycleNodes = new Set<string>()
+  // 0 = on the path currently being walked, 1 = fully explored.
+  const state = new Map<string, 0 | 1>()
```
seaerchin: comment: either do `true/false` or give a string literal type ('EXPLORED' | 'WALKING')

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+    // Stopping on a node still marked "on the current path" closed a cycle; every
+    // node from it to the end of the path lies on that cycle.
+    if (current !== null && state.get(current) === 0) {
+      for (const node of path.slice(path.indexOf(current))) {
+        cycleNodes.add(node)
+      }
+    }
+    for (const node of path) {
+      state.set(node, 1)
+    }
```
seaerchin, line 818: comment: can we make this stricter? in order for it to have a fully formed path and not a cycle, the `current === null` has to be true so we should assert that condition.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+
+// Finds every node that lies on a cycle in the functional redirect graph `edges`
+// (each source maps to at most one next source). One traversal over the whole
+// graph — O(total edges) — instead of re-walking the chain from each row, which
+// is O(N^2) and lets a long uploaded chain (up to ~100k rows within the 1MB cap)
+// block the single-threaded event loop. A node is on a cycle iff following its
+// chain returns to it, so membership in the returned set is exactly the old
+// per-row "chain from `start` returns to `start`" check — a source on the tail
+// leading INTO a cycle is (correctly) not included.
+const findCycleNodes = (edges: Map<string, string | null>): Set<string> => {
```
seaerchin, line 786: clarification: just for my understanding, we walk each path to completion (following the edges passed in); at completion, we just check if the exit is a terminal node or if we've seen it previously.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+          throw new BulkRedirectRaceError()
+        }
+      }
+
+      const insertedRows: typeof existingRows = []
+      for (
+        let i = 0;
+        i < valuesToInsert.length;
+        i += BULK_REDIRECT_INSERT_CHUNK_SIZE
+      ) {
```
seaerchin: suggestion: could we ocnsider using lodash.chunk? https://lodash.com/docs/4.18.1 seems to be what we are doing here and easier to understand

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+      const graphAfter = new Map<string, string | null>()
+      for (const redirect of liveAfterInsert) {
+        graphAfter.set(
+          redirect.source,
+          resolvedAfter.get(redirect.destination) ?? null,
+        )
+      }
+      const cycleNodes = findCycleNodes(graphAfter)
+      if (sources.some((source) => cycleNodes.has(source))) {
+        throw new BulkRedirectRaceError()
```
seaerchin, line 1204: comment: i suggest creating a new `BulkRedirectCycleError` and in the constructor, taking the nodes that make the cycle. then we don't have to recompute + can match on the error

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -568,6 +579,627 @@ export const createRedirect = async ({
...
+          .execute()
+        insertedRows.push(...inserted)
+      }
+
+      // One returned row per input is expected (new rows insert, soft-deleted rows
+      // revive). A shortfall means a source went live between validation and the
+      // insert (the upsert skipped it) — abort the txn so a partial batch never
+      // publishes; bulkCreate catches this and re-validates for fresh row errors.
+      if (insertedRows.length !== valuesToInsert.length) {
+        throw new BulkRedirectRaceError()
```
seaerchin, line 1174: comment: suggest providing somemore info to this error so that we can report errors

apps/studio/src/server/modules/redirect/redirect.service.ts
seaerchin: just fyi, a node that points to a node inside a cycle but itself is not part of the cycle will be marked as OK. 

i'm ok w/ this because once the cycle is fixed, all subsequent issues deriving from the cycle will also be fixed but wondering if we should report this back to the user or nah. 

eg: a -> b -> c is a cycle and will be reported; d -> a will lead into the cycle but will **not** be reported

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: good catch, that was an actual NUL byte (`0x00`), not intentional. there were three of them: this separator, the matching one in the lookup below, and one that had snuck into the comment. swapped them all for `/`. a `parentId` is only digits, so the first `/` still delimits it from the permalink, and i fixed the comment to match.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: sure. `pg_advisory_xact_lock(k1, k2)` grabs a transaction-scoped lock on the pair `(0x5244, siteId)`. any other txn that calls it with the same pair blocks until this one commits or rolls back, then it releases on its own. so it just serialises redirect writes per site: the loop/shadow rechecks see a stable table because no other writer can slip in between. the two keys are only there so it can't clash with some future advisory lock elsewhere in the app.

if there's somewhere that wants the same guard i can pull it into a helper like `acquireRedirectWriteLock(tx, siteId)`. no second caller yet so i left it inline, but easy to extract.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: moved the two bulk-only message strings (malformed row, duplicate source) into `constants/redirect.ts` next to `REDIRECT_MESSAGES`, since that's where the rest of the redirect copy sits. left the interfaces here though: they're only used in this file (the client sees them through `RouterOutput` inference, not a direct import), so a types file would just be indirection for now.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: `parseDestination` is already doing the classifying. it splits a destination into reference / external / internalPath, same as single-create uses. a protocol check would only tell external URLs from paths. it wouldn't skip an existing `[resource:...]` reference, and it wouldn't handle the `?`/`#` case right below (an internal path with a query or fragment has to stay literal so it doesn't follow page moves). so the type check plus the `?`/`#` guard is catching stuff a protocol sniff can't.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: not deletion, no. `null` means nothing resolves to that permalink at all right now: a typo, or a page that was never created (or isn't indexed). those stay as the literal path, same as single-create, where a destination like that just doesn't follow future moves. added a comment on the second pass saying so.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: it's not actually defensive. the first pass sets the literal destinations (external, query/fragment) and the guard was skipping those on the second pass, so only the internal paths were left to resolve. but yeah, it reads like the guard should be dead code, so i reworked it: first pass collects the internal paths, second pass loops over just those. no re-walking the whole `Set`, no `has()` check. safe because `parseDestination` returns `value === destination` for internal paths, so the path is the key.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: done, it's `Map<string, "walking" | "explored">` now, so the `0`/`1` legend comment went away too.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: two things here:

on asserting `current === null`: don't think we can, it's not the only non-cycle exit. a walk can also stop because it hit an already-`explored` node, e.g. `d -> a` where `a` got fully resolved on an earlier walk. not a cycle, but `current` is `a` (not null) there, so the assert would throw on valid input. instead i wrote the three exit cases out in a comment above the check (`null` = terminal, `explored` = merged into a resolved chain, `walking` = cycle).

on the tail-into-cycle thing (`d -> a` where `a -> b -> c -> a` loops): right, `d` doesn't get flagged, and that's on purpose. it keeps the loop semantics single-create already has, and once you break the cycle the tail sorts itself out. decided to leave it rather than report tail nodes, so bulk stays consistent with single-create and we don't over-report.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: yep, that's it. walk forward from each unvisited node, mark them `walking` and push onto the current path, until you hit `null`, an already-`explored` node, or a node still marked `walking`. only that last one's a cycle, and the path from that node onward is the cycle. after the walk everything on the path gets marked `explored` so we don't re-walk it.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: done, this loop and the audit-insert loop below both use `chunk` from `lodash-es` now (already used elsewhere, e.g. `features/settings/utils.ts`).

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: looked at this but i think re-validating is the better path. when any of these rechecks abort, the catch re-runs `runBulkValidation` against the current table and hands back per-row errors that match the inline form, so the client already gets exact row-level errors without us mapping cycle nodes back to rows and re-deriving messages. putting the nodes on the error gives us a second place errors get derived, for a rare race, and the recompute it saves is tiny on that path.

what i did add is the case your comment is really getting at: if the race clears before we re-validate (say the conflicting redirect got deleted), re-validation comes back clean and we'd return `ok: false` with zero rows, i.e. an errors screen listing nothing. that now throws `CONFLICT` instead and the modal asks for a retry, which works. lmk if you'd still rather have the typed error.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: same as the cycle recheck below: the catch re-validates against the current table and returns fresh per-row errors, so a shortfall shows up as the specific offending row(s) in the errors file rather than a generic fail. i'd keep that single error-derivation path instead of threading detail onto the error. did add a guard for when the race has already cleared by the time we re-validate: that returns a transient `CONFLICT` ("please try again") instead of an empty errors screen.

apps/studio/src/server/modules/redirect/redirect.service.ts
seaerchin: can't tell if this comment is AI wtf

seaerchin: question: should we have a timeout? wary of deadlocks in this situation since the second writer might mindlessly wait for the first one to commit (which might never happen)

happy to opt for a 15s timeout and abort

dcshzj: Sure – I understand your confusion, and I can assure you that I am committed to helping you understand the intent behind this change. Would you like me to elaborate more?

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -456,6 +465,14 @@ export const createRedirect = async ({
   const byUser = await getByUser(byUserId)
 
   const created = await db.transaction().execute(async (tx) => {
+    // Serialise redirect writes for this site (the same lock bulkCreateRedirects
+    // takes), so a single create and a concurrent bulk create can't each miss
+    // the other's uncommitted row in the loop/shadow guards below and both
+    // publish a loop. Released automatically at commit/rollback.
+    await sql`SELECT pg_advisory_xact_lock(${REDIRECT_WRITE_LOCK_NAMESPACE}, ${siteId})`.execute(
+      tx,
```
seaerchin: fyi: the advisory locks uses the same mem pool as normal pg locks, which is defined by `max_locks_per_transaction` + `max_connections`
	 - once this pool is exhausted, the server will not grant any locks so we might wanna check how often users create the redirect but given that this is held on a `siteId` level, i don't think it's important

seaerchin (approved):

apps/studio/src/server/modules/redirect/redirect.router.ts
```diff
@@ -105,6 +108,42 @@ export const redirectRouter = router({
...
+  // nears the size cap (a connection reset). Still read-only and gated on "read"
+  // like the single-redirect `validate` preflight above.
+  bulkValidate: protectedProcedure
+    .input(bulkRedirectsCsvSchema)
+    .mutation(async ({ ctx, input }) => {
+      await validateUserPermissionsForSite({
+        siteId: input.siteId,
+        userId: ctx.user.id,
+        action: "read",
+      })
```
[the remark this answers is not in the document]
dcshzj: this isn't specific to the two new procedures. every procedure in `redirect.router.ts` (list, count, validate, resolveReferences, getBySource, countByDestinationResource, create, delete) already calls `validateUserPermissionsForSite` in the router, so matching `bulkValidate`/`bulkCreate` to the existing 8 keeps the module consistent. agree the checks would read better in the services, but that's a module-wide refactor on its own. doing 2 of 10 in the service and leaving 8 in the router is more confusing than what's there now, so i'd rather keep it uniform here and file the move as a separate follow-up.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -455,117 +465,795 @@ export const createRedirect = async ({
...
+  }
+  for (const row of rows) {
+    if (
+      row.error === null &&
+      row.normalizedSource !== null &&
+      (sourceCounts.get(row.normalizedSource) ?? 0) > 1
+    ) {
+      row.error = BULK_DUPLICATE_SOURCE_MESSAGE
+    }
+  }
```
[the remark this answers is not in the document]
dcshzj: this is intentional, and it can't actually let a duplicate through. only error-free rows go into `valuesToInsert`, and step 2 flags every source that appears more than once among those, so `valuesToInsert` never holds two rows with the same source. if one instance has a format error, the other valid row is created, but that's still a single live redirect for the source (and the `(siteId, source)` unique index plus the on-table recheck back that up). the gap is only first-pass feedback: the user sees the format error, fixes it, re-uploads, and then the duplicate / already-redirected check flags the rest. i'd rather not normalize the source separately for errored rows just to surface that one pass earlier, since it complicates the one-error-per-row precedence. can revisit if that first-pass UX turns out to matter.

dcshzj:
### Merge activity

* **Jul 23, 1:31 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806).
* **Jul 23, 1:31 AM UTC**: [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806) couldn't merge this pull request because a downstack PR #2805 failed to merge.
* **Jul 23, 1:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806).
* **Jul 23, 1:32 AM UTC**: [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806) couldn't merge this pull request because a downstack PR #2805 failed to merge.
* **Jul 23, 1:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806).
* **Jul 23, 1:32 AM UTC**: [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806) couldn't merge this pull request because a downstack PR #2805 failed to merge.
* **Jul 23, 1:45 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806).
* **Jul 23, 1:45 AM UTC**: [Graphite](https://app.graphite.com/github/pr/opengovsg/isomer/2806) couldn't merge this PR because it failed for an unknown reason.

PR #2887, opened 2026-07-22
🔖 fix(studio): enforce tagCategories admin gate server-side on page.updatePageBlob

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 2 | +58 | -1 |
| 🧪 Test | 1 | +190 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Collection Filters (`tagCategories`) were only gated in the UI. The drawer and JsonForms control hid the fields behind a client permission check, but `trpc.page.updatePageBlob` never checked on the server. It validated AJV shape and generic page-update permission, then wrote whatever you sent. Anyone who could update a page could POST a mutated `tagCategories` array and skip the UI gate. Hacktron flagged this as medium.

This stacks on #2865, which moves the client gate from Isomer Admin to site Admin (`canManageCollectionFilters`: CASL `can("create", { parentId: null })`). The server check here uses that same permission, not platform IsomerAdminRole, so it matches where #2865 is taking the feature.

## Solution

Added `assertTagCategoriesUnchangedForNonSiteAdmin` in `page.service.ts`. It diffs incoming `page.tagCategories` against the stored blob and only runs the permission check when they differ, so Editors and Publishers can still save ordinary content edits.

The permission check is the same CASL "create at site root" call the client uses: `bulkValidateUserPermissionsForResources({ action: "create", resourceIds: [null] })`. Site Admins and Isomer Admins (who get implicit Admin abilities) can change Filters. Everyone else gets `FORBIDDEN`.

Wired into `updatePageBlob` in `page.router.ts` after the old blob loads and before the write.

`page.tags` is legacy and unused, so we leave it alone.

**Breaking Changes**

**Bug Fixes**

- `page.updatePageBlob` now rejects `tagCategories` changes from non-site-Admins.

## Tests

Added `describe("updatePageBlob - tagCategories admin enforcement", ...)` in `page.router.test.ts`:

- 403s a non-admin site Editor who modifies `tagCategories`, and leaves the blob untouched.
- Lets a non-admin site Editor save when `tagCategories` is unchanged.
- Lets a site Admin modify `tagCategories`.
- Lets an Isomer Admin (Core/Migrator) with no explicit site permission modify `tagCategories`, matching the implicit-admin behavior #2865's e2e test relies on.

The first test fails on the pre-fix code (mutation succeeds instead of `FORBIDDEN`) and passes after the fix.

**Manual Verification Steps**

- [ ] As a site Editor, call `trpc.page.updatePageBlob` with a modified `tagCategories` on a collection index page. Expect `FORBIDDEN`.
- [ ] As a site Admin, create/edit/save Filters end-to-end.
- [ ] `pnpm test:unit -- src/server/modules/page/__tests__/page.router.test.ts`

**New scripts**: None
**New dependencies**: None

This PR adds a server-side gate for collection filter edits in `page.updatePageBlob`. The main changes are:

- Adds `assertTagCategoriesUnchangedForNonSiteAdmin` to compare old and new `page.tagCategories` values.
- Requires site-root `create` permission when collection `tagCategories` changes.
- Runs the guard before writing the updated draft blob.
- Adds tests for editor rejection, unchanged editor saves, site admin updates, Isomer admin updates, and non-collection page saves.
</details>

Safe to merge with low residual risk.

The guard runs before the blob write, uses the same root-create permission model as the client gate, preserves ordinary content edits, and includes focused tests for denied and allowed paths.

</details>

**What T-Rex did**
- The targeted unit test for Page Update Tagcategories was executed, and the log shows the full command, working directory, exit code, and the Testcontainers blocker stack.
- A fallback typecheck by tsgo --noEmit was attempted, but it terminated due to a timeout with exit code 124.
- A fallback lint run with type-aware oxlint on the changed files aborted with SIGABRT.
- A fallback lint run with non-type-aware oxlint on the changed files also aborted with SIGABRT.

<a href="https://app.greptile.com/trex/runs/15535446/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/page/page.service.ts | Adds a server-side `tagCategories` diff guard that requires site-root create permission before allowing collection filter changes. |
| apps/studio/src/server/modules/page/page.router.ts | Wires the `tagCategories` authorization guard into `updatePageBlob` after loading the existing blob and before persisting updates. |
| apps/studio/src/server/modules/page/__tests__/page.router.test.ts | Adds tests for non-admin rejection, unchanged saves, normal non-collection edits, site-admin updates, and Isomer-admin override behavior. |

</details>

```mermaid
sequenceDiagram
participant Caller
participant Router as page.updatePageBlob
participant DB as Blob/Resource DB
participant Guard as assertTagCategoriesUnchangedForNonSiteAdmin
participant Perms as bulkValidateUserPermissionsForResources

Caller->>Router: Submit updated page blob
Router->>Perms: Validate generic page update permission
Router->>DB: Load page resource and old blob
Router->>Guard: Compare old vs new page.tagCategories
alt Non-collection or tagCategories unchanged
    Guard-->>Router: Allow without extra check
else tagCategories changed
    Guard->>Perms: Validate create on site root resourceId null
    Perms-->>Guard: Allow site Admin / Isomer Admin or throw FORBIDDEN
    Guard-->>Router: Allow when authorized
end
Router->>DB: Update draft blob and write audit log
Router-->>Caller: Return input content
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;collection-tags-master&#39; in..."](https://github.com/opengovsg/isomer/commit/92fc70e78ee36094446e27ba76f0f2a2d9c8168e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46181392)</sub>


apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -65,6 +68,48 @@ export const createDefaultPage = ({
...
+ * Collection Filters (`tagCategories`) are UI-gated to site Admins via
+ * `canManageCollectionFilters`. Callers with page-update access can still hit
+ * `page.updatePageBlob` directly, so when `page.tagCategories` changes, run
+ * the same CASL check the client uses (`can("create", { parentId: null })`).
+ * Throws FORBIDDEN if the caller fails it.
+ *
+ * No-op for non-collection layouts. `tagCategories` only exists on the
+ * collection layout's `page` schema.
+ */
+export const assertTagCategoriesUnchangedForNonSiteAdmin = async ({
```
dcshzj, line 81: issue: wah hmm I think this makes our permissions model a bit more complicated, cos we no longer can say that you can edit and publish anything in the site as a non-admin. We would need to carve our an exception in our phrasing.

Can we push back on this restriction so it's simpler?

adriangohjw: oh even before this feature, only admin can edit the tags. it's just that it was only enforced in the frontend and not backend. so this is more of security hardening

dcshzj: Are we able to simplify it? Seems a bit unusual to have a separate permissions boundary within another

adriangohjw: @dcshzj  thought about it and considered a few patterns, but feels like the current implementation still make the most sense (for now). the following ideas were rejected: 

1. **moving into `bulkValidateUserPermissionsForResources`** - its used by many procedures, and this check is unique only for page blob and not other mutations e.g. updating collection link
2. **moving it out of the transaction after `bulkValidateUserPermissionsForResources`** -  follow closer to the existing pattern of assertion at the top, but would mean that the fetching of the old blob is out of the transaction. technically more incorrect as what we check against and what is being processed might be different theoretically 
3. **having a separate endpoint for updating collection index page (something like updatePageMeta)** -  additional endpoint thats likely duplicating many operations which can drift. extracting into shared helper also seem like an overkill (for now)

dcshzj: Sorry omg I didn't mean to simplify the code itself, I meant can we push back with product on the need to gate tags to only admins? I felt that it is weird if say I'm conducting a training and it can be quite difficult for users to understand why there is a special admin-only feature when I'm allowed to edit a page

adriangohjw: oh LOL okay

dcshzj: Discussed during team sync, since the permissions will be relooked in the future, we don't need to implement this hardening at this point in time

PR #2888, opened 2026-07-22
chore: set content disposition on publish

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 6 | +91 | -12 |
| 🧪 Test | 4 | +223 | -12 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 1 | +6 | -0 |

## Problem
We want the gazettes' downloaded filename to be different from the object key that the gazette was saved to in s3. As  an example, consider teh gazette 26ggExample.pdf - the key is 26ggExample, but the download name should be Government Gazettes Example. 

## Solution
Set the content disposition at time of publish. This contains the `filename` hint to the browser that this file should be saved as `filename`. At publish time, the title has already been set so we can do this safely with the actual title of the gazette. 

## Tests
- [ ] Navigate to https://staging-studio.isomer.gov.sg/sites/33/gazettes
- [ ] Schedule a gazette to be published but ensure that hte file id and the title are different
- [ ] Go to https://staging.d32trrzjyhalo0.amplifyapp.com/egazette-browse/ 
- [ ] Click on your gazette and download it
- [ ] Ensure that hte filename that shows on teh download is the title

As an example: https://egazettes-user-content-stg.by.gov.sg/2026/Government%20Gazette/Appointments/26gg4057ffffasdfdddd.pdf

The filekey is `26gg4057ffffasdfdddd.pdf`, but the file name on download is `the gazette title should be this`
image

## Risks
This involves a self-copy to the same object; there might be latency concerns if the gazette is large but known upper bounds are ~40MB, which should be well within AWS' limits

The PR updates scheduled gazette publishing to set a title-based download filename through an S3 self-copy before applying object retention.
- Adds reusable Content-Disposition generation using the gazette title and original file extension.
- Preserves content type and user metadata during the S3 self-copy.
- Encodes each CopySource key segment and applies the same helper to existing file-copy operations.
- Adds regression coverage for reserved key characters, retry behavior, metadata preservation, and titles containing path separators.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/lib/s3.ts | Adds encoded CopySource construction and an idempotent metadata-preserving self-copy before retention; both previously reported encoding paths are addressed. |
| apps/studio/src/server/cron/jobs/schedulePushDocumentJob.ts | Threads the gazette title into publication and supplies the generated Content-Disposition when making the S3 asset public. |
| apps/studio/src/server/modules/asset/asset.service.ts | Centralizes standards-compliant Content-Disposition generation and sanitizes path separators before deriving title-based filenames. |
| apps/studio/src/lib/__tests__/s3.test.ts | Covers metadata preservation, operation ordering, retries, and encoded CopySource values for both affected copy operations. |
| apps/studio/src/server/modules/asset/__tests__/asset.service.test.ts | Covers title extensions, international filenames, separator replacement, and valid header formatting. |

</details>

```mermaid
sequenceDiagram
  participant Cron as Scheduled publishing job
  participant S3
  participant Search as Gazette search index
  Cron->>S3: Fetch PDF
  Cron->>S3: HEAD object metadata
  Cron->>S3: Self-copy with title-based Content-Disposition
  Cron->>S3: Apply retention lock and remove scheduled tag
  Cron->>Search: Publish extracted document content
```
</details>

<sub>Reviews (9): Last reviewed commit: ["chore: format"](https://github.com/opengovsg/isomer/commit/0505d13392bf7740ba2fc9c06636a80ece5fa3b7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46187515)</sub>

- Knowledge Base — [Gazette](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/gazette.md)
- Knowledge Base — [Cron Jobs and pg-boss Queue](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/cron-and-jobs.md)
- Knowledge Base — [Redirects, SearchSG, and Asset Uploads](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/redirects-search-assets.md)
</details>

### DIFF
--- a/apps/studio/src/lib/s3.ts
+++ b/apps/studio/src/lib/s3.ts
@@ -22,6 +22,12 @@ import { env } from "~/env.mjs"
 const DELETE_TAG = "deletedAt"
 const EGAZETTE_COMPLIANCE_HOLD_IN_DAYS = 10000
 
+// Unlike Key params (which the SDK URL-encodes), CopySource is sent verbatim
+// as the x-amz-copy-source header, so keys with spaces or reserved characters
+// (e.g. "2026/Government Gazette/...") must be encoded per path segment here.
+const getEncodedCopySource = (Bucket?: string, Key?: string) =>
+  `${Bucket}/${Key?.split("/").map(encodeURIComponent).join("/")}`
+
 // R2 credentials are only set for preview, but the choice of backend is
 // driven by their presence rather than the environment name. Exported so
 // other modules don't have to re-derive this from the raw env vars.
@@ -196,17 +202,21 @@ export const setAssetAsPublished = async ({
   // tags carry over via the default TaggingDirective (COPY).
   if (ContentDisposition) {
     const head = await storage.send(new HeadObjectCommand({ Bucket, Key }))
-    await storage.send(
-      new CopyObjectCommand({
-        Bucket,
-        CopySource: `${Bucket}/${Key}`,
-        Key,
-        MetadataDirective: "REPLACE",
-        ContentType: head.ContentType,
-        Metadata: head.Metadata,
-        ContentDisposition,
-      }),
-    )
+    // Skip the (paid) self-copy when the disposition is already correct,
+    // e.g. on a pg-boss retry after an earlier attempt already rewrote it.
+    if (head.ContentDisposition !== ContentDisposition) {
+      await storage.send(
+        new CopyObjectCommand({
+          Bucket,
+          CopySource: getEncodedCopySource(Bucket, Key),
+          Key,
+          MetadataDirective: "REPLACE",
+          ContentType: head.ContentType,
+          Metadata: head.Metadata,
+          ContentDisposition,
+        }),
+      )
+    }
   }
 
   // NOTE: Lock first to preserve guarantee that once published
@@ -268,7 +278,7 @@ export const copyFile = async ({
   return storage.send(
     new CopyObjectCommand({
       Bucket,
-      CopySource: `${Bucket}/${SourceKey}`,
+      CopySource: getEncodedCopySource(Bucket, SourceKey),
       Key: DestKey,
     }),
   )
--- a/apps/studio/src/lib/s3.ts
+++ b/apps/studio/src/lib/s3.ts
@@ -147,15 +147,19 @@ export const deleteFile = async ({
   )
 }
 
-// NOTE: In order to set the asset as published, we have to do 2 things:
-// 1. we have to set the object lock retention
-// 2. we have to remove the scheduledAt tag
+// NOTE: In order to set the asset as published, we have to do 3 things:
+// 1. we rewrite the Content-Disposition (when given) so downloads are named
+//    after the gazette title rather than the raw S3 key
+// 2. we have to set the object lock retention
+// 3. we have to remove the scheduledAt tag
 // this is required to guarantee that the gazettes
 // can be seen and cannot be deleted
 export const setAssetAsPublished = async ({
   Key,
   Bucket,
-}: Pick<PutObjectTaggingCommandInput, "Key" | "Bucket">) => {
+  ContentDisposition,
+}: Pick<PutObjectTaggingCommandInput, "Key" | "Bucket"> &
+  Pick<PutObjectCommandInput, "ContentDisposition">) => {
   // Skip on R2: the COMPLIANCE-mode Object Lock below is irreversible for
   // ~27 years — applying it to a shared preview bucket would permanently
   // lock every test upload. The GuardDuty malware-scan tag check is also
@@ -182,6 +186,27 @@ export const setAssetAsPublished = async ({
   )
   if (hasFailedScan) {
     throw new Error("Cannot publish asset with failed malware scan")
+  }
+
+  // NOTE: S3 object metadata is immutable, so rewriting Content-Disposition
+  // requires a self-copy with MetadataDirective REPLACE. This must happen
+  // before the retention lock below — once the object lock is applied the
+  // object can no longer be overwritten. REPLACE drops all existing metadata,
+  // so ContentType and user metadata are read back and re-supplied; object
+  // tags carry over via the default TaggingDirective (COPY).
+  if (ContentDisposition) {
+    const head = await storage.send(new HeadObjectCommand({ Bucket, Key }))
+    await storage.send(
+      new CopyObjectCommand({
+        Bucket,
+        CopySource: `${Bucket}/${Key}`,
+        Key,
+        MetadataDirective: "REPLACE",
+        ContentType: head.ContentType,
+        Metadata: head.Metadata,
+        ContentDisposition,
+      }),
+    )
   }
 
   // NOTE: Lock first to preserve guarantee that once published
--- a/apps/studio/src/lib/s3.ts
+++ b/apps/studio/src/lib/s3.ts
@@ -25,7 +25,7 @@ const EGAZETTE_COMPLIANCE_HOLD_IN_DAYS = 10000
 // Unlike Key params (which the SDK URL-encodes), CopySource is sent verbatim
 // as the x-amz-copy-source header, so keys with spaces or reserved characters
 // (e.g. "2026/Government Gazette/...") must be encoded per path segment here.
-const getEncodedCopySource = (Bucket?: string, Key?: string) =>
+const getEncodedCopySource = (Bucket: string, Key: string) =>
   `${Bucket}/${Key?.split("/").map(encodeURIComponent).join("/")}`
 
 // R2 credentials are only set for preview, but the choice of backend is
@@ -171,6 +171,9 @@ export const setAssetAsPublished = async ({
   // lock every test upload. The GuardDuty malware-scan tag check is also
   // moot, since GuardDuty is an AWS-only service that never scans R2 objects.
   if (isR2Configured) return
+  if (!Bucket) throw new Error("Bucket must be defined")
+  if (!Key) throw new Error("Key must be defined")
+
   const objectTag = await storage.send(
     new GetObjectTaggingCommand({
       Bucket,
@@ -275,6 +278,8 @@ export const copyFile = async ({
   SourceKey: string
   DestKey: string
 }) => {
+  if (!Bucket) throw new Error("Bucket must be defined")
+
   return storage.send(
     new CopyObjectCommand({
       Bucket,
--- a/apps/studio/src/lib/s3.ts
+++ b/apps/studio/src/lib/s3.ts
@@ -27,7 +27,7 @@ const EGAZETTE_COMPLIANCE_HOLD_IN_DAYS = 10000
 // Unlike Key params (which the SDK URL-encodes), CopySource is sent verbatim
 // as the x-amz-copy-source header, so keys with spaces or reserved characters
 // (e.g. "2026/Government Gazette/...") must be encoded per path segment here.
-const getEncodedCopySource = (Bucket?: string, Key?: string) =>
+const getEncodedCopySource = (Bucket: string, Key: string) =>
   `${Bucket}/${Key?.split("/").map(encodeURIComponent).join("/")}`
 
 // R2 credentials are only set for preview, but the choice of backend is
@@ -174,6 +174,9 @@ export const setAssetAsPublished = async ({
   // lock every test upload. The GuardDuty malware-scan tag check is also
   // moot, since GuardDuty is an AWS-only service that never scans R2 objects.
   if (isR2Configured) return
+  if (!Bucket) throw new Error("Bucket must be defined")
+  if (!Key) throw new Error("Key must be defined")
+
   const objectTag = await storage.send(
     new GetObjectTaggingCommand({
       Bucket,
@@ -299,6 +302,8 @@ export const copyFile = async ({
   SourceKey: string
   DestKey: string
 }) => {
+  if (!Bucket) throw new Error("Bucket must be defined")
+
   return storage.send(
     new CopyObjectCommand({
       Bucket,
--- a/apps/studio/src/lib/s3.ts
+++ b/apps/studio/src/lib/s3.ts
@@ -150,15 +150,19 @@ export const deleteFile = async ({
   )
 }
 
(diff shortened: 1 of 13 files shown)


apps/studio/src/lib/s3.ts
```diff
@@ -184,6 +188,27 @@ export const setAssetAsPublished = async ({
...
+  // requires a self-copy with MetadataDirective REPLACE. This must happen
+  // before the retention lock below — once the object lock is applied the
+  // object can no longer be overwritten. REPLACE drops all existing metadata,
+  // so ContentType and user metadata are read back and re-supplied; object
+  // tags carry over via the default TaggingDirective (COPY).
+  if (ContentDisposition) {
+    const head = await storage.send(new HeadObjectCommand({ Bucket, Key }))
+    await storage.send(
+      new CopyObjectCommand({
+        Bucket,
```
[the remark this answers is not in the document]
seaerchin: Isn't the key already URL encoded via `getContentDispositionForTitle`, specifically:

```ts
  const encoded = encodeURIComponent(`${title}${extension}`)
  return `inline; filename*=UTF-8''${encoded}`
```

apps/studio/src/lib/s3.ts
```diff
@@ -184,6 +188,27 @@ export const setAssetAsPublished = async ({
...
+        CopySource: `${Bucket}/${Key}`,
```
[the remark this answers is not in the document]
seaerchin: Isn't the key already URL encoded via getContentDispositionForTitle, specifically:

```ts
  const encoded = encodeURIComponent(`${title}${extension}`)
  return `inline; filename*=UTF-8''${encoded}`
```

apps/studio/src/lib/s3.ts
```diff
@@ -184,6 +188,27 @@ export const setAssetAsPublished = async ({
...
+        Key,
+        MetadataDirective: "REPLACE",
+        ContentType: head.ContentType,
+        Metadata: head.Metadata,
+        ContentDisposition,
+      }),
+    )
+  }
```
adriangohjw, line 226: nit(non-blocking): if `pushDocumentsForIngestion` fails, then pg job will fail and retry. But as we alreay lock the object due to compliance mode, subsequent retry will fail.

To me, its a nit because its hardening and it depends on transient failure in `pushDocumentsForIngestion`

```md
Cron fires schedule-push-document job (pg-boss, retryLimit: 3)
│
▼
Query PushDocumentJob rows where scheduledAt <= now
│
▼
For each gazette resource ─────────────────────────────────────────────┐
│                                                                       │
│  extractResourceData()                                                │
│    │                                                                  │
│    ├─► fetch index page blob from DB          ✓                       │
│    ├─► fetch PDF from S3                      ✓                       │
│    │                                                                  │
│    └─► setAssetAsPublished()                                          │
│          │                                                            │
│          ├─► HeadObject + CopyObject (self-copy)   ✓  sets Content-Disposition
│          │                                                            │
│          ├─► PutObjectRetention (COMPLIANCE)       ✓  object is now HARD-LOCKED
│          │         RetainUntilDate = now + 10000 days                 │
│          │         (cannot overwrite or delete — even root)           │
│          │                                                            │
│          └─► PutObjectTagging (remove scheduledAt) ✓                  │
│                                                                       │
│    ├─► parse index page content                 ✓                     │
│    ├─► parseFullTextFromPDF()                   ✗ FAILS               │
│    │       (or SearchSG ingestion throws later)                       │
│    │                                                                  │
│    └──► error caught → resource skipped, NOT added to documents[]     │
│                                                                       │
◄───────────────────────────────────────────────────────────────────────┘
│
▼
pushDocumentsForIngestion(documents)              ✗ THROWS (or job dies here)
│
│   deleteProcessedJobs()  ◄── NEVER REACHED
│
▼
pg-boss marks job failed → schedules RETRY (attempt 2 of 3)


═══════════════════════════════════════════════════════════════════════
                       RETRY (attempt 2)
═══════════════════════════════════════════════════════════════════════

Same PushDocumentJob row still in DB  (scheduledAt <= now)
│
▼
extractResourceData() again
│
└─► setAssetAsPublished()
      │
      ├─► HeadObject                              ✓  object exists, already locked
      │
      └─► CopyObject (self-copy, same Key)        ✗ FAILS
            │
            │   S3: object under COMPLIANCE retention
            │       → overwrite/copy-in-place is rejected
            │
            └──► entire extractResourceData throws
                   │
                   └──► retry 2 fails … retry 3 fails … job permanently dead


═══════════════════════════════════════════════════════════════════════
                       END STATE (stuck)
═══════════════════════════════════════════════════════════════════════

S3 object          PushDocumentJob row       Search index
─────────          ───────────────────       ────────────
PUBLISHED          STILL PRESENT             NEVER INDEXED
(locked,           (kept because job
 viewable)          never reached deleteProcessedJobs)
```

apps/studio/src/lib/s3.ts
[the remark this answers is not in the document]
adriangohjw: btw the same issue happens here too - can fix together

https://github.com/opengovsg/isomer/blob/1861a7ecc51abe30afcfa6f881247f3319a3abe5/apps/studio/src/lib/s3.ts#L235-L250

apps/studio/src/lib/s3.ts
```diff
@@ -184,6 +188,27 @@ export const setAssetAsPublished = async ({
...
+      new CopyObjectCommand({
+        Bucket,
+        CopySource: `${Bucket}/${Key}`,
+        Key,
+        MetadataDirective: "REPLACE",
+        ContentType: head.ContentType,
+        Metadata: head.Metadata,
+        ContentDisposition,
+      }),
+    )
```
adriangohjw: nit(non-blocking): save an API call (and $0.000005 LOLz) if content disposition is the same

```suggestion
    // Skip the (paid) self-copy when the disposition is already correct.
    if (head.ContentDisposition !== ContentDisposition) {
      await storage.send(
        new CopyObjectCommand({
          Bucket,
          CopySource: `${Bucket}/${Key}`,
          Key,
          MetadataDirective: "REPLACE",
          ContentType: head.ContentType,
          Metadata: head.Metadata,
          ContentDisposition,
        }),
      )
    }
```

apps/studio/src/server/modules/asset/asset.service.ts
```diff
@@ -63,6 +63,23 @@ export const getContentDispositionForKey = (key: string): string => {
...
+ * the right application.
+ */
+export const getContentDispositionForTitle = (
+  title: string,
+  key: string,
+): string => {
+  const segment = key.split("/").pop() ?? ""
+  const extension = segment.includes(".")
+    ? segment.substring(segment.lastIndexOf("."))
+    : ""
```
adriangohjw: fyi (suggestion): same logic happening across files more than twice, can refactor them to shared helper

```md
# Asset key parsing refactor

## Done — `apps/studio/src/server/modules/asset/asset.service.ts`

Three functions duplicated the same S3 key parsing:

| Helper | Extracts |
|---|---|
| `getFilenameFromKey(key)` | last segment via `key.split("/").pop()` |
| `getExtensionFromFilename(filename)` | extension via `lastIndexOf(".")` |

Used by:

- `getContentTypeFromKey` — lowercase filename first (Unicode-safe MIME lookup)
- `getContentDispositionForKey` — raw filename
- `getContentDispositionForTitle` — raw extension appended to title
```

adriangohjw (commented):
additional remarks on top of greptile/hacktron comments - good to approve once their comment is addressed

apps/studio/src/lib/s3.ts
seaerchin: i actually don't think this is correct - i tried this manually with the [AWS UI ](https://058264420411-3kof4mcu.ap-southeast-1.console.aws.amazon.com/s3/object/egazette-assets-private-341f27b?region=ap-southeast-1&prefix=2026/Government+Gazette/Notices+under+other+Acts/111111.pdf&tab=versions)

but you can observe that i did a self copy below: 
image

original had compliance lock:
image

new one doesn't: 
image

the lock itself is a property of the specified object + version; doing a self version **creates a new version at that same key** - the existing version still has compliance lock

apps/studio/src/lib/s3.ts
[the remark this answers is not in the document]
seaerchin: !fixed — `CopySource` is now URL-encoded per path segment (in both `setAssetAsPublished` and `copyFile`), with a regression test using a key containing spaces and `#`. Fixed in d4a9c0460.

apps/studio/src/lib/s3.ts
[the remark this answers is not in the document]
seaerchin: Fixed in d4a9c0460 — extracted a `getEncodedCopySource` helper that encodes each path segment (keeping `/` delimiters) and applied it to both `setAssetAsPublished` and `copyFile`. Added regression tests with keys containing spaces and `#`.

apps/studio/src/lib/s3.ts
seaerchin: Applied in d4a9c0460 — the self-copy is now skipped when `head.ContentDisposition` already matches, with a test covering the skip path. As a bonus this also covers your retry concern on the other thread: a retry after a successful copy goes straight to the retention/tagging steps.

apps/studio/src/server/modules/asset/asset.service.ts
seaerchin: Extracted `getFilenameFromKey` and `getExtensionFromFilename` helpers in `asset.service.ts` in d4a9c0460 — all three functions now share them.

apps/studio/src/lib/s3.ts
seaerchin: Additionally hardened in d4a9c0460: the self-copy is now skipped when the disposition already matches, so a retry after a successful copy no longer re-copies at all (regardless of lock semantics).

apps/studio/src/server/modules/asset/asset.service.ts
```diff
@@ -43,23 +43,39 @@ export const generateTagsQueryString = (
...
+ * the right application.
+ */
+export const getContentDispositionForTitle = (
+  title: string,
+  key: string,
+): string => {
+  const extension = getExtensionFromFilename(getFilenameFromKey(key))
+  const encoded = encodeURIComponent(`${title}${extension}`)
   return `inline; filename*=UTF-8''${encoded}`
 }
```
adriangohjw, line 83: tldr; download still works but browsers might not follow the suggested filenames for these edge cases.

nit and non-blocking for me since IMO its edge cases and does not break download - but might be worth a fix if we want to avoid agency from bringing up the same issue and having to re-investigate in the future

## Summary

- We uses `encodeURIComponent` which is mostly correct but RFC 5987 `attr-char` excludes `*`, `'`, `(`, `)`. `encodeURIComponent` leaves those alone.
- We emit invalid `filename*` values like:

```http
inline; filename*=UTF-8''O'Brien.pdf
inline; filename*=UTF-8''Report%20(final).pdf
```

- `( )` / `'` / `*` are fine in the filename itself, and in quoted `filename="…"`.
- The bug is only raw `'()`/`*` inside the `filename*` value after `UTF-8''`.
- We send only `filename*` (no `filename=` fallback), so a bad browser parse has nothing else to use.

## Console demo

Paste into the browser console:

```js
const names = [
  "O'Brien.pdf",
  "Report (final).pdf",
  "file*.pdf",
  "Report (final)*'draft'.pdf",
]

console.table(
  names.map((name) => {
    const buggy = encodeURIComponent(name)
    const fixed = buggy.replace(
      /['()*]/g,
      (c) => `%${c.charCodeAt(0).toString(16).toUpperCase()}`,
    )
    return {
      name,
      buggyHeader: `inline; filename*=UTF-8''${buggy}`,
      fixedHeader: `inline; filename*=UTF-8''${fixed}`,
      leftRaw: [...name].filter((c) => "()'*" .includes(c)).join("") || "—",
    }
  }),
)
```

- Compare `buggyHeader` vs `fixedHeader`.
- Any raw `'()`/`*` after `UTF-8''` is the bug.

## What users actually hit

- Symptom: wrong, truncated, or ignored download filename
- Not a 500, not a corrupt file

### Worst plausible case

1. Publish a gazette titled `O'Brien's Notice (No. 1)`.
2. Cron rewrites S3 `Content-Disposition` from that title.
3. Chromium ignores `filename*` on download.
4. Save As uses the key basename instead of the human title.

## Proposed fix: `content-disposition`

```ts
import { create } from "content-disposition"

create(`${title}${extension}`, { type: "inline" })
```

Why this one:

- Zero deps, small
- Standard HTTP helper for this header (same lineage Next/Express already ship)
- We stop owning RFC 5987 encoding

adriangohjw (approved):

apps/studio/src/server/modules/asset/asset.service.ts
```diff
@@ -43,24 +44,38 @@ export const generateTagsQueryString = (
...
+ * Build Content-Disposition using a human-readable title as the download
+ * filename, keeping the key's extension so the saved file still opens in
+ * the right application.
+ */
+export const getContentDispositionForTitle = (
+  title: string,
+  key: string,
+): string => {
+  const extension = getExtensionFromFilename(getFilenameFromKey(key))
+  return contentDisposition(`${title}${extension}`, { type: "inline" })
```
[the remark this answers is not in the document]
seaerchin: Fixed in 31cf619cc — `getContentDispositionForTitle` now replaces path separators (`/`, `\`) in the title with `-` before calling `contentDisposition`, so `path.basename` no longer truncates it (`"A/B\\C"` → `filename="A-B-C.pdf"`). Uses `-` to match this file's existing `filenamify(..., { replacement: "-" })` convention. Added a regression test covering the slash/backslash case.

seaerchin:
not merging first, will let IMDA/Toppan UAT till they ok

PR #2937, opened 2026-07-24
test(e2e): add per-site isolation for mutating tests

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 1 | +0 | -2 |
| 🧪 Test | 14 | +331 | -112 |
| 📄 Doc | 1 | +30 | -1 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

E2E tests shared seed site ID `1`, so parallel runs could flake when settings, creates, and invites collided on the same DB rows. The initial fix for this (`provisionE2ESite`/`teardownE2ESite`) also had its own correctness gaps: cleanup missed data from repeated publishes, ran redundant queries, and its options type allowed a call shape the runtime couldn't actually handle.

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- None

**Improvements**:

- Add `provisionE2ESite` / `teardownE2ESite` in `fixtures/site.ts` so every test file gets a dedicated site (`beforeAll` / `afterAll`), not just mutating ones
- Migrate all test files off `getSeedSiteId()` — including the last read-only caller (`site/admin.test.ts`) — and remove `getSeedSiteId()` entirely
- Extend `resetSiteAgencySettings` to accept the provisioned site name
- Add `conventions/e2e-tests.md` (living E2E conventions doc) with a per-site isolation section, updated to drop the now-obsolete read-only/mutating split
- Increase visibility timeouts on settings tests for cold-start tRPC latency

**Bug Fixes**:

- `teardownE2ESite` only deleted the `Version`/`Blob` rows referenced by each resource's *current* `publishedVersionId`/`draftBlobId`. Since publishing always inserts a new `Version` row without deleting the superseded one, any resource published more than once during a test left orphaned `Version`/`Blob` rows that teardown never touched. Cleanup is now scoped by `resourceId IN resourceIds` so it catches every version, not just the latest.
- `teardownE2ESite` also ran two resourceId-scoped deletes (`CodeBuildJobs`, `ResourcePermission`) that were fully redundant with the siteId-scoped deletes immediately after them — removed, since both tables' `siteId` is always set consistently by the app.
- `provisionE2ESite({})` type-checked despite no caller ever doing this, but would have silently attributed the root page to the publisher's user id without granting the publisher any permission on the site. The opts type is now a union requiring at least one of `admin`/`editor`/`publisher`.
- Fixed a pre-existing type error in `resetSiteAgencySettings` (`config` was cast to `Record<string, unknown>`, losing the branded `IsomerSiteConfigProps` shape the `Site.config` column expects) and an `interface`-vs-`type` lint violation on `ProvisionedSite`.

## Before & After Screenshots

**BEFORE**:

N/A — test infrastructure only

**AFTER**:

N/A — test infrastructure only

## Tests

- `pnpm exec playwright test site/settings-agency site/settings-notification --workers=2` — 3 passed (parallel mutating files)
- `pnpm exec playwright test --grep-invert singpass` — 23 passed
- `pnpm typecheck` and `pnpm lint` — clean (only pre-existing, unrelated warnings elsewhere in the repo)

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

The PR replaces the shared seed-site dependency with independently provisioned sites so E2E files can run without mutating the same site records.
- Adds a reusable site-provisioning fixture with role-specific permissions and required root/search resources.
- Migrates collection, page, folder, site, settings, and invitation tests to provisioned site IDs.
- Moves database initialization into Playwright global setup and removes the separate CI seed step.
- Documents the per-site isolation convention and makes collection fixtures require explicit site IDs.
</details>

<details><summary><h3>Confidence Score: 3/5</h3></summary>

The PR does not appear safe to merge until the destructive reset verifies that its connection targets the disposable E2E database instance rather than merely a database named `test`.

The current guard accepts any connection URL whose pathname is `test`, then truncates all public application tables with cascading identity reset; the previously reported host or instance validation gap therefore remains.

**Files Needing Attention:** apps/studio/tests/e2e/global-setup.ts
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/tests/e2e/global-setup.ts | Adds run-scoped database reset and self-contained E2E role seeding before authentication state is generated. |
| apps/studio/tests/e2e/fixtures/site.ts | Introduces per-file site provisioning with explicit role grants and required initial page resources. |
| apps/studio/tests/e2e/fixtures/seed.ts | Removes the fixed seed-site ID and provisions whitelist, users, roles, and elevated administrators during global setup. |
| apps/studio/tests/e2e/fixtures/collection.ts | Makes site ownership explicit across collection fixture creation and lookup operations. |
| .github/workflows/ci.yml | Removes the redundant database seed command now that Playwright global setup initializes E2E data. |

</details>

```mermaid
flowchart LR
  Setup[Playwright global setup] --> Reset[Reset E2E database]
  Reset --> Seed[Seed canonical test users]
  Seed --> Files[Test files]
  Files --> Provision[Provision dedicated site]
  Provision --> Permissions[Grant requested roles]
  Permissions --> Resources[Create root and search pages]
  Resources --> Tests[Run tests against isolated site]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["chore: format fix"](https://github.com/opengovsg/isomer/commit/36c6da12dcb3e83c15b6cd7472d16311a5d8ff5b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54278444)</sub>

### DIFF
--- a/apps/studio/tests/e2e/fixtures/seed.ts
+++ b/apps/studio/tests/e2e/fixtures/seed.ts
@@ -6,9 +6,6 @@ import { TEST_EMAILS } from "./auth"
 
 const SEED_SITE_ID = 1
 
-/** @deprecated for mutating tests — use provisionE2ESite() */
-export const getSeedSiteId = () => SEED_SITE_ID
-
 /**
  * Idempotent: inserts user if missing, then ensures a ResourcePermission
  * with `role` on the seed site exists (re-activating if soft-deleted).
--- a/apps/studio/tests/e2e/fixtures/seed.ts
+++ b/apps/studio/tests/e2e/fixtures/seed.ts
@@ -6,6 +6,7 @@ import { TEST_EMAILS } from "./auth"
 
 const SEED_SITE_ID = 1
 
+/** @deprecated for mutating tests — use provisionE2ESite() */
 export const getSeedSiteId = () => SEED_SITE_ID
 
 /**
--- a/apps/studio/tests/e2e/fixtures/seed.ts
+++ b/apps/studio/tests/e2e/fixtures/seed.ts
@@ -1,14 +1,13 @@
 import { createId } from "@paralleldrive/cuid2"
+import { setUpWhitelist, setupSite } from "tests/integration/helpers/seed"
 import { db } from "~/server/modules/database"
 import { IsomerAdminRole, RoleType } from "~prisma/generated/generatedEnums"
 
 import { TEST_EMAILS } from "./auth"
 
-const SEED_SITE_ID = 1
-
 /**
  * Idempotent: inserts user if missing, then ensures a ResourcePermission
- * with `role` on the seed site exists (re-activating if soft-deleted).
+ * with `role` on `siteId` exists (re-activating if soft-deleted).
  *
  * The unique constraint on ResourcePermission is
  * (userId, siteId, resourceId, deletedAt) NULLS NOT DISTINCT.
@@ -18,6 +17,7 @@ const SEED_SITE_ID = 1
 const ensureUserWithRole = async (
   email: string,
   role: (typeof RoleType)[keyof typeof RoleType] | null,
+  siteId: number,
 ) => {
   const user = await db
     .insertInto("User")
@@ -50,7 +50,7 @@ const ensureUserWithRole = async (
     .insertInto("ResourcePermission")
     .values({
       userId: user.id,
-      siteId: SEED_SITE_ID,
+      siteId,
       role,
       resourceId: null,
     })
@@ -70,8 +70,9 @@ const ensureUserWithRole = async (
 const ensureGodModeAdmin = async (
   email: string,
   role: (typeof IsomerAdminRole)[keyof typeof IsomerAdminRole],
+  siteId: number,
 ) => {
-  const user = await ensureUserWithRole(email, null)
+  const user = await ensureUserWithRole(email, null, siteId)
 
   await db
     .insertInto("IsomerAdmin")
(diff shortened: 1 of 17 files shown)


apps/studio/tests/e2e/global-setup.ts
```diff
@@ -40,6 +67,7 @@ const signInOnce = async (role: keyof typeof TEST_EMAILS, baseURL: string) => {
 const globalSetup = async (config: FullConfig) => {
   const baseURL = config.projects[0]?.use.baseURL ?? "http://127.0.0.1:3000"
 
+  await resetE2EDatabase()
```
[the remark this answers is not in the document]
adriangohjw: does this fix it? https://github.com/opengovsg/isomer/pull/2937/commits/27699adf1161243b7a94e8fec1fa73972c46f4ba

adriangohjw:
@cursor review

apps/studio/tests/e2e/fixtures/seed.ts
```diff
@@ -85,11 +84,24 @@ const ensureGodModeAdmin = async (
...
-  await ensureUserWithRole(TEST_EMAILS.publisher, RoleType.Publisher)
-  await ensureGodModeAdmin(TEST_EMAILS.core, IsomerAdminRole.Core)
-  await ensureGodModeAdmin(TEST_EMAILS.migrator, IsomerAdminRole.Migrator)
+  // protectedProcedure's authMiddleware requires the caller's email to be
+  // whitelisted (see whitelist.service.ts's isEmailWhitelisted), so every
+  // TEST_EMAILS user (all @open.gov.sg) needs this domain whitelisted for
+  // any tRPC call to succeed across the whole e2e suite. prisma/seed.ts
+  // used to provide this row via CI's now-removed "Seed testing db" step;
+  // it's self-provisioned here instead.
+  await setUpWhitelist({ email: "@open.gov.sg" })
```
dcshzj, line 93: suggestion(non-blocking): might want to consider making the email address more explicit, such as `@whitelisted.gov.sg` so it's clear from the fixtures that we are dealing with someone that is whitelisted

apps/studio/tests/e2e/global-setup.ts
```diff
@@ -1,12 +1,53 @@
...
 
+// The e2e suite's DATABASE_URL points at a `test` database that has no
+// purpose other than e2e fixtures (a separate logical database from local
+// dev's `app` database, inside the same docker-compose Postgres container),
+// so wiping it completely at the start of every run is safe. The table list
+// is derived dynamically from information_schema so this doesn't silently
+// go stale as the schema evolves.
+const E2E_DATABASE_NAME = "test"
+
+const resetE2EDatabase = async (): Promise<void> => {
```
dcshzj, line 19: question: this seems to be destructive to the entire database, how does the per-site isolation get achieved here? If I'm reading correctly from the changes, there are 2 behaviours which can happen here:

1. This `resetE2EDatabase` is only run for the first time the e2e tests are started, which then means that each test has their own site (e.g. the site ID can become ID 999 if there are 999 e2e tests) → This should be ok, just want your confirmation
2. The `resetE2EDatabase` runs before every test (I vaguely saw it mentioned somewhere), then we might not be able to run the e2e tests concurrently

adriangohjw: @dcshzj  unless im understanding "run the e2e tests concurrently" wrongly, like you meant having multiple worktrees or terminal running them?

dcshzj: The terminal one, within a test file there would be multiple test fixtures so I am not sure if they are running sequentially or concurrently?

If concurrently, then it seems like they might step on one another cos the database would keep getting reset, ps not quite understanding this thing

adriangohjw: @dcshzj  within a file, tests can run in parallel because of `fullyParallel: true`

you are correct that they might step on one another, but we create isolated site and reference them by siteID (same for resources). if we want it to be serial, we have to set `test.describe.configure({ mode: "serial" })` (what we does for site settings e2e downstream)

apps/studio/tests/e2e/global-setup.ts
```diff
@@ -1,12 +1,53 @@
...
+  // dotenv (used to load .env.test) does not override an already-set
+  // DATABASE_URL by default, so a shell that already has the dev DATABASE_URL
+  // exported would otherwise cause this to silently truncate the dev
+  // database instead of the disposable e2e one.
+  const databaseName = new URL(env.DATABASE_URL).pathname.replace(/^\//, "")
+  if (databaseName !== E2E_DATABASE_NAME) {
```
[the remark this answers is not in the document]
adriangohjw: nah too niche and overthinking, current checks are sufficient. thanks for pointing out either way

PR #2941, opened 2026-07-24
test(e2e): role-based Playwright projects and feature-sharded CI

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 12 | +147 | -37 |
| 🧪 Test | 15 | +858 | -742 |
| 📄 Doc | 2 | +58 | -2 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

E2E auth was wired per file via `test.use({ storageState })`, which doesn't scale as the suite grows across roles. We need role-scoped Playwright projects so tests can be filtered and authenticated consistently.

On top of that, CI ran the full E2E suite as 4 numeric shards (`shard: [1,2,3,4]`) regardless of what changed, with no way to tell from a shard number which feature broke, each shard signing in all 6 roles regardless of which the shard's tests actually need, and no gate to skip E2E entirely for changes that can't affect it (e.g. a README edit).

Part of the e2e scale-and-coverage stack (PR-3); depends on per-site isolation (PR-2).

## Solution

**Breaking Changes**

**Features**:

- N/A (test infrastructure only)

**Improvements**:

- Replace the single `e2e` Playwright project with `unauthenticated`, `singpass`, and one project per role (`admin`, `editor`, `publisher`, `nomember`, `core`, `migrator`)
- Role projects supply `storageState` and select tests via `grep` on `@role` tags
- Add typed `roleTag()` so describe tags must be members of `ROLES`
- Migrate existing suites off `test.use({ storageState })` / ad-hoc browser contexts onto `{ tag: roleTag(...) }`, including the `collection/` suites (`filters-permission`, `required-tags-validation`) which had no role tag at all and so never actually ran in CI
- Make `site/admin.test.ts` exhaustive over `Role` via `Record<Role, "allowed" | "denied">`
- Move E2E auth bootstrap into Playwright `global-setup` (`fixtures/seed.ts`): whitelist `@open.gov.sg` + canonical test users, with no hardcoded seed site or dependency on `prisma/seed.ts`
- Remove the CI `db:seed` step — `setup:test` + `test:e2e` is now the same path locally and in CI
- Migrate `site/list.test.ts` to `provisionE2ESite` (assert on provisioned `siteName`, not "Sample Site")
- Consolidate `page/`, `resource/`, and `collection/` test files into a single `resource/` directory, matching the domain model (pages, folders, and collections are all `Resource` types)
- Replace CI's numeric Playwright sharding (`--shard=N/M`) with named, feature-based CI jobs — one per `tests/e2e/<feature>` directory (`site`, `resource`, `user`, `godmode`, plus a `root` job for the untagged `smoke`/`singpass` files) — so a failing job's name tells you what broke instead of an arbitrary shard number. `apps/studio/scripts/e2eCiCoverage.test.ts` fails `pnpm test:unit` if a `tests/e2e/` directory isn't wired into a CI job
- Gate the `e2e-tests` CI job on a new `studio_e2e_relevant` path filter (`apps/studio/**`, the workspace packages Studio depends on at runtime, the lockfile, and the CI config itself) so PRs that can't affect Studio's runtime behavior skip E2E entirely instead of always running the full suite
- Scope `global-setup`'s role sign-in to only the roles a given CI job's `tests/e2e/<feature>` paths actually exercise (resolved via Playwright's own `--list --reporter=json`, since loop-generated `roleTag` calls make static parsing unreliable), falling back to signing in all roles for local runs or if resolution fails. The `root` job (smoke + singpass) now signs in zero roles instead of all six
- Skip generating production browser source maps and the `output: "standalone"` dependency trace/copy for the E2E CI build specifically (`SKIP_PRODUCTION_SOURCE_MAPS`, `SKIP_STANDALONE_OUTPUT` env vars, validated in `src/env.mjs`) — both exist only for real deploys (Datadog RUM, Docker) that E2E's `next start` never exercises, and generating them is a meaningful chunk of build time. Production/staging/UAT builds and local `pnpm build` are unaffected — the env vars default to preserving today's behavior
- Add "list" alongside "github" in the CI Playwright reporter so CI logs show each test's pass/fail line with its duration, not just error annotations
- Parallelise global-setup role sign-ins with `Promise.all`
- Document the convention in `tests/e2e/README.md` and `e2e-tests.md`

**Bug Fixes**:

- Fix a CI job-skip bug: `e2e-tests` previously had no dependency on the (now-removed) shared-build job's success, and separately fix the `changes` job's new path filter to actually gate `e2e-tests` (both introduced and then corrected within this branch, called out here since they touched shipped CI behavior)

## Before & After Screenshots

**BEFORE**:

N/A (Playwright config / CI workflow / e2e tests only)

**AFTER**:

N/A

## Tests

- [ ] `cd apps/studio && pnpm exec dotenv -e .env.test -- ./node_modules/.bin/playwright test --project=admin --list` lists only `@admin` / `roleTag("admin")` suites
- [ ] `pnpm exec dotenv -e .env.test -- ./node_modules/.bin/playwright test --project=unauthenticated --list` lists smoke only
- [ ] `pnpm setup:test && pnpm test:e2e` (no `db:seed`) — active e2e suite green across role projects
- [ ] `pnpm exec dotenv -e .env.test -- ./node_modules/.bin/playwright test --list tests/e2e/resource` lists 16 tests spanning `create-page`, `create-folder`, `filters-permission`, and `required-tags-validation`
- [ ] On this PR, confirm the `e2e-tests` CI job runs as 5 named jobs (`site`, `resource`, `user`, `godmode`, `root`) instead of 4 numbered shards, and each uploads its own `e2e-test-results-*` artifact
- [ ] Push a commit that only touches `README.md` and confirm the `e2e-tests` job is skipped entirely (not run-then-pass)
- [ ] Confirm a real deploy build (staging/UAT/production, or local `pnpm build`) still produces `.next/standalone` and browser source maps as before — `SKIP_STANDALONE_OUTPUT`/`SKIP_PRODUCTION_SOURCE_MAPS` are unset outside the e2e CI build, so `next.config.mjs` should resolve identically to before this PR

**New scripts**:

- `apps/studio/scripts/e2eCiCoverage.test.ts`: fails `pnpm test:unit` if a `tests/e2e/` directory or root test file isn't wired into a CI `e2e-tests` matrix entry

**New dependencies**:

- N/A

**New dev dependencies**:

- N/A

The PR replaces numeric Playwright shards with role-aware projects and feature-based CI jobs while avoiding unnecessary authentication and build work.
- Adds role-tagged Playwright projects with project-level authentication state.
- Splits E2E execution into named feature jobs and gates them using Studio runtime dependency paths.
- Moves authentication bootstrap into global setup and signs in only the roles required by each feature job.
- Skips deploy-only standalone output and browser source maps during E2E builds.
- Migrates existing E2E suites to role tags and adds CI coverage checks for feature paths.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/ci.yml | Introduces the Studio-relevant path gate and named feature matrix; the prior validators omission is fixed and all current runtime workspace dependencies are covered. |
| apps/studio/playwright.config.ts | Replaces the single E2E project with unauthenticated, Singpass, and role-scoped projects using role tags and project storage state. |
| apps/studio/tests/e2e/global-setup.ts | Resolves the roles required by each feature path and performs role sign-ins concurrently with an all-role fallback. |
| apps/studio/scripts/e2eCiCoverage.test.ts | Adds a unit check ensuring every E2E feature directory and root test file appears exactly once in the CI matrix. |
| apps/studio/next.config.mjs | Makes standalone output and production browser source maps optional for the E2E-specific build. |
| apps/studio/src/env.mjs | Validates and exposes the two opt-out flags used by the E2E build. |
| apps/studio/tests/e2e/fixtures/auth.ts | Adds a typed helper for generating role tags used by Playwright project filtering. |

</details>

```mermaid
flowchart LR
  Changes[Changed paths] --> Filter{Studio E2E relevant?}
  Filter -->|No| Skip[Skip E2E]
  Filter -->|Yes| Matrix[Feature matrix]
  Matrix --> Site[site]
  Matrix --> Resource[resource]
  Matrix --> User[user]
  Matrix --> Godmode[godmode]
  Matrix --> Root[root]
  Site --> Roles[Resolve required role projects]
  Resource --> Roles
  User --> Roles
  Godmode --> Roles
  Root --> NoRoles[No role sign-in]
  Roles --> Setup[Global setup and role sign-in]
  Setup --> Tests[Run tagged Playwright tests]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["ci(e2e): include packages/validators in ..."](https://github.com/opengovsg/isomer/commit/26bfc6bc33d1eeaa608638f8bd3bc3dd878f568a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54308719)</sub>


apps/studio/playwright.config.ts
```diff
@@ -8,23 +10,40 @@ const opts = {
...
+    },
+    // singpass.test.ts stays skipped; isolated project so it never runs under role greps
+    {
+      name: "singpass",
+      testMatch: /singpass\.test\.ts/,
+      use: { ...baseUse },
     },
+    ...ROLES.map((role) => ({
+      name: role,
+      grep: new RegExp(`@${role}\\b`),
```
dcshzj, line 45: suggestion: can we make this more extensible by instead defining it as `@role:${role}`? So it looks like `@role:admin` rather than just a pure `@admin`

apps/studio/tests/e2e/resource/create-page.test.ts
```diff
@@ -0,0 +1,219 @@
...
+const createSeedFolder = () =>
+  db
+    .insertInto("Resource")
+    .values({
+      permalink: `e2e-test-folder-${crypto.randomUUID().slice(0, 8)}`,
+      siteId,
+      parentId: null,
+      title: "E2E Test Folder",
+      draftBlobId: null,
+      state: ResourceState.Draft,
```
dcshzj: issue(non-blocking): this should be Published since we don't have a concept of Draft folders

apps/studio/tests/e2e/resource/create-page.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    expect(created?.parentId).toBeNull()
+  })
+})
+
+test.describe("publisher", { tag: roleTag("publisher") }, () => {
+  test.beforeEach(async () => {
+    await ensureUserOnboarded(TEST_EMAILS.publisher)
+  })
+
+  test("publisher does not see the Create new button", async ({ page }) => {
```
dcshzj: issue: test description should be clearer that this pertains to just root pages, not quite clear at first glance that it is only relevant for root pages

apps/studio/tests/e2e/resource/required-tags-validation.test.ts
```diff
@@ -0,0 +1,197 @@
...
+
+// Shared across every test in this file: one required tag category with a
+// single option, so both drawers have something to validate against.
+const TAG_CATEGORY_ID = crypto.randomUUID()
+const TAG_OPTION_ID = crypto.randomUUID()
+const TAG_CATEGORY_LABEL = "Topic"
+const TAG_OPTION_LABEL = "Technology"
+
+test.describe(
+  "collection link — required tag categories",
```
dcshzj, line 40: suggestion: should we rename this into something like collection.test.ts? I think we can adopt a convention of e2e test naming where it is linked to each route rather than the individual intentions

No strong opinion on this, lmk what you think but ideally we can standardise

apps/studio/playwright.config.ts
dcshzj: Ps @adriangohjw is it possible to make these changes on this PR directly instead of downstream, since rebasing is quite cheap to do now

apps/studio/tests/e2e/resource/required-tags-validation.test.ts
dcshzj: Same for this also

apps/studio/tests/e2e/resource/create-page.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    expect(created?.parentId).toBeNull()
+  })
+})
+
+test.describe("publisher", { tag: roleTag("publisher") }, () => {
+  test.beforeEach(async () => {
+    await ensureUserOnboarded(TEST_EMAILS.publisher)
+  })
+
+  test("publisher does not see the Create new button on the site homepage", async ({
```
dcshzj: nit: `root` and not `homepage`

adriangohjw: @cursoragent update

PR #2953, opened 2026-07-25
test(e2e): refactor fixtures, page objects, and convention alignment

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 3 | +8 | -12 |
| 🧪 Test | 110 | +2186 | -1907 |
| 📄 Doc | 2 | +122 | -298 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

The e2e suite had drifted from team conventions. Fixtures lived in flat files. Tests called `page.*` directly. Several flows never checked DB state after a mutation. `tests/e2e/README.md` and `e2e-tests.md` disagreed on structure and on whether `page.*` belongs in test files.

## Solution

Restructure fixtures and page objects to match the conventions doc. No production behavior changes.

**Breaking Changes**

  - Details ...

**Features**

- `~e2e/*` path alias with subpath barrels (`~e2e/fixtures/po`, `~e2e/fixtures/resource`, …). No root `fixtures/index.ts`.
- Entity folders (`resource/`, `site/`, `user/`, `whitelist/`, `role/`) split into `seed.ts`, `db.ts`, and `expect.ts`.
- New POs: `SitesListPO`, `SiteAdminPO`, plus collection/dashboard/editor coverage.
- Shared helpers for wizards, invites, site reset, and `expect.poll` DB checks.

**Improvements**

- `page.*` calls live in POs/helpers only (allowlist in conventions).
- Singpass tests sit in the `unauthenticated` project behind `test.skip`. Dropped the separate `singpass` Playwright project.
- Godmode publishing runs real `site.publish` and asserts the audit log row.
- README and `e2e-tests.md` trimmed to one set of rules.
- Fixture/PO comments cut to short footgun notes. Mockpass login comment lives in one place on `LoginPage`.
- Removed dead PO methods: `SitePO.openSite`, `openSettings`, and unused `PageEditorPO` add-text / meta-settings helpers.
- AAA markers on permission-gate tests.

**Bug Fixes**

- `resetUserSingpassState` per test instead of a singpass `beforeEach` that wiped every user.
- DB polls added for collection edits, schedule-publish, resend-invite, and godmode publishing.
- `LoginPage.waitForAppHomeUrl` throws when `NEXT_PUBLIC_APP_URL` is unset.
- Godmode's `clickPublishForSite` matches its row by site ID again instead of the CodeBuild ID text, since the publishing test clears that ID right before clicking — avoided a race where a future table refetch could make the button unfindable.
- Dropped a redundant post-Act assertion in the resend-invite test that only re-checked a role `resendInvite` never mutates (it just re-sends an email).

## Before & After Screenshots

N/A. Tests and docs only.

## Tests

Run from `apps/studio`:

- `pnpm test:e2e` (needs test DB; `pnpm setup:test` first)
- `pnpm typecheck` (covers `~e2e/*` imports)

**New scripts**: none

**New dependencies**: none


apps/studio/tests/e2e/fixtures/resource/expect.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export const expectResourcePresent = (resourceId: string) =>
+  expect.poll(async () => {
+    const row = await db
+      .selectFrom("Resource")
+      .where("id", "=", resourceId)
+      .select("id")
+      .executeTakeFirst()
+    return row?.id ?? null
+  })
```
dcshzj, line 24: question: think I asked this previously but these two functions are duplicated?

apps/studio/tests/e2e/fixtures/resource/expect.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export const expectResourceParentId = (resourceId: string) =>
+  expect.poll(async () => {
+    const row = await db
+      .selectFrom("Resource")
+      .where("id", "=", resourceId)
+      .select("parentId")
+      .executeTakeFirst()
```
dcshzj, line 32: question: pertains to the entire file but curious why do we inline the DB query for this file but not for the `site` one?

PR #2954, opened 2026-07-26
ci(e2e): cache Playwright browsers, retry flaky tests, and skip standalone tracing for E2E builds

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 3 | +114 | -21 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

E2E CI runs were slower and less reliable than they needed to be:
- Each job that needed Playwright reinstalled the Chromium browser from scratch instead of reusing a cache.
- The E2E build job produced a full standalone Next.js server (file tracing + source maps), even though E2E tests only need `next start`.
- `setup:test` ran a redundant `generate` (Prisma client generation) on top of the one already run earlier in the job, and container setup/teardown wasn't scoped to the `isomer-studio` package.
- Flaky E2E tests failed CI outright with no retry and no video to help debug the failure.

## Solution

**Breaking Changes**

  - Details ...

**Improvements**:

- Added a reusable `.github/actions/playwright-install` composite action that caches Playwright browsers by OS + Playwright version and installs Chromium once; both the Studio and Components jobs in `ci.yml` now call it instead of duplicating the raw install step.
- Introduced an `E2E_BUILD` env var (`apps/studio/src/env.mjs`) so the E2E build step in CI can opt out of `output: "standalone"` / `outputFileTracingRoot` / `productionBrowserSourceMaps` in `next.config.mjs` — E2E uses `next start`, not the standalone server, so this work was unnecessary overhead. Production builds are unaffected since `E2E_BUILD` defaults to `false`.
- Scoped `setup:test`/`teardown` in CI to run via `turbo ... --filter=isomer-studio` instead of the root `pnpm run` scripts, and removed the redundant `generate` step from the `setup:test` script chain in `apps/studio/package.json` since Prisma generation is already handled earlier in the job.
- Added CI-only Playwright retry/video config: `retries: 1` and `video: "on-first-retry"` when `process.env.CI` is set, so a flaky test gets one retry with video captured on that retry (local runs keep `retain-on-failure` behavior unchanged).

## Tests

No manual verification needed — this PR only touches CI workflow/config and is exercised by CI itself on this PR.


adriangohjw:
Closing this — the single-build-then-shard approach doesn't pay off here. Uploading the build artifact once and downloading it in each shard job costs ~1.5 min collectively (upload + N downloads), which eats the time saved by sharding and defeats the purpose of this change.

PR #3021, opened 2026-08-03
fix(studio): allow Backspace to remove empty paragraph above table

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 2 | +48 | -1 |
| 🧪 Test | 2 | +286 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Users could not remove empty space above a table in a Text block using Backspace. After clearing text above a table, an empty paragraph (`{ type: "paragraph", attrs: { dir: null } }`) remained and could only be removed via JSON mode.

Closes https://linear.app/ogp/issue/ISOM-2475/studio-tiptap-unable-to-remove-space-from-1st-line

## Solution

TipTap's default Backspace keymap calls `joinBackward()` but not `deleteCurrentNode()`, so an empty paragraph at the start of a prose block cannot be joined into a preceding table sibling. Delete already worked via `deleteCurrentNode()`, but Backspace did not.

This PR adds `deleteEmptyTextblockBeforeTable`, chained into `IsomerTable`'s Backspace shortcut (before the table extension's own handlers), so Backspace removes an empty textblock when the next sibling is a table.

**Breaking Changes**

**Bug Fixes**:

- Backspace now removes an empty paragraph directly above a table in Text block editors

## Before & After Screenshots

**BEFORE**:

Empty paragraph above table persists after Backspace; only removable via JSON mode.

**AFTER**:

Backspace on the empty paragraph removes it, leaving the table as the first block.

https://github.com/user-attachments/assets/47540394-a9ed-46a7-9f36-b9d32b80b095

## Tests

- [ ] Open a page with a Text block containing a table
- [ ] Add text above the table, then Backspace until the paragraph is empty
- [ ] Press Backspace once more — the empty paragraph should be removed and the table should move up with no extra gap
- [ ] Confirm normal Backspace/typing inside table cells still works
- [ ] Confirm Mod-a cell selection and table deletion via the table picker still work

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

[Slack Thread](https://opengovproducts.slack.com/archives/C06R4DX966P/p1785727994344469?thread_ts=1785727994.344469&cid=C06R4DX966P)

<div><a href="https://cursor.com/agents/bc-381347bd-e01c-5ef7-a257-1c87cd47376c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-381347bd-e01c-5ef7-a257-1c87cd47376c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

The PR fixes the Studio rich-text editor so Backspace can remove an empty first paragraph immediately above a table.
- Adds a guarded command that deletes only an empty paragraph whose next sibling is a table.
- Chains the command ahead of the table extension’s existing Backspace behavior.
- Adds unit and browser-level coverage for deletion, fallback joining, and empty-heading behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/features/editing-experience/hooks/useTextEditor/deleteEmptyTextblockBeforeTable.ts | Adds narrowly guarded deletion of an empty top-level paragraph immediately preceding a table. |
| apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts | Integrates the new deletion command into table Backspace shortcuts while retaining fallback table behavior. |
| apps/studio/src/features/editing-experience/hooks/useTextEditor/__tests__/deleteEmptyTextblockBeforeTable.test.ts | Covers successful deletion and the principal conditions where the helper must decline handling. |
| apps/studio/src/features/editing-experience/hooks/useTextEditor/__tests__/deleteEmptyTextblockBeforeTable.browser.test.ts | Exercises the real EditorView shortcut flow, including backward joining and heading demotion. |

</details>

```mermaid
flowchart TD
  A[Backspace shortcut] --> B{Empty selection in an empty paragraph?}
  B -- No --> F[Delegate to table Backspace handler]
  B -- Yes --> C{First block and next sibling is a table?}
  C -- No --> F
  C -- Yes --> D[Delete current paragraph]
  D --> E[Table becomes first block]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(studio): join backward past content ..."](https://github.com/opengovsg/isomer/commit/94a30c4e5e0c1097e67984c54318b15d7d000910) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49575750)</sub>

**Context used:**

- Knowledge Base — [Resource Tree and Page Editing](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/resource-page-editing.md)

### DIFF
--- a/apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
+++ b/apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
@@ -127,7 +127,7 @@ export const IsomerTable = Table.extend({
 
     const handleBackspace = () =>
       deleteEmptyTextblockBeforeTable(this.editor) ||
-      deleteTableWhenAllCellsSelected?.() ||
+      deleteTableWhenAllCellsSelected?.({ editor: this.editor }) ||
       false
 
     return {
--- a/apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
+++ b/apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
@@ -29,6 +29,7 @@ import {
   createTableSelectionBorderPlugin,
   getHtmlWithRelativeReferenceLinks,
 } from "../../utils"
+import { deleteEmptyTextblockBeforeTable } from "./deleteEmptyTextblockBeforeTable"
 import { selectTableCellContent } from "./selectTableCellContent"
 
 export { TableRow } from "@tiptap/extension-table-row"
@@ -121,10 +122,21 @@ export const IsomerTable = Table.extend({
     }
   },
   addKeyboardShortcuts() {
+    const parentShortcuts = this.parent?.() ?? {}
+    const deleteTableWhenAllCellsSelected = parentShortcuts.Backspace
+
+    const handleBackspace = () =>
+      deleteEmptyTextblockBeforeTable(this.editor) ||
+      deleteTableWhenAllCellsSelected?.() ||
+      false
+
     return {
-      ...this.parent?.(),
+      ...parentShortcuts,
       "Mod-a": () =>
         selectTableCellContent(this.editor) || this.editor.commands.selectAll(),
+      Backspace: handleBackspace,
+      "Mod-Backspace": handleBackspace,
+      "Shift-Backspace": handleBackspace,
     }
   },
   addProseMirrorPlugins() {
--- a/apps/studio/src/features/editing-experience/hooks/useTextEditor/deleteEmptyTextblockBeforeTable.ts
+++ b/apps/studio/src/features/editing-experience/hooks/useTextEditor/deleteEmptyTextblockBeforeTable.ts
@@ -1,0 +1,28 @@
+import type { Editor } from "@tiptap/react"
+import { Table } from "@tiptap/extension-table"
+
+/** Remove an empty textblock when Backspace cannot join backward into a table. */
+export const deleteEmptyTextblockBeforeTable = (editor: Editor): boolean => {
+  const { state } = editor
+  const { empty, $anchor } = state.selection
+
+  if (!empty) {
+    return false
+  }
+
+  const parent = $anchor.parent
+  if (!parent.type.isTextblock || parent.content.size > 0) {
+    return false
+  }
+
+  const parentDepth = $anchor.depth - 1
+  const nextSibling = $anchor
+    .node(parentDepth)
+    .maybeChild($anchor.index(parentDepth) + 1)
+
+  if (nextSibling?.type.name !== Table.name) {
+    return false
+  }
+
+  return editor.commands.deleteCurrentNode()
+}
--- a/apps/studio/src/features/editing-experience/hooks/useTextEditor/deleteEmptyTextblockBeforeTable.ts
+++ b/apps/studio/src/features/editing-experience/hooks/useTextEditor/deleteEmptyTextblockBeforeTable.ts
@@ -1,4 +1,5 @@
 import type { Editor } from "@tiptap/react"
+import { Paragraph } from "@tiptap/extension-paragraph"
 import { Table } from "@tiptap/extension-table"
 
 /** Remove an empty textblock when Backspace cannot join backward into a table. */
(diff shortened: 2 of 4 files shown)


apps/studio/src/features/editing-experience/hooks/useTextEditor/constants.ts
```diff
@@ -121,10 +122,21 @@ export const IsomerTable = Table.extend({
     }
   },
   addKeyboardShortcuts() {
+    const parentShortcuts = this.parent?.() ?? {}
+    const deleteTableWhenAllCellsSelected = parentShortcuts.Backspace
+
+    const handleBackspace = () =>
+      deleteEmptyTextblockBeforeTable(this.editor) ||
+      deleteTableWhenAllCellsSelected?.({ editor: this.editor }) ||
```
adriangohjw, line 130: Why is this needed: we override Backspace on IsomerTable, so the table extension’s default handler (delete the whole table when all cells are selected) would otherwise be lost. We call it explicitly as a fallback after our empty-paragraph fix.

apps/studio/src/features/editing-experience/hooks/useTextEditor/deleteEmptyTextblockBeforeTable.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  const parentDepth = $anchor.depth - 1
+  const nextSibling = $anchor
+    .node(parentDepth)
+    .maybeChild($anchor.index(parentDepth) + 1)
+
+  if (nextSibling?.type.name !== Table.name) {
+    return false
+  }
```
dcshzj, line 32: This guard admits two cases it shouldn't, and both are fixed here.

**1. No previous-sibling check → Backspace regresses when content precedes the empty paragraph.**

`@tiptap/extension-table` declares `priority: 101` against core `keymap`'s `100`, so this new `Backspace` binding *always* preempts core's `joinBackward` chain — it isn't a tie broken by declaration order. For `[paragraph("before"), empty paragraph, table]` all three guards pass, so `deleteCurrentNode()` runs and maps the caret into the table's first cell. The document outcome is identical to `joinBackward` either way — that empty paragraph gets deleted regardless — so the only thing this changes is where the caret ends up.

Verified in real Chromium (real `EditorView`, real `KeyboardEvent` dispatched through `view.someProp("handleKeyDown")`), pressing Backspace and then typing `X`:

| | `main` | this PR |
|---|---|---|
| first block | `"beforeX"` | `"before"` |
| table text | `"Header"` | `"XHeader"` ← the keystroke lands in the table |

**2. `isTextblock` also admits headings.**

For `[empty h2, table]`, `main` runs core's `clearNodes()` and downgrades the heading to a paragraph; this PR deletes the block outright — blocks go `["paragraph", "table"]` → `["table"]`.

Returning `false` in both cases is what keeps core's full chain (`undoInputRule` → `clearNodes` → `deleteSelection` → `joinBackward`) intact. Calling `joinBackward()` from inside this helper instead would *not* — at priority 101 it would skip core's earlier steps.

```suggestion
  const parent = $anchor.parent
  if (parent.type.name !== Paragraph.name || parent.content.size > 0) {
    return false
  }

  const parentDepth = $anchor.depth - 1
  const index = $anchor.index(parentDepth)

  // Only take over when Backspace has nowhere to join backward into. Otherwise
  // defer to the default chain so the caret stays in the preceding block.
  if (index > 0) {
    return false
  }

  const nextSibling = $anchor.node(parentDepth).maybeChild(index + 1)

  if (nextSibling?.type.name !== Table.name) {
    return false
  }
```

This needs `import { Paragraph } from "@tiptap/extension-paragraph"` alongside the existing `Table` import — mirroring the `Table.name` idiom already used below. Note that importing `PARAGRAPH_TYPE` from `./constants` would be circular, since `constants.ts` imports this module.

Both scenarios would also be worth locking in as a `.browser.test.ts`. The current test calls this helper directly and never dispatches a key through a real view, which is why neither case is caught today — and there's already `selectTableCellContent.browser.test.ts` in the same folder to model it on.

adriangohjw: Verified both issues against the codebase — `defaultPriority` in `@tiptap/core` is `100`, confirming `IsomerTable`'s `priority: 101` always preempts core's Backspace chain, and `IsomerHeading`'s `content: "text*"` does make it a ProseMirror textblock.

Applied the suggested diff in 94a30c4e5: guard is now `parent.type.name !== Paragraph.name` (not `isTextblock`) plus the `index > 0` check so it only takes over when there's nothing to join backward into.

Also added `deleteEmptyTextblockBeforeTable.browser.test.ts`, modeled on `selectTableCellContent.browser.test.ts`, dispatching a real `Backspace` `KeyboardEvent` through the `EditorView` to lock in both regressions (caret joining backward past `"before"`, and heading demotion instead of deletion) — the existing unit test calls the helper directly and doesn't exercise either case.

dcshzj (approved):

adriangohjw:
@cursor pull from main (or rebase) and deconflict

PR #2912, opened 2026-07-22
feat(studio): deliver audit exports via sealed download tokens

seaerchin:
### 📊 PR diff breakdown

| Bucket | Files | +Lines | -Lines |
|---|---|---|---|
| ⚙️ App | 13 | +455 | -89 |
| 🧪 Test | 4 | +516 | -14 |
| 📄 Doc | 5 | +161 | -1 |
| 🚫 Ignore | 1 | +50 | -3 |

## Problem

The export worker presigns a "3-day" S3 GET URL and emails it, but SigV4 presigned URLs are capped by the **lifetime of the signing credentials** — the ECS task role's session credentials last ~1 hour, so the emailed link silently dies within the hour. Dead on arrival for anyone reading email later that day.

## Solution

Stop presigning at export time. The ready email now carries a **sealed Download Token** (iron-session `sealData`, AES-256-CBC + HMAC — encrypted *and* unforgeable) containing only `{ purpose: "audit-log-export", requestId }`, pointing at a new unauthenticated redemption route:

```
GET /api/audit-log-exports/download?token=...
```

On click, Studio unseals the token, re-reads the request row (source of truth), and only if the row is `Done` and within its **Download Window** (`completedAt` + `AUDIT_LOG_EXPORT_URL_EXPIRY_DAYS`) mints a fresh ~5-minute presigned URL and 302s to it — comfortably inside any credential session. Every failure mode (forged/tampered/expired token, unknown row, not Done, window elapsed) yields one indistinguishable redirect to a friendly "link expired" page.

Full rationale in `docs/adr/0006-sealed-download-tokens-for-audit-exports.md` (supersedes ADR 0004). Highlights:

- **Window anchors to the request's `completedAt`, not the CSV's age** — under Complete-Artifact reuse (#2832) a fresh request may reuse an older CSV; anchoring to object creation would email dead links.
- **Sealed with the existing session password map** (extracted to a shared `getIronPassword()`), with a `purpose` discriminator + strict shape validation to prevent cross-purpose confusion with session cookies. No new secret to provision; versioned rotation keeps in-flight links alive.
- **Revocable**: the token is a claim ticket, not the goods — flip the row or delete the object and outstanding links die instantly.
- No secrets in the email: no S3 key, no signature, just ciphertext. Email copy unchanged ("expires in 3 days" stays true, now enforced by the window).
[...]

PR #3048, opened 2026-08-04
chore(studio): dogfooding fixes for audit log export

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 13 | +232 | -59 |
| 🧪 Test | 10 | +301 | -73 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 1 | +6 | -0 |

## Problem

Dogfooding the audit log export feature (stacked on #2912) surfaced a handful of rough edges before this ships:

- The month picker let admins request months before their site even existed, which can only ever return empty exports and is confusing.
- Download emails gave no sense of file size, so admins had no way to tell a large export apart from a small one before clicking.
- A few UI details (checkbox hover fill, the "to date" caveat showing for past months, month picker density, card padding) didn't match design polish expectations.
- A duplicated `pg-cursor` key in `package.json` was silently shadowing the real entry.

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Add `audit.getExportWindow` query and `getAuditLogExportWindow`/`getMaxExportableMonths` in `auditLogExport.service.ts`: caps how many months back the export picker offers to `min(AUDIT_LOG_EXPORT_MAX_MONTHS, months since the site was created)`, so a young site never offers months that predate it.
- Wire that cap into `getMonthOptions` (`AuditLogExport/utils.ts`) and `AuditLogExportSection`, so the month `SingleSelect` is populated from the site's real export window instead of always assuming the full 12 months.
- Report the generated/reused CSV's byte size in the "export ready" email (`Download {label} review logs for {month} [.csv, {size}MB]`), computed from the existing S3 `getFileSize` call and logged server-side alongside the object key for observability.
- Pin the Access ("User access review logs") export's date range to the *current* month even when a "Both" request's Activity month is set to a past month — the Access card has no month picker, so it must always reflect current access rather than silently inheriting whatever month was picked for Activity.

**Improvements**:

- Only show the "this will include activities from 1st day of the month to date" caveat when the current (partial) month is selected — past months in the picker are always complete calendar months, so the caveat didn't apply to them.
- Checkbox hover/focus fill on the report-type cards now reaches the card's edges (padding moved onto the `Checkbox` itself instead of a wrapping `Box`); dropped the focus box-shadow ring.
- Denser month picker (`size="xs"`, non-searchable) and `size="sm"` cards to match the rest of the Logs settings design.
- Tightened the "export ready" email markup: fixed a misplaced `<br/>`/`<p>` ordering around the download link, dropped a stray `<br/>` from the failure email, and stopped double-escaping `siteName` (it's already escaped once upstream).
- Fixed a duplicated `pg-cursor` key in `apps/studio/package.json` (was shadowing the real dependency entry) and added the missing `@types/pg-cursor` dev dependency.

**Bug Fixes**:

- N/A beyond the above (this PR is itself a set of fixes to the export feature branch, per dogfooding feedback).

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- `apps/studio/src/features/settings/AuditLogExport/__tests__/utils.test.ts` — `getMonthOptions` respects a custom `maxMonths`.
- `apps/studio/src/features/settings/AuditLogExport/__tests__/AuditLogExportSection.test.tsx` — picker reflects the queried export window; current-month caveat only shows for the current month.
- `apps/studio/src/server/modules/audit/__tests__/auditLogExport.service.test.ts` — `getMaxExportableMonths` clamps to the standard window and to a young site's age.
- `apps/studio/src/server/modules/audit/__tests__/auditLogExport.dedupe.test.ts` — Access export stays pinned to the current month for a "Both" request with a past Activity month.
- `apps/studio/src/features/mail/templates/__tests__/auditLogExport.test.ts` — download link label includes file size; email markup ordering.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- `@types/pg-cursor` : types for the existing `pg-cursor` dependency (was missing due to the duplicate-key bug above)


apps/studio/src/features/mail/templates/templates.ts
```diff
@@ -292,11 +294,11 @@ const accountDeactivationTemplate = (
 const auditLogExportReadyTemplate = (
   data: AuditLogExportReadyEmailTemplateData,
 ): EmailTemplate => {
-  const { recipientEmail, siteName, month, link } = data
+  const { recipientEmail, siteName, month, link, sizeInBytes } = data
 
   const logName = link.label === "access" ? "Access" : "Audit"
 
-  const downloadLink = `<a href="${link.url}">${getDownloadLinkLabel(link.label, month)}</a>`
+  const downloadLink = `<a href="${link.url}">${getDownloadLinkLabel(link.label, month, sizeInBytes ? (sizeInBytes / ONE_MB_IN_BYTES).toFixed(2) : "-")}</a>`
```
adriangohjw, line 301: nit (non-blocking): i think this logic should be handled by `getDownloadLinkLabel`

adriangohjw (approved):

apps/studio/src/features/mail/templates/templates.ts
seaerchin: that was my initial approach but i went with this because i wanted `getDownloadLinkLabel` to just format without additional logic.

PR #3054, opened 2026-08-04
🧹 (DO NOT MERGE) feat(scripts): publish never-published collection index pages

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 6 | +750 | -0 |
| 🧪 Test | 2 | +836 | -0 |
| 📄 Doc | 1 | +193 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Collections created through `collection.create` never get their index page published — the `// TODO: Create the index page for the collection and publish it` at `collection.router.ts:184` is still open. So every such Collection has an `IndexPage` with `publishedVersionId IS NULL`.

The earlier `page.category` → `page.tagCategories` migration wrote `tagCategories` into both published and draft index-page blobs. For these never-published pages that landed **only in the draft blob** — and the site build reads published blobs exclusively.

On rebuild:

1. `tooling/build/scripts/publishing/queries.ts` joins `Version` on `publishedVersionId`, then `Blob` on `Version.blobId`. **`draftBlobId` appears nowhere in `tooling/build/`** — there is no draft fallback.
2. A null `publishedVersionId` means null `content`, so `index.ts` skips the index page: no `schema/<collection>/_index.json`, no sitemap entry, therefore no `collectionPagePageProps.tagCategories`.
3. The build **fabricates a stub** via `getCollectionIndexPageContents` — no `tagCategories`, no `sortOrder`, no `subtitle`.
4. `getCollectionItems.ts` gates on `tagCategories && item.tagged`, so every item gets `tags: undefined`. `getTagFilters.ts` returns `[]`, and `CollectionClient.tsx` hides the filter rail.
5. Articles inside the collection lose header tags (`Article.tsx` reads `parent.collectionPagePageProps?.tagCategories` from the sitemap), as do homepage `CollectionBlock`s.
6. Search ingestion: `schedulePushDocumentJob.ts` INNER JOINs on `publishedVersionId` with `executeTakeFirstOrThrow()` and throws. The throw is swallowed per-row and the job row is deleted regardless, so it is never retried.

Studio hides this because editor preview paths (`getPublishedIndexBlobByParentId`) fall back to the draft.

Closes [insert issue #]

## Solution

A one-off data script at `apps/studio/prisma/scripts/publish-collection-index-pages/`. It publishes a canonical index page for eligible Collection index pages so the build can read `tagCategories` from a real published blob.

**Scope guardrails** — only collections that are already partially live:

- `IndexPage` with `publishedVersionId IS NULL`
- Parent is a `Collection`
- **At least one published `CollectionPage` or `CollectionLink` child** (`publishedVersionId IS NOT NULL`)

Empty collections (no published children) are excluded — they are not on the live site today and should not be taken live by this script.

**Title resolution** matches the site build's dangling-directory stub (`tooling/build/scripts/publishing/index.ts`): Collection `Resource.title` → IndexPage `Resource.title` → permalink slug. Draft `page.title` is **not** read (it can be stale after renames; the stub never uses it).

**Apply flow**

```mermaid
flowchart TD
  A["findNeverPublishedCollectionIndexPages()<br/>IndexPage unpublished · parent=Collection · ≥1 published child"] --> B{"for each row"}

  B --> C["classifyRow()<br/>tagCategories from draft<br/>title: parent → resource → permalink"]
  C --> D["BEGIN TRANSACTION"]

  D --> E{"for each row"}
  E --> F["publishNewBlobVersion()<br/>INSERT Blob + Version<br/>UPDATE Resource"]
  F --> E
  E -->|done| G["COMMIT"]

  G --> H["write report.json"]
  H --> I["site rebuild reads publishedVersionId → Version → Blob"]
```

`publishNewBlobVersion` re-checks `publishedVersionId` under a row lock — that is only a concurrency guard (editor publish or a second script run between query and write), not part of the normal path.

Fixing the draft blob alone would achieve nothing — the build never reads drafts. The fix has to create a published `Version`.

**Breaking Changes**

  - Details ...

No schema change, no migration, no app behaviour change. The script is not run by `migrate:<env>` — it is executed manually.

**Features**:

- New script `publishCollectionIndexPages.ts`. Per eligible row: `createCollectionIndexJson(title)` plus `page.tagCategories` from the draft (verbatim). Inserts a **new** `Blob` and **new** `Version`; sets `Resource.publishedVersionId` and `state = Published`.
- **`draftBlobId` is never touched.** The draft survives byte-for-byte; the dashboard Draft badge stays correct.
- Dry-run and apply modes via prompt. Apply always plans first, then confirms. Optional site ID (blank = all sites). Apply prompts for an **IsomerAdmin email** (`Version.publishedBy`).
- One `report.json` per run in `.out/` (gitignored): totals, `variantFlipCount`, `publishedRows` (rollback data).
- **One transaction** for the full apply run. A mid-run failure rolls back that run; re-run apply to resume (already-published rows drop out of the query).

**Improvements**:

- `state` set to `Published` so Studio "live on end site" queries (`folder.router.ts`, `resource.service.ts` nested-resources CTE) match what the build publishes.
- Idempotency: `publishedVersionId` removes a row from the target query; a second run selects zero rows.
- `NOTE` on `createCollectionIndexJson` in `collection.service.ts` (comment-only production touch).

**Bug Fixes** (after site rebuild):

- Restores collection item categories, filter rail, in-collection article header tags, and `CollectionBlock` tags for affected collections.
- Unblocks search ingestion **except** rows whose draft had no `tagCategories` — `collectionIndexPageContentSchema` still requires that array; those need the earlier category migration re-run.

### Accepted live-site changes

Publishing a real index page **replaces the build's fabricated stub**:

| Field | Live today (stub) | After this script |
| --- | --- | --- |
| `subtitle` | blank | `"Read up-to-date news articles, speeches, and press releases here."` |
| `variant` | from `CollectionMeta.content.variant` | absent ⇒ default `"collection"` = 1-column |

Review `variantFlipCount` in the dry-run report before applying — blog-variant collections flip from 2-column to 1-column.

### Rollback

For each entry in `publishedRows`:

```sql
UPDATE "Resource" SET "publishedVersionId" = NULL, state = '<previousState>' WHERE id = <resourceId>;
DELETE FROM "Version" WHERE id = <newVersionId>;
DELETE FROM "Blob"    WHERE id = <newBlobId>;
```

The draft blob was never modified.

### Notes

- **No audit logging** — trail is the report's `publishedRows` plus this PR.
- **No redirects** — collection URL already resolves via the stub; no permalink changes.
- **Needed again** until `collection.router.ts:184` is fixed — new Collections re-enter the target set.
- Imports `createCollectionIndexJson` directly from `collection.service.ts` (no local copy to drift).

## Before & After Screenshots

Not applicable — data script and a code comment. Observable effect is on published sites after rebuild (see verification steps).

## Tests

`helpers.test.ts` — transform: `tagCategories` verbatim, title resolution (parent → resource → permalink, draft `page.title` ignored), variant flip counting.

`__tests__/runBackfill.test.ts` — integration against Postgres: target query filters (including published-child requirement), write invariants (`draftBlobId` untouched), idempotency, site scoping, dry-run writes nothing.

- [ ] `pnpm run jump:staging`, point `DATABASE_URL` at port 5433, export required vars from `apps/studio/.env`.
- [ ] Run the README cross-check SQL (eligible count with published-child `EXISTS`). Confirm it matches dry-run "Eligible rows".
- [ ] Run the second README query (published index blobs missing `tagCategories`). Confirm **0** — non-zero means re-scope before applying.
- [ ] `source .env && pnpm exec tsx prisma/scripts/publish-collection-index-pages/publishCollectionIndexPages.ts` — `dry-run`, single site ID. Read `.out/<timestamp>-site-<id>.report.json`.
- [ ] Review `variantFlipCount`. Confirm 2-column → 1-column blast radius is acceptable.
- [ ] Re-run in `apply` mode for the same site; supply an IsomerAdmin email. Confirm published count matches `toPublish`.
- [ ] In Studio: affected collection index page still shows **Draft** badge; collection item tag picker lists categories.
- [ ] Trigger site rebuild. On live site: filter rail renders, item categories and article header tags appear.
- [ ] Confirm collection `subtitle` on live page (expected default copy — flag to content team if wrong for that site).
- [ ] Re-run `apply` for same site → **0 eligible rows**.
- [ ] All-sites run: dry-run → review → apply → rebuild. Pair with another engineer for production.
- [ ] Keep final `report.json` for rollback.

**New scripts**:

- `apps/studio/prisma/scripts/publish-collection-index-pages/publishCollectionIndexPages.ts` — one-off backfill; idempotent, dry-runnable; not run by `migrate:<env>`.

**New dependencies**:

None.

**New dev dependencies**:

None.


adriangohjw:
@greptileai review pls

PR #3055, opened 2026-08-04
chore: harden supply-chain security across pnpm, CI, and containers

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 16 | +115 | -6 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

N/A - not tied to a specific issue. Proactive supply-chain hardening: most real-world compromises (Shai-Hulud, s1ngularity, the chalk/debug attack) execute via an unpinned lifecycle-script invocation, a mutable CI reference, or a stolen maintainer credential — this repo had several of these gaps despite already having good foundations (SHA-pinned Actions, Dependabot cooldowns, `blockExoticSubdeps`, `allowBuilds`).

## Solution

**Breaking Changes**

  - Details ...
  - One behavioral note: `trustPolicy: no-downgrade` and `minimumReleaseAge` will now fail `pnpm install` if a future dependency bump resolves to a version with a lower publish-trust level than its own prior releases, or one that hasn't cleared the 24h cooldown. This is the intended effect of the hardening, but it means an occasional dependency-bump PR may need a documented `trustPolicyExclude`/`minimumReleaseAgeExclude` entry after investigation (see precedent added in this PR).

**Improvements**:

- Removed an unpinned `pnpm dlx uuid` call from the devcontainer's `postCreateCommand` — replaced with Node's built-in `crypto.randomUUID()`, removing a third-party dependency entirely instead of just pinning it.
- `pnpm-workspace.yaml`: added `registries.default` (so a dev's global config/env var can't silently redirect resolution), `minimumReleaseAgeIgnoreMissingTime: false` (closes a cooldown bypass for registries with missing publish-time metadata), and `trustPolicy: no-downgrade` (fails install if a package's publish-trust level drops versus its own prior releases — the signal a stolen maintainer credential produces).
  - Enabling `trustPolicy` flagged `algoliasearch@4.27.0` (+ its `@algolia/*` deps) and `tailwind-merge@2.6.1` as trust downgrades. Investigated both against upstream GitHub commit/tag history before excluding — neither is a compromise:
    - The `algoliasearch` 4.x line (deliberately pinned here already for API-compatibility with the legacy egazette integration) has never used provenance across its whole history, published by long-standing listed maintainers. The "downgrade" is only relative to 5.x's newer OIDC-based publishing.
    - `tailwind-merge@2.6.1` is a real commit by the actual maintainer (verified commit SHA + GitHub tag), a genuine backport bugfix to the old 2.x line, published from a personal npm token the same day the maintainer migrated their publish workflow to GitHub Actions OIDC (hence no provenance on that one release while every release since has it).
  - Both are documented in `trustPolicyExclude` with the verification findings, pinned to the exact flagged version (not bare package names).
- Added a `supply-chain` CI job (`ci.yml`) that runs `pnpm audit signatures` and generates a CycloneDX SBOM artifact — turns "were we exposed, and in which build?" into an artifact lookup instead of lockfile archaeology.
- Added least-privilege `permissions:` blocks to jobs across `ci.yml`, `bundle-size.yml`, `chromatic.yml`, `neon_cleanup.yml`, and `pr-breakdown.yml` that previously had none (defaulting to the repo's broader GITHUB_TOKEN grant). Most just needed `contents: read` for checkout; the Graphite CI-optimizer and Neon branch-cleanup jobs got `permissions: {}` after verifying against their `action.yml` that neither ever receives a `github_token` — they only use a Graphite/Neon-specific secret.
- Pinned `apps/studio/Dockerfile`'s base image (`node:24-alpine`) by digest instead of a mutable tag, plus the devcontainer Dockerfile and local/test-only `docker-compose.yml` images (`postgres`, `mockpass`) for consistency. Added a Dependabot `docker` ecosystem entry so these digest pins get refreshed instead of going stale.
- Marked 4 internal-only tooling packages (`seed-from-repo`, `migrate-tags`, `export-import`, `site-launch`) as `"private": true` — they had no `publishConfig` and no public API, unlike every other non-published workspace package.

**Bug Fixes**:

- N/A (see unpinned `dlx` removal above, filed under Improvements since it's hardening rather than a functional bug)

## Before & After Screenshots

N/A - this PR only touches CI/CD configuration, `pnpm-workspace.yaml`, Dockerfiles, and `package.json` metadata. No UI changes.

## Tests

Verified locally before opening this PR:

- `pnpm install --frozen-lockfile` - passes cleanly with `trustPolicy`/`minimumReleaseAge` enabled (after the two documented exclusions above)
- `pnpm run lint:ws` (sherif) - no issues
- `pnpm run lint` - passes
- `pnpm run typecheck` - passes
- `pnpm run format` - passes (one unrelated pre-existing failure in `packages/db`'s generated files, confirmed via clean `git diff` to be untouched by this PR)
- All edited workflow YAML files validated for syntax
- `docker compose config` validated on both edited `docker-compose.yml` files with the new digest pins
- Confirmed the pinned image digests (`node:24-alpine`, the devcontainer base image, `postgres:15.10`, `opengovsg/mockpass:4.5.1`) resolve correctly via `docker pull`/`docker manifest inspect`, and validated the `FROM ...@sha256:... AS base` + comment syntax against a scratch Dockerfile (a same-line trailing comment breaks Docker's parser on a multi-arg `FROM`; a comment on its own line above does not)
- Did **not** get a clean full build of `apps/studio/Dockerfile` locally — the build environment ran out of disk space mid-session (host-level issue, unrelated to this change) before that could complete. Worth a deployer running `docker build -f apps/studio/Dockerfile .` once before merge as an extra check, since the base-image line itself is the only part of that file this PR touches

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None


adriangohjw:
@cursor pull from main (or rebase) and resolve conflict

PR #3057, opened 2026-08-05
feat(db): add concurrent index on Version(resourceId, versionNum)

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 3 | +32 | -0 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Summary
Adds a concurrent index on `Version(resourceId, versionNum DESC)` to support `getLatestVersionByResourceId`, which today runs a full table scan on every publish (and will run on unpublish/re-lock once those ship).

`Version` is large, append-only, and shared across every site, so the index is built via `CREATE INDEX CONCURRENTLY` (hand-written in `packages/db/prisma/custom/migration.sql`, since Prisma's schema DSL has no way to request `CONCURRENTLY`) to avoid locking writers during the build.

## Test plan
- [ ] Confirm `pnpm generate` / `prisma migrate deploy` picks up the new migration cleanly in a fresh DB
- [ ] Confirm `EXPLAIN ANALYZE` on `getLatestVersionByResourceId`'s query uses the new index post-deploy

### DIFF
--- a/packages/db/prisma/custom/migration.sql
+++ b/packages/db/prisma/custom/migration.sql
@@ -17,3 +17,16 @@ DROP INDEX "User_email_deletedAt_key";
 CREATE UNIQUE INDEX IF NOT EXISTS "User_email_deletedAt_key" ON "User"("email", "deletedAt") NULLS NOT DISTINCT;
 
 ---------------------------------
+
+
+---------------------------------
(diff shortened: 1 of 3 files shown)


Ssatu:

PR #3060, opened 2026-08-05
feat(studio): add getLatestVersionByResourceId helper for version service

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 1 | +14 | -0 |
| 🧪 Test | 1 | +201 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Summary
Adds `getLatestVersionByResourceId`, which fetches a resource's latest `Version` row regardless of current publish state. Unlike following `Resource.publishedVersionId`, this distinguishes "never published" from "published before, now unpublished" — the latter still has historical `Version` rows even though `publishedVersionId` is null. This is the helper e3's unpublish/re-lock logic depends on.

Stacked on #3059 (e2).

## Test plan
- [x] Unit tests cover: no versions, single version, highest-versionNum-not-most-recent, unpublished-with-history, and resource scoping

### DIFF
--- a/apps/studio/src/server/modules/version/__tests__/version.service.test.ts
+++ b/apps/studio/src/server/modules/version/__tests__/version.service.test.ts
@@ -1,0 +1,167 @@
+import { resetTables } from "tests/integration/helpers/db"
+import {
+  setupBlob,
+  setupPageResource,
+  setupUser,
+} from "tests/integration/helpers/seed"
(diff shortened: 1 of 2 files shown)


Ssatu:

PR #3066, opened 2026-08-05
feat(studio): implement unpublishPageResource and archiveDraftResource as separate service functions

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 6 | +304 | -3 |
| 🧪 Test | 1 | +542 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Summary
Backend core of the unpublish/archive feature. Two independent, single-purpose functions rather than one branching on prior resource state — the changelog's explicit decision, since the two paths' risk profiles differ meaningfully:

- **`unpublishPageResource`** (heavy — `Published -> Archived`): guards on `publishedVersionId` already null, copy-on-write backfills `draftBlobId` from the published Version's content (never points `draftBlobId` directly at a Version's blob — `updateBlobById` mutates in place, which would corrupt immutable published history), clears `publishedVersionId`, sets `state = Archived`, logs `AuditLogEvent.Unpublish`, then rebuilds the site outside the transaction.
  - Takes a `preserveContent: "draft" | "published"` choice, required whenever a WIP draft already exists at call time: `"published"` overwrites it with the published content (the only prior behavior), `"draft"` leaves it untouched. Ignored when no draft exists. Recorded in the audit log's `metadata`.
- **`archiveDraftResource`** (light — `Draft -> Archived`, new): just flips `state`, nothing else — covers both "archive a never-published draft" and "cancel Edit this page" as the identical mutation, no history-lookup gating needed.

New `archiveDraft` router mutation alongside the existing `unpublishPage`, both gated on the same `"unpublish"` base permission action.

Known gap, flagged inline (not fixed here — see risk #1 in the effort budget): neither function's initial resource read is `.forUpdate()`-locked, so concurrent publish/unpublish/archive on the same resource can race. This needs a coordinated fix across every Resource-mutating path, not a partial one in this PR.

Stacked on #3060 (pr2).

## Test plan
- [x] Unit tests: guard rejections (never-published, already-archived), successful unpublish (blob clone correctness, Version untouched), `preserveContent` choice (published/draft/conflict-without-choice), `archiveDraft` (never-published draft, post-unpublish "cancel edit" case), blob-corruption regression (edit-after-unpublish doesn't touch the historical Version's blob)

### DIFF
--- a/apps/studio/src/schemas/page.ts
+++ b/apps/studio/src/schemas/page.ts
@@ -90,6 +90,21 @@ export const publishPageSchema = z.object({
   siteId: z.number().min(1),
 })
 
+// Takes a live page down: Published -> Archived.
+export const unpublishPageSchema = z.object({
+  pageId: z.number().min(1),
+  siteId: z.number().min(1),
+  // Required if an in-progress draft exists: keep it, or overwrite with the
+  // published content. Ignorable otherwise.
+  preserveContent: z.enum(["draft", "published"]).optional(),
+})
+
+// Shelves a Draft page that was never live: Draft -> Archived.
+export const archiveDraftSchema = z.object({
+  pageId: z.number().min(1),
+  siteId: z.number().min(1),
+})
+
 export const createCollectionPageFormSchema = z
   .discriminatedUnion("type", [
     z.object({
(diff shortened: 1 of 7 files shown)


Ssatu:

PR #3067, opened 2026-08-05
feat(page): add editThisPage mutation and guard updatePageBlob against editing archived pages

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 2 | +92 | -1 |
| 🧪 Test | 1 | +205 | -1 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Summary
The only exit from `Archived`: `editThisPage` flips `state` back to `Draft`, leaving `publishedVersionId`/`draftBlobId` untouched — once flipped, the page behaves like an ordinary Draft page.

`updatePageBlob` now rejects edits on an Archived page with `PRECONDITION_FAILED`, with a non-fatal contradiction-check log if `publishedVersionId` is unexpectedly non-null in that state (should be structurally impossible).

Also exposes `state` on `readPageAndBlob`'s output — the main page-editor route sources its data exclusively from this query, and previously had no way to know a page was Archived without an extra round-trip. Lets the frontend gate the editor UI proactively instead of only discovering it via `updatePageBlob`'s guard on save.

Stacked on #3066 (e3).

## Test plan
- [x] Unit tests: `editThisPage` (401/403/404, reject on non-Archived state, successful Archived -> Draft flip with untouched publishedVersionId/draftBlobId), `updatePageBlob` guard (blocks edits when Archived), `readPageAndBlob` returns `state`

### DIFF
--- a/apps/studio/src/schemas/page.ts
+++ b/apps/studio/src/schemas/page.ts
@@ -198,3 +198,6 @@ export const getPrefillSchema = z.object({
   siteId: z.number().min(1),
   resourceId: z.string().regex(/^\d+$/),
 })
+
+// "Edit this page": the only exit from Archived (Archived -> Draft).
+export const editThisPageSchema = basePageSchema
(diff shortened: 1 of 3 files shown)


Ssatu:

PR #3068, opened 2026-08-05
fix(studio): key incrementVersion off Version history, not publishedVersionId

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 1 | +6 | -12 |
| 🧪 Test | 1 | +135 | -1 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Summary
`incrementVersion` derived `newVersionNum`/`previousVersion` from `page.publishedVersionId`, which is null both for a resource that's never been published and one that was published before but has since been unpublished. On a republish in that second case it would wrongly reset `versionNum` back to 1 and drop `previousVersion`, colliding with the resource's real Version history.

Fix: use `getLatestVersionByResourceId` (keyed on the resource's actual `Version` rows) instead, so "never published" and "published before, now archived" are distinguished correctly.

Fixes `publishPageResource`'s audit-log `delta.before` field for free, since it's derived directly from `incrementVersion`'s `previousVersion` return value with no other logic in between.

Removes the now-unused `getVersionById` helper (and its `db` import), superseded by `getLatestVersionByResourceId`.

Stacked on #3067 (e4).

## Test plan
- [x] Unit test: republishing after an unpublish continues the real version history instead of resetting to 1

### DIFF
--- a/apps/studio/src/server/modules/version/__tests__/version.service.test.ts
+++ b/apps/studio/src/server/modules/version/__tests__/version.service.test.ts
@@ -7,7 +7,10 @@ import {
 import { ResourceType } from "~prisma/generated/prisma/client"
 
 import { db, ResourceState } from "../../database"
-import { getLatestVersionByResourceId } from "../version.service"
+import {
+  getLatestVersionByResourceId,
+  incrementVersion,
+} from "../version.service"
 
 describe("version.service", () => {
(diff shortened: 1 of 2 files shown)


Ssatu:

PR #3079, opened 2026-08-06
chore: add posthog events for key user actions

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 23 | +143 | -23 |
| 🧪 Test | 2 | +38 | -2 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

We had no visibility into how users interact with key flows in Studio — content creation (pages, folders, collections, gazettes), publishing, settings, redirects, and user management. Without these events, we can't measure adoption of these features or diagnose drop-off in the flows that matter most to editors and admins.

## Solution

Instrumented PostHog `capture` calls at the point of user intent (button/menu click) across the app, rather than only on request success, so we can see when users *attempt* an action, not just when it completes.

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- Added PostHog events for content creation: `page_create_modal_opened`, `folder_create_modal_opened`, `collection_create_modal_opened`, `collection_page_create_modal_opened`, `gazette_create_modal_opened` — each tagged with `site_id` (and `parent_type` where the same action can be triggered from both the site root and a folder)
- Added `publish_modal_opened` and `scheduled_publish_modal_opened` events on the publish button
- Added `page_changes_saved` event fired on successful page blob save, added consistently across all editor state drawers (Root, Complex, Collection, Database, Metadata, Hero, RawJson, SiderailOrdering)
- Added `settings_saved` event on the settings header save button, tagged with `site_id` and `settings_section` (derived from the route)
- Added `redirect_bulk_upload_modal_opened` event on the redirects bulk-upload CTA
- Added `add_user_modal_opened` and `user_invite_resent` events in user management
- Added `user_logged_in` (tagged by `method`: `email` or `singpass`) and `user_logged_out` events in the auth flow
- Added `site_selected` event when a user clicks into a site from the dashboard site list

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

Manually verified events fire in the PostHog debug panel when clicking through each instrumented action (site selection, page/folder/collection/gazette creation, publish, scheduled publish, settings save, redirect bulk upload, add user, resend invite, login, logout).

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

### DIFF
--- a/apps/studio/src/features/dashboard/SiteList.tsx
+++ b/apps/studio/src/features/dashboard/SiteList.tsx
@@ -12,6 +12,7 @@ import {
 } from "@chakra-ui/react"
 import { Link } from "@opengovsg/design-system-react"
 import NextLink from "next/link"
+import posthog from "posthog-js"
 import { NoResultIcon } from "~/components/Svg/NoResultIcon"
 import { ISOMER_SUPPORT_LINK } from "~/constants/misc"
 import { withSuspense } from "~/hocs/withSuspense"
@@ -31,7 +32,11 @@ const Site = ({
 }): JSX.Element => {
   return (
     <LinkBox cursor="pointer" role="group">
-      <LinkOverlay href={`/sites/${siteId}`} as={NextLink}>
+      <LinkOverlay
+        href={`/sites/${siteId}`}
+        as={NextLink}
+        onClick={() => posthog.capture("site_selected", { site_id: siteId })}
+      >
         <Flex key={siteId} flexDirection="column" gap="1rem" width="100%">
           <Box position="relative">
             <Image
--- a/apps/studio/src/features/auth/LoginStateContext.tsx
+++ b/apps/studio/src/features/auth/LoginStateContext.tsx
@@ -55,6 +55,11 @@ const PostHogIdentity = () => {
   const { data: user } = trpc.me.get.useQuery(undefined, {
     enabled: hasLoginStateFlag,
   })
+  // Site-wide (siteId, role) pairs for cohorting users in PostHog by the
+  // permissions they hold — see `site.list` for how roles are resolved.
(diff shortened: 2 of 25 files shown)


apps/studio/src/features/dashboard/SiteList.tsx
```diff
@@ -31,7 +32,11 @@ const Site = ({
 }): JSX.Element => {
   return (
     <LinkBox cursor="pointer" role="group">
-      <LinkOverlay href={`/sites/${siteId}`} as={NextLink}>
+      <LinkOverlay
+        href={`/sites/${siteId}`}
+        as={NextLink}
+        onClick={() => posthog.capture("site_selected", { site_id: siteId })}
```
dcshzj, line 38: suggestion: I'm ok with this in principle but I think it might be quite useful if we have an intermediate layer to capture these events which then does the pass-through to posthog. Imagine something like this where we do `() => captureEvent("site_selected", { site_id: siteId })`, then the implementation of `captureEvent` is to just pass the props to posthog.

Reason being if in the future we intend to expand the events collection to pass to a new provider (Intercom or DataDog), we can just make use of the same call site and propagate this cheaply. We don't get too tied into a single provider as well.

dcshzj (approved):

apps/studio/src/features/dashboard/SiteList.tsx
seaerchin: actually i agree - wdyt of us having a button abstraction that does this?

dcshzj: I think button no need bah, we not too sure yet to what extent we can capture the events, can explore in future PRs.

PR #3087, opened 2026-08-07
fix(redirects): hide write controls from non-admins, polish wildcard hint

dcshzj:
> The folder destination-warning fix that was originally part of this PR has moved to #3124, stacked on top of this one.

## Problem

Three issues raised in the "Bulk upload & wildcard redirects" dogfooding session, plus two robustness problems found while fixing them.

1. **Editors saw redirect controls they cannot use.** `redirect.create` / `redirect.delete` require `create`/`delete` on `Site`, which only site admins hold. The settings page rendered the add-redirect form, the bulk-upload link, the `.csv` template and the row delete button to everyone, so a non-admin could fill in the form and download the template only to be met with `FORBIDDEN` on submit.
2. **"Section" is not a term the CMS uses.** The wildcard hint read "redirect a whole section", which does not map onto anything in the site tree.
3. **The wildcard hint sat flush against the input**, reading as part of the field rather than as a caption.
4. **A malformed `Resource.parentId` chain could hang a request.** `getResourceFullPermalinks` recursed with `unionAll` and the in-memory ancestor walk it feeds had no visited guard, so a redirect destination stored as a `[resource:...]` reference pointing into a cycle hung `resolveReferences` outright.
5. **A flaky audit-log story** was failing Chromatic on this branch.

Closes N/A — items 1–3 were raised on the dogfooding feedback board rather than as GitHub issues.

## Solution

**Breaking Changes**

**Improvements**:

- The add-redirect card and the delete column render only for site admins, gated on a `RedirectManagement` ability of their own rather than borrowing `UserManagement`. The two resolve the same way today, but they are separate concerns and a named context says what the check means. Read access still sees the full table.
- The roles query behind that gate turned "loading" and "failed" into an empty ability, which is indistinguishable from a genuine read-only role — so an admin saw a read-only page transiently on load, and permanently on failure, with nothing explaining why. The context now surfaces `isPending` and `isError`: a skeleton while roles load, an explicit message when they fail.
- Table rows are pinned to a fixed height, so dropping the delete column no longer makes a non-admin's rows about 20px shorter than an admin's.
- The wildcard hint now reads "Add /\* at the end to redirect all pages within a folder or collection (e.g. /old-news/\*)", spaced off the input to match the label-to-input gap. The spacing has to go through `sx`: the design system's `FormHelperText` merges the theme's `helperText` styles into `sx`, and Chakra's `sx` outranks style props, so an `mt` prop was being silently dropped.

**Bug Fixes**:

- `getResourceFullPermalinks` is now cycle-safe on both counts — the CTE uses `union` and the ancestor walk stops at the first id seen twice, matching `getResourcePermalinkTree` and `withResourceSubtree`, which already dedupe for this reason. Nothing creates a cycle today (moving a folder into its own descendant is rejected), so this is defence against malformed data.
- The audit-log `ExportRequested` story asserted `toBeVisible()` on a toast with a 3s duration immediately after finding it, so on a loaded CI machine the toast could begin tearing down between the find and the assertion. It now asserts presence only, matching the convention already used for the same reason in `Redirects.stories.tsx`. That story is unrelated to this branch — it only started being snapshotted here because the permissions util changed, pulling it into Chromatic's TurboSnap set.

## Before & After Screenshots

**BEFORE**: every role saw the "Add new redirects" card, the bulk-upload link and the per-row delete icons; the wildcard hint read "redirect a whole section" and sat flush against the input.

**AFTER**: non-admins see only the redirects table, at the same row height an admin sees; admins see the card, with a skeleton while permissions load. The hint reads "all pages within a folder or collection" with a 0.75rem gap above it.

## Tests

`RedirectsSettings.browser.test.tsx` covers the admin and non-admin views, the hint copy, row-height parity between the two roles, and the roles-query loading and error states. It runs in Browser Mode because it asserts real layout, and because the design system's CJS `FormHelperText` and Chakra's ESM `FormControl` resolve to different module instances under jsdom. `redirect.router.test.ts` adds a test proving a cyclic parent chain terminates instead of hanging.

- [ ] As a **site admin**, open Settings → Redirects. Confirm the "Add new redirects" card, the "bulk upload with a .csv instead" link and the per-row delete icons are all present.
- [ ] As an **Editor or Publisher** on the same site, open the same page. Confirm the add card, the bulk-upload link and the entire delete column are gone, and that the table still lists every redirect.
- [ ] Compare the two views and confirm the table rows are the same height.
- [ ] Reload as an admin on a slow connection (throttle in devtools) and confirm a skeleton appears where the add card goes, rather than the page briefly presenting itself as read-only.
- [ ] With `is-advanced-redirects-enabled` on, confirm the hint under "When someone visits" reads "Add /* at the end to redirect all pages within a folder or collection (e.g. /old-news/*)" and has a clear gap above it.
- [ ] Confirm adding and deleting a redirect as a site admin still works end to end.

**New scripts**: none

**New dependencies**: none

The redirect-management update adds clearer permission-state handling and safer resource traversal for redirect resolution.
</details>

No blocking failure remains.

No accepted blocking failure remains.
</details>

**What T-Rex did**
- Initial state verified: Admin roles load successfully and the Redirects page shows a Delete control for /old-news.
- After attempting a background refetch, the Admin roles are retained but the page displays that adding/removing redirects is unavailable, while the Delete redirect for /old-news remains rendered.
- The evaluation confirms the P2 hypothesis based on the after-state observations.

<a href="https://app.greptile.com/trex/runs/18445923/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (4): Last reviewed commit: ["test(audit-log): stop the export story r..."](https://github.com/opengovsg/isomer/commit/ce42b21a1a9124a4166e098a617b62763e495543) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51082155)</sub>

### DIFF
--- a/apps/studio/src/features/settings/Redirects/RedirectManagementContext.tsx
+++ b/apps/studio/src/features/settings/Redirects/RedirectManagementContext.tsx
@@ -1,0 +1,81 @@
+import type { PropsWithChildren } from "react"
+import type { RedirectManagementAbility } from "~/server/modules/permissions/permissions.type"
+import { Ability } from "@casl/ability"
+import { createContext, useContext, useMemo } from "react"
+import { buildRedirectManagementPermissions } from "~/server/modules/permissions/permissions.util"
+import { trpc } from "~/utils/trpc"
+
+interface RedirectManagement {
+  ability: RedirectManagementAbility | Ability
+  // Whether the roles behind `ability` are still loading, or failed to load.
+  // Both leave `ability` permitting nothing, which is indistinguishable from a
+  // genuine read-only role unless the consumer can tell them apart — so it is
+  // surfaced rather than collapsed into the ability.
+  isPending: boolean
+  isError: boolean
+}
+
+export const RedirectManagementContext = createContext<RedirectManagement>({
+  // A dummy ability that permits nothing, so a consumer mounted outside the
+  // provider falls back to read-only rather than to full access.
+  ability: new Ability(),
+  isPending: false,
+  isError: false,
+})
+
+interface RedirectManagementProviderProps {
+  siteId: number
+}
+
+// Reads the same site-wide roles query the other permission providers use, so
+// it costs no extra request — React Query serves it from the shared cache.
+export const RedirectManagementProvider = ({
+  siteId,
+  children,
+}: PropsWithChildren<RedirectManagementProviderProps>) => {
+  const {
+    data: roles,
+    isPending,
+    isError,
+  } = trpc.resource.getRolesFor.useQuery({
+    siteId,
+    resourceId: null,
+  })
+
+  const value = useMemo(
+    () => ({
+      ability: roles
+        ? buildRedirectManagementPermissions(roles)
+        : new Ability(),
+      isPending,
+      isError,
+    }),
+    [roles, isPending, isError],
+  )
+
+  return (
+    <RedirectManagementContext.Provider value={value}>
+      {children}
+    </RedirectManagementContext.Provider>
+  )
+}
+
+interface UseRedirectManagementResult {
+  // Whether the current user may add or remove redirects. The server enforces
+  // the same rule on `redirect.create` / `redirect.bulkCreate` /
+  // `redirect.delete`; this only decides whether we offer the controls at all.
+  // False while the roles are unknown, so callers must check `isPending` and
+  // `isError` before presenting the page as read-only.
+  canManageRedirects: boolean
+  isPending: boolean
+  isError: boolean
+}
+
+export const useRedirectManagement = (): UseRedirectManagementResult => {
+  const { ability, isPending, isError } = useContext(RedirectManagementContext)
+  return {
+    canManageRedirects: ability.can("manage", "RedirectManagement"),
+    isPending,
+    isError,
+  }
+}
--- a/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
+++ b/apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
@@ -44,6 +44,7 @@ import {
   useListRedirects,
   useResolveRedirectReferences,
 } from "../api"
+import { useRedirectManagement } from "../RedirectManagementContext"
 import {
   formatAddedAt,
   getDestinationDisplay,
@@ -64,6 +65,16 @@ const MISSING_PAGE_LABEL = "Page no longer exists"
 // specific segment visible ("/promo/really-long…/end.html") instead of losing
 // the tail to a plain end-ellipsis.
 const TRUNCATION_TAIL_LENGTH = 10
+
+// Body rows are pinned to this height so they don't depend on which columns are
+// rendered. The delete IconButton (2.5rem) is what sets the row height for a
+// site admin; without that column, rows for everyone else would come out ~20px
+// shorter. That's 2.5rem plus the theme's 0.5rem vertical cell padding top and
+// bottom — and one more pixel for the collapsed row border, which the cell's
+// border-box height has to cover too. Without that pixel the button no longer
+// fits and the delete cell alone grows 1px, putting the two views back out of
+// step.
+const ROW_HEIGHT = "calc(3.5rem + 1px)"
 
 const splitForMiddleTruncation = (value: string) => {
   const splitAt = Math.max(0, value.length - TRUNCATION_TAIL_LENGTH)
@@ -281,9 +292,13 @@ function SourceCell({ source }: { source: string }): JSX.Element {
   )
 }
 
+// The delete column is only built for a user who may actually delete — deleting
+// is site-admin-only server-side, so showing the button to anyone else only
+// yields a FORBIDDEN toast.
 const getColumns = (
   onDeleteClick: (row: RedirectRow) => void,
   infoByDestination: Map<string, ResolvedDestination>,
+  canDelete: boolean,
 ) => [
   columnsHelper.accessor("source", {
     minSize: 250,
@@ -330,26 +345,36 @@ const getColumns = (
       </Text>
     ),
   }),
-  columnsHelper.display({
-    id: "delete",
-    size: 80,
-    enableSorting: false,
-    header: () => (
-      <TableHeader textStyle="subhead-2" color="base.content.medium">
-        <VisuallyHidden>Actions</VisuallyHidden>
-      </TableHeader>
-    ),
-    cell: ({ row }) => (
-      <IconButton
-        aria-label={`Delete redirect for ${row.original.source}`}
-        icon={<BiTrash />}
-        variant="clear"
-        colorScheme="critical"
-        size="sm"
-        onClick={() => onDeleteClick(row.original)}
-      />
-    ),
-  }),
+  ...(canDelete
+    ? [
+        columnsHelper.display({
+          id: "delete",
+          size: 80,
+          enableSorting: false,
+          header: () => (
+            <TableHeader textStyle="subhead-2" color="base.content.medium">
+              <VisuallyHidden>Actions</VisuallyHidden>
+            </TableHeader>
+          ),
+          // The button is inline-flex, so on a text baseline it claims a hair
+          // more than its own 2.5rem and rounds the row up past ROW_HEIGHT.
+          // A flex wrapper takes it out of the line box, keeping this row
+          // exactly as tall as one without the column.
+          cell: ({ row }) => (
+            <Box display="flex" alignItems="center" h="2.5rem">
+              <IconButton
+                aria-label={`Delete redirect for ${row.original.source}`}
+                icon={<BiTrash />}
+                variant="clear"
+                colorScheme="critical"
+                size="sm"
+                onClick={() => onDeleteClick(row.original)}
+              />
+            </Box>
+          ),
+        }),
+      ]
+    : []),
 ]
 
 interface RedirectsTableProps {
@@ -359,6 +384,11 @@ interface RedirectsTableProps {
 export const RedirectsTable = ({
   siteId,
 }: RedirectsTableProps): JSX.Element => {
+  // While the roles are unknown we can't tell whether the delete column
+  // belongs, so the table stays in its loading state rather than rendering a
+  // column set it may have to change a moment later.
+  const { canManageRedirects: canDelete, isPending: isPermissionPending } =
+    useRedirectManagement()
   const toast = useToast(BRIEF_TOAST_SETTINGS)
   const { mutate: deleteRedirect, isPending } = useDeleteRedirect()
   // Newest redirects first, matching the design's default sort on "Added"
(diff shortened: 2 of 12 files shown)


apps/studio/src/features/settings/Redirects/index.tsx
```diff
@@ -10,14 +12,24 @@ interface RedirectsSettingsProps {
...
-  <Stack spacing="1.5rem" px="2rem" py="1.5rem" w="full">
-    <RedirectsHeader />
+}: RedirectsSettingsProps): JSX.Element => {
+  // Adding and removing redirects are site-wide actions the server grants only
+  // to site admins (`create`/`delete` on Site). Mirror that here off the same
+  // admin ability the sidenav uses, so a non-admin never gets inputs, a .csv
+  // template, or a delete button whose submit can only come back FORBIDDEN.
+  // Read-only access still sees the table.
+  const ability = useContext(UserManagementContext)
+  const canManageRedirects = ability.can("manage", "UserManagement")
```
adriangohjw: issue: this works but i think it's very much worth defining a new context so that we 1) don't couple these 2 together (even though right now its the same) and 2) it's much more readable and save us the trouble of a comment explaining so

i think its cheap enough to add/maintain a separate RedirectManagementContext

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
```diff
@@ -359,6 +384,11 @@ interface RedirectsTableProps {
 export const RedirectsTable = ({
   siteId,
 }: RedirectsTableProps): JSX.Element => {
+  // Same site-admin ability that gates the add-redirect card (see index.tsx) —
+  // read here rather than passed down so the table owns which columns it can
+  // render.
+  const ability = useContext(UserManagementContext)
+  const canDelete = ability.can("manage", "UserManagement")
```
adriangohjw: same as the other comment on having its own context

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -175,19 +180,117 @@ const getPublishedStateByResourceIds = async (
...
+    .where("Resource.id", "in", notLiveIds)
+    .where("Resource.type", "in", [
+      ResourceType.Folder,
+      ResourceType.Collection,
+    ])
+    .select("Resource.id")
+    .execute()
+  if (undecided.length === 0) {
+    return state
+  }
```
adriangohjw: question: why is this needed ah? if the goal is to pre-filter and avoid the costly CTE downstream in getContainerIdsHoldingPublishedPages, then we can update getContainerIdsHoldingPublishedPages to filter for Folder and Collection then we can have the same outcome? Will be more readable (more direct too)

apps/studio/src/features/settings/Redirects/index.tsx
dcshzj: Done — added `RedirectManagementContext` (+ provider and a `useCanManageRedirects()` hook) in the Redirects feature, with its own `RedirectManagement` subject and `buildRedirectManagementPermissions` alongside the user-management one. The provider reuses the same `resource.getRolesFor` query, so React Query serves it from cache and it costs no extra request. Both the card and the table now read the hook, and the comments explaining the borrow are gone.

apps/studio/src/features/settings/Redirects/components/RedirectsTable.tsx
dcshzj: Same change here — this now reads `useCanManageRedirects()` from the new `RedirectManagementContext`.

apps/studio/src/server/modules/redirect/redirect.service.ts
dcshzj: You're right, it was redundant. The pre-filter existed only to keep non-containers out of the `IN` list, which the walk could do itself. Moved it into `getContainerIdsHoldingPublishedPages` — the base case now selects parents from a subquery constrained to Folder/Collection, so the caller passes its candidate ids straight through and one round-trip disappears.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -175,19 +180,117 @@ const getPublishedStateByResourceIds = async (
...
+const getContainerIdsHoldingPublishedPages = async (
+  siteId: number,
+  containerIds: string[],
+): Promise<Set<string>> => {
+  if (containerIds.length === 0) {
+    return new Set()
+  }
+  // `rootId` carries the container each row descends from down the recursion,
+  // so one walk answers every container at once. A resource's parent chain is
+  // intrinsic to its id, so no row can be attributed to the wrong root.
```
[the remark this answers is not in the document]
dcshzj: Good catch — the comment was wrong. Ids are autoincrementing and encode nothing about ancestry. Reworded to say what actually holds: rows are only ever reached through `parentId` joins, so a row can only carry a root it genuinely descends from, and where two candidates are nested a page under both is emitted once per root (deduped by the `DISTINCT`).

apps/studio/src/features/settings/Redirects/index.tsx
dcshzj: Btw separately (not AI reply), I think we have been very lax about this permissions right now, reusing where we can. We should eventually move towards each action being their own permission item, so if more roles were to come out in the future it would be much easier to reason about them.

Right now I think things like audit logs also reuse this, our whole editing also is one single permission and is not granular enough (like differentiating between a published and unpublished page).

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -175,19 +180,112 @@ const getPublishedStateByResourceIds = async (
...
+          fb
+            // Recursive case: walk down into each child's own children
+            .selectFrom("Resource")
+            .innerJoin("Descendant", "Descendant.id", "Resource.parentId")
+            .where("Resource.siteId", "=", siteId)
+            .select([
+              "Resource.id",
+              "Descendant.rootId",
+              "Resource.publishedVersionId",
+            ]),
```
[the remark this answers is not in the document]
dcshzj: Valid, and the fix was already in the codebase — I'd duplicated something that exists. `resource.service.ts` has `withResourceSubtree`, whose comment reads *"`union` (not `unionAll`) dedupes rows so a malformed parent chain with a cycle can't drive the recursion forever"*, and it already exports `hasPublishedDescendant` — literally "true when `resourceId` or any descendant is published". So the bespoke CTE is gone and that helper is called instead: cycle-safe, and it stops at the first published row rather than materialising the subtree.

Worth flagging what I found while writing a regression test for this. The runaway in this scenario isn't actually the traversal you pointed at — it's `getResourceFullPermalinks`, which resolves reference destinations on the same path and has **two** pre-existing cycle bugs on `main`: its CTE used `unionAll`, and the in-memory ancestor walk it feeds had no visited guard, so it would loop even once the query terminated. With a `[resource:...]` destination pointing into a cycle, `resolveReferences` hung outright — my first version of the test timed out at 5s and took the vitest worker down with it. Both are fixed, and the test now passes in ~4s.

Cycles aren't reachable through the app today (`resource.router.ts:493` rejects moving a folder into its own descendant), so this is defence against malformed data rather than a live path.

apps/studio/src/features/settings/Redirects/RedirectManagementContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  children,
+}: PropsWithChildren<RedirectManagementProviderProps>) => {
+  const { data: roles } = trpc.resource.getRolesFor.useQuery({
+    siteId,
+    resourceId: null,
+  })
+
+  const ability = useMemo(
+    () => (roles ? buildRedirectManagementPermissions(roles) : new Ability()),
+    [roles],
```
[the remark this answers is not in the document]
dcshzj: Valid — this was a real gap and I'd flagged the same thing internally without fixing it. Collapsing "loading" and "failed" into an empty ability made both indistinguishable from a genuine read-only role, so an admin saw the page without their controls (transiently on load, permanently on failure) with nothing explaining why.

The context now carries `isPending` and `isError` alongside the ability rather than folding them into it, and `useRedirectManagement()` returns all three:

- **loading** → the add card renders as a skeleton, and the table stays in its existing loading state instead of committing to a column set it may have to change a moment later
- **error** → an explicit `Infobox`: "We couldn't check your permissions, so adding and removing redirects is unavailable. Refresh the page to try again."
- **loaded** → as before

Two browser-mode tests cover the loading and error states. This also removes the layout flash where the card popped in after the roles landed.

apps/studio/src/features/settings/Redirects/index.tsx
dcshzj: Agreed, and this PR nudges that way rather than fixing it: `RedirectManagement` is now its own subject with its own `buildRedirectManagementPermissions`, so redirects no longer ride on `UserManagement` even though both currently resolve to "is site admin". Audit logs still borrow `UserManagement`, and resource editing is still one coarse permission — neither touched here, both would be worth their own pass.

apps/studio/src/server/modules/redirect/redirect.service.ts
```diff
@@ -175,19 +181,64 @@ const getPublishedStateByResourceIds = async (
...
+      // Reuses the resource module's existing subtree walk rather than a second
+      // one here: it stops at the first published row instead of materialising
+      // the subtree, and its CTE is a `union` so a malformed parent chain with
+      // a cycle terminates instead of recursing forever.
+      isLive: await hasPublishedDescendant(db, {
+        siteId,
+        resourceId: String(id),
+      }),
+    })),
+  )
```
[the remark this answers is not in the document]
dcshzj: The reported attack isn't reachable — the premise that the batch is unbounded is incorrect.

`resolveRedirectReferencesSchema` caps `references` at `MAX_REDIRECT_REFERENCES` (100) and tRPC validates it **before** the procedure body runs, so a request carrying "5,000 to 10,000 references" never reaches `getDestinationLiveStateByResourceIds` — it is rejected with `BAD_REQUEST` / "Too many references". There was no test pinning that, so I added two: one over the cap is rejected, and a batch right at the cap (every entry unresolvable, so every one takes the liveness path) resolves normally. Both pass.

That said, the underlying shape was worth fixing. Within the cap a caller can make all 100 entries undecided by pointing them at unpublished resources, so one request could put 100 concurrent queries on the pool where `main` did 2 — a poor neighbour even if not a DoS. The lookups now run in windows of `LIVENESS_LOOKUP_CONCURRENCY` (5). The table only ever has a page's worth in flight, so it's unaffected.

On your regression-test criterion: the cap tests cover the control that actually bounds this. The concurrency window itself has no observable output to assert — it changes only how many queries overlap, not the result — so I haven't faked a test for it. Verified by construction and by the 121 redirect tests passing.

Severity looks overstated to me: authenticated-Editor-only, and the stated exploit path is closed at input validation.

Ssatu (approved):
looks good!

PR #3110, opened 2026-08-11
fix(studio): key incrementVersion off Version history, not publishedVersionId

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 4 | +43 | -10 |
| 🧪 Test | 1 | +101 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

`incrementVersion` (`apps/studio/src/server/modules/version/version.service.ts`) derives the previous version and next `versionNum` from `Resource.publishedVersionId`. This works today because `publishedVersionId` is only ever cleared by... nothing — there's currently no way to un-publish a page.

We're about to add an unpublish action that clears `publishedVersionId` while leaving `draftBlobId`/Version history untouched (so the draft the user was working on isn't lost). Once that lands, republishing an unpublished page would hit `incrementVersion`, see `publishedVersionId === null`, and reset `newVersionNum` to `1` — silently colliding with/orphaning the existing Version history for that resource.

## Solution

- Add `getLatestVersionByResourceId`, which looks up the most recent `Version` row for a resource by `versionNum`, independent of `Resource.publishedVersionId`.
- `incrementVersion` now derives `previousVersion`/`newVersionNum` from Version history via this helper instead of `publishedVersionId`.
- Add a `CREATE INDEX CONCURRENTLY` on `Version(resourceId, versionNum)` (via the custom migration mechanism, since Version can grow large) to keep this lookup cheap.

No behavior change for any page that has never been unpublished — for those, "latest Version" and "Version pointed to by publishedVersionId" are always the same row.

**Breaking Changes**

**Bug Fixes**:

- Fixes version-numbering collision/orphaning that would occur once unpublish (clearing `publishedVersionId`) ships, by keying `incrementVersion` off Version history instead.

## Tests

- Added `apps/studio/src/server/modules/version/__tests__/version.service.test.ts`:
  - `incrementVersion` returns `null` when there's no draft to publish.
  - Version numbering starts at 1 for a never-published page.
  - Version numbering continues correctly (`versionNum` 2, `previousVersion` = versionNum 1) even after `publishedVersionId` has been cleared, simulating an unpublish.
- Verified the new regression test fails against the pre-fix code (`newVersionNum` resets to 1) and passes with the fix.
- Full unit suite (`pnpm test:unit`) passes.

### DIFF
--- a/apps/studio/src/server/modules/version/__tests__/version.service.test.ts
+++ b/apps/studio/src/server/modules/version/__tests__/version.service.test.ts
@@ -1,0 +1,101 @@
+import { resetTables } from "tests/integration/helpers/db"
+import {
+  setupBlob,
+  setupPageResource,
+  setupUser,
+} from "tests/integration/helpers/seed"
+import { ResourceType } from "~prisma/generated/prisma/client"
+
+import { db, ResourceState } from "../../database"
+import { incrementVersion } from "../version.service"
+
+describe("version.service", () => {
+  beforeEach(async () => {
+    await resetTables("Resource", "Version", "Blob", "User", "Site")
+  })
+
+  describe("incrementVersion", () => {
+    it("should return null if there is no draft to publish", async () => {
+      // Arrange
+      const user = await setupUser({})
+      const { site, page } = await setupPageResource({
+        resourceType: ResourceType.Page,
+        state: ResourceState.Published,
+        userId: user.id,
+      })
+
+      // Act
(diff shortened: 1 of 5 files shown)


Ssatu:
Closing to reopen as a draft — branch was rebased to drop stale validators/lockfile commits that were already merged to main separately, which were cluttering the diff.

PR #3120, opened 2026-08-12
⏪ [Unpublishing] feat(studio): implement unpublishPage mutation and enforce unpublish-before-delete

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 8 | +315 | -2 |
| 🧪 Test | 3 | +1018 | -1 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

Stacked on #3119.

📖 [Explainer: Turning off the lights without deleting the room: an unpublish service](https://app.notion.com/p/3c777dbba7888170997ad25ade4e8bdd)

## Problem

There's currently no way to take a live page back to not-live — the only way to stop a page from serving is to delete it outright, which is destructive (loses the resource, its draft, and its Version history). We need an unpublish action that reverses publish without touching the draft a user may be working on.

## Solution

- Add `unpublishPageResource` (`resource.service.ts`): clears `Resource.publishedVersionId` and flips `state` back to `Draft`, leaving Version history and any in-progress draft untouched. Emits an `Unpublish` audit event recording the version that was live, then triggers a site rebuild so the page's output is actually removed from the live site (mirrors what deletion does today).
- Add a single `unpublishPage` mutation (`page.router.ts`), gated behind the `unpublish` CASL action added in #3119 and dark-launched behind `IS_UNPUBLISH_ENABLED_FEATURE_KEY`. It accepts Page-like resource ids as well as Folder/Collection ids — a Folder/Collection has no content of its own, so `unpublishPageResource` resolves the given id to its child IndexPage internally (via `getFullPageById`) and unpublishes that instead. There are no separate `unpublishFolder`/`unpublishCollection` mutations; the type allow-list (`UNPUBLISHABLE_RESOURCE_TYPES_WITH_CONTAINERS`) excludes FolderMeta/CollectionMeta/RootPage from being passed directly while still accepting them as containers to resolve through.
- Enforce unpublish-before-delete (`resource.router.ts`): deleting a page that's still `publishedVersionId !== null` now throws `BAD_REQUEST` instead of silently deleting a live page. For Folder/Collection deletes, the guard checks the entire subtree (not just the container's own row, which never carries a `publishedVersionId`, or only its immediate IndexPage) — `Resource.parentId` is `onDelete: Cascade`, so deleting a container also wipes every descendant, and any of them being live blocks the delete.
- Extend `logPublishEvent`'s `eventType` to also accept `"Unpublish"`, and add `BlobUnpublishEventLogProps` (delta `before: VersionPointer, after: null`) since unpublish has no new version to point to.

**Breaking Changes**

**Features**:

- Adds a single `unpublishPage` mutation allowing Publishers/Admins to take a live page (or a folder/collection's landing page, resolved internally to its child IndexPage) back to draft.

**Bug Fixes**:

- Deleting a live page, or a Folder/Collection with a live page anywhere in its subtree, is now blocked with a clear error instead of silently removing content that's still serving traffic.

## Tests

- `page.router.test.ts`: covers `unpublishPage` — success path clears `publishedVersionId`/flips state to Draft and logs the audit event, for Pages as well as Folder/Collection ids (resolved to their child IndexPage while the container's own row is left untouched), plus failure cases (page not found, unsupported resource type, page not currently published, insufficient permissions, feature flag off).
- `resource.router.test.ts`: covers the new delete guard — deleting a published page is rejected; deleting a Folder/Collection is rejected if its IndexPage, a direct child page, or a page nested inside a subfolder is still live; deleting a Folder/Collection with nothing live anywhere in its subtree still succeeds.

- [ ] As a Publisher/Admin, publish a page, then call the new unpublish action on it; confirm the page's `publishedVersionId` clears, its state shows as Draft in the CMS, and the live site rebuild removes the page from the public site.
- [ ] Confirm any draft edits made before unpublishing are still present after unpublishing (i.e. unpublish did not touch the draft/Version history).
- [ ] Attempt to delete a currently-published page and confirm it's rejected with a "must be unpublished before it can be deleted" error; unpublish it first, then confirm delete succeeds.
- [ ] Attempt to delete a Folder/Collection with a live page nested inside it and confirm it's rejected; unpublish the nested page, then confirm delete succeeds.
- [ ] Confirm an Editor (no `unpublish` permission) is denied when attempting to unpublish a page.

The PR adds a feature-gated unpublish operation and prevents live resources from being deleted before they are unpublished.
- Adds page and container landing-page unpublishing while preserving drafts and version history.
- Records unpublish audit events and triggers a site rebuild.
- Checks container subtrees for live descendants before unpublishing landing pages or deleting containers.
- Adds integration coverage for permissions, feature flags, scheduling, draft preservation, container resolution, and deletion guards.
</details>

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/resource/resource.service.ts | Implements the transactional publication-state transition, container landing-page resolution, draft preservation, audit logging, and post-transaction rebuild. |
| apps/studio/src/server/modules/page/page.router.ts | Adds the feature-gated, permission-checked unpublish mutation and its supported resource-type boundary. |
| apps/studio/src/server/modules/resource/resource.router.ts | Adds live-resource and live-descendant deletion guards while retaining legacy deletion behavior when the feature is disabled. |
| apps/studio/src/server/modules/audit/audit.service.ts | Extends publication audit logging to represent an Unpublish transition with a null resulting version. |
| apps/studio/src/schemas/page.ts | Adds the input schema consumed by the new unpublish procedure. |

</details>

```mermaid
sequenceDiagram
    actor User
    participant API as page.unpublishPage
    participant DB as Resource transaction
    participant Audit as Audit log
    participant Build as Site publisher
    User->>API: Unpublish resource
    API->>API: Check permission, flag, and type
    API->>DB: Lock and resolve target page
    DB->>DB: Validate publication and schedule state
    DB->>DB: Preserve or backfill draft
    DB->>Audit: Record Unpublish event
    DB-->>API: Commit Draft state
    API->>Build: Rebuild published site
    Build-->>User: Unpublish completes
```
</details>

<sub>Reviews (20): Last reviewed commit: ["perf(studio): skip delete&#39;s rebuild when..."](https://github.com/opengovsg/isomer/commit/13cdc3d7c03cf18fb94b2530e91bb60b5e5a7df1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53300363)</sub>

- Knowledge Base — [Content API domains](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/content-api-domains.md)
- Knowledge Base — [Audit and privileged operations](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/audit-and-privileged-operations.md)
- Knowledge Base — [Studio runtime integrations](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/studio-runtime-integrations.md)
</details>

### DIFF
--- a/apps/studio/src/schemas/page.ts
+++ b/apps/studio/src/schemas/page.ts
@@ -90,7 +90,6 @@ export const publishPageSchema = z.object({
   siteId: z.number().min(1),
 })
 
-// TODO: siteId should be taken from user's context (not input)
 export const unpublishPageSchema = z.object({
   pageId: z.number().min(1),
   siteId: z.number().min(1),
--- a/apps/studio/src/schemas/page.ts
+++ b/apps/studio/src/schemas/page.ts
@@ -90,6 +90,12 @@ export const publishPageSchema = z.object({
   siteId: z.number().min(1),
 })
 
+// TODO: siteId should be taken from user's context (not input)
+export const unpublishPageSchema = z.object({
+  pageId: z.number().min(1),
+  siteId: z.number().min(1),
+})
+
 export const createCollectionPageFormSchema = z
   .discriminatedUnion("type", [
     z.object({
--- a/apps/studio/src/server/modules/audit/audit.service.ts
+++ b/apps/studio/src/server/modules/audit/audit.service.ts
@@ -253,7 +253,7 @@ interface PublishEventLogProps<
     before: Before extends null ? null : WithoutMeta<Before>
     after: After extends null ? null : WithoutMeta<After>
   }
-  eventType: Extract<AuditLogEvent, "Publish">
+  eventType: Extract<AuditLogEvent, "Publish" | "Unpublish">
   ip?: string
   metadata: Meta
   siteId: Site["id"]
@@ -264,6 +264,14 @@ interface PublishEventLogProps<
 type BlobPublishEventLogProps = PublishEventLogProps<
   null | VersionPointer,
   VersionPointer,
+  BlobPublishEvent
+>
+
+// NOTE: Unpublish clears Resource.publishedVersionId while leaving Version
+// history and any draft untouched, so there is no new version to point to.
+type BlobUnpublishEventLogProps = PublishEventLogProps<
+  VersionPointer,
+  null,
   BlobPublishEvent
 >
 
@@ -284,6 +292,7 @@ type RepublishEventLogProps = PublishEventLogProps<
 
 export const logPublishEvent: AuditLogger<
   | BlobPublishEventLogProps
+  | BlobUnpublishEventLogProps
   | ResourcePublishEventLogProps
   | ConfigPublishEventLogProps
   | RepublishEventLogProps
--- a/apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
+++ b/apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
@@ -18,6 +18,7 @@ import {
   setupEditorPermissions,
   setupFolder,
   setupPageResource,
+  setupPublisherPermissions,
   setupSite,
   setupUser,
 } from "tests/integration/helpers/seed"
@@ -1233,6 +1234,107 @@ describe("collection.router", async () => {
         .where("id", "=", collection.id)
         .executeTakeFirst()
       expect(result).toMatchObject(expected!)
+    })
+  })
+
+  describe("unpublishCollection", () => {
+    it("should throw 401 if not logged in", async () => {
+      const result = unauthedCaller.unpublishCollection({
+        siteId: 1,
+        resourceId: 1,
+      })
+
+      await expect(result).rejects.toThrow(
+        new TRPCError({ code: "UNAUTHORIZED" }),
+      )
+    })
+
+    it("should throw 403 if user does not have unpublish access", async () => {
+      const { site, collection } = await setupCollection()
+      await setupEditorPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+
+      const result = caller.unpublishCollection({
+        siteId: site.id,
+        resourceId: Number(collection.id),
+      })
+
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "FORBIDDEN",
+          message:
+            "You do not have sufficient permissions to perform this action",
+        }),
+      )
+    })
+
+    it("should throw if the collection's IndexPage is not currently published", async () => {
+      const { site, collection } = await setupCollection()
+      await setupPageResource({
+        siteId: site.id,
+        parentId: collection.id,
+        resourceType: ResourceType.IndexPage,
+        state: ResourceState.Draft,
+      })
+      await setupPublisherPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+
+      const result = caller.unpublishCollection({
+        siteId: site.id,
+        resourceId: Number(collection.id),
+      })
+
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "PRECONDITION_FAILED",
+          message: "This page is not currently published",
+        }),
+      )
+    })
+
+    it("should unpublish the collection's IndexPage, leaving the collection's own row untouched", async () => {
+      // Arrange
+      const { site, collection } = await setupCollection()
+      const { page: indexPage } = await setupPageResource({
+        siteId: site.id,
+        parentId: collection.id,
+        resourceType: ResourceType.IndexPage,
+        state: ResourceState.Published,
+        userId: session.userId,
+      })
+      await setupPublisherPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+
+      // Act
+      await caller.unpublishCollection({
+        siteId: site.id,
+        resourceId: Number(collection.id),
+      })
+
+      // Assert — the IndexPage is unpublished
+      const updatedIndexPage = await db
+        .selectFrom("Resource")
+        .where("id", "=", indexPage.id)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(updatedIndexPage.publishedVersionId).toBeNull()
+      expect(updatedIndexPage.state).toEqual(ResourceState.Draft)
+      expect(updatedIndexPage.draftBlobId).not.toBeNull()
+
+      // Assert — the Collection's own row is untouched
+      const updatedCollection = await db
+        .selectFrom("Resource")
+        .where("id", "=", collection.id)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(updatedCollection.publishedVersionId).toBeNull()
+      expect(updatedCollection.state).toEqual(ResourceState.Draft)
     })
   })
 
--- a/apps/studio/src/server/modules/collection/collection.router.ts
+++ b/apps/studio/src/server/modules/collection/collection.router.ts
@@ -2,6 +2,7 @@ import type { UnwrapTagged } from "type-fest"
 import { TRPCError } from "@trpc/server"
 import { get, pick } from "lodash-es"
 import { INDEX_PAGE_PERMALINK } from "~/constants/sitemap"
+import { ENABLE_CODEBUILD_JOBS } from "~/lib/growthbook"
 import {
   countTagOptionsUsageSchema,
   createCollectionSchema,
@@ -10,6 +11,7 @@ import {
   getCollectionTagsSchema,
   readCollectionSchema,
   readLinkSchema,
+  unpublishCollectionSchema,
 } from "~/schemas/collection"
 import { readFolderSchema } from "~/schemas/folder"
 import { createCollectionPageSchema } from "~/schemas/page"
@@ -32,6 +34,7 @@ import {
   getBlobOfResource,
   getSiteResourceById,
   publishResource,
+  unpublishPageResource,
   updateBlobById,
 } from "../resource/resource.service"
 import { validateUserPermissionsForSite } from "../site/site.service"
@@ -66,6 +69,26 @@ export const collectionRouter = router({
       }
       return resource
     }),
+  unpublishCollection: protectedProcedure
+    .input(unpublishCollectionSchema)
+    .mutation(
+      async ({ ctx: { user, gb, logger }, input: { siteId, resourceId } }) => {
+        await bulkValidateUserPermissionsForResources({
+          siteId,
+          action: "unpublish",
+          userId: user.id,
+        })
+        await unpublishPageResource({
+          logger,
+          siteId,
+          resourceId: String(resourceId),
+          userId: user.id,
+          sitePublish: {
+            enableCodebuildJobs: gb.isOn(ENABLE_CODEBUILD_JOBS),
+          },
+        })
+      },
+    ),
   create: protectedProcedure
     .input(createCollectionSchema)
     .mutation(
--- a/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
+++ b/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
@@ -2418,6 +2418,48 @@ describe("page.router", async () => {
         .execute()
       expect(auditLogs.length).toEqual(1)
     })
+
+    it("should backfill a draft from the published content when there is no pending draft", async () => {
+      // Arrange — a published page with no pending draft (draftBlobId is
+      // null, as setupPageResource leaves it after publishing)
+      const { site, page, blob } = await setupPageResource({
+        resourceType: ResourceType.Page,
+        state: ResourceState.Published,
+        userId: session.userId ?? undefined,
+      })
+      await setupPublisherPermissions({
+        userId: session.userId ?? undefined,
+        siteId: site.id,
+      })
+      const before = await db
+        .selectFrom("Resource")
+        .where("id", "=", page.id)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(before.draftBlobId).toBeNull()
+
+      // Act
+      await caller.unpublishPage({ siteId: site.id, pageId: Number(page.id) })
+
+      // Assert — a new draft Blob was created with the published content,
+      // rather than reusing the Version's Blob
+      const updated = await db
+        .selectFrom("Resource")
+        .where("id", "=", page.id)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(updated.publishedVersionId).toBeNull()
+      expect(updated.state).toEqual(ResourceState.Draft)
+      expect(updated.draftBlobId).not.toBeNull()
+      expect(updated.draftBlobId).not.toEqual(blob.id)
+
+      const draftBlob = await db
+        .selectFrom("Blob")
+        .where("id", "=", updated.draftBlobId)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(draftBlob.content).toEqual(blob.content)
+    })
   })
 
   describe("updateMeta", () => {
--- a/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
+++ b/apps/studio/src/server/modules/page/__tests__/page.router.test.ts
@@ -14,6 +14,7 @@ import {
 } from "tests/integration/helpers/iron-session"
 import {
   setupAdminPermissions,
+  setupBlob,
   setupCollection,
   setupEditorPermissions,
   setupFolder,
@@ -2310,6 +2311,112 @@ describe("page.router", async () => {
         .where("source", "=", "/old-url")
         .executeTakeFirstOrThrow()
       expect(redirect.destination).toEqual(`[resource:${site.id}:${folder.id}]`)
+    })
+  })
+
+  describe("unpublishPage", () => {
+    it("should throw 401 if not logged in", async () => {
+      const unauthedSession = applySession()
+      const unauthedCaller = createCaller(createMockRequest(unauthedSession))
+
+      const result = unauthedCaller.unpublishPage({ siteId: 1, pageId: 1 })
+
+      await expect(result).rejects.toThrow(
+        new TRPCError({ code: "UNAUTHORIZED" }),
+      )
+    })
+
+    it("should throw 403 if user does not have unpublish access to the page", async () => {
+      // Arrange
+      const { site, page } = await setupPageResource({
+        resourceType: ResourceType.Page,
+        state: ResourceState.Published,
+        userId: session.userId ?? undefined,
+      })
+      await setupEditorPermissions({
+        userId: session.userId ?? undefined,
+        siteId: site.id,
+      })
+
+      // Act
+      const result = caller.unpublishPage({
+        siteId: site.id,
+        pageId: Number(page.id),
+      })
+
+      // Assert
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "FORBIDDEN",
+          message:
+            "You do not have sufficient permissions to perform this action",
+        }),
+      )
+    })
+
+    it("should throw if the page is not currently published", async () => {
+      // Arrange
+      const { site, page } = await setupPageResource({
+        resourceType: ResourceType.Page,
+      })
+      await setupPublisherPermissions({
+        userId: session.userId ?? undefined,
+        siteId: site.id,
+      })
+
+      // Act
+      const result = caller.unpublishPage({
+        siteId: site.id,
+        pageId: Number(page.id),
+      })
+
+      // Assert
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "PRECONDITION_FAILED",
+          message: "This page is not currently published",
+        }),
+      )
+    })
+
+    it("should unpublish a live page while retaining its draft", async () => {
+      // Arrange — a published page with a pending draft
+      const { site, page } = await setupPageResource({
+        resourceType: ResourceType.Page,
+        state: ResourceState.Published,
+        userId: session.userId ?? undefined,
+      })
+      await setupPublisherPermissions({
+        userId: session.userId ?? undefined,
+        siteId: site.id,
+      })
+      const draftBlob = await setupBlob()
+      await db
+        .updateTable("Resource")
+        .where("id", "=", page.id)
+        .set({ draftBlobId: draftBlob.id })
+        .execute()
+
+      // Act
+      await caller.unpublishPage({ siteId: site.id, pageId: Number(page.id) })
+
+      // Assert — DB (Resource)
+      const updated = await db
+        .selectFrom("Resource")
+        .where("id", "=", page.id)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(updated.publishedVersionId).toBeNull()
+      expect(updated.state).toEqual(ResourceState.Draft)
+      expect(updated.draftBlobId).toEqual(draftBlob.id)
+
+      // Assert — DB (AuditLog)
+      const auditLogs = await db
+        .selectFrom("AuditLog")
+        .where("eventType", "=", AuditLogEvent.Unpublish)
+        .selectAll()
+        .execute()
+      expect(auditLogs.length).toEqual(1)
     })
   })
 
--- a/apps/studio/src/server/modules/page/page.router.ts
+++ b/apps/studio/src/server/modules/page/page.router.ts
@@ -32,6 +32,7 @@ import {
   publishPageSchema,
   readPageOutputSchema,
   reorderBlobSchema,
+  unpublishPageSchema,
   updatePageBlobSchema,
   updatePageMetaSchema,
 } from "~/schemas/page"
@@ -62,6 +63,7 @@ import {
   getResourcePermalinkTree,
   publishPageResource,
   publishResource,
+  unpublishPageResource,
   updateBlobById,
   updatePageById,
 } from "../resource/resource.service"
@@ -667,6 +669,27 @@ export const pageRouter = router({
             publisherEmail: user.email,
           })
         }
+      },
+    ),
+
+  unpublishPage: protectedProcedure
+    .input(unpublishPageSchema)
+    .mutation(
+      async ({ ctx: { user, gb, logger }, input: { siteId, pageId } }) => {
+        await bulkValidateUserPermissionsForResources({
+          siteId,
+          action: "unpublish",
+          userId: user.id,
+        })
+        await unpublishPageResource({
+          logger,
+          siteId,
+          resourceId: String(pageId),
+          userId: user.id,
+          sitePublish: {
+            enableCodebuildJobs: gb.isOn(ENABLE_CODEBUILD_JOBS),
+          },
+        })
       },
     ),
 
--- a/apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
+++ b/apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
@@ -2974,6 +2974,82 @@ describe("resource.router", async () => {
       expect(result).toEqual(page)
     })
 
+    it("should block deleting a page that is still published", async () => {
+      // Arrange
+      const { page, site } = await setupPageResource({
+        resourceType: "Page",
+        state: ResourceState.Published,
+        userId: session.userId,
+      })
+      const auditSpy = vitest.spyOn(auditService, "logResourceEvent")
+      await setupAdminPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+
+      // Act
+      const result = caller.delete({
+        resourceId: page.id,
+        siteId: site.id,
+      })
+
+      // Assert
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "BAD_REQUEST",
+          message: "This page must be unpublished before it can be deleted",
+        }),
+      )
+      expect(auditSpy).not.toHaveBeenCalled()
+      const actual = await db
+        .selectFrom("Resource")
+        .where("id", "=", page.id)
+        .executeTakeFirst()
+      expect(actual).not.toBeUndefined()
+    })
+
+    it("should still allow deleting a published folder (no unpublish path yet)", async () => {
+      // Arrange
+      const { site, folder } = await setupFolder({})
+      await setupAdminPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+      const blob = await setupBlob()
+      const version = await db
+        .insertInto("Version")
+        .values({
+          versionNum: 1,
+          resourceId: folder.id,
+          blobId: blob.id,
+          publishedBy: session.userId as string,
+        })
+        .returning("id")
+        .executeTakeFirstOrThrow()
+      await db
+        .updateTable("Resource")
+        .where("id", "=", folder.id)
+        .set({
+          state: ResourceState.Published,
+          publishedVersionId: version.id,
+        })
+        .execute()
+
+      // Act
+      const result = await caller.delete({
+        resourceId: folder.id,
+        siteId: site.id,
+      })
+
+      // Assert
+      expect(result).toBeDefined()
+      const actual = await db
+        .selectFrom("Resource")
+        .where("id", "=", folder.id)
+        .executeTakeFirst()
+      expect(actual).toBeUndefined()
+    })
+
     it("should soft-delete redirects pointing to the deleted page", async () => {
       // Arrange — a live redirect whose destination references the page
       const { page, site } = await setupPageResource({ resourceType: "Page" })
--- a/apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
+++ b/apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
@@ -3008,32 +3008,55 @@ describe("resource.router", async () => {
       expect(actual).not.toBeUndefined()
     })
 
-    it("should still allow deleting a published folder (no unpublish path yet)", async () => {
+    it("should block deleting a folder whose IndexPage is still published", async () => {
+      // Arrange — the folder's own row never carries a publishedVersionId;
+      // it's the child IndexPage that's actually live.
+      const { site, folder } = await setupFolder({})
+      await setupAdminPermissions({
+        userId: session.userId,
+        siteId: site.id,
+      })
+      await setupPageResource({
+        siteId: site.id,
+        parentId: folder.id,
+        resourceType: ResourceType.IndexPage,
+        state: ResourceState.Published,
+        userId: session.userId ?? undefined,
+      })
+
+      // Act
+      const result = caller.delete({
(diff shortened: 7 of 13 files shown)


apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+  if (sitePublish)
+    await publishSite(logger, {
+      siteId,
+      codebuildJob: sitePublish.enableCodebuildJobs
+        ? {
+            resourceWithUserIds: [{ resourceId: targetResourceId, userId }],
+            isScheduled: false,
+          }
+        : undefined,
+    })
```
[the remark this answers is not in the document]
Ssatu: Filed as a follow-up: [ISOM-2505](https://linear.app/ogp/issue/ISOM-2505/fix-unpublish-is-unrecoverable-if-the-post-commit-site-rebuild-fails) — this is the same commit-then-rebuild pattern `publishPageResource` already uses today, so the ticket also covers deciding whether the fix should apply to both. Not blocking this PR; tracked separately.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+    await updatePageById(
+      {
+        id: Number(targetResourceId),
+        siteId,
+        publishedVersionId: null,
+        draftBlobId,
+        state: ResourceState.Draft,
+      },
+      tx,
+    )
```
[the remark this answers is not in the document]
Ssatu: Verified — this appears to already be closed by `getFullPageById`'s existing `.forUpdate()`. Both `publishPageResource` and `unpublishPageResource` call it first inside their transaction, and it's a bare `FOR UPDATE` (no `OF <table>` restriction), so it locks the `Resource` row itself, not just the joined Blob. Two concurrent calls on the same resourceId serialize: the second blocks until the first commits, then re-reads post-commit state (not a stale snapshot).

One residual gap I found while checking: `incrementVersion` does a second, unlocked re-read (`getPageById`, no `.forUpdate()`) before writing — safe today only because it always runs after the locked read in the same transaction, but not independently protected against a future refactor. No deterministic concurrency test exists yet to pin this down either.

Full writeup and the residual gap are tracked on the pre-existing ticket for this exact question: [ISOM-2414](https://linear.app/ogp/issue/ISOM-2414/fix-row-lock-resource-in-publishpageresource-select-for-update). Not blocking this PR.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+          "Please ensure you are attempting to unpublish a page that exists",
+      })
+    }
+
+    if (!fullResource.publishedVersionId) {
+      throw new TRPCError({
+        code: "PRECONDITION_FAILED",
+        message: "This page is not currently published",
+      })
+    }
```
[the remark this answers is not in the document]
Ssatu: Fixed — now `fullResource.publishedVersionId === null`.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -753,6 +753,28 @@ export const resourceRouter = router({
...
+                siteId: Number(siteId),
+                resourceId,
+              })
+            : before.publishedVersionId !== null
+
+        if (isLive) {
+          throw new TRPCError({
+            code: "BAD_REQUEST",
+            message: "This page must be unpublished before it can be deleted",
+          })
```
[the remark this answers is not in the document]
Ssatu: Fixed — message now branches on container vs. page: "This folder/collection has live pages inside it — unpublish them before deleting" vs. the original page message.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -753,6 +753,28 @@ export const resourceRouter = router({
...
+        // removal is always preceded by an audited Unpublish event. Folder
+        // and Collection rows never carry their own publishedVersionId —
+        // and `parentId` is `onDelete: Cascade`, so deleting a container also
+        // wipes every descendant — so check the whole subtree for anything
+        // still live, not just the container's own (always-null) field.
+        const isLive =
+          before.type === ResourceType.Folder ||
+          before.type === ResourceType.Collection
+            ? await hasPublishedDescendant(tx, {
+                siteId: Number(siteId),
```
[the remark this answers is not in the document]
Ssatu: Fixed — PR description rewritten to match the implemented subtree-check behavior.

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2313,6 +2314,154 @@ describe("page.router", async () => {
...
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "FORBIDDEN",
+          message:
+            "You do not have sufficient permissions to perform this action",
+        }),
+      )
+    })
+
+    it("should throw if the page is not currently published", async () => {
```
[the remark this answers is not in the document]
Ssatu: Added — a "page not found" test, plus a "wrong resource type" test alongside it.

apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
```diff
@@ -1236,6 +1237,107 @@ describe("collection.router", async () => {
...
+
+      // Assert — the Collection's own row is untouched
+      const updatedCollection = await db
+        .selectFrom("Resource")
+        .where("id", "=", collection.id)
+        .selectAll()
+        .executeTakeFirstOrThrow()
+      expect(updatedCollection.publishedVersionId).toBeNull()
+      expect(updatedCollection.state).toEqual(ResourceState.Draft)
+    })
```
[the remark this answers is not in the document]
Ssatu: Added — asserts the Unpublish AuditLog row, matching the folder test's pattern.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+    await publishSite(logger, {
+      siteId,
+      codebuildJob: sitePublish.enableCodebuildJobs
+        ? {
+            resourceWithUserIds: [{ resourceId: targetResourceId, userId }],
+            isScheduled: false,
+          }
+        : undefined,
+    })
+}
```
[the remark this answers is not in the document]
Ssatu: !accepted_risk Shipping the Lite (no-cascade) scope is intentional for v1 — cascading unpublish to descendants is real follow-up work, now tracked at ISOM-2506 (https://linear.app/ogp/issue/ISOM-2506/feat-cascade-unpublish-to-descendants-foldercollection).

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -253,7 +253,7 @@ interface PublishEventLogProps<
     before: Before extends null ? null : WithoutMeta<Before>
     after: After extends null ? null : WithoutMeta<After>
   }
-  eventType: Extract<AuditLogEvent, "Publish">
+  eventType: Extract<AuditLogEvent, "Publish" | "Unpublish">
```
dcshzj, line 256: question: hmm what's this change for? Seems like the original type is intended to only refer to the act of publishing and not unpublishing

apps/studio/src/server/modules/audit/audit.service.ts
```diff
@@ -267,6 +267,14 @@ type BlobPublishEventLogProps = PublishEventLogProps<
   BlobPublishEvent
 >
 
+// NOTE: Unpublish clears Resource.publishedVersionId while leaving Version
+// history and any draft untouched, so there is no new version to point to.
+type BlobUnpublishEventLogProps = PublishEventLogProps<
+  VersionPointer,
+  null,
+  BlobPublishEvent
```
dcshzj, line 275: question(non-blocking, clarifying understanding): what is this BlobPublishEvent and why is it possible for us to use the same event shape for both the publish and unpublish event?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+
+    if (!fullResource) {
+      throw new TRPCError({
+        code: "PRECONDITION_FAILED",
+        message:
+          "Please ensure you are attempting to unpublish a page that exists",
+      })
+    }
+
+    if (fullResource.publishedVersionId === null) {
```
dcshzj, line 1420: question(non-blocking, clarifying understanding): is it possible for the `fullResource.state` to be `Published` but the `fullResource.publishVersionId` is NULL?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+    // historical content.
+    const draftBlobId =
+      fullResource.draftBlobId ??
+      (
+        await tx
+          .insertInto("Blob")
+          .values({ content: jsonb(fullResource.content) })
+          .returning("Blob.id")
+          .executeTakeFirstOrThrow()
+      ).id
```
dcshzj: suggestion: prefer we don't do inline so that it is easier to read. We should check for the existence of an existing draft blob first, then update the ID if a new one is created in separate steps

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1353,114 @@ export const publishPageResource = async ({
...
+
+    targetResourceId = fullResource.id
+
+    await updatePageById(
+      {
+        id: Number(targetResourceId),
+        siteId,
+        publishedVersionId: null,
+        draftBlobId,
+        state: ResourceState.Draft,
```
dcshzj, line 1470: question(non-blocking, clarifying understanding): how do we differentiate a resource that has always been in draft, vs a resource that was published previously?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -670,6 +672,50 @@ export const pageRouter = router({
...
+        // can't be used to unpublish a container's IndexPage under a page-only
+        // contract (getFullPageById would otherwise silently resolve it).
+        const page = await db
+          .selectFrom("Resource")
+          .where("Resource.id", "=", String(pageId))
+          .where("Resource.siteId", "=", siteId)
+          .where("Resource.type", "not in", [
+            ResourceType.Folder,
+            ResourceType.Collection,
+          ])
```
dcshzj: issue: this would also mean that we allow users to unpublish the RootPage, which can cause build errors since we assume the home page is always present

Also, other non-publishable resources also exist like FolderMeta and CollectionMeta, we should exclude those as well

apps/studio/src/server/modules/collection/collection.router.ts
```diff
@@ -66,6 +69,44 @@ export const collectionRouter = router({
...
+          throw new TRPCError({
+            code: "NOT_FOUND",
+            message: "This collection does not exist",
+          })
+        }
+
+        await unpublishPageResource({
+          logger,
+          siteId,
+          resourceId: String(resourceId),
```
dcshzj, line 114: question: this `resourceId` is it referring to the Collection or the IndexPage?

apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
```diff
@@ -1236,6 +1243,161 @@ describe("collection.router", async () => {
...
+
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "NOT_FOUND",
+          message: "This collection does not exist",
+        }),
+      )
+    })
+
+    it("should throw 404 if resourceId is not a Collection", async () => {
```
dcshzj, line 1334: suggestion: we should also test that this endpoint throws 404 if the provided resourceId is a folder and not a collection

apps/studio/src/server/modules/audit/audit.service.ts
Ssatu: gd qn! unpublishing under the hood is essentially "publishing" an update as well, just that the update is that the resource's live version is now _**null**_  and thus "unpublished"

hence, the decision to keep it in the same audit log is to preserve the mental model that all changes (including unpublishing) must be "published", but still be able to distinguish between Publish and Unpublish

apps/studio/src/server/modules/audit/audit.service.ts
Ssatu: same reasoning as above reply to line 256

apps/studio/src/server/modules/collection/__tests__/collection.router.test.ts
Ssatu: since the function checks for the resource **being** a collection (rather than checking for **not** a folder/page etc.) i'm in favour of having the test be as-is

apps/studio/src/server/modules/collection/collection.router.ts
Ssatu: the resourceId will be referring to the IndexPage!

apps/studio/src/server/modules/page/page.router.ts
Ssatu: agreed, thanks for spotting this - the BE should enforce a guard on these

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: it isn't guarded against in any schema level/DB level assertion, nor are there any checks on this

but for all flows that currently update publishedVersionId/draftBlobId, state is updated together in the same transaction so there would be no real drift

**however, the pre-existing behaviour is indeed fragile and nothing stops us from introducing drift in future - could be up for consideration/review outside of unpublishing PRs**

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: agreed

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: purely by checking whether a version of that resource previously exists

just fyi: there is no functionality in the current scope that requires us to check this yet

Ssatu (commented):
responses to PR feedback

Changes:
* improve unpublishing guard on resource types that can't/shouldn't be unpublished
* improve readability of draftBlobId backfill

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Yup a drift can happen, but we can reconstruct back the missing data based on the resource ID information in the Version table.

Also, it is possible to have a legal Resource that has the `state` to be `Published` and yet the `publishVersionId` is NULL. All our Folder and Collection resources have this behaviour, so there is no concept of a draft folder/collection, that construct is only applied to the corresponding index page.

apps/studio/src/server/modules/collection/collection.router.ts
dcshzj: Hmm sorry a bit confused now, at above the `resourceId` is used to find the actual Collection (lines 86-90), so this resourceId wouldn't be referring to the Collection's ID, or am I missing something

Ssatu: oh sorry, you're right, it would be the actual collection's ID in this case!

we resolve it to the indexPage via getFullPageById from resource.service.ts

Ssatu: FYI: but we would not really have a user flow that utilises unpublishCollection/unpublishFolder yet (given that there is no "page"/"UI" that users will be publishing/unpublishing folders directly - only their indexPage)

Ssatu:
## Changes since last review

Follow-ups addressing gaps found while re-auditing `unpublishPage`'s resourceType coverage against all `ResourceType` values and production data:

- **`34d9ef83c`** – Allow unpublishing a `CollectionLink` via `unpublishPage`. It shares the same (unguarded) `publishPage` path as regular pages (see `LinkEditNavbar`'s use of `PublishButton`) and is built into a real sitemap entry by the static-site build, so it needs a symmetric unpublish path — previously it had none, meaning a published `CollectionLink` could permanently block deleting its parent Collection via the `hasPublishedDescendant` guard.
- **`bac8b377b`** – Exclude `FolderMeta`/`CollectionMeta` from `hasPublishedDescendant`/`getPublishedDescendantResourceIds`. Some rows of these types carry a stray `publishedVersionId` in production, but neither type is ever built into a visitor-facing page — without this exclusion, a stray value would incorrectly block deleting/moving an otherwise fully-unpublished folder or collection.
- **`47cf263d7`** – Doc-only: explains the FolderMeta/CollectionMeta exclusion above, and notes both types are themselves candidates for future deprecation.
- **`9cf945363`** – Allow unpublishing the `RootPage` via `unpublishPage`. It shares the same publish path as every other page (`RootpageRow` links to the same `/pages/[pageId]` editor), and unlike delete, unpublishing it isn't destructive — the static-site build already falls back to a generic homepage when it has no `publishedVersionId`.
- **`fa7dcc5e3`** – Extracted `PUBLISHABLE_RESOURCE_TYPES` (`~/constants/resources.ts`) as the single named source of truth for `unpublishPage`'s allow-list, replacing the inline array, so the contract is grep-able and easier to extend if another publishable type is added later.

Net effect: `unpublishPage` now accepts exactly the resource types that actually have a real publish path in the frontend today (`Page`, `CollectionPage`, `IndexPage`, `CollectionLink`, `RootPage`), and the delete/move published-descendant guards no longer false-positive on internal-only metadata types. `unpublishFolder`/`unpublishCollection` are unchanged.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -670,6 +673,48 @@ export const pageRouter = router({
...
+        // Allow-list rather than deny-list: only PUBLISHABLE_RESOURCE_TYPES
+        // are real, independently unpublishable content pages — see that
+        // constant for why Folder/Collection/FolderMeta/CollectionMeta are
+        // excluded (Folder/Collection ids belong to the dedicated
+        // unpublishFolder/unpublishCollection mutations instead).
+        const page = await db
+          .selectFrom("Resource")
+          .where("Resource.id", "=", String(pageId))
+          .where("Resource.siteId", "=", siteId)
+          .where("Resource.type", "in", PUBLISHABLE_RESOURCE_TYPES)
```
dcshzj: issue: publishable resource types would be a superset of unpublishable ones, cos we allow RootPage to be published but not unpublished

We can maybe add a branch after this to check if page.type is RootPage, then throw a 400 error

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1373,116 @@ export const publishPageResource = async ({
...
+        .values({ content: jsonb(fullResource.content) })
+        .returning("Blob.id")
+        .executeTakeFirstOrThrow()
+
+      draftBlobId = clonedBlob.id
+    }
+
+    targetResourceId = fullResource.id
+
+    await updatePageById(
```
dcshzj, line 1464: suggestion: for completeness, I think we can make sure the page is not scheduled to be published when doing this unpublish step?

Ssatu: agreed, would be good to enforce this on the BE alongside the FE gates

apps/studio/src/server/modules/page/page.router.ts
Ssatu: thanks for catching this, noted that RootPage although being publishable should not be allowed to be unpublished (nor deleted)

will add a comment elsewhere to gate this on the FE when implementing

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1374,130 @@ export const publishPageResource = async ({
...
+    // draft blob. Make the caller cancel the schedule first.
+    if (fullResource.scheduledAt) {
+      throw new TRPCError({
+        code: "PRECONDITION_FAILED",
+        message: `This page is scheduled to be published at ${format(
+          fullResource.scheduledAt,
+          "yyyy-MM-dd HH:mm",
+        )}. Cancel the schedule before unpublishing.`,
+      })
+    }
```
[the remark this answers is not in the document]
Ssatu: This issue is noted and accepted as an issue to be tackled outside of unpublishing - https://linear.app/ogp/issue/ISOM-2505/fix-publishunpublish-actions-are-unrecoverable-if-the-post-commit-site

apps/studio/src/schemas/page.ts
```diff
@@ -90,6 +90,11 @@ export const publishPageSchema = z.object({
   siteId: z.number().min(1),
 })
 
+export const unpublishPageSchema = z.object({
+  pageId: z.number().min(1),
+  siteId: z.number().min(1),
+})
```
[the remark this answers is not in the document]
seaerchin: valid but should always be within constraints unless user pings our API with session cookie lol

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2313,6 +2317,510 @@ describe("page.router", async () => {
...
+        await expect(result).rejects.toThrow(
+          new TRPCError({
+            code: "NOT_FOUND",
+            message: "This page either does not exist or cannot be unpublished",
+          }),
+        )
+      })
+    })
+
+    it("should throw if the page is not currently published", async () => {
```
seaerchin, line 2402: question: why not just handle it gracefully? unpublishing a draft page can just be a no po.

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2313,6 +2317,510 @@ describe("page.router", async () => {
...
+      // Assert
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "PRECONDITION_FAILED",
+          message: "This page is not currently published",
+        }),
+      )
+    })
+
+    it("should unpublish a Folder's IndexPage, leaving the Folder's own row untouched", async () => {
```
seaerchin, line 2515: issue: this implementation of flipping the index page might not be sufficient as the folder itself auto generates an index page at build time. this can be seen under `processDanglingDirectories`. hence, in order to unpublish a folder/collection, we'd first need to ensure that no children under them are published. this is not simply a first level check (ie parent -> child), but we need to extend this guard to all leaf nodes also

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
```diff
@@ -2313,6 +2317,510 @@ describe("page.router", async () => {
...
+        .executeTakeFirstOrThrow()
+      expect(updatedCollection.publishedVersionId).toBeNull()
+      expect(updatedCollection.state).toEqual(ResourceState.Draft)
+    })
+
+    it.each([
+      ResourceType.RootPage,
+      ResourceType.FolderMeta,
+      ResourceType.CollectionMeta,
+    ])("should throw 404 if pageId refers to a %s", async (resourceType) => {
```
seaerchin, line 2806: clarification: what does %s mean here hahahah

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -670,6 +678,62 @@ export const pageRouter = router({
...
+        await bulkValidateUserPermissionsForResources({
+          siteId,
+          action: "unpublish",
+          userId: user.id,
+        })
+
+        // Dark-launched: same NOT_FOUND as the type-guard below, so a caller
+        // can't distinguish "flag off" from "wrong resource type" and infer
+        // the feature exists before it's rolled out.
+        if (!gb.isOn(IS_UNPUBLISH_ENABLED_FEATURE_KEY)) {
```
seaerchin, line 703: comment: we should soft-launch for Isomer admins?

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -670,6 +678,62 @@ export const pageRouter = router({
...
+          throw new TRPCError({
+            code: "NOT_FOUND",
+            message: "This page either does not exist or cannot be unpublished",
```
seaerchin: nit: this error/error message is repeated, maybe can centralise?

apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
```diff
@@ -2974,6 +2977,329 @@ describe("resource.router", async () => {
...
+        resourceId: page.id,
+        siteId: site.id,
+      })
+
+      // Assert
+      await expect(result).rejects.toThrow(
+        new TRPCError({
+          code: "BAD_REQUEST",
+          message: "This page must be unpublished before it can be deleted",
+        }),
```
seaerchin, line 3004: comment: this seems like a bad bet? users already have the mental model (because it is existing behaviour) that deletion implies unpublication. doing this as a 2 step is kinda unintuitive haha

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -753,6 +756,40 @@ export const resourceRouter = router({
...
+        // make every currently-published resource permanently undeletable —
+        // fall back to the pre-unpublish behaviour (delete unconditionally).
+        if (ctx.gb.isOn(IS_UNPUBLISH_ENABLED_FEATURE_KEY)) {
+          const isContainer =
+            before.type === ResourceType.Folder ||
+            before.type === ResourceType.Collection
+
+          const isLive = isContainer
+            ? await hasPublishedDescendant(tx, {
+                siteId: Number(siteId),
```
seaerchin, line 772: question(non-blocking): the intent here seems to be that we want to always do a unpublish -> delete so that there's no breakage. 

to that end, collections _can_ be referenced from the `RootPage` via a component, wondering if we should check on that _prior_ to unpublishing a collection? not a blocker for me because we alreayd allow deletes for referenced collections on production

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -827,6 +840,10 @@ export const hasPublishedDescendant = async (
     .selectFrom("subtree")
     .innerJoin("Resource", "Resource.id", "subtree.id")
     .where("Resource.publishedVersionId", "is not", null)
+    .where("Resource.type", "not in", [
```
seaerchin: comment: actually since we already have a constant might as well just use it here? this also misses `RootPage` but no impact

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1353,6 +1374,130 @@ export const publishPageResource = async ({
...
+      throw new TRPCError({
+        code: "PRECONDITION_FAILED",
+        message: `This page is scheduled to be published at ${format(
+          fullResource.scheduledAt,
+          "yyyy-MM-dd HH:mm",
+        )}. Cancel the schedule before unpublishing.`,
+      })
+    }
+
+    const previousVersionId = fullResource.publishedVersionId
```
seaerchin, line 1452: question: this is abit different from how `incrementVersion` does it? should we keep the same impl across?

seaerchin (commented):
in here, we adopt the following guards:

1. to unpublish a page, it needs to be already published
2. to unpublish a page, it needs to have no scheduled publish

we also don't touch any redirects/blobs. however, in our db, redirects to internal resources are keyed by their resource ids. what happens if a published page has a redirect to page A then was then subsequently unpublished? 

additionally, condition #2 might not hold; under `scheduledPublishingJob.publishedScheduledResources`, we `resetScheduledAtForPublishedResources` _prior_ to actually processing teh job. so it's entirely possible that:

1. schedule publishing job runs
2. reset scheduled at 
3. user unpublishes
4. new page is published

that being said, the chances of htis is minute so it doesn't stand out as a must-fix to me. if you wanna fix, can try taking a pg advisory lock to avoid this.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -670,6 +678,62 @@ export const pageRouter = router({
...
+        // UNPUBLISHABLE_RESOURCE_TYPES_WITH_CONTAINERS for why FolderMeta/
+        // CollectionMeta/RootPage are excluded.
+        const page = await db
+          .selectFrom("Resource")
+          .where("Resource.id", "=", String(pageId))
+          .where("Resource.siteId", "=", siteId)
+          .where(
+            "Resource.type",
+            "in",
+            UNPUBLISHABLE_RESOURCE_TYPES_WITH_CONTAINERS,
```
dcshzj, line 721: nit: I think it's a bit confusing that this function is handling both a page resource and a container resource, but the naming of the function indicates that it is unpublishing the provided resource ID, might not be clear that there is some transformation going on to convert the container ID to the underlying IndexPage ID. Maybe can add a comment

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
Ssatu: i think the frontend (when implemented in future) should not allow a user to invoke an unpublish

currently, users rely on whether the publish button is enabled or disabled to infer the status of the page (draft/published). I suppose those used to this will also similarly rely on whether the button is disabled/enabled to infer the status of the page!

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
Ssatu: yup, agreed! we currently check this already

image

apps/studio/src/server/modules/page/__tests__/page.router.test.ts
Ssatu: its a parameterised test! runs the test for each of the types to give:

- "should throw 404 if pageId refers to a RootPage"
- "should throw 404 if pageId refers to a FolderMeta"
- "should throw 404 if pageId refers to a CollectionMeta"

apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts
Ssatu: agreed that this is the current mental model! but one of the problems ops team raised is that they see frequent accidental deletions of live folders/pages and seek us out to "restore" them.

gating delete on unpublishing should hopefully introduce meaningful friction to deleting something that is **live**. ideally, users should only **delete** content that's no longer needed (unpublished)

hopefully this at least reduces the number of "accidental deletions" of live pages

apps/studio/src/server/modules/resource/resource.router.ts
Ssatu: thanks for raising this, yes it would introduce some stale (404) state! 

i think we can consider fixing this outside of unpublishing (together with deletion) as this might require some deeper thinking to solve well

https://linear.app/ogp/issue/ISOM-2539/check-collection-references-before-unpublishing

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: the use case of `incrementVersion` is slightly different (finding out the latest version - which could be currently published or a previously published version) to what we're hoping to achieve here (tracking the currently published version)

Ssatu:
> in here, we adopt the following guards:
> 
> 1. to unpublish a page, it needs to be already published
> 2. to unpublish a page, it needs to have no scheduled publish
> 
> we also don't touch any redirects/blobs. however, in our db, redirects to internal resources are keyed by their resource ids. what happens if a published page has a redirect to page A then was then subsequently unpublished?
> 
> additionally, condition #2 might not hold; under `scheduledPublishingJob.publishedScheduledResources`, we `resetScheduledAtForPublishedResources` _prior_ to actually processing teh job. so it's entirely possible that:
> 
> 1. schedule publishing job runs
> 2. reset scheduled at
> 3. user unpublishes
> 4. new page is published
> 
> that being said, the chances of htis is minute so it doesn't stand out as a must-fix to me. if you wanna fix, can try taking a pg advisory lock to avoid this.

captured here as part of a pre-existing bug https://linear.app/ogp/issue/ISOM-2505/fix-publishunpublish-actions-are-unrecoverable-if-the-post-commit-site

apps/studio/src/server/modules/resource/resource.service.ts
seaerchin: when will the latest published version not be the currently published version?

Ssatu: when something has been unpublished!

seaerchin: but it looks like we're unpublishing by unsetting the `publishedVersionId`? meaning that:

1. page A has version 1, state: published, publishedVersionId: 1
2. admin says unpublish A
3. page A has version 1, state: draft, publishedVersionId: null

not sure where the new version/latest published version is updated because the "last published version" is still 1 here?

Ssatu: maybe i'm not understanding your question, can lmk:

1. Q1 (currently published VS previously published version)

in my understanding `publishedVersionId` represents what is "currently published"

in the scenario above, what is "currently published" would be `null` (representing _not published_ - we don't record this step in the Version table)
but the last published version is 1, which incrementVersion needs to link back to on a re-publish to maintain the audit trail for this resource

2. Q2 (why not retrieve the version id from the version table, like in incrementVersion?)
> `previousVersionId ` is used here for audit logging to track what is the version that is being unpublished, in this case the information is readily available to us from the resource's publishedVersionId. because we already have the info, we won't need to dig into the Version table to find out

> in **incrementVersion**, we do need to reach into the version table and can't just rely on `publishedVersionId` as it will be null for an unpublished resource.

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -753,6 +756,34 @@ export const resourceRouter = router({
           })
         }
 
+        // Block deleting anything still live. Deletion cascades (`parentId`
+        // is `onDelete: Cascade`), so a Folder/Collection needs its whole
+        // subtree checked, not just its own (always-null) publishedVersionId.
+        // Gated on the flag: with unpublish unreachable, a live resource
+        // could never become deletable, so skip the guard entirely rather
+        // than lock it out permanently.
+        if (ctx.gb.isOn(IS_UNPUBLISH_ENABLED_FEATURE_KEY)) {
```
dcshzj, line 765: suggestion: now that deletion only happens when the page is not published, we can drop the `publishResource` call so that we do not incur an unnecessary CodeBuild build, since nothing actually changes on the live site

dcshzj (approved):

PR #3130, opened 2026-08-13
[Unpublishing] feat(studio): live status badges, publish/unpublish UX, and dashboard status filter

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 47 | +2867 | -648 |
| 🧪 Test | 13 | +1740 | -24 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

Based on `remove-autogen`.

## Problem

This branch started as a single Live/Not-Live dashboard filter, but grew to cover the rest of the "status" surface area that was left dangling once unpublishing existed as a real state: the dashboard and editing views had no consistent way to show Live/Draft/Scheduled status, the publish and unpublish flows were two independently-built modal trees with drifting copy, and there was no guard against unpublishing a folder/collection landing page out from under still-live descendants (or moving a published page into a container that's about to go dark).

## Solution

**Status badges (dashboard + editing view)**
- Added a shared `LiveStatusBadges`/`HasDraftIndicator` pair and standardized every place that showed live/draft/scheduled state (dashboard resource table, root page row, index page row, page editing view) onto it, replacing several one-off `Badge`/`BadgeLeftIcon` implementations.
- `liveTemplate` (a Folder/Collection whose own landing page isn't published but has a live descendant) now visually reads as "Published", same as `live` — no separate pip.
- Added a `lastPublishedAt` correlated subquery (`resource.service.ts`) intended to back a "Last published on …" tooltip on the badge — data plumbing only; not yet wired up to `LiveStatusBadges` or any other component, so the badge itself currently has no such tooltip.
- Fixed a `Tooltip`-positioning bug (design-system `Badge` doesn't forward refs) and a stale-cache bug where dashboard live-status queries didn't refresh after publishing.

**Dashboard status filter**
- Added a multi-select status filter (Published / Unpublished / Scheduled to publish / Scheduled to unpublish / Has draft) to the dashboard resource table, replacing the single Live/Not-Live filter this PR started with.
- Fixed "Clear filter" rendering as permanently disabled, and added a filter-aware empty state ("No pages match X. Try a different filter." + Clear filter link).

**Publish/unpublish UX**
- `EditingLockedOverlay`'s Publish/Unpublish Now and Cancel Schedule actions now open real confirmation modals instead of acting immediately.
- Merged the previously-separate publish and unpublish modal trees (`PublishModal`/`PublishOptionsInput`/`ScheduledPublishDetails` and `UnpublishModal`/`UnpublishOptionsInput`/`UnpublishDateTimeFields`) into one shared `PublishOrUnpublishModal` (+ `ActionOptionsInput`/`ScheduleDateTimeFields`/`ScheduleBanner`), so the publish and unpublish flows are the same UI end to end. Removed the now-redundant `QuickSelectTimeSection`.
- Added a "Can't unpublish this folder/collection yet" informational modal for when unpublishing a container's landing page would strand other published descendants, with a "Go to folder/collection" shortcut.
- Added a read-only `resource.getMoveLockInfo` query mirroring the `move` mutation's existing unpublish-lock check, so the Move modal warns as soon as a destination is picked instead of only after "Move here" errors out.

## Breaking Changes

## Features

- Live/Draft/Scheduled status badges, consistent across the dashboard and editing views.
- Multi-select status filter (with filter-aware empty state) on the dashboard resource table.
- Confirmation modals for all four Publish/Unpublish Now and Cancel Schedule actions.
- A single shared publish/unpublish modal instead of two parallel implementations.
- A blocking modal when unpublishing a folder/collection landing page would strand live descendants.
- Early warning (on destination select, not just on submit) when moving a page into a container scheduled to unpublish.

## Tests

- `resource.router.test.ts`: status filter option combinations, `getMoveLockInfo` (unauthenticated, blocked/unblocked/root-destination cases).
- `folder.router.test.ts`: unpublish-block info surfaced via `folder.getIndexpage`.
- `schedule.test.ts`: schedule schema coverage.

- [ ] On the dashboard, toggle the Status filter through each option and confirm the table and empty state update correctly, and that Clear filter re-enables/works.
- [ ] From the editing view, trigger each of Publish Now, Unpublish Now, Cancel Scheduled Publish, and Cancel Scheduled Unpublish and confirm the right confirmation copy shows and the action completes.
- [ ] Attempt to unpublish a folder/collection's landing page while another page inside it is still live; confirm the blocking modal appears and "Go to folder/collection" navigates correctly.
- [ ] Attempt to move a published page into a folder/collection scheduled to unpublish; confirm the warning appears as soon as the destination is selected.
- [ ] Confirm a Folder/Collection with a live descendant but no published landing page of its own shows "Published" (not a separate "Live · Template" pip).

## Out of scope

Folder/collection unpublish/scheduling business logic itself (covered in `pr5-scheduled-unpublish`), any further status-badge redesign beyond consolidating existing implementations.

This PR consolidates live, draft, and scheduled status presentation while introducing unified publish and unpublish workflows.
- Adds shared status badges and a composable dashboard status filter.
- Unifies immediate and scheduled publish/unpublish confirmation interfaces.
- Blocks unsafe container unpublishing and warns before moves into containers scheduled to go offline.
- Refreshes affected dashboard and editing queries after publication-state changes.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/resource/resource.service.ts | Implements composable status filtering and correctly derives container draft and schedule state from child IndexPages. |
| apps/studio/src/server/modules/resource/resource.router.ts | Returns filtered resources with consistent status metadata and a total count computed over the same predicate. |
| apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/PublishOrUnpublishModal.tsx | Consolidates immediate and scheduled publishing actions into a shared modal workflow. |
| apps/studio/src/features/editing-experience/components/PageMoreActionsButton.tsx | Adds unpublish and schedule-cancellation actions with descendant-aware blocking behavior. |
| apps/studio/src/features/dashboard/components/ResourceTable/ResourceTable.tsx | Integrates multi-select status filtering, filtered pagination totals, shared badges, and filter-aware empty states. |

</details>

```mermaid
sequenceDiagram
    actor Editor
    participant Studio as Studio UI
    participant API as tRPC API
    participant Domain as Resource/Page Service
    participant DB as Database

    Editor->>Studio: Select publish or unpublish action
    Studio->>API: Load page and descendant status
    API->>Domain: Resolve publication constraints
    Domain->>DB: Read page, schedule, and descendant state
    DB-->>Domain: Current state
    Domain-->>API: Action availability / blocking information
    API-->>Studio: Show confirmation or blocking modal
    Editor->>Studio: Confirm allowed action
    Studio->>API: Publish, unpublish, or schedule
    API->>Domain: Apply state transition
    Domain->>DB: Persist publication state
    API-->>Studio: Success
    Studio->>Studio: Refresh page and dashboard status queries
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(studio): container-aware status filt..."](https://github.com/opengovsg/isomer/commit/06e24193b88d494aacfc104ac2b8f7ff8bbcd75c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59443458)</sub>


apps/studio/src/components/Datatable/EmptyTablePlaceholder.tsx
```diff
@@ -20,7 +29,20 @@ export const EmptyTablePlaceholder = ({
...
+              <>
+                <NoSearchResultSvgr mb="0.5rem" />
+                <Text textStyle="subhead-4">No {entityName}s found.</Text>
+                <Text textStyle="body-2">
+                  No {entityName}s match "{activeFilterLabels.join(", ")}". Try
+                  a different filter.
+                </Text>
+                <Link textStyle="body-2" onClick={onClearFilter}>
+                  Clear filter
+                </Link>
```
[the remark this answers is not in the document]
dcshzj: @Ssatu think this is a valid point, can consider changing this to a Button and a clear variant (or link variant but I think not that appropriate here)

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -103,6 +108,128 @@ export const applyResourceOrderBy = <O>(
...
+                eb("Resource.scheduledAction", "=", ScheduledAction.Publish),
+              ]),
+            ])
+          case "scheduledToUnpublish":
+            return eb.and([
+              eb("Resource.scheduledAt", "is not", null),
+              eb("Resource.scheduledAction", "=", ScheduledAction.Unpublish),
+            ])
+          case "hasDraft":
+            return eb("Resource.draftBlobId", "is not", null)
```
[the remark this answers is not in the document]
Ssatu: Fixed — the scheduled/draft filters were reading the container's own row, but that state lives on its IndexPage child. Extended the same container-aware pattern the live/notLive filter already used to cover hasDraft/scheduledToPublish/scheduledToUnpublish too, and the dashboard badges now pull from the IndexPage as well. See 06e24193b.

apps/studio/src/components/Datatable/EmptyTablePlaceholder.tsx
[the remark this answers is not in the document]
Ssatu: Good catch, fixed — swapped it for a Button with the link variant so it's keyboard-reachable and doesn't fake-navigate. Thanks for catching this @dcshzj

apps/studio/src/features/editing-experience/components/EditingLockedOverlay/EditingLockedOverlay.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const cancelScheduleDisclosure = useDisclosure()
+
+  if (!currPage.scheduledAt || !currPage.scheduledAction) {
+    return null
+  }
+
+  const { verb, actionLabel } = COPY[currPage.scheduledAction]
+  const isScheduledToPublish =
+    currPage.scheduledAction === ScheduledAction.Publish
+
```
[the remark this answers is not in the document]
Ssatu: Confirmed, fixed — now defaults scheduledAction to Publish when null, same as PublishButton/PageMoreActionsButton, so legacy scheduled-publish pages lock correctly.

apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/ScheduleBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+        ) : (
+          <>
+            This page will become unpublished between{" "}
+            <Text display="inline" textStyle="subhead-2">
+              {format(scheduledAt, "h:mm")} –{" "}
+              {format(
+                add(scheduledAt, { minutes: UNPUBLISH_WINDOW_MINUTES }),
+                "h:mm a",
+              )}
+            </Text>
```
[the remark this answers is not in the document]
Ssatu: Fixed, added the AM/PM suffix to the start time too.

apps/studio/src/features/dashboard/components/ResourceTable/ResourceFilterMenu.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import { RESOURCE_TABLE_STATUS_FILTER_OPTIONS } from "./constants"
+
+// Decorative only — the surrounding Menu.Item's onClick drives the actual
+// toggle. Drawn as a raw SVG (matching Figma's checkbox glyph exactly)
+// rather than the design system's Checkbox component, since that
+// component's own label/container padding doesn't collapse to zero even
+// with no label — it was eating into the row's width and wrapping the
+// longer option labels onto two lines.
+const FilterCheckbox = ({ isChecked }: { isChecked: boolean }) => (
```
dcshzj: question: can I understand a bit more on this? Was the `sx` prop able to remove the extra padding?

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTable.tsx
```diff
@@ -64,42 +82,45 @@ export const ResourceTable = ({
 }: ResourceTableProps): JSX.Element => {
   const [sortOption, setSortOption] =
     useState<ResourceOrderByOption>("updated-desc")
+  const [statusFilter, setStatusFilter] = useState<
+    ResourceStatusFilterOption[]
```
dcshzj, line 85: suggestion: would using a `Set` be easier here?

apps/studio/src/components/HasDraftIndicator.tsx
```diff
@@ -0,0 +1,29 @@
+import { Box, Text, Tooltip } from "@chakra-ui/react"
+import { Badge, BadgeLeftIcon } from "@opengovsg/design-system-react"
+import { BiSolidCircle } from "react-icons/bi"
+
+interface HasDraftIndicatorProps {
+  draftBlobId: string | null
+}
+
+export const HasDraftIndicator = ({
```
dcshzj: super nit: we can name this as just `DraftIndicator`, the `Has` part indicates that it is a boolean rather than an actual component

apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/PublishOrUnpublishModal.tsx
```diff
@@ -0,0 +1,401 @@
...
+    Promise.all([
+      utils.page.readPage.refetch({ pageId, siteId }),
+      utils.site.getLocalisedSitemap.invalidate({
+        resourceId: pageId,
+        siteId,
+      }),
+      // Publishing/unpublishing changes this resource's liveStatus, which the
+      // dashboard tables/index-page row derive from — refresh whichever of
+      // these is currently mounted (folder, collection item list, or index
+      // page).
```
dcshzj, line 113: super nit: I think this comment can remove, self-explanatory

apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/PublishOrUnpublishModal.tsx
```diff
@@ -0,0 +1,401 @@
...
+    trpc.page.schedulePage.useMutation({
+      onSettled: () => {
+        void invalidateAfterMutation()
+        onClose()
+      },
+      onSuccess: () => {
+        fireContentEditSurveyEvent(PUBLISHED_AFTER_EDITING_EVENT)
+        toast({
+          status: "success",
+          title: "Page scheduled successfully",
```
dcshzj, line 162: nit: should we be more specific on this copy that it is a scheduled publish to distinguish it from the unpublish variant?

apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/PublishOrUnpublishModal.tsx
```diff
@@ -0,0 +1,401 @@
...
+            : "Failed to schedule page unpublish. Please contact Isomer support.",
+        ...BRIEF_TOAST_SETTINGS,
+      })
+    },
+  })
+
+  const isSubmitting =
+    action === "publish"
+      ? isPublishingNow || isSchedulingPublish
+      : isUnpublishingNow || isSchedulingUnpublish
```
dcshzj: super nit: this is an overall OR condition, so I think we can lump everything together:

```javascript
const isSubmitting = isPublishingNow || isSchedulingPublish || isUnpublishingNow || isSchedulingUnpublish
```

apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/ScheduleDateTimeFields.tsx
```diff
@@ -0,0 +1,105 @@
...
+      <FormControl isInvalid={!!errors[dateField]} flexGrow={1}>
+        <FormLabel isRequired>Date</FormLabel>
+        <Controller
+          name={dateField}
+          control={control}
+          render={({ field }) => (
+            <DatePicker
+              {...field}
+              size="sm"
+              shouldSetDateOnTodayButtonClick={true}
```
dcshzj: super nit: can drop `={true}` part since the existence of the prop is truthy

apps/studio/src/features/editing-experience/components/PublishOrUnpublishModal/PublishOrUnpublishModal.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+  // Validate the schedule form as it changes so the info banner can be shown
+  // once it's complete. Do NOT use trigger() since that surfaces error messages.
+  useEffect(() => {
+    const validateForm = () => {
+      setIsScheduleValid(schema.safeParse(methods.getValues()).success)
+    }
+    const subscription = methods.watch(() => validateForm())
+    return () => subscription.unsubscribe()
+  }, [methods, schema])
```
dcshzj: suggestion: we can probably derive `scheduledAt` directly from `schema.safeParse(methods.watch())` during render, which would remove the effect, `isScheduleValid` state, and duplicate date/time assembly while preserving when validation errors are shown. Something like this:

```
const result = schema.safeParse(methods.watch())
const scheduledAt = result.success ? result.data.scheduledAt : null
```

apps/studio/src/features/editing-experience/components/EditingLockedOverlay/EditingLockedOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+  },
+}
+
+const SuspendableEditingLockedOverlay = ({
+  pageId,
+  siteId,
+}: EditingLockedOverlayProps): JSX.Element | null => {
+  const [currPage] = trpc.page.readPage.useSuspenseQuery({ pageId, siteId })
+  const actionDisclosure = useDisclosure()
+  const cancelScheduleDisclosure = useDisclosure()
```
dcshzj, line 38: super nit: we deconstruct the disclosure directly here, so it becomes something like this:

```
const {
  isOpen: isActionModalOpen,
  onOpen: onActionModalOpen,
  onClose: onActionModalClose
} = useDisclosure()
```

apps/studio/src/features/editing-experience/components/PageMoreActionsButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+                action="unpublish"
+                pageId={pageId}
+                siteId={siteId}
+                hasDraftChanges={currPage.draftBlobId !== null}
+                containerType={
+                  isIndexPage ? parentIndexPageInfo?.parentType : undefined
+                }
+                {...unpublishModalDisclosure}
+              />
+            )}
```
dcshzj, line 118: question: do you need to force a re-render? Otherwise would be more performant if we just pass in the `isOpen` as a prop rather than doing conditional rendering

apps/studio/src/features/editing-experience/components/MoveResourceModal/MoveResourceModal.tsx
```diff
@@ -179,6 +192,12 @@ const MoveResourceContent = withSuspense(
               existingResource={movedItem ?? undefined}
               onChange={(resourceId) => setCurResourceId(resourceId)}
             />
+            {isMoveBlocked && (
+              <Infobox variant="warning" size="sm" w="full">
+                This destination (or a folder/collection above it) is scheduled
+                to be unpublished, so a published page can't be moved here.
+              </Infobox>
+            )}
```
dcshzj, line 200: praise: I think this is a good catch on a potential edge case, thanks for being so thorough!

apps/studio/src/features/editing-experience/components/PageMoreActionsButton.tsx
```diff
@@ -0,0 +1,212 @@
...
+                              flexShrink={0}
+                              isDisabled={
+                                !!disabledReason || isBlockInfoLoading
+                              }
+                              isLoading={isBlockInfoLoading}
+                              leftIcon={<Icon as={BiHide} boxSize="1rem" />}
+                              onClick={() => {
+                                onClose()
+                                if (isBlockedByLiveDescendants) {
+                                  cantUnpublishModalDisclosure.onOpen()
```
dcshzj, line 201: question: clarifying my understanding here, due to how we derive `isBlockedByLiveDescendants`, we are intentionally not allowing users to schedule unpublish the landing page, even if the descendants have a scheduled unpublish?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -103,6 +108,194 @@ export const applyResourceOrderBy = <O>(
...
+const CONTAINER_TYPES = [ResourceType.Folder, ResourceType.Collection]
+
+// Folder/Collection ids that count as each status tag, derived from the same
+// child-status map the Status column badges use — see getChildLiveStatusMap.
+// A container is "live" (whether fully live or just live-template) iff
+// something in its subtree is published; otherwise it's "not live". The
+// scheduled/draft tags key off the container's own IndexPage child, since
+// that's the only place a Folder/Collection's schedule/draft state actually
+// lives.
+export const splitContainerIdsByStatus = (
```
dcshzj, line 156: suggestion: I think this function does the job, but it is a bit hard to read and reason about. Can consider hand-writing this so that it is easier for a human to understand how each ID lands on each individual bucket.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1245,6 +1438,163 @@ export const assertResourceNotLive = async (
...
+          resourceId === null ? "is" : "=",
+          resourceId,
+        )
+        .select([
+          "Resource.id",
+          "Resource.id as branchId",
+          sql<number>`0`.as("depth"),
+        ])
+        // `union` (not `unionAll`) dedupes rows so a malformed parent chain with
+        // a cycle can't drive the recursion forever.
```
dcshzj: question(non-blocking): sorry can explain this? We have app-level guarantees that cycles do not form, and I think we can make a reasonable assumption that there won't be any cycles present

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -625,9 +632,39 @@ export const resourceRouter = router({
...
+      async ({
+        ctx,
+        input: { siteId, movedResourceId, destinationResourceId },
+      }) => {
+        await bulkValidateUserPermissionsForResources({
+          action: "read",
+          resourceIds: [
+            movedResourceId,
+            ...(destinationResourceId ? [destinationResourceId] : []),
+          ],
```
dcshzj, line 652: quiz: do you know why we need to validate resource IDs? Isn't it such that if someone has a collaborator role for the site, they can see everything?

apps/studio/src/server/modules/resource/resource.router.ts
```diff
@@ -714,8 +806,56 @@ export const resourceRouter = router({
...
+            scheduledAction: status?.indexPageScheduledAction ?? null,
+          } as const
+        })
+
+        const totalCount = rows[0]
+          ? Number(rows[0].totalCount)
+          : Number(
+              (
+                await query
+                  .clearOrderBy()
```
dcshzj, line 852: question(non-blocking): why do we need this if we are just counting the number of rows?

apps/studio/src/schemas/schedule.ts
```diff
@@ -49,6 +49,49 @@ export const schedulePublishClientSchema = basePageSchema
     }
   })
 
+/**
+ * This schema includes the unpublish date and time for the scheduled unpublication
+ */
+export const scheduleUnpublishClientSchema = basePageSchema
```
dcshzj, line 62: suggestion: I think this is almost a duplicate of `schedulePublishClientSchema`. Can combine the two of them in one, then expose the two separate schemas as aliases

apps/studio/src/features/editing-experience/components/EditingLockedOverlay/EditingLockedOverlay.tsx
```diff
@@ -0,0 +1,123 @@
...
+            scheduled time.
+          </Text>
+        </VStack>
+        <Flex gap="0.75rem">
+          <Button variant="reverse" onClick={actionDisclosure.onOpen}>
+            {actionLabel}
+          </Button>
+          <Button onClick={cancelScheduleDisclosure.onOpen}>
+            Cancel schedule
+          </Button>
```
dcshzj: issue: I think these buttons will also appear for editors? I don't think they should be able to see these buttons if they don't have the relevant permissions

apps/studio/src/components/HasDraftIndicator.tsx
Ssatu: agreed, will do so!

apps/studio/src/features/dashboard/components/ResourceTable/ResourceFilterMenu.tsx
Ssatu: thanks for catching this, overriding chakra's checkbox works for this and is much more maintainable than hand-rolling

will update it!

apps/studio/src/features/dashboard/components/ResourceTable/ResourceTable.tsx
Ssatu: good consideration! for me an array is sufficient since we have max 5 items in this list

amongst other stuff like needing to .map() over statusFilter/ convert to an array for tRPC serialization

apps/studio/src/features/editing-experience/components/EditingLockedOverlay/EditingLockedOverlay.tsx
Ssatu: thanks a lot for catching this! i forgot to consider conditional rendering for this component based on permissions (the BE would still catch but it would appear as an error toast rather than preventing the behaviour outright)

apps/studio/src/features/editing-experience/components/PageMoreActionsButton.tsx
Ssatu: valid qn, didn't look too deeply into this initially so thanks for raising this

after digging a bit I think we pay the re-render cost when the modal is open/closed then opened again which i feel is rare enough to accept as a tradeoff rather than manually managing the state of the form in the scheduling modal!

apps/studio/src/server/modules/resource/resource.router.ts
Ssatu: (commenting without consulting claude first 😆 ) when i was studying the RBAC setup in the codebase i noticed we used to consider the potential of extending our permissions checking to per-resource rather than per-site! i assume this is to support that

Ssatu: after claude enlightment: there's also a check for if the resourceId doesn't actually belong to the site 🫨

apps/studio/src/server/modules/resource/resource.router.ts
Ssatu: minor optimisation since counting doesn't need the orderby

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: agreed! took a while for me to understand this as well when i was reviewing it

will re-look it

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: i think this is a defensive guard in case anything breaks this invariant in future! but i agree that the risk is low, will remove this check

apps/studio/src/features/editing-experience/components/PageMoreActionsButton.tsx
Ssatu: great catch, it should allow scheduled unpublishing but disallow instant unpublishing (we didn't account for this flow in the UI design either) 

will update the branching logic and cook something up for the UI to represent this state correctly

Ssatu (commented):
Naming/Cleanup nits
- HasDraftIndicator → DraftIndicator (the "Has" implied a boolean, not a component)
- Dropped shouldSetDateOnTodayButtonClick={true} → just the prop name
- Flattened isSubmitting into one OR chain instead of an action-gated ternary

UI/component fixes
- ResourceFilterMenu's filter checkbox now reuses the design system's Checkbox style config instead of a hand-drawn SVG
- EditingLockedOverlay now wraps its publish/unpublish action buttons in a Can permission check — editors without the right permission no longer see actionable buttons on a locked/scheduled page

Schema cleanup
- schedulePublishClientSchema/scheduleUnpublishClientSchema: pulled shared validation/transform logic into helpers instead of duplicating it, while keeping each schema's shape concrete (a generic factory hit a Zod v4 typing wall)

Backend logic
- splitContainerIdsByStatus: rewritten with explicit if/else buckets instead of ternary chains, plus a clearer comment on why 5 independent id-lists are needed (dashboard filter checkboxes OR together, so a container can match multiple tags at once)
- Removed a redundant cycle-safety guard (.union() → .unionAll()) in getChildLiveStatusMap's recursive CTE, since the app already guarantees the resource tree can't cycle (enforced at move-time)

Real bug fix: scheduled-unpublish on folders/collections
- Previously, a folder/collection landing page couldn't be scheduled to unpublish if any descendant was currently live, even if that descendant already had its own unpublish scheduled.
- Added unschedulableDescendantCount (backend) to distinguish "blocked right now" from "blocked forever no matter what date you pick."
- PageMoreActionsButton now only hard-blocks scheduling when a descendant is truly unschedulable; otherwise it opens the schedule flow with a disabled "now" option and an explanatory reason.
- Updated CantUnpublishModal copy and PublishOrUnpublishModal/ActionOptionsInput to support disabling just the "now" choice.

Render-cycle simplification
- PublishOrUnpublishModal now derives scheduledAt directly from schema.safeParse(methods.watch()) during render, removing a separate useEffect + isScheduleValid state and the hand-rolled date/time reassembly — the schema's own transform already does that work.

apps/studio/src/components/Datatable/EmptyTablePlaceholder.tsx
```diff
@@ -1,14 +1,24 @@
 import { Flex, Stack, Td, Text, Tr } from "@chakra-ui/react"
```
dcshzj: test

apps/studio/src/components/HasDraftIndicator.tsx
dcshzj: Btw the file is still called HasDraftIndicator.tsx, so it needs a rename

apps/studio/src/features/editing-experience/components/PageMoreActionsButton.tsx
dcshzj: Makes sense yup, didn't look too deeply originally about this form state but that sounds like a valid trade-off.

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: Nice thanks! I think the updated code is a lot cleaner and easier to understand

apps/studio/src/server/modules/resource/resource.router.ts
dcshzj: Yup this is like one of those unopened roads that we built in case we need to swap to a full resource-based permissions model, although I think realistically it might be quite difficult to explain to users how such a permissions model would work, so up to debate if we want to remove it.

apps/studio/src/server/modules/resource/resource.router.ts
dcshzj: Ohhh makes sense, I read it the other way round where we were adding the ORDER BY.

dcshzj (approved):
Approving this first but trouble you to rename HasDraftIndicator.tsx to just DraftIndicator.tsx! The rest of the unresolved threads are good to have so leave it up to you

PR #3139, opened 2026-08-14
[Unpublishing] feat(studio): scheduled unpublish

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 63 | +4505 | -997 |
| 🧪 Test | 19 | +3662 | -275 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 2 | +10 | -0 |

📖 [Explainer: Scheduling a page to unpublish itself later](https://app.notion.com/p/3c677dbba78881c9b3aad1a25554cc39)

## Summary
- Adds a `scheduledAction` enum (`Publish` | `Unpublish`) to `Resource`. Scheduling was previously publish-only; the cron job's due-resource query has no `state`/`publishedVersionId` filter at all, so without this discriminator it would silently **republish** a resource that was actually scheduled for unpublish.
- New `scheduleUnpublish`/`cancelScheduleUnpublish` mutations, mirroring `schedulePage`/`cancelSchedulePage`'s shape and gated on the existing `unpublish` CASL action (no new permission needed).
- `schedulePublishingJob.ts`'s cron handler now branches per-resource on `scheduledAction`, calling `unpublishPageResource` instead of `publishPageResource` where appropriate. Reset step now also clears `scheduledAction`.
- New `AuditLogEvent.ScheduleUnpublish` / `CancelScheduleUnpublish` values, excluded from the activity-report query the same way their publish counterparts already are.
- `unpublishPageResource` now refuses to manually unpublish a page while it has *any* pending schedule (publish or unpublish) — closing the race where a manual unpublish could beat a scheduled one — and the precondition message correctly names which action is actually pending, instead of always saying "published". `schedulePage`'s "already scheduled" message was made equally action-neutral.
- A `failedUnpublish` email covers a scheduled/manual unpublish that itself fails. Separately, a new `failedSiteRebuild` email covers the case where the page-level publish/unpublish *succeeded* but the follow-up site rebuild failed — using the `failedPublish`/`failedUnpublish` copy there was wrong, since it told the reader to retry an action that had already succeeded (and for unpublish, retrying throws `PageAlreadyUnpublishedError`).
- The `alreadyUnpublished` email/cron special-case from an earlier commit on this branch was removed: the guard above makes "someone manually unpublished it before the schedule fired" unreachable in practice, so `PageAlreadyUnpublishedError` now just falls through to the ordinary failed-unpublish handling.

## Explicitly out of scope (tracked as follow-ups)
- **Block edits while a page has a pending scheduled unpublish**: separate standalone PR per prior discussion, since it depends on this PR's `scheduledAction` field but shouldn't ship bundled with it. (Note: manually *unpublishing* while any schedule is pending is now blocked, per above — this item covers other edits, e.g. content changes.)
- **Folder/collection scheduled unpublish**: out of scope, same as pr3/pr4's page-only scoping.
- **Read-then-write race in `scheduleAction`/`cancelScheduleAction`**: two concurrent schedule requests can both pass the "not already scheduled" check before either commits, since the read isn't row-locked and the update isn't conditioned on `scheduledAt IS NULL`. Flagged by review; accepted as a known, narrow-window issue rather than blocking this PR.

## Test plan
- [x] `pnpm typecheck` clean
- [x] `pnpm lint` — 0 errors (pre-existing unrelated warnings only)
- [x] `pnpm format` clean
- [x] Full unit suite passes (1762 passed, 0 failures — one unrelated local Playwright-browser-binary error, not a test failure)
- [x] New tests: `scheduleUnpublish`/`cancelScheduleUnpublish` mutations (401/403/precondition/happy path), cron job branches correctly on `scheduledAction` (unpublishes instead of republishing, resets fields, sends the correct failure email), site-rebuild failure sends `failedSiteRebuild` rather than `failedPublish`/`failedUnpublish`
- [x] Updated existing `schedulePage`/`cancelSchedulePage` tests and fixtures for the new required field

Stacked on `pr3-unpublish-service`.

The PR adds scheduled page unpublishing by sharing the existing scheduling fields and dispatching either publish or unpublish work from the cron job.
- Adds the `scheduledAction` database discriminator, migrations, schemas, and generated types.
- Adds schedule/cancel-unpublish mutations, audit events, notifications, and failure handling.
- Updates scheduled processing to execute the selected action and clear all scheduling state afterward.
- Separates page-action failures from subsequent site-rebuild failures and expands regression coverage.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| packages/db/prisma/schema.prisma | Adds the nullable scheduled-action discriminator used to distinguish publish and unpublish schedules. |
| packages/db/prisma/migrations/20260814160000_add_scheduled_action/migration.sql | Introduces the scheduled-action enum and resource column. |
| packages/db/prisma/migrations/20260814160001_backfill_scheduled_action/migration.sql | Backfills existing scheduled resources as publish actions for compatibility. |
| apps/studio/src/server/modules/page/page.router.ts | Exposes schedule and cancellation mutations with router-level authorization and unpublish-specific validation. |
| apps/studio/src/server/modules/page/page.service.ts | Implements publish and unpublish scheduling state transitions, audit logging, and schedule preconditions. |
| apps/studio/src/server/cron/jobs/schedulePublishingJob.ts | Branches due work by scheduled action, clears scheduling state, and coordinates rebuild and failure notifications. |
| apps/studio/src/server/modules/resource/resource.service.ts | Adds pending-schedule protection and unpublish-specific failure handling to page resource operations. |
| apps/studio/src/features/mail/templates/templates.ts | Adds scheduled-unpublish and rebuild-failure emails through the existing escaping wrapper. |
| apps/studio/src/server/cron/jobs/__test__/schedulePublishingJob.test.ts | Covers publish/unpublish dispatch, state reset, action failures, and rebuild failures. |

</details>

```mermaid
sequenceDiagram
  participant Editor
  participant Router
  participant ResourceDB as Resource DB
  participant Cron
  participant PageService
  participant Rebuild as Site Rebuild

  Editor->>Router: Schedule publish or unpublish
  Router->>Router: Validate action permission
  Router->>ResourceDB: Store scheduledAt, scheduledBy, scheduledAction
  Cron->>ResourceDB: Load due resources
  alt scheduledAction is Unpublish
    Cron->>PageService: unpublishPageResource
  else scheduledAction is Publish
    Cron->>PageService: publishPageResource
  end
  PageService-->>Cron: Page action completed
  Cron->>ResourceDB: Clear scheduling fields
  Cron->>Rebuild: Rebuild affected site
  Rebuild-->>Editor: Completion or failure notification
```
</details>

<sub>Reviews (19): Last reviewed commit: ["refactor(studio): rename resolution/asse..."](https://github.com/opengovsg/isomer/commit/68f74262254d99a37632202d150d64ae6082c4fc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56145540)</sub>


apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -1,17 +1,21 @@
 import type { Resource } from "~/server/modules/database"
 import { env } from "~/env.mjs"
-import { sendFailedPublishEmail } from "~/features/mail/service"
+import {
+  sendFailedPublishEmail,
+  sendFailedUnpublishEmail,
+} from "~/features/mail/service"
```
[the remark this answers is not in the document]
Ssatu: Fixed — updated the PR description to reflect that the failed-unpublish email is implemented.

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -372,123 +382,72 @@ export const pageRouter = router({
...
         scheduledAt,
-        recipientEmail: by.email,
-      })
-    }),
+        action: ScheduledAction.Publish,
+        permissionAction: "publish",
+        requiresPublished: false,
+        auditEvent: AuditLogEvent.SchedulePublish,
+        alreadyScheduledMessage: (at) =>
+          `Page is already scheduled to be updated at ${format(at, "yyyy-MM-dd HH:mm")}`,
```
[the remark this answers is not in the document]
Ssatu: Fixed — made the message action-neutral, matching `scheduleUnpublish`'s wording.

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -174,19 +210,22 @@ export const publishScheduledSites = async (
...
           })
           logger.warn(
-            `Sent failed publish email to ${resource.email} for resource: ${resource.id}, since site publish failed for site ${siteId}`,
+            `Sent failed ${handler.verb} email to ${resource.email} for resource: ${resource.id}, since site publish failed for site ${siteId}`,
           )
         } catch (emailError) {
           logger.error(
             { error: emailError },
-            `Failed to send failed publish email to ${resource.email} for resource: ${resource.id}, since site publish failed for site ${siteId}`,
+            `Failed to send failed ${handler.verb} email to ${resource.email} for resource: ${resource.id}, since site publish failed for site ${siteId}`,
```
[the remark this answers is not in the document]
Ssatu: Fixed — added a distinct `failedSiteRebuild` email/template for this case instead of reusing failedPublish/failedUnpublish.

packages/db/prisma/schema.prisma
```diff
@@ -71,11 +71,14 @@ model Resource {
...
-  scheduledBy String?
-  scheduler   User?     @relation(fields: [scheduledBy], references: [id])
+  // This is only set if the resource is scheduled for a publish/unpublish at
+  // a later date. If set, scheduledAt indicates when the action should run,
+  // and scheduledAction indicates which action (Publish or Unpublish) — the
+  // cron job branches on this rather than assuming publish.
+  scheduledAt     DateTime?
+  scheduledBy     String?
+  scheduledAction ScheduledAction?
+  scheduler       User?            @relation(fields: [scheduledBy], references: [id])
```
seaerchin, line 81: comment: honestly at this point it feels like we should be shifting this into a new table rather than tacking it on

apps/studio/src/server/modules/page/page.router.ts
```diff
@@ -378,121 +389,89 @@ export const pageRouter = router({
...
-      if (isBefore(scheduledAt, new Date())) {
+        siteId,
+        pageId,
+        action: ScheduledAction.Publish,
+        permissionAction: "publish",
+        auditEvent: AuditLogEvent.CancelSchedulePublish,
+        notScheduledMessage:
+          "Unable to cancel schedule for a page that is not scheduled to be published",
+        failedMessage: "Failed to cancel page schedule",
+        sendConfirmationEmail: sendCancelSchedulePageEmail,
```
seaerchin: comment: actually one of the biggest value that i found in doing the `router/service` separation is that there's a clear separation of concerns between what the router does (validation, logging, sending emails etc) vs what the service does (actual business logic of the module)

doing it in this fashion forces colocation without a similar bundling of concerns. my take is that code that lives together should have a similar purpose but this isn't quite through here 

happy to hear your thoughts

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -80,3 +91,205 @@ export const createFolderIndexPage = (title: string) => {
...
+  sendConfirmationEmail: (args: {
+    resource: Resource
+    scheduledAt: Date
+    recipientEmail: string
+  }) => Promise<void>
+}
+
+// Shared body for schedulePage/scheduleUnpublish: they differ only in the
+// scheduled action, permission, published-state precondition, audit event,
+// error copy, and confirmation email.
```
seaerchin: comment: rather than doing it this way, i think we might wanna just have individual service methods doing each of the steps

if we really wanna be clever about it and ensure code reuse + similar structure, can go wtih either [builder pattern](https://www.baeldung.com/java-builder-pattern) or [pipeline pattern](https://www.baeldung.com/java-pipeline-design-pattern) but i think those are pre-mature and i'd rather have straight forward to read code before committing to a design pattern across the codebase

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -80,3 +91,205 @@ export const createFolderIndexPage = (title: string) => {
...
+export const scheduleAction = async ({
```
seaerchin: comment: didn't reveiw too closely, assumed it's just shifting of code

apps/studio/src/server/modules/page/page.service.ts
```diff
@@ -80,3 +91,205 @@ export const createFolderIndexPage = (title: string) => {
...
+        code: "NOT_FOUND",
+        message: "Resource not found",
+      })
+    }
+    // A resource scheduled before scheduledAction existed has it as null;
+    // the cron treats that null the same as Publish (see
+    // schedulePublishingJob.ts), so cancellation must match that convention
+    // rather than requiring an exact enum match.
+    const effectiveScheduledAction =
+      resource.scheduledAction ?? ScheduledAction.Publish
```
seaerchin, line 317: comment: this reads like we're using app logic to patch our data? in this case, would we also have a corresponding PR to remove since past a certain point, we'd already have all the jobs using the new code and hence having the required `ScheduledAction.Publish`

alternatively can just backfill?

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1245,6 +1256,24 @@ export const publishPageResource = async ({
...
+    let clearsSchedule = false
+    if (fullResource.scheduledAt) {
+      const scheduledAction =
+        fullResource.scheduledAction ?? ScheduledAction.Publish
+      if (scheduledAction === ScheduledAction.Unpublish) {
+        throw new ScheduledActionConflictError(
+          "unpublished",
+          fullResource.scheduledAt,
+        )
+      }
```
seaerchin: question: when will we actually run into this case? it happens in the case of publishing a resource now, with a scheduled unpublish _in the future_

if that is so, shouldn't we allow the publish anyway? it could be, for example, amending a mistake that's live on their site and the page is only invalid in the future.

hypothetical example: 

1. agency has mistakenly put their number as 999 on their contact us page
2. they scheduled an unpublish to their contact us page in the future cos they will be shifting it over to new subsite (just an example lolol)
3. they amend the number to 995 now and publish - blocked

IMO, we should always allow a publish **even if** there is a scheduled unpublish action in the future because the time delta means that there could be a useful update for that period.

seaerchin: comment: can we just simplify this?

```ts
if (fullResource.scheduledAt && fullResource.scheduledAction === Unpublish) {
  // throw error
}

clearScheduledResource(...)
```

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1376,6 +1418,35 @@ interface UnpublishPageResourceArgs {
...
+  db: SafeKysely,
+  { resourceId, siteId }: { resourceId: number; siteId: number },
+) => {
+  const page = await db
+    .selectFrom("Resource")
+    .where("Resource.id", "=", String(resourceId))
+    .where("Resource.siteId", "=", siteId)
+    .where("Resource.type", "in", UNPUBLISHABLE_RESOURCE_TYPES)
+    .select("Resource.id")
+    .executeTakeFirst()
```
seaerchin, line 1538: question: don't we also need `"Resource.publishedVersionId", "is", null`?

seaerchin:
from codex: 

> - High — Feature-flag rollback disables cancellation but not execution. [page.router.ts (line 455)](/Users/chin/isomer/apps/studio/src/server/modules/page/page.router.ts:455) gates cancelScheduleUnpublish behind IS_UNPUBLISH_ENABLED_FEATURE_KEY, while [schedulePublishingJob.ts (line 54)](/Users/chin/isomer/apps/studio/src/server/cron/jobs/schedulePublishingJob.ts:54) executes pending unpublishes without checking that flag. If the feature is disabled because of an incident, existing schedules continue unpublishing pages, and users lose the API needed to cancel them. Keep cancellation available when the flag is off and make the worker honor an explicit execution/kill-switch flag—or define and test intentional drain behavior. Confidence: 0.99.

i don't really care about this - once we launch the ff is as good as gone so this probably doesn't matter.

> - High — Cancellation can succeed while the cron still executes the stale unpublish. [schedulePublishingJob.ts (line 118)](/Users/chin/isomer/apps/studio/src/server/cron/jobs/schedulePublishingJob.ts:118) reads due resources into memory, then separately clears schedules at [schedulePublishingJob.ts (line 129)](/Users/chin/isomer/apps/studio/src/server/cron/jobs/schedulePublishingJob.ts:129) before executing that snapshot. A cancellation occurring after the read clears the database fields and reports success, but the cron still invokes unpublishPageResource from the stale row. This can remove a live page after the user received cancellation confirmation. Atomically claim each schedule using a conditional update/UPDATE … RETURNING that matches its ID, timestamp, action, and scheduler; execute only successfully claimed rows, with explicit cancellation-versus-claim semantics and a concurrency regression test. Confidence: 0.98.

looks like this is valid, but it's so rare then i think we can take the hit and do it manually lol.

apps/studio/src/server/modules/page/page.router.ts
Ssatu: strongly agreed on this, i think i accidentally violated this when trying hard to coerce the publish/unpublish methods in the interest of code reuse.

will re-look this to ensure we follow router/service separation

apps/studio/src/server/modules/page/page.service.ts
Ssatu: **tldr: will split the service methods**

agreed! i did grapple with the tradeoff between readability and forcing code reuse, and i agree that we should prioritise readability

also, thanks for pointing out that the absence of a design pattern that supports code reuse well is what would help solve the tension - and the insight that this would be a codebase level decision (to commit to one design pattern over another)

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: the action we want to guard against here is someone publishing without the awareness that the resource will be unpublished in future (and vice versa someone unpublishing thinking their page is no longer visible but in the future it ends up visible)

the scenario raised is valid, and the current implementation does not block the user via the following flow: 
- cancel the scheduled action,
- make the edit, 
- publish
- reschedule

this flow is useful in ensuring that the user is aware that their edit will eventually be unpublished

apps/studio/src/server/modules/page/page.service.ts
Ssatu: yup, this is a safeguard to ensure backward compat in case someone schedules a publish action in between backfill and the studio deployment.

thanks for pointing this out! i'll create a linear ticket to remove this & all instances of stale logic after backfill

[ISOM-2543](https://linear.app/ogp/issue/ISOM-2543/remove-scheduledaction-null-fallback-backward-compat-code-after)

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: Good question — turns out we can't add it here though. This check runs on the raw `pageId`, before it gets resolved from a Folder/Collection id to its child IndexPage. A container's own `publishedVersionId` is always null regardless of whether its landing page is live, so filtering on it at this point would reject every valid Folder/Collection id. The publish-state check happens after resolution instead, inside `scheduleUnpublish` (`if (!resource.publishedVersionId)`).

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1283,6 +1405,10 @@ export const publishPageResource = async ({
       return
     }
 
+    if (fullResource.scheduledAt) {
+      await clearScheduledResource(tx, { id: Number(resourceId), siteId })
+    }
```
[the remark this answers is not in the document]
Ssatu: !fp incrementVersion (pre-existing, unrelated to this PR) fetches the resource via getPageById before this line runs, and getPageById's type allow-list excludes Folder/Collection. Reproduced directly: calling publishPage with a Folder's own id throws TRPCError NOT_FOUND "Page not found" at incrementVersion, and the transaction rolls back before this clearScheduledResource line is ever reached. So resourceId only ever equals fullResource.id by the time this line executes — the reported resourceId-vs-fullResource.id mismatch can't occur in practice.

Ssatu:
## Summary of changes since last review (seaerchin, `a9eb182a5`)

7 commits, mostly review-driven fixes plus one hardening pass:

- **Split `scheduleAction`/`cancelScheduleAction` into four plain functions** (`schedulePublish`, `scheduleUnpublish`, `cancelSchedulePublish`, `cancelScheduleUnpublish`) — each spells out its own steps instead of sharing one body via callback params, matching the existing `publishPage`/`unpublishPage` split.
- **Simplified the schedule-conflict checks** and **gated cron execution on the unpublish feature flag** — previously the flag was only checked at schedule time, so disabling it after a resource was already scheduled wouldn't stop the cron from unpublishing it; a due Unpublish is now skipped (not failed) when the flag is off.
- **Split the `scheduledAction` migration backfill into its own migration**, so the `ADD COLUMN`'s table lock isn't held for the duration of the backfill `UPDATE` against the hot `Resource` table.
- **Added a container-siblings guard at schedule time**: scheduling a Folder/Collection's landing page for unpublish now blocks if another page in the same container won't reliably be down by the scheduled time (mirrors the existing execution-time guard, just with earlier feedback).
- **Ordered scheduled-action cron execution by `scheduledAt`**, so same-tick actions run in a deterministic sequence — needed for the new container-siblings guard's execution-time counterpart to hold.
- **Fixed schedule/cancel mutations to resolve Folder/Collection ids** the same way `publishPage`/`unpublishPage` already do — previously passing a container id to any of the four schedule/cancel mutations would 404.
- **Naming/doc cleanup**: renamed `resolveContainerToIndexPageId` → `resolveEffectiveResourceId` and `assertPageIsUnpublishable` → `assertUnpublishableResourceType` to match what they actually check; fixed a doc comment that had gone stale after the container-resolution fix; added inline comments to the container-siblings guard's SQL conditions.

Also triaged the outstanding Hacktron finding on `clearScheduledResource`'s `resourceId` usage as a false positive — reproduced that `incrementVersion` already rejects a Folder/Collection id earlier in the same transaction, so that line is unreachable with a mismatched id in practice.

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -863,6 +885,77 @@ export const getPublishedDescendantResourceIds = async (
...
+            eb("Resource.scheduledAt", "is", null),
+            // An unpublish is scheduled, but at/after `scheduledAt` — not
+            // guaranteed to land first.
+            eb("Resource.scheduledAt", ">=", scheduledAt),
+            // A schedule exists with no action recorded — legacy rows
+            // (pre-scheduledAction) default to Publish, so this is not an
+            // Unpublish.
+            eb("Resource.scheduledAction", "is", null),
+            // A schedule exists, but it's a Publish, not an Unpublish.
+            eb("Resource.scheduledAction", "!=", ScheduledAction.Unpublish),
```
dcshzj: super nit: standardise to `eb("Resource.scheduledAction", "=", ScheduledAction.Publish)`? To keep it similar to line 949

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -863,6 +885,77 @@ export const getPublishedDescendantResourceIds = async (
...
+          ]),
+        ]),
+        // Unsafe case 2: not live right now, but scheduled to become live
+        // again before `scheduledAt` fires.
+        eb.and([
+          eb("Resource.publishedVersionId", "is", null),
+          // A schedule exists...
+          eb("Resource.scheduledAt", "is not", null),
+          // ...and it fires strictly before `scheduledAt`...
+          eb("Resource.scheduledAt", "<", scheduledAt),
```
dcshzj: issue: I think as long as there is a scheduled publish regardless of time, we shouldn't allow it. We don't have a guarantee that the index page would be published by the time the child page is published

I can't remember if we are tackling the issue where the default template index page would go live if the child is schedule-published after the index page becomes unpublished, but looking at your unsafe case 1, the intent is that we don't want to be able to schedule unpublish an index page if a child page is scheduled to be published

apps/studio/src/server/modules/resource/resource.service.ts
```diff
@@ -1211,6 +1304,18 @@ export const getResourceIdsByPermalinks = async (
...
 
+// Clears a resource's pending schedule, since a manual publish/unpublish
+// that matches the schedule's direction makes it redundant (see the
+// same-direction-proceeds comments in publishPageResource/unpublishPageResource).
+const clearScheduledResource = (
+  tx: Transaction<DB>,
+  { id, siteId }: { id: number; siteId: number },
+) =>
+  updatePageById(
+    { id, siteId, scheduledAt: null, scheduledBy: null, scheduledAction: null },
```
dcshzj, line 1316: issue: do we need to also guarantee the reverse dependency in which we do not allow:

1. Cancelling a scheduled publish of a index page, if the child page is scheduled to be published
2. Cancelling a scheduled unpublish of a child page, if the index page is scheduled to be unpublished

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
```diff
@@ -170,23 +256,32 @@ export const publishScheduledSites = async (
...
+        // failed. Use the site-rebuild-specific email, not the "we couldn't
+        // {verb} your page" one: that would tell the user to retry an action
+        // that already succeeded (and for unpublish, retrying would just
+        // throw PageAlreadyUnpublishedError).
+        const { verb } = getScheduledActionHandler(
+          resource.scheduledAction ?? ScheduledAction.Publish,
+        )
         try {
-          await sendFailedPublishEmail({
+          await sendFailedSiteRebuildEmail({
```
dcshzj, line 273: issue: [webhook.utils.ts:137-145](<https://github.com/opengovsg/isomer/blob/68f74262254d99a37632202d150d64ae6082c4fc/apps/studio/src/server/modules/webhook/webhook.utils.ts#L137-L145>) assumes a scheduled publish only, so that needs to be updated as well

apps/studio/src/server/modules/resource/resource.service.ts
dcshzj: I think for this seems to be covered in downstack PR #3256, cmiw

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: agreed! we want to gate all scheduled unpublish on
1. published child (that isn't unpublished by then)
2. scheduled-for-publish child

apps/studio/src/server/modules/resource/resource.service.ts
Ssatu: for index pages, the constraint we enforce is " **unpublishing an index page should only be allowed if the folder's contents are unpublished** "

unpublish index page -> all child pages are unpublished

To guarantee this: child page cannot be published/cancel its scheduled unpublish if index page is scheduled for unpublishing

so i think only (2) is enforced

apps/studio/src/server/cron/jobs/schedulePublishingJob.ts
Ssatu: thanks for pointing this out, didn't realise we have an EventBridge that invokes a webhook for sending emails post-code build

since site builds can consist of both newly published and unpublished resources, I've updated the email wording for webhook.utils.ts to use "updated" instead of published/unpublished

dcshzj (approved):

PR #3151, opened 2026-08-17
Split audit-log export by surface and remove the Both combined-report flow

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 11 | +299 | -421 |
| 🧪 Test | 13 | +285 | -430 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Audit-log export lived entirely on the Settings page as a single form with two selectable log types — "User access review logs" (Access) and "Audit logs" (Activity) — plus a "Both" combined option that fanned out into two separate export rows server-side. This made a one-off, quick "who has access right now" export unnecessarily heavy (navigate to Settings, tick a box, wait), and kept the "Both" flow's fan-out/dedupe/race-handling logic alive in the service layer even though it added real complexity for a case few people used.

## Solution

Split the two report types across their natural surfaces, then removed the now-unused "Both" flow entirely.

**Breaking Changes**

  - Details ...
  - The `reportType` input no longer accepts `"Both"`. Nothing in this codebase submits it any more, so this is safe, but note it if anything external ever called `audit.createExportRequest` directly with `"Both"`.

**Features**:

- Added a one-click "Export access logs" button on the Users page (`ExportAccessLogsButton`) for Site Admins — requests an Access report for the current month with no extra form. Hidden behind the `is-audit-log-enabled` GrowthBook flag and only rendered for users who can manage users.

**Improvements**:

- Simplified the Settings → Audit logs page down to a single-purpose Activity-log export: removed the two-card checkbox selector (`LogTypeCard`, `toggleReportType`) and the "select a log type" empty state, and added an `Infobox` pointing admins to User Management for access-history exports instead.
- Removed the `Both` combined-report type from `AuditLogExportRequestedReportType`, `createAuditLogExportRequestSchema`, and the `useCreateAuditLogExportRequest` PostHog capture logic.
- Simplified `createAuditLogExportRequest` on the server: it no longer loops over a fan-out array to insert up to two `AuditLogExportRequest` rows and log up to two audit events per ask — every request now inserts (or idempotently reuses) exactly one row and logs exactly one `AuditLogExportCreate` event.
- Extracted `useCreateAuditLogExportRequest` as a shared hook so both the Settings form and the new Users-page button get identical success/error toasts and PostHog captures from one place.

**Bug Fixes**:

- Fixed a whitespace bug where the space between "...export it on" and the "User management" link inside the new `Infobox` was silently swallowed by the browser — `Infobox` renders its children as direct flex items, so passing text/link/text as three separate children split them into three flex boxes that each trimmed their own leading/trailing whitespace. Fixed by wrapping the sentence in a single `<Text>` child.
- Centered the "help" tooltip icon next to the month-picker label using `TouchableTooltip`'s `wrapperStyles`, which was previously sitting off-center against the ambient line-height.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- `pnpm typecheck`, `pnpm lint`, and `pnpm format` all pass.
- Full `apps/studio` unit/integration suite passes (1740 tests).
- Updated/added coverage:
  - `ExportAccessLogsButton.test.tsx` — feature-flag gating, admin-only visibility, payload construction.
  - `AuditLogExportSection.test.tsx` / `.stories.tsx` — reworked for the single-purpose Activity form (no more checkbox interactions).
  - `audit.test.ts` (schema) — added a regression test asserting `"Both"` is now rejected.
  - `audit.router.test.ts` / `auditLogExport.dedupe.test.ts` — removed the tests whose entire premise was the `Both` fan-out (two-row insert, partial-overlap dedupe, race-loser-on-one-half); kept and retargeted the tests that only incidentally used `"Both"` as a value.

- [ ] As a Site Admin with the `is-audit-log-enabled` flag on, visit a site's Users page and confirm "Export access logs" appears, requests successfully, and disappears for non-admins.
- [ ] Visit Settings → Audit logs, confirm the page now only offers the Activity export with a month picker, and that the Infobox link correctly navigates to that site's Users page.
- [ ] Submit an export from both surfaces and confirm the "Export requested" toast and the emailed download link both arrive correctly.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

The PR separates audit exports by surface and removes the combined `Both` request flow.
- Adds a one-click Access export to User Management for authorized administrators.
- Restricts the Settings form to month-scoped Activity exports.
- Simplifies request validation, persistence, analytics, and tests to handle one concrete report type per request.
</details>

<details><summary><h3>Confidence Score: 4/5</h3></summary>

The PR is not yet safe to merge because an administrator whose browser reports a later Singapore calendar month than the server cannot request the otherwise-valid Access export.

The Access button submits a browser-derived month, the server rejects that value when it appears to be in the future, and Access processing does not otherwise use the submitted month because its range is derived server-side.

**Files Needing Attention:** apps/studio/src/features/users/components/ExportAccessLogsButton.tsx
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/features/users/components/ExportAccessLogsButton.tsx | Adds the administrator-only Access export action, but its browser-derived month can still cause a spurious server rejection. |
| apps/studio/src/features/settings/AuditLogExport/AuditLogExportSection.tsx | Simplifies the Settings surface to submit only month-scoped Activity exports and directs Access exports to User Management. |
| apps/studio/src/features/settings/AuditLogExport/useCreateAuditLogExportRequest.ts | Centralizes mutation toasts and per-report analytics for both export surfaces. |
| apps/studio/src/schemas/audit.ts | Removes `Both` from the accepted report-type contract while retaining the existing month validation. |
| apps/studio/src/server/modules/audit/auditLogExport.service.ts | Replaces combined-report fan-out with one request row and one audit event per accepted request. |
| apps/studio/src/server/modules/audit/auditLogExport.query.ts | Updates report-query behavior and terminology for the split Access and Activity export flows. |

</details>

```mermaid
sequenceDiagram
  participant Admin
  participant Users as User Management
  participant Settings as Audit Settings
  participant API as Audit Export API
  participant Worker as Export Worker
  Admin->>Users: Request Access export
  Users->>API: Access request
  Admin->>Settings: Select month and request Activity export
  Settings->>API: Activity request
  API->>Worker: Queue one concrete report
  Worker-->>Admin: Email download link
```
</details>

<sub>Reviews (6): Last reviewed commit: ["chore: update tests and stories"](https://github.com/opengovsg/isomer/commit/d8412c63760d9d4c9dbe204640910e0569769dae) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53898944)</sub>

**Context used:**

- Knowledge Base — [Audit and privileged operations](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/audit-and-privileged-operations.md)
- Knowledge Base — [Site settings workflows](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/site-settings-workflows.md)


apps/studio/src/features/settings/AuditLogExport/AuditLogExportSection.tsx
```diff
@@ -202,151 +120,80 @@ export const AuditLogExportSection = ({
...
+        <Infobox
+          variant="info"
+          size="sm"
+          w="full"
+          border="1px solid"
+          borderColor="base.divider.brand"
+          borderRadius="4px"
+        >
+          <Text>
+            Looking for access history instead? You can export it on{" "}
```
dcshzj, line 136: nit: not really access history? The user management export will just be a snapshot of the list of users at the point of request

The audit logs themselves do have the access history based on the PermissionCreate and PermissionDelete

apps/studio/src/features/users/components/ExportAccessLogsButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onClick={() =>
+        createExportRequest({
+          siteId,
+          month: getCurrentSingaporeMonth(),
+          reportType: AuditLogExportRequestedReportType.Access,
+        })
+      }
+      {...buttonProps}
+    >
+      Export access logs
```
dcshzj, line 62: question: similar to the comment above, I think calling it access logs is a bit unintuitive since it is a snapshot of the list of users at the point of request, unless I got a misunderstanding of the access report?

apps/studio/src/features/users/components/__tests__/ExportAccessLogsButton.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { beforeEach, describe, expect, it, vi } from "vitest"
+import { UserManagementContext } from "~/features/users"
+import { getCurrentSingaporeMonth } from "~/schemas/audit"
+import { buildUserManagementPermissions } from "~/server/modules/permissions/permissions.util"
+import { theme } from "~/theme"
+import { RoleType } from "~prisma/generated/generatedEnums"
+
+import { ExportAccessLogsButton } from "../ExportAccessLogsButton"
+
+const SITE_ID = 42
```
dcshzj: super nit: maybe can consider putting this as some shared constant somewhere as part of the test fixtures? Seem to see this number everywhere, but nbd if not important

apps/studio/src/schemas/audit.ts
```diff
@@ -47,15 +47,11 @@ export const validateIsMonthInPastYear = (
...
-// the DB enum has no Both member.
-export const AuditLogExportRequestedReportType = {
-  ...AuditLogExportReportType,
-  Both: "Both",
-} as const
-export type AuditLogExportRequestedReportType =
-  (typeof AuditLogExportRequestedReportType)[keyof typeof AuditLogExportRequestedReportType]
+// What the user may ask for — matches the DB enum 1:1. Kept as a distinct
+// alias (rather than importing `AuditLogExportReportType` everywhere) so
+// client/service code isn't coupled directly to the Prisma-generated enum.
```
dcshzj, line 52: question(non-blocking): actually why not couple it lmao, any changes to the enum would require changes to the usages of the const and type anyway

apps/studio/src/server/modules/audit/auditLogExport.service.ts
```diff
@@ -105,74 +85,54 @@ export const createAuditLogExportRequest = async ({
...
+      .where("reportType", "=", reportType)
+      .where("status", "in", IN_FLIGHT_STATUSES)
+      .selectAll()
+
+    // Fast path: idempotent-accept the common (non-racing) duplicate.
+    const existing = await inFlightRowQuery.executeTakeFirst()
 
-      const inserted = await tx
+    const row =
+      existing ??
```
dcshzj, line 128: suggestion: I think we should split this into different branches rather than have a ternary operator here for better readability?

Something like a `let row` then if-else branch

apps/studio/src/features/settings/AuditLogExport/AuditLogExportSection.tsx
seaerchin: agreed, but i also dont' have a good name for this. maybe `User review`?

apps/studio/src/schemas/audit.ts
seaerchin: right now it's the same, but next time potentially more + diff combi (eg: `BOTH`)

apps/studio/src/server/modules/audit/auditLogExport.service.ts
seaerchin: i dont think it's a ternary? null coalescing reads left to right so should be quite readable and dont'' have to figure out if/else

apps/studio/src/features/users/components/__tests__/ExportAccessLogsButton.test.tsx
seaerchin: Good call — extracted into `SITE_ID` at `apps/studio/src/lib/testing/constants.ts` and wired it through this test, `AuditLogExportSection.test.tsx`, and `auditSettingsPage.test.tsx` in d0639b033.

apps/studio/src/features/settings/AuditLogExport/AuditLogExportSection.tsx
seaerchin: cc @sehyunidaaa - `User review` or `User overview` migth fit better?

sehyunidaaa: Hows this guys!

Reviewing who has access to manage your site? You can view and export it on {" "}

seaerchin: updated, see ss!
image

dcshzj (approved):
Btw CI tests are failing

PR #3162, opened 2026-08-18
fix(components): preserve visible link accessible names

dcshzj:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 6 | +19 | -132 |
| 🧪 Test | 3 | +143 | -111 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Closes [ISOM-2449](https://linear.app/ogp/issue/ISOM-2449/template-a11y-fix-link-accessible-name-mismatch-for-visible-text).

Visible link text was replaced by generated `aria-label` values, causing WCAG label-in-name mismatches; CSS-generated external arrows could also enter accessible names.

## Solution

Links now derive accessible names from rendered content, append new-tab context for assistive technology, preserve explicit labels for icon-only links, and render decorative indicators as `aria-hidden` DOM content.

The publishing test harness now waits for the final PostgreSQL server to become ready, preventing the CI startup race exposed by this change.

**Breaking Changes**

**Features**:

- None

**Improvements**:

- External links consistently announce when they open in a new tab.
- Decorative external-link icons are excluded from accessible names.
- Publishing tests no longer connect during PostgreSQL's temporary initialization phase.

**Bug Fixes**:

- Visible URL, email, and nested-component text is no longer replaced by a mismatched generated accessible name.

## Before & After Screenshots

**BEFORE**:

Not applicable; visual presentation is unchanged.

**AFTER**:

Not applicable; visual presentation is unchanged.

## Tests

- `pnpm exec vitest run` — 50 files and 436 tests passed
- `tsgo --noEmit --emitDeclarationOnly false`
- Targeted type-aware Oxlint and Oxfmt checks
- `pnpm exec turbo test-ci:unit --filter=publishing --env-mode=loose` — 39 tests passed
- `pnpm exec turbo test-ci:build --filter=publishing --env-mode=loose` — full template build and 4 assertions passed

- [ ] Verify visible URL and email link text remains the accessible name.
- [ ] Verify external links append `(opens in new tab)`.
- [ ] Verify decorative external-link arrows are visible but absent from the accessibility tree.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

This PR changes published-site links to derive accessible names from rendered content while preserving explicit labels for icon-only links.
- Adds screen-reader-only new-tab context to external links.
- Replaces CSS-generated external indicators with accessibility-hidden DOM content.
- Adds browser-level accessible-name coverage.
- Tightens publishing-test PostgreSQL readiness detection.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| packages/components/src/templates/next/components/internal/Link/Link.tsx | Uses rendered children for accessible names, appends new-tab context, and hides decorative external indicators. |
| packages/components/src/templates/next/components/internal/Link/Link.browser.test.tsx | Covers visible, nested, explicit, URL-like, and external-link accessible-name behavior. |
| packages/components/src/templates/next/components/internal/BaseParagraph/BaseParagraph.tsx | Delegates external-arrow rendering to Link instead of generated CSS content. |
| packages/components/src/templates/next/components/internal/Footer/Footer.tsx | Marks the external-link SVG icon as decorative. |
| packages/components/src/templates/next/components/internal/LinkButton/LinkButton.tsx | Excludes the external-link icon from the accessibility tree. |
| tooling/build/scripts/publishing/tests/global-setup.ts | Waits for PostgreSQL's final readiness message rather than its temporary initialization listener. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(test): wait for publishing database ..."](https://github.com/opengovsg/isomer/commit/0a777f8f86cf99f519298c01e32ea786ffc70491) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54215667)</sub>

**Context used:**

- Knowledge Base — [@opengovsg/isomer-components](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/components-package.md)


packages/components/src/templates/next/components/internal/Link/Link.tsx
```diff
@@ -58,14 +44,18 @@ export const Link = ({
...
       aria-current={current}
       data-current={!!current || undefined}
       disabled={isDisabled}
-    />
+    >
+      {children}
+      {showExternalIcon && <span aria-hidden="true"> ↗</span>}
+      {isExternal && !label && (
+        <span className="sr-only"> (opens in new tab)</span>
+      )}
```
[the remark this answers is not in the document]
dcshzj: Thanks for flagging this. I checked the old and new implementations against the flexbox behavior and in Chromium. The generated `::after` box was already an in-flow child/flex item of an `inline-flex` link, so `gap-1` applied to it as well (see the [CSS Flexbox specification](https://www.w3.org/TR/css-flexbox-1/#flex-items)). Both implementations also contain the same leading space: the old `content: " ↗"` and the new `<span> ↗</span>`.

A Chromium comparison produced identical widths for both implementations in flex-with-gap and regular inline layouts, so the span does not introduce additional spacing. Keeping the leading space is also necessary for non-flex inline links, where removing it would render the arrow directly against the link text.

adriangohjw, line 58: question: i think the "opens in new tab" should be before the icon? feels visually weird if its after?

dcshzj: Hmm they serve different purposes. The icon is visual-only whereas the other one is for screen readers only, so the order doesn't matter exactly

adriangohjw (approved):

adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

- **Scope (confirmed)** — `tooling/build/scripts/publishing/tests/global-setup.ts:47`: This PR bundles an unrelated flaky-test infra fix in with the a11y fix. The file changes the Testcontainers wait strategy for the publishing test's Postgres container from `Wait.forListeningPorts()` to `Wait.forLogMessage(..., 2)`, and lives in its own commit (`fix(test): wait for publishing database readiness`) separate from the accessible-name fix commit. It has nothing to do with ISOM-2449 and makes the diff harder to review/revert in isolation — consider splitting into its own PR.
- **Correctness (plausible, low severity)** — `packages/components/src/templates/next/components/internal/Link/Link.tsx:47`: Removing `generateAriaLabel` drops the old `mailto:` special-case with no replacement. If rich-text content (rendered via `BaseParagraph`/Interweave) has an anchor whose visible text is literally `mailto:jane@agency.gov.sg` (a common copy-paste pattern), a screen reader will now read the raw string verbatim instead of the previous friendlier "Send an email to jane@agency.gov.sg". This is arguably consistent with the PR's own goal of always matching the accessible name to visible text, so it's likely an accepted trade-off — but it isn't called out in the PR description or covered by the new tests.

Everything else checked out: the new `Link.browser.test.tsx` suite (9 tests) passes, the full `packages/components` unit suite (436 tests) passes, `tsc`/`tsgo` typecheck is clean, and `oxlint`/`oxfmt` are clean on all changed files. I traced the `showExternalIcon`/`isExternal`/`label` logic across all current `Link` call sites (Footer, BaseParagraph, ContactMethod, LinkButton, Breadcrumb, CreditsSection) and found no orphaned or mismatched usages — the core refactor (moving from a computed `aria-label` string to letting visible children + a `sr-only` "(opens in new tab)" suffix drive the accessible name) is sound and directly addresses the stated "accessible name mismatch" bug.

PR #3164, opened 2026-08-18
feat(studio): validate and normalize AskGov widget input

dcshzj:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 8 | +106 | -8 |
| 🧪 Test | 3 | +190 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

The AskGov ID field in Site settings → Integrations accepted any string. It expects only the agency ID (e.g. `mha`), but agencies commonly paste the full URL from their browser address bar (e.g. `https://ask.gov.sg/mha/questions/123`). That value was stored verbatim and spread straight onto the widget element as `data-agency`, so the AskGov script silently failed to load on the published site.

Closes ISOM-2340

## Solution

Validate the field at the schema level so only a bare ID or an `ask.gov.sg` URL is accepted, and normalize accepted URLs down to the agency ID on the server before the site config is persisted, so the stored value is always in the form the widget script expects.

**Breaking Changes**

  - The new `pattern` on `data-agency` applies to the whole `SiteConfigSchema`, which `updateSiteConfig` validates in full. The Name and agency and Logo settings pages both submit the entire stored config, so a site whose existing `data-agency` does not match the pattern will fail to save on those pages with a generic validation error until the value is corrected on the Integrations page. A stored value fails if it is an empty string, a bare `https://ask.gov.sg` root URL, a URL on any other host, or anything other than a slug (`[A-Za-z0-9_-]+`) once URLs are excluded. Publishing is unaffected — `SiteConfigSchema` is not validated in the build pipeline. Please run `SELECT id, config->'askgov'->>'data-agency' FROM "Site" WHERE config->'askgov' IS NOT NULL;` before release and normalize any offending rows.

**Features**:

- `getAskgovIdFromString` in `@opengovsg/isomer-components` returns a plain agency ID unchanged and extracts the ID from an ask.gov.sg link, returning `null` for anything else.

**Improvements**:

- The AskGov ID field now rejects unsupported input with the message "must be an ID or a valid ask.gov.sg URL". Subdomains (e.g. `staging.ask.gov.sg`), other hosts (e.g. `example.com/mha`), non-HTTP(S) schemes, and root URLs with no agency ID are all rejected.
- `updateSiteConfig` and `updateSiteIntegrations` normalize the AskGov config before writing, so a pasted URL is stored as the bare agency ID.
- The schema `pattern` and the extractor are both derived from one set of constants in `utils/validation.ts`, alongside the other validation patterns in the package. A unit test asserts the two agree on every input in the fixture, so the form cannot accept a value the mutation then rejects. The scheme and host match case-insensitively and surrounding whitespace is trimmed.

**Bug Fixes**:

- Pasting an AskGov URL into the ID field no longer produces a broken widget on the published site.
- The Integrations form now resets from the mutation response after saving, so the field shows the normalized ID instead of leaving the typed URL on screen with the page stuck in an unsaved state.

## Before & After Screenshots

No layout changes. The only visual differences are the inline validation error on unsupported input, and the field showing the normalized agency ID after a successful save.

**BEFORE**:

**AFTER**:

## Tests

Unit tests cover the schema pattern (`packages/components/src/interfaces/internal/__tests__/Askgov.test.ts`) and the ID extraction helper (`packages/components/src/utils/__tests__/getAskgovIdFromString.test.ts`), including accepted IDs, accepted URL forms with and without a scheme, and rejected hosts, schemes and root URLs. tRPC router tests cover normalization through both `updateSiteConfig` and `updateSiteIntegrations`.

- [ ] Go to Site settings → Integrations, enable AskGov, enter `https://ask.gov.sg/mha/questions/question-id?from=widget` and save. Confirm the success toast appears and the field then shows `mha`.
- [ ] Reload the page and confirm the field still shows `mha`, and that the Save button is disabled (no unsaved changes).
- [ ] Enter a bare ID such as `mha` and save. Confirm it is stored unchanged.
- [ ] Enter `https://ask.gov.sg` and confirm the inline error "must be an ID or a valid ask.gov.sg URL" appears and Save is disabled. Repeat with `https://staging.ask.gov.sg/mha` and `https://example.com/mha`.
- [ ] Publish the site and confirm the AskGov widget loads and opens on the published page.
- [ ] On a site that already has AskGov configured, save the Name and agency page and the Logo page, and confirm both still succeed.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

The PR validates AskGov integration values and normalizes supported AskGov URLs to agency IDs before persistence.
- Adds shared schema patterns and an AskGov ID extraction utility.
- Normalizes AskGov configuration through both site-update mutations.
- Resets the integrations form from the normalized mutation response.
- Adds schema, utility, and router coverage for accepted, rejected, and normalized inputs.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| packages/components/src/utils/validation.ts | Defines matching case-insensitive patterns for bare AskGov IDs and supported AskGov URLs. |
| packages/components/src/utils/getAskgovIdFromString.ts | Trims valid input and extracts the first AskGov URL path segment as the agency ID. |
| packages/components/src/interfaces/internal/Askgov.ts | Applies the shared AskGov validation pattern and field-specific error message. |
| apps/studio/src/server/modules/site/site.service.ts | Normalizes AskGov configuration and rejects values that cannot produce an agency ID. |
| apps/studio/src/server/modules/site/site.router.ts | Applies normalization before persisting configuration through both affected update procedures. |
| apps/studio/src/pages/sites/[siteId]/settings/integrations.tsx | Resets integration form state from the normalized server response after a successful save. |
| packages/components/CLAUDE.md | Corrects the documented test convention to match the package’s existing `.test.ts` usage. |

</details>

<sub>Reviews (2): Last reviewed commit: ["refactor(components): derive AskGov vali..."](https://github.com/opengovsg/isomer/commit/09060109e3b20ee510a875b2abcebb132c46c7a9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54225309)</sub>

### DIFF
--- a/apps/studio/src/server/modules/site/site.router.ts
+++ b/apps/studio/src/server/modules/site/site.router.ts
@@ -42,6 +42,7 @@ import {
   getNotification,
   getSiteConfig,
   getSiteTheme,
+  normalizeAskgovConfig,
   resolveEgazetteAlgoliaSearchConfig,
   setSiteNotification,
   validateUserPermissionsForSite,
@@ -136,21 +137,29 @@ export const siteRouter = router({
         .executeTakeFirstOrThrow()
 
       const { config } = site
+      const normalizedConfig = normalizeAskgovConfig({ ...rest, siteName })
 
       const updatedConfig = await db.transaction().execute(async (tx) => {
         // Preserve the existing clientId from DB - only admins can update it via setSiteConfigByAdmin
         const searchConfig =
-          rest.search?.type === "searchSG" && config.search?.type === "searchSG"
-            ? { ...rest.search, clientId: config.search.clientId }
+          normalizedConfig.search?.type === "searchSG" &&
+          config.search?.type === "searchSG"
+            ? {
+                ...normalizedConfig.search,
+                clientId: config.search.clientId,
+              }
             : // egazette-algolia is admin-managed; site admins cannot switch
               // to/from it or tamper with its Algolia credentials.
-              resolveEgazetteAlgoliaSearchConfig(config.search, rest.search)
+              resolveEgazetteAlgoliaSearchConfig(
+                config.search,
+                normalizedConfig.search,
+              )
 
         const updatedSite = await tx
           .updateTable("Site")
           .set({
             name: siteName,
-            config: jsonb({ ...rest, siteName, search: searchConfig }),
+            config: jsonb({ ...normalizedConfig, search: searchConfig }),
           })
           .where("id", "=", siteId)
           .returningAll()
@@ -201,6 +210,7 @@ export const siteRouter = router({
         .where("id", "=", ctx.user.id)
         .selectAll()
         .executeTakeFirstOrThrow()
+      const normalizedData = normalizeAskgovConfig(data)
 
       return await db.transaction().execute(async (tx) => {
         const site = await tx
@@ -214,7 +224,7 @@ export const siteRouter = router({
         // a site admin from switching back to localSearch once SearchSG is set.
         if (
           site.config.search?.type === "searchSG" &&
-          data.search?.type === "localSearch"
+          normalizedData.search?.type === "localSearch"
         ) {
           throw new TRPCError({
             code: "BAD_REQUEST",
@@ -227,12 +237,12 @@ export const siteRouter = router({
         // it or tamper with its Algolia credentials.
         const search = resolveEgazetteAlgoliaSearchConfig(
           site.config.search,
-          data.search,
+          normalizedData.search,
         )
 
         const updatedSite = await tx
           .updateTable("Site")
-          .set({ config: jsonb({ ...data, search }) })
+          .set({ config: jsonb({ ...normalizedData, search }) })
           .where("id", "=", siteId)
           .returningAll()
           .executeTakeFirstOrThrow()
--- a/packages/components/src/interfaces/internal/Askgov.ts
+++ b/packages/components/src/interfaces/internal/Askgov.ts
@@ -2,12 +2,25 @@ import type { Static } from "@sinclair/typebox"
 import type { IsomerSiteProps } from "~/types"
 import { Type } from "@sinclair/typebox"
 
+// Accepts existing IDs such as `mha`, plus AskGov URLs with or without an
+// HTTP(S) scheme, for example `https://ask.gov.sg/mha` and
+// `www.ask.gov.sg/mha/questions/123`. The ID branch remains free-form but
+// excludes URL-like values and all AskGov subdomains so only `ask.gov.sg` and
+// `www.ask.gov.sg` can use the URL form. Root URLs without an ID are rejected.
+const ASKGOV_DOMAIN_PATTERN = "(?:[A-Za-z0-9-]+\\.)*ask\\.gov\\.sg"
+const ASKGOV_HOST_PATTERN = "(?:www\\.)?ask\\.gov\\.sg"
+const ASKGOV_ID_OR_URL_PATTERN = `^(?:(?!${ASKGOV_DOMAIN_PATTERN}(?:[./?#]|$)|.*://).+|(?:https?://)?${ASKGOV_HOST_PATTERN}/[^/?#]+.*)$`
+
 export const AskgovSchema = Type.Object(
   {
     "data-agency": Type.String({
       title: "AskGov ID",
       description:
         'This is what comes after ask.gov.sg. For example, for https://ask.gov.sg/help, "help" is the ID.',
+      pattern: ASKGOV_ID_OR_URL_PATTERN,
+      errorMessage: {
+        pattern: "must be an ID or a valid ask.gov.sg URL",
+      },
     }),
     "data-topic": Type.Optional(
       Type.String({
--- a/packages/components/src/interfaces/internal/Askgov.ts
+++ b/packages/components/src/interfaces/internal/Askgov.ts
@@ -1,15 +1,7 @@
 import type { Static } from "@sinclair/typebox"
 import type { IsomerSiteProps } from "~/types"
 import { Type } from "@sinclair/typebox"
-
-// Accepts existing IDs such as `mha`, plus AskGov URLs with or without an
-// HTTP(S) scheme, for example `https://ask.gov.sg/mha` and
-// `www.ask.gov.sg/mha/questions/123`. The ID branch remains free-form but
-// excludes URL-like values and all AskGov subdomains so only `ask.gov.sg` and
-// `www.ask.gov.sg` can use the URL form. Root URLs without an ID are rejected.
-const ASKGOV_DOMAIN_PATTERN = "(?:[A-Za-z0-9-]+\\.)*ask\\.gov\\.sg"
-const ASKGOV_HOST_PATTERN = "(?:www\\.)?ask\\.gov\\.sg"
-const ASKGOV_ID_OR_URL_PATTERN = `^(?:(?!${ASKGOV_DOMAIN_PATTERN}(?:[./?#]|$)|.*://).+|(?:https?://)?${ASKGOV_HOST_PATTERN}/[^/?#]+.*)$`
+import { ASKGOV_ID_OR_URL_REGEX } from "~/utils/validation"
 
 export const AskgovSchema = Type.Object(
   {
@@ -17,7 +9,7 @@ export const AskgovSchema = Type.Object(
       title: "AskGov ID",
       description:
         'This is what comes after ask.gov.sg. For example, for https://ask.gov.sg/help, "help" is the ID.',
-      pattern: ASKGOV_ID_OR_URL_PATTERN,
+      pattern: ASKGOV_ID_OR_URL_REGEX,
       errorMessage: {
         pattern: "must be an ID or a valid ask.gov.sg URL",
       },
--- a/packages/components/src/interfaces/internal/__tests__/Askgov.test.ts
+++ b/packages/components/src/interfaces/internal/__tests__/Askgov.test.ts
@@ -1,42 +1,67 @@
 import { Value } from "@sinclair/typebox/value"
 import { describe, expect, it } from "vitest"
+import { getAskgovIdFromString } from "~/utils/getAskgovIdFromString"
 
 import { AskgovSchema } from "../Askgov"
 
 const isValidAgencyInput = (value: string) =>
   Value.Check(AskgovSchema, { "data-agency": value })
 
+const ACCEPTED = [
+  "mha",
+  "help2",
+  "agency_id",
+  " mha ",
+  "https://ask.gov.sg/mha",
+  "http://ask.gov.sg/mha/",
+  "https://www.ask.gov.sg/help/questions/question-id",
+  "HTTPS://WWW.ASK.GOV.SG/mha",
+  "ask.gov.sg/mha",
+  "www.ask.gov.sg/help/questions/question-id",
+  "  https://ask.gov.sg/mha  ",
+]
+
+const REJECTED = [
+  "",
+  "   ",
+  "custom/agency",
+  "example.com/mha",
+  "https://example.com/mha",
+  "https://ask.gov.sg",
+  "https://ask.gov.sg/",
+  "https://www.ask.gov.sg?topic=scams",
+  "https://staging.ask.gov.sg/mha",
+  "https://foo.www.ask.gov.sg/mha",
+  "https://ask.gov.sg.example.com/mha",
+  "ask.gov.sg",
+  "www.ask.gov.sg/",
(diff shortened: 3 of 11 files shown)


packages/components/src/utils/__tests__/getAskgovIdFromString.test.ts
```diff
@@ -0,0 +1,45 @@
+import { describe, expect, it } from "vitest"
```
seaerchin, line 1: question: how come we need this? the test config should automatically include

packages/components/src/utils/getAskgovIdFromString.ts
```diff
@@ -0,0 +1,33 @@
+const ASKGOV_HOSTNAMES = new Set(["ask.gov.sg", "www.ask.gov.sg"])
+const SCHEMELESS_ASKGOV_DOMAIN_PATTERN =
+  /^(?:[a-z0-9-]+\.)*ask\.gov\.sg(?:[./?#]|$)/i
```
seaerchin: comment: this is case insensitive so we need to normalise prior to passing it in

packages/components/src/utils/getAskgovIdFromString.ts
```diff
@@ -0,0 +1,33 @@
...
+      : null
+
+  if (!valueToParse) return value
+
+  try {
+    const url = new URL(valueToParse)
+
+    if (
+      !["http:", "https:"].includes(url.protocol) ||
+      !ASKGOV_HOSTNAMES.has(url.hostname)
```
seaerchin: question: this is abit odd because the regex admits subdomains like `testing.ask.gov.sg` but the valiation doesn;t?

packages/components/src/utils/getAskgovIdFromString.ts
```diff
@@ -0,0 +1,33 @@
...
+const SCHEMELESS_ASKGOV_DOMAIN_PATTERN =
+  /^(?:[a-z0-9-]+\.)*ask\.gov\.sg(?:[./?#]|$)/i
+
+/**
+ * Returns an AskGov agency ID unchanged, or extracts it from a supported URL.
+ * URL inputs must use HTTP(S), have an accepted hostname, and include the
+ * agency ID as the first path segment.
+ */
+export const getAskgovIdFromString = (value: string): string | null => {
+  const valueToParse = value.includes("://")
```
seaerchin: comment: feels like there's a fixed format - `www.ask.gov.sg/agency/3` and we should just validate against this format? this seems liek a very round-about way of doing so tbh

adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

- **[Correctness] `packages/components/src/interfaces/internal/Askgov.ts:12`** — The `data-agency` JSON-schema `pattern` is case-sensitive for the URL scheme/host, but the sibling normalizer `getAskgovIdFromString` (and its own unit tests) explicitly accept mixed-case scheme/host. The two "is this a valid AskGov ID/URL" implementations have drifted apart.
  - **Failure scenario:** I compiled the actual ajv/TypeBox schema and confirmed `validate({"data-agency": "HTTPS://WWW.ASK.GOV.SG/MHA/questions/question-id"})` returns `false` ("must be an ID or a valid ask.gov.sg URL"). This same schema backs both the live client-side form validator in `integrations.tsx` (blocks saving entirely) and the actual tRPC input validation for `updateSiteConfig`/`updateSiteIntegrations` (`apps/studio/src/schemas/site.ts`'s `isomerSiteConfigValidator`) — so the request never reaches `normalizeAskgovConfig`/`getAskgovIdFromString`, which would have handled it correctly (its own test asserts `getAskgovIdFromString("HTTPS://WWW.ASK.GOV.SG/MHA/questions/question-id")` → `"MHA"`). A perfectly valid AskGov URL gets rejected purely due to casing (e.g. `HTTPS://` or an uppercased host).

- **[Correctness, plausible] `apps/studio/src/schemas/site.ts:83`** — The new stricter `data-agency` pattern is enforced via the whole-object `isomerSiteConfigValidator` used by both mutations, but this field previously had *no* format restriction, and the PR ships no migration/backfill for already-persisted values.
  - **Failure scenario:** Any site whose stored `config.askgov["data-agency"]` doesn't match the new pattern (e.g. a previously-saved root URL like `https://ask.gov.sg` with no ID, or a value hitting the case-sensitivity bug above) will now fail validation on *every* future settings save — not just AskGov edits. `agency.tsx` and `logo.tsx` both spread the existing config (`...rest`, including the stale `askgov` value) unmodified into `updateSiteConfig.mutate(...)`, so an admin could be blocked from saving the site's name or logo until someone separately fixes the AskGov field on the integrations page. Worth confirming no such data exists in production before merging.

apps/studio/src/server/modules/site/site.router.ts
```diff
@@ -136,21 +137,29 @@ export const siteRouter = router({
...
+                ...normalizedConfig.search,
+                clientId: config.search.clientId,
+              }
             : // egazette-algolia is admin-managed; site admins cannot switch
               // to/from it or tamper with its Algolia credentials.
-              resolveEgazetteAlgoliaSearchConfig(config.search, rest.search)
+              resolveEgazetteAlgoliaSearchConfig(
+                config.search,
+                normalizedConfig.search,
+              )
```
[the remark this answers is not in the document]
adriangohjw: i think unrelated to this PR but valid and worth fixing. while `If a site is enabling SearchSG for the first time` is rare, it could be cases where we have yet to issue SearchSG for a given site

packages/components/src/interfaces/internal/Askgov.ts
```diff
@@ -2,12 +2,25 @@ import type { Static } from "@sinclair/typebox"
...
 import { Type } from "@sinclair/typebox"
 
+// Accepts existing IDs such as `mha`, plus AskGov URLs with or without an
+// HTTP(S) scheme, for example `https://ask.gov.sg/mha` and
+// `www.ask.gov.sg/mha/questions/123`. The ID branch remains free-form but
+// excludes URL-like values and all AskGov subdomains so only `ask.gov.sg` and
+// `www.ask.gov.sg` can use the URL form. Root URLs without an ID are rejected.
+const ASKGOV_DOMAIN_PATTERN = "(?:[A-Za-z0-9-]+\\.)*ask\\.gov\\.sg"
+const ASKGOV_HOST_PATTERN = "(?:www\\.)?ask\\.gov\\.sg"
+const ASKGOV_ID_OR_URL_PATTERN = `^(?:(?!${ASKGOV_DOMAIN_PATTERN}(?:[./?#]|$)|.*://).+|(?:https?://)?${ASKGOV_HOST_PATTERN}/[^/?#]+.*)$`
```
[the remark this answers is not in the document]
dcshzj: Fixed. The schema `pattern` and `getAskgovIdFromString` are now both built from shared constants in `utils/validation.ts`, and the scheme and host spell their casing out as character classes (`[Hh][Tt][Tt][Pp][Ss]?://`) instead of relying on the `i` flag, since ajv compiles JSON Schema `pattern` with the `u` flag only. `HTTPS://WWW.ASK.GOV.SG/mha` is now accepted by both layers, and a test asserts the two agree on every input in the fixture.

packages/components/src/utils/__tests__/getAskgovIdFromString.test.ts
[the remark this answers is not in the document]
dcshzj: Not valid here. The convention in this package is `.test.ts` inside a `__tests__/` folder — 52 `.test.ts` files, zero `.spec.ts` — and `vitest.config.ts` discovers them with the default glob, so nothing is missed. The `.spec.ts` line in `packages/components/CLAUDE.md` was stale, and `main` has since corrected it to `.test.ts` / `.test.tsx`.

apps/studio/src/server/modules/site/site.router.ts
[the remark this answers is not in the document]
dcshzj: Agreeing with @adriangohjw: pre-existing on `main`, not introduced by this PR. The only change here to these lines is renaming `rest` to `normalizedConfig` — the `clientId` preservation branch and the absence of any equivalent check in `updateSiteIntegrations` are both unchanged. Leaving it for a separate PR with a regression test rather than widening the scope of this one.

packages/components/src/interfaces/internal/Askgov.ts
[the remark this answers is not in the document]
dcshzj: Fixed. The ID branch is now `[A-Za-z0-9_-]+`, so a value containing dots or slashes can only match through the AskGov host branch. `example.com/mha` and `https://example.com/mha` are both rejected, and both are in the test fixtures.

packages/components/src/interfaces/internal/__tests__/Askgov.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    "ask.gov.sg",
+    "www.ask.gov.sg/",
+    "staging.ask.gov.sg/mha",
+    "foo.www.ask.gov.sg/mha",
+    "ask.gov.sg.example.com/mha",
+    "ftp://ask.gov.sg/mha",
+    "https://",
+  ])("rejects %j", (value) => {
+    expect(isValidAgencyInput(value)).toBe(false)
+  })
```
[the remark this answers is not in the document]
dcshzj: Added. `example.com/mha` and `https://example.com/mha` are now in the rejected fixture for both the schema suite and the extractor suite. There is also a new test asserting the schema and `getAskgovIdFromString` agree on every input in the fixture, so the form cannot accept a value that the mutation then rejects.

packages/components/src/utils/__tests__/getAskgovIdFromString.test.ts
dcshzj: `globals: true` in `vitest.config.ts` covers the runtime, but the package tsconfig does not register `vitest/globals` under `types`, so typecheck fails without the import — dropping it gives `TS2593: Cannot find name 'describe'`. All 48 existing `.test.ts` files in the package import from `vitest` explicitly for the same reason, so this file follows suit. Happy to add `vitest/globals` to the tsconfig and drop the imports package-wide as a follow-up if you would prefer that.

packages/components/src/utils/getAskgovIdFromString.ts
dcshzj: Agreed, and the mismatch is now gone. Case-insensitivity lives in the one shared pattern as character classes rather than the `i` flag on a separate regex, so the JSON Schema `pattern` (which ajv compiles with the `u` flag only) and the extractor accept exactly the same set of inputs. The value is trimmed as well, and a test pins the two layers together across the whole fixture.

packages/components/src/utils/getAskgovIdFromString.ts
dcshzj: That was deliberate but confusing, agreed — the schemeless regex admitted subdomains only so they would reach the hostname allowlist and be rejected there, rather than falling through and being stored verbatim as an ID. The two-step is gone: there is a single host pattern, `(www.)?ask.gov.sg`, and `testing.ask.gov.sg` now fails it directly.

packages/components/src/utils/getAskgovIdFromString.ts
dcshzj: Agreed — rewritten that way. There is now one anchored pattern for the fixed format and `getAskgovIdFromString` is a single match against it, so the `includes("://")` check, the `https://` prepend, the `new URL()` + try/catch and the hostname `Set` are all gone. The pattern sits in `utils/validation.ts` alongside the other validation patterns, and the JSON Schema `pattern` is derived from the same constants. Your example resolves as expected: `www.ask.gov.sg/agency/3` gives `agency`.

apps/studio/src/server/modules/site/site.router.ts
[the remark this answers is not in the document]
dcshzj: Follow-up: this is now fixed in #3189, off `main` so it does not wait on this PR.

Root cause was slightly broader than the report: `resolveEgazetteAlgoliaSearchConfig` returns the incoming config verbatim for every non-egazette search type, so it never guarded SearchSG at all — hence no protection in `updateSiteIntegrations` and only the already-on-searchSG case covered in `updateSiteConfig`. Both mutations now go through one `resolveSearchConfig` that always takes the `clientId` from the DB and rejects enabling SearchSG, since the editor renders that control disabled and legitimate traffic only echoes back the stored config. Three regression tests, each confirmed failing before the fix and passing after.

Not marking `!fixed` yet — #3189 is still open.

seaerchin (approved):

PR #3165, opened 2026-08-18
fix(studio): hide default Search page and block its custom blocks

dcshzj:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 8 | +109 | -116 |
| 🧪 Test | 2 | +213 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

The default Search page (permalink `/search`, no parent) is created automatically for every site and is not user-editable, but it was still listed in the site root resource table and in the directory sidebar. Clicking into it and pressing "Add block" crashed Studio: `ComponentSelector` throws `Unsupported page layout: search` because the `search` layout has no allowed-blocks entry for `ResourceType.Page`. This is a follow-up to #2080, which locked down edit/move/delete for the same page but left it reachable and its block editor unguarded.

Closes [ISOM-2511](https://linear.app/ogp/issue/ISOM-2511/studio-search-page-is-clickable-and-crashes-studio-when-add-block-is)

## Solution

Hide the page from the two lists that present the site's own content tree, and make the drawer treat `search` the same way it already treats `collection`.

**Breaking Changes**

  - Details ...

**Features**:

- N/A

**Improvements**:

- `resource.getChildrenOf` accepts a new optional `includeSearchPage` input (defaults to `true`, so existing callers are unaffected). The directory sidebar passes `false`; the link editor and redirect destination pickers keep listing `/search` so editors can still link to it.
- The reserved `search` permalink is now a single `SEARCH_PAGE_PERMALINK` constant in `~/constants/sitemap`, next to the existing `INDEX_PAGE_PERMALINK`, replacing the literal across site creation, the move/delete/settings guards, and the new listing filters. Tests keep the literal on purpose so a rename has to be acknowledged.
- Removed `SearchPageMenuItems` and the `isSearchPage` branch from `ResourceTableMenu`. Hiding the page from `listWithoutRoot` makes that branch unreachable — `ResourceTable` is its only consumer and inside a folder `parentId` is never `null` — so this is the same cleanup #2080 applied to `CollectionTableMenu`. The server-side guards on move, delete, and `updateSettings` are untouched and remain the real enforcement.

**Bug Fixes**:

- Pressing "Add block" on the Search page no longer crashes Studio. `canAddBlocks` in `RootStateDrawer` now excludes the `search` layout as well as `collection`, which hides both the "Add block" button and the empty-state "Add a new block" button.
- The Search page no longer appears in the site root resource table or the directory sidebar. `resource.countWithoutRoot` and `resource.listWithoutRoot` exclude `permalink = 'search'` when listing the site root only, so identically-named resources nested inside folders are untouched.
- The drawer's "Custom blocks — Use blocks to display your content." heading is now gated on `canAddBlocks`, so layouts that cannot take custom blocks no longer show an empty section inviting the user to add them. Previously this affected the Search page; `collection` was already excluded.

## Tests

- `pnpm test:unit -- src/server/modules/resource/__tests__/resource.router.test.ts` — 157 passed, 14 skipped, including three new cases covering `getChildrenOf` with and without `includeSearchPage`, and the Search page's exclusion from `countWithoutRoot` and `listWithoutRoot`.
- `pnpm exec vitest run --project browser src/features/editing-experience/components/Drawer/__tests__/RootStateDrawer.browser.test.tsx` — 2 passed. New browser test asserting the Search page hides the add-block affordances and the "Custom blocks" heading, plus a `content` layout case asserting they are still shown.
- After the review follow-up: `pnpm test:unit` over `resource.router.test.ts`, `page.router.test.ts`, and `site/__tests__` — 373 passed, 21 skipped; the full browser project — 8 files, 47 passed; `pnpm typecheck` and `pnpm lint` clean on every touched file.

- [ ] Open a site in Studio and confirm the root resource table no longer has a "Search" row, and that the pagination count dropped by one accordingly.
- [ ] Confirm the left directory sidebar no longer lists `/search` under Home, and that folders, collections, and pages still appear and expand normally.
- [ ] Edit any page, add a link via the link editor, and confirm `/search` is still selectable in the resource picker.
- [ ] Go to Settings → Redirects, add a redirect, and confirm `/search` is still selectable as a destination page.
- [ ] Open the Search page's editor directly via its URL (`/sites/<siteId>/pages/<searchPageId>`) and confirm the drawer shows "Fixed blocks" but no "Custom blocks" section, no "Add block" button, and no crash.
- [ ] Open a regular content page and confirm "Custom blocks" and "Add block" still work as before.
- [ ] On the root resource table and inside a folder, open the "..." menu on a page, a folder, and a collection, and confirm "Edit settings" / "Edit folder settings", "Move to...", and "Delete" all still appear and work (the dead Search-page menu branch was removed).

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

The PR hides the system-managed root Search page from Studio’s content-tree surfaces while preserving it in link-oriented resource pickers, and prevents Search pages from exposing unsupported custom-block controls.

- Adds root-only Search-page filtering to resource list and count procedures.
- Adds an opt-out flag so the directory sidebar hides Search while existing picker callers continue including it.
- Reuses a shared Search permalink constant across creation and mutation guards.
- Extends drawer tests and resource-router tests for the new behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/resource/resource.router.ts | Applies the Search-page exclusion only to root queries while preserving nested resources and default picker behavior. |
| apps/studio/src/features/editing-experience/components/Drawer/RootStateDrawer.tsx | Consistently gates the custom-block heading and both add-block affordances for collection and Search layouts. |
| apps/studio/src/features/dashboard/components/DirectorySidebar/DirectorySidebarContent.tsx | Explicitly opts the navigation tree out of receiving the system Search page. |
| apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx | Removes a menu branch rendered unreachable by the root resource query, while server-side mutation guards remain. |
| apps/studio/src/schemas/resource.ts | Adds a backward-compatible includeSearchPage input whose default preserves existing callers. |
| apps/studio/src/features/editing-experience/components/Drawer/__tests__/RootStateDrawer.browser.test.tsx | Covers hidden Search-page controls and retained controls for ordinary content pages. |
| apps/studio/src/server/modules/resource/__tests__/resource.router.test.ts | Covers picker/sidebar inclusion behavior and root list/count exclusion. |

</details>

```mermaid
flowchart TD
    Root[Root resource request] --> Surface{Calling surface}
    Surface -->|Resource table| Filter[Exclude root /search]
    Surface -->|Directory sidebar| Flag[includeSearchPage=false]
    Surface -->|Link or redirect picker| Default[includeSearchPage=true]
    Filter --> Content[Show editable content]
    Flag --> Content
    Default --> Search[Keep /search selectable]
    Search --> Editor[Direct Search editor]
    Editor --> Guard[Hide custom-block controls]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["refactor(studio): extract SEARCH\_PAGE\_PE..."](https://github.com/opengovsg/isomer/commit/ef94fa85238a916592292be469145f19d57338df) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54227073)</sub>

### DIFF
--- a/apps/studio/src/constants/sitemap.ts
+++ b/apps/studio/src/constants/sitemap.ts
@@ -1,1 +1,5 @@
 export const INDEX_PAGE_PERMALINK = "_index"
+
+// The system-managed Search page sits at the root of every site under this
+// permalink, so it is only that page when `parentId === null` as well.
+export const SEARCH_PAGE_PERMALINK = "search"
--- a/apps/studio/src/features/dashboard/components/DirectorySidebar/DirectorySidebarContent.tsx
+++ b/apps/studio/src/features/dashboard/components/DirectorySidebar/DirectorySidebarContent.tsx
@@ -55,6 +55,7 @@ export const DirectorySidebarContent = ({
         resourceId,
         siteId,
         limit: 25,
+        includeSearchPage: false,
       },
       {
         getNextPageParam: (lastPage) => lastPage.nextOffset,
--- a/apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
+++ b/apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx
@@ -28,39 +28,6 @@ interface ResourceTableMenuProps {
   parentId: ResourceTableData["parentId"]
 }
 
-// The default Search page (permalink /search, no parent) is a system-managed
-// page used to render SearchSG results. Its slug is hardcoded into the site
-// renderer, so users must not be able to edit its settings, move it, or
-// delete it — any of those would break search on the published site.
-const SearchPageMenuItems = () => {
-  return (
-    <>
-      <MenuItem
-        isDisabled
-        icon={<BiCog fontSize="1rem" />}
-        tooltip="This is a default page and its settings cannot be edited."
-      >
-        Edit settings
-      </MenuItem>
-      <MenuItem
-        isDisabled
-        icon={<BiFolderOpen fontSize="1rem" />}
-        tooltip="This is a default page that cannot be moved."
-      >
-        Move to...
-      </MenuItem>
-      <MenuItem
-        isDisabled
-        colorScheme="critical"
-        icon={<BiTrash fontSize="1rem" />}
-        tooltip="This is a default page that cannot be removed."
-      >
-        Delete
-      </MenuItem>
-    </>
-  )
-}
-
 export const ResourceTableMenu = ({
   resourceId,
   title,
@@ -75,7 +42,6 @@ export const ResourceTableMenu = ({
   const setResourceModalState = useSetAtom(deleteResourceModalAtom)
   const setFolderSettingsModalState = useSetAtom(folderSettingsModalAtom)
   const setPageSettingsModalState = useSetAtom(pageSettingsModalAtom)
-  const isSearchPage = permalink === "search" && parentId === null
 
   return (
     <Menu isLazy size="sm">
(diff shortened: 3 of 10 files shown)


adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

- **Low · Maintainability** — `apps/studio/src/features/dashboard/components/ResourceTable/ResourceTableMenu.tsx:31-78`: This PR's new unconditional `Resource.permalink != "search"` filter in `listWithoutRoot`/`countWithoutRoot` (`apps/studio/src/server/modules/resource/resource.router.ts:657`, `:691`) — the only queries backing `ResourceTable.tsx`, the sole consumer of `ResourceTableMenu` — means the Search page can no longer appear as a row in that table at all. That makes the existing `isSearchPage`/`SearchPageMenuItems` special-casing (disabled "Edit settings"/"Move to..."/"Delete" items) permanently unreachable dead code. Not a runtime bug, but worth a cleanup pass or at least a note, since a future reader would reasonably assume the search page still surfaces there (just disabled).
- **Low · Maintainability** — `apps/studio/src/server/modules/resource/resource.router.ts:657` (and `:691`, `:256`): the literal `"search"` used to identify the system-managed Search page is now duplicated across 5+ call sites (`page.router.ts`, `resource.router.ts` x3, `ResourceTableMenu.tsx`) with no shared constant. Pre-existing pattern that this PR extends rather than introduces — low risk today, but a future rename/change to the reserved permalink would require hunting down every occurrence by hand.

**Verification performed:** ran `pnpm typecheck` (clean), `pnpm lint` (0 errors, pre-existing unrelated warnings only), and the touched test suites (`resource.router.test.ts`, `RootStateDrawer.browser.test.tsx`) — all pass (1751 passed, 0 failed, 41 pre-existing skips).

seaerchin (commented):
this `getChildrenOf` is used only in 1 place and it seems like we should just make this the default behaviour rather than adding a new prop

dcshzj:
Both points addressed in ef94fa852.

**Dead `isSearchPage` branch** — removed `SearchPageMenuItems` and the `isSearchPage` ternary from `ResourceTableMenu`. Confirmed unreachable: `ResourceTable` is the only consumer, it is fed solely by `listWithoutRoot`, and inside a folder `parentId` is never `null`. This is the same cleanup #2080 applied to `CollectionTableMenu` for the same reason, so removing rather than annotating seemed the more consistent call. Enforcement still lives server-side in the `move`, `delete`, and `page.updateSettings` guards, which are untouched.

**Duplicated `"search"` literal** — extracted `SEARCH_PAGE_PERMALINK` into `~/constants/sitemap`, next to the existing `INDEX_PAGE_PERMALINK`, and replaced the literal at all six production call sites (site creation, the three guards, and the two listing filters). The test fixtures deliberately keep the raw `"search"` string so a future rename has to be acknowledged rather than silently following the constant.

Verified after the change: `resource.router.test.ts` + `page.router.test.ts` + `site/__tests__` → 373 passed, 21 skipped; full browser project → 47 passed; `pnpm typecheck` and `pnpm lint` clean on every touched file.

adriangohjw (approved):

dcshzj:
@seaerchin `getChildrenOf` has three query consumers, not one:

- `DirectorySidebarContent.tsx:53` — the side tree nav, which is the one passing `includeSearchPage: false`
- `LinkEditorModal.tsx:55` → `ResourceSelector` → `useResourceQuery.tsx:24`
- `SelectDestinationPageModal.tsx:62` (Settings → Redirects) → same path

`ResourceSelector` only routes to `getFolderChildrenOf` when `interactionType === "move"` (`ResourceSelector.tsx:100`), so both link pickers land on `getChildrenOf`.

That split is why it is a prop rather than the default. ISOM-2511 asks for the page to be hidden from the dashboard and side tree nav but explicitly still selectable in the link editor — "I should still be able to link to /search even though it is hidden from dashboard and side tree nav." Making exclusion the default would also drop `/search` from the link editor and the redirects destination picker. The `should only hide the default Search page when requested` test in `resource.router.test.ts` pins both halves of that.

PR #3169, opened 2026-08-18
fix(audit): log inactive-user permission removals under a system actor

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 3 | +67 | -1 |
| 🧪 Test | 1 | +97 | -1 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

**NOTE: already created user on staging/uat/prod**

## Problem

Closes ISOM-2376

The daily inactive-user job (`bulkDeactivateInactiveUsers` → `deactivateUsers` in `apps/studio/src/server/modules/user/inactiveUsers.service.ts`) soft-deletes a user's `ResourcePermission` rows but never wrote an `AuditLog` entry for the removal. This meant permission access could disappear with no audit trail, so there was no way to answer "why was this user's access removed" after the fact — the gap that originally surfaced from an oncall request to explain a specific access removal.

## Solution

`deactivateUsers` now writes a `PermissionDelete` audit log entry for every `ResourcePermission` it soft-deletes, inside the same transaction as the deletion. The entry is attributed to a dedicated system user (looked up by `SYSTEM_USER_EMAIL = "system@isomer.gov.sg"`) rather than a human actor, and tagged with `metadata: { reason: "inactivity" }`, so system-initiated removals are distinguishable from user-initiated ones in the audit trail and exports.

The system user row is **not** created by this PR — it must be inserted manually per environment (name/email/phone only; `id` can use its normal default). If the row is missing, the job now fails fast (checked before any deletion happens) rather than silently deleting permissions with no audit trail.

**Breaking Changes**

  - Details ...
  - Requires an operational step before deploy: insert the system user row (see below) in each environment, or the inactive-user cron job will start failing.

**Features**:

- None (this is a bug fix for missing audit coverage, no new user-facing behaviour).

**Improvements**:

- `AuditLog` `PermissionDelete` entries are now created whenever the inactive-user job revokes access, matching the same `logPermissionEvent` pattern used by human-initiated permission removals elsewhere in the codebase. These automatically surface in the existing audit log export/activity report (`auditLogExport.query.ts`) with the system user's name/email as the actor and the affected user's email in the description — no changes needed there.

**Bug Fixes**:

- Fixed: permission removals performed by the inactive-user cron job left no audit trail, making it impossible to verify audit integrity or investigate why a specific user lost access.

## Before & After Screenshots

Not applicable — backend-only change with no UI surface.

## Tests

- `apps/studio/src/server/modules/user/__tests__/inactiveUsers.service.test.ts` — added:
  - an audit log entry is created with the correct event type, actor, siteId, and metadata for a single removed permission
  - one audit log entry is created per removed permission when a user has access across multiple sites
  - the job fails and rolls back (no permissions removed, no emails sent) if the system user row does not exist
- Full existing suite still passes (`pnpm test:unit`), `pnpm typecheck`, and `pnpm lint` are clean.

**Manual step required before/at deploy** — insert the system user row in each environment, e.g.:

```sql
INSERT INTO "User" (id, name, email, phone)
VALUES (gen_random_uuid()::text, 'System', 'system@isomer.gov.sg', '00000000');
```

**New scripts**: none

**New dependencies**: none

The PR records inactive-user permission removals as audit events attributed to a dedicated system actor and prevents OTP issuance for that actor.
- Adds configurable system-actor email handling.
- Creates or reuses the system actor transactionally during inactive-user deactivation.
- Writes one `PermissionDelete` audit event per removed permission with inactivity metadata.
- Adds integration coverage for audit creation, multi-site removals, and system-user recreation.
</details>

The PR appears safe to merge because no blocking failure remains within the eligible follow-up-review scope.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/env.mjs | Adds the optional server-side system-user email setting with a lowercase default. |
| apps/studio/src/server/modules/auth/email/email.router.ts | Suppresses OTP creation and delivery when a login request targets the configured system actor. |
| apps/studio/src/server/modules/user/inactiveUsers.service.ts | Creates or reuses the system actor and transactionally records an audit event for every permission soft-deletion. |
| apps/studio/src/server/modules/user/__tests__/inactiveUsers.service.test.ts | Covers audit attribution, per-site event creation, and recovery when the system actor is absent. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: omit saving otp to db when user is ..."](https://github.com/opengovsg/isomer/commit/42fc697bc0f38c8ab09a2485d11b3168e350f4df) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54253056)</sub>

- Knowledge Base — [Authentication flow (sign-in)](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/auth-flow.md)
- Knowledge Base — [Dashboard, Users, Godmode, and Me](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/dashboard-and-admin-surfaces.md)
- Knowledge Base — [Publishing Pipeline](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/publishing-pipeline.md)
</details>

### DIFF
--- a/apps/studio/src/server/modules/user/constants.ts
+++ b/apps/studio/src/server/modules/user/constants.ts
@@ -1,1 +1,6 @@
 export const MAX_DAYS_FROM_LAST_LOGIN = 90 // hardcoded to 90 as per IM8 requirement
+
+// Actor used to attribute audit log entries created by automated jobs (e.g.
+// inactive user removal) rather than a human. This row is not created by
+// application code — it must be inserted manually in each environment.
+export const SYSTEM_USER_EMAIL = "system@isomer.gov.sg"
--- a/apps/studio/src/server/modules/user/constants.ts
+++ b/apps/studio/src/server/modules/user/constants.ts
@@ -1,6 +1,1 @@
 export const MAX_DAYS_FROM_LAST_LOGIN = 90 // hardcoded to 90 as per IM8 requirement
-
-// Actor used to attribute audit log entries created by automated jobs (e.g.
-// inactive user removal) rather than a human. This row is not created by
-// application code — it must be inserted manually in each environment.
-export const SYSTEM_USER_EMAIL = "system@isomer.gov.sg"
--- a/apps/studio/src/server/modules/user/inactiveUsers.service.ts
+++ b/apps/studio/src/server/modules/user/inactiveUsers.service.ts
@@ -6,12 +6,14 @@ import {
   sendAccountDeactivationWarningEmail,
 } from "~/features/mail/service"
 import { createBaseLogger } from "~/lib/logger"
+import { AuditLogEvent } from "~prisma/generated/generatedEnums"
 
 import type { ResourcePermission, Site, User } from "../database"
 import type { BulkSendAccountDeactivationWarningEmailsProps } from "./types"
+import { logPermissionEvent } from "../audit/audit.service"
 import { db, RoleType, sql } from "../database"
 import { PG_ERROR_CODES } from "../database/constants"
-import { MAX_DAYS_FROM_LAST_LOGIN } from "./constants"
+import { MAX_DAYS_FROM_LAST_LOGIN, SYSTEM_USER_EMAIL } from "./constants"
 
 const logger = createBaseLogger({
   path: "server/modules/user/inactiveUsers.service",
@@ -137,13 +139,50 @@ const deactivateUsers = async ({ userIds }: DeactivateUsersProps) => {
       .transaction()
       .setIsolationLevel("serializable") // for idempotency
       .execute(async (tx) => {
-        return tx
+        // Fetch first so a missing system user row fails the job
+        // immediately, even on a run with nothing to deactivate.
+        const systemUser = await tx
+          .selectFrom("User")
+          .where("email", "=", SYSTEM_USER_EMAIL)
+          .selectAll()
+          .executeTakeFirstOrThrow()
+
+        const permissionsToDelete = await tx
+          .selectFrom("ResourcePermission")
+          .where("userId", "in", userIds)
+          .where("deletedAt", "is", null)
+          .selectAll()
+          .execute()
+
+        if (permissionsToDelete.length === 0) return []
+
+        const updated = await tx
           .updateTable("ResourcePermission")
           .where("userId", "in", userIds)
           .where("deletedAt", "is", null)
           .set({ deletedAt: new Date() })
           .returningAll()
           .execute()
+
+        for (const after of updated) {
+          const before = permissionsToDelete.find((p) => p.id === after.id)
+          // Not expected: same tx/isolation level as the read above.
+          if (!before) {
+            throw new Error(
+              `Could not find pre-update state for ResourcePermission ${after.id}`,
+            )
+          }
+
+          await logPermissionEvent(tx, {
+            eventType: AuditLogEvent.PermissionDelete,
+            by: systemUser,
+            delta: { before, after },
+            siteId: after.siteId,
+            metadata: { reason: "inactivity" },
+          })
+        }
+
+        return updated
       })
   } catch (error) {
     if (
--- a/apps/studio/src/server/modules/user/inactiveUsers.service.ts
+++ b/apps/studio/src/server/modules/user/inactiveUsers.service.ts
@@ -1,3 +1,4 @@
+import { createId } from "@paralleldrive/cuid2"
 import { startOfDay, subDays } from "date-fns"
 import { toZonedTime } from "date-fns-tz"
 import { env } from "~/env.mjs"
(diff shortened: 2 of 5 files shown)


adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

No issues found at medium effort.

Verified: full `pnpm test:unit` suite (1749 tests) passes, `tsc --noEmit` and `oxlint` are clean on the touched files. The new audit-logging path runs inside the existing serializable transaction (before permission soft-delete is committed), correctly rolls back with no permissions removed and no emails sent when the system-user lookup fails, and follows the existing `logPermissionEvent` convention used elsewhere. The one operational dependency worth being aware of — the `system@isomer.gov.sg` actor row must exist in each environment or the cron job fails fast — is already called out and handled in the PR description, so it isn't flagged here as a defect.

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -137,13 +139,50 @@ const deactivateUsers = async ({ userIds }: DeactivateUsersProps) => {
...
       .setIsolationLevel("serializable") // for idempotency
       .execute(async (tx) => {
-        return tx
+        // Fetch first so a missing system user row fails the job
+        // immediately, even on a run with nothing to deactivate.
+        const systemUser = await tx
+          .selectFrom("User")
+          .where("email", "=", SYSTEM_USER_EMAIL)
+          .selectAll()
+          .executeTakeFirstOrThrow()
```
dcshzj, line 161: question: actually curious why not do an upsert instead so that we can guarantee this user exists?

Consideration is that we may spin up local environments and would not work out of the box

apps/studio/src/server/modules/user/constants.ts
```diff
@@ -1 +1,6 @@
 export const MAX_DAYS_FROM_LAST_LOGIN = 90 // hardcoded to 90 as per IM8 requirement
+
+// Actor used to attribute audit log entries created by automated jobs (e.g.
+// inactive user removal) rather than a human. This row is not created by
+// application code — it must be inserted manually in each environment.
+export const SYSTEM_USER_EMAIL = "system@isomer.gov.sg"
```
dcshzj: suggestion: ps I am thinking we can put this as an env var instead of being hard-coded in the codebase, but nbd, can ignore if too troublesome

apps/studio/src/server/modules/user/inactiveUsers.service.ts
seaerchin: oh hmm fair enough - lemme update to upsert

apps/studio/src/server/modules/user/constants.ts
seaerchin: i'll do a env var and default this to `system@isomer.gov.sg`!

apps/studio/src/server/modules/user/inactiveUsers.service.ts
```diff
@@ -137,13 +140,62 @@ const deactivateUsers = async ({ userIds }: DeactivateUsersProps) => {
...
-        return tx
+        // Upsert so the system user is guaranteed to exist for audit-log
+        // attribution, rather than relying on it being created manually.
+        // The no-op doUpdateSet forces RETURNING to yield the existing row
+        // on conflict, rather than doNothing (which returns nothing).
+        const systemUser = await tx
+          .insertInto("User")
+          .values({
+            id: createId(),
+            email: env.SYSTEM_USER_EMAIL,
```
dcshzj, line 151: issue: oh ya sorry I remembered about 2 potential issues:

1. Flagged below but would the system user be subject to the 90-day inactivity rule? I not sure if the last login date is unset would the inactivity cron skip it
2. It is technically possible for someone to login to the system user account. Very minor but we can consider preventing logins from happening if the email address matches the configured system user email

seaerchin: 1 - i think we're subject to the rule but we don't actually delete users, we just remove their `ResourcePermissions` 

```ts
    deletedPermissions = await db
      .transaction()
      .setIsolationLevel("serializable") // for idempotency
      .execute(async (tx) => {
        return tx
          .updateTable("ResourcePermission")
          .where("userId", "in", userIds)
          .where("deletedAt", "is", null)
          .set({ deletedAt: new Date() })
          .returningAll()
          .execute()
      })
```

this is in `inactiveUsers.service.ts` -> `deactiveUsers`

dcshzj: Perfect, sounds good, thanks!

dcshzj (approved):

apps/studio/src/server/modules/user/inactiveUsers.service.ts
seaerchin: for 2 - sounds good, lemme skip generating OTP if the user is the system user

PR #3171, opened 2026-08-18
RTE: link bubble menu (Edit/Remove) + fix empty-link validation & link colour (ISOM-2363)

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 6 | +142 | -24 |
| 🧪 Test | 1 | +139 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Summary
- Adds a contextual bubble menu on RTE links with Edit / Remove actions (using the official `@tiptap/react/menus` `BubbleMenu`), and a "Remove link" button inside the edit modal itself — both undoable via the editor's existing Ctrl-Z history, no confirmation dialog.
- Fixes validation so a link can no longer be saved with an empty destination:
  - `linkHref` is now required in the modal's zod schema (was `.optional()` as a workaround for the FileAttachment flow).
  - Fixes `LinkHrefEditor.tsx`'s External/Email `onChange` handlers, which previously always overwrote a cleared field with a bare `"https://"`/`"mailto:"` instead of `""`, silently bypassing validation.
  - Adds a schema-level `.refine()` backstop rejecting bare `"https://"`/`"mailto:"` values.
- Improves the empty "Link text" Zod error message to "Link text cannot be empty." instead of the raw Zod message.
- Applies the `interaction.links.default`/`.hover` colour tokens to rendered links in the RTE (previously unstyled).

Linear: ISOM-2363

Filed as a separate follow-up (not in scope here): [ISOM-2521](https://linear.app/ogp/issue/ISOM-2521/link-editor-modal-switching-link-type-doesnt-clearsync-href-silently) — switching link type in the modal doesn't clear/sync `href`, so it's possible to silently save a stale destination under the wrong displayed type.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `oxlint` / `oxfmt --check` clean on touched files
- [x] `pnpm test:unit` — 1704 passed, 0 new failures
- [ ] Manual: click into an existing link → bubble menu appears with Edit/Remove
- [ ] Manual: Remove → link mark disappears, Ctrl-Z restores it
- [ ] Manual: Edit → opens the same modal as the toolbar button, prefilled correctly
- [ ] Manual: clearing the External/Email field disables Save with a clear inline error
- [ ] Manual: link colour in rendered content matches `interaction/links/default`
- [ ] Manual: FileAttachment upload-a-file-as-link flow still works with the tightened validation

The PR adds link editing and removal controls to the rich-text editor while tightening destination validation and applying the standard interaction-link colors.
- Adds a contextual link bubble menu with Edit and Remove actions.
- Supports removing links from the edit modal.
- Rejects empty, malformed, and unsafe link destinations.
- Corrects External and Email field clearing behavior.
- Styles rendered links with the default and hover interaction tokens.
</details>

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/components/PageEditor/LinkEditorModal.tsx | Adds shared destination validation and modal-level link removal without leaving an eligible blocking follow-up issue. |
| apps/studio/src/components/PageEditor/MenuBar/TiptapLinkBubbleMenu.tsx | Adds contextual Edit and Remove actions and now uses Chakra theme-scale values for spacing and radius. |
| apps/studio/src/components/PageEditor/MenuBar/TiptapLinkEditorModal.tsx | Connects modal removal to Tiptap's link-mark removal command. |
| apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx | Prevents cleared External and Email fields from being replaced with bare schemes. |
| apps/studio/src/styles/tiptap.scss | Applies interaction-link default and hover color tokens to rich-text links. |
| apps/studio/src/components/PageEditor/LinkEditorModal.test.ts | Covers accepted destinations and empty, malformed, and unsafe href validation cases. |

</details>

<sub>Reviews (8): Last reviewed commit: ["fix(rte): add type-aware z.url() guard f..."](https://github.com/opengovsg/isomer/commit/eb52359ffccf7475750748dbf1928797729a175c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54263492)</sub>

### DIFF
--- a/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
+++ b/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
@@ -46,6 +46,31 @@ import { AttachmentData } from "../AttachmentData"
 import { ResourceSelector } from "../ResourceSelector"
 import { FileAttachment } from "./FileAttachment"
 
+// Explicit, tested backstop alongside Tiptap's own `isAllowedUri` link-mark
+// guard — keeps the allowlist visible here rather than relying solely on
+// upstream library defaults.
+const DISALLOWED_HREF_SCHEMES = ["javascript:", "data:", "vbscript:"]
+
+export const linkEditorSchema = z.object({
+  linkText: z.string().min(1, "Link text cannot be empty."),
+  linkHref: z
+    .string()
+    .min(1, "Link destination cannot be empty.")
+    // A cleared External/Email field still lands here as a bare scheme
+    // ("https://"/"mailto:"), not "", so `.min(1)` alone won't catch it.
+    .refine(
+      (href) => !["https://", "mailto:"].includes(href),
+      "Link destination cannot be empty.",
+    )
+    .refine(
+      (href) =>
+        !DISALLOWED_HREF_SCHEMES.some((scheme) =>
+          href.trim().toLowerCase().startsWith(scheme),
+        ),
+      "Link destination is not allowed.",
+    ),
+})
+
 interface PageLinkElementProps {
   value: string
   onChange: (value: string) => void
@@ -98,18 +123,7 @@ const LinkEditorModalContent = ({
     formState: { errors },
   } = useZodForm({
     mode: "onChange",
-    schema: z.object({
-      linkText: z.string().min(1, "Link text cannot be empty."),
-      linkHref: z
-        .string()
-        .min(1, "Link destination cannot be empty.")
-        // A cleared External/Email field still lands here as a bare scheme
-        // ("https://"/"mailto:"), not "", so `.min(1)` alone won't catch it.
-        .refine(
-          (href) => !["https://", "mailto:"].includes(href),
-          "Link destination cannot be empty.",
-        ),
-    }),
+    schema: linkEditorSchema,
     defaultValues: {
       linkText: strippedLinkText,
       linkHref,
--- a/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
+++ b/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
@@ -67,7 +67,19 @@ export const linkEditorSchema = z.object({
     .refine(
       (href) => DOMPurify.isValidAttribute("a", "href", href),
       "Link destination is not allowed.",
-    ),
+    )
+    // Page/File hrefs aren't real URLs ([resource:...], /uuid/path) -- only
+    // External/Email are expected to be well-formed URLs. Catches
+    // malformed-but-scheme-safe input (e.g. a stray space in the host)
+    // that DOMPurify doesn't check, since it only validates scheme safety,
+    // not URL structure.
+    .refine((href) => {
+      const type = getLinkHrefType(href)
+      if (type === LINK_TYPES.External || type === LINK_TYPES.Email) {
+        return z.url().safeParse(href).success
+      }
+      return true
+    }, "Link destination is not a valid URL."),
 })
 
 interface PageLinkElementProps {
--- a/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
+++ b/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
@@ -51,6 +51,19 @@ import { FileAttachment } from "./FileAttachment"
 // upstream library defaults.
 const DISALLOWED_HREF_SCHEMES = ["javascript:", "data:", "vbscript:"]
 
+// Browsers strip C0 control chars (tabs, newlines, CRs) and various unicode
+// whitespace from URLs before parsing the scheme, so "java\tscript:" is
+// interpreted as "javascript:". Same char class Tiptap's own isAllowedUri
+// uses (borrowed from DOMPurify) — must strip it too or this check is a
+// no-op against that bypass.
+const URL_WHITESPACE_REGEX =
+  /[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205F\u3000]/g
+
+const hasDisallowedHrefScheme = (href: string) => {
+  const normalized = href.replace(URL_WHITESPACE_REGEX, "").toLowerCase()
+  return DISALLOWED_HREF_SCHEMES.some((scheme) => normalized.startsWith(scheme))
+}
+
 export const linkEditorSchema = z.object({
   linkText: z.string().min(1, "Link text cannot be empty."),
   linkHref: z
@@ -63,10 +76,7 @@ export const linkEditorSchema = z.object({
       "Link destination cannot be empty.",
     )
     .refine(
-      (href) =>
-        !DISALLOWED_HREF_SCHEMES.some((scheme) =>
-          href.trim().toLowerCase().startsWith(scheme),
-        ),
+      (href) => !hasDisallowedHrefScheme(href),
       "Link destination is not allowed.",
     ),
 })
--- a/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
+++ b/apps/studio/src/components/PageEditor/LinkEditorModal.tsx
@@ -3,12 +3,14 @@ import type { LinkTypes } from "~/features/editing-experience/components/LinkEdi
 import {
   Box,
   FormControl,
+  HStack,
   Modal,
   ModalBody,
   ModalContent,
   ModalFooter,
   ModalHeader,
   ModalOverlay,
+  Spacer,
 } from "@chakra-ui/react"
 import {
   Button,
@@ -71,7 +73,7 @@ const PageLinkElement = ({ value, onChange }: PageLinkElementProps) => {
 
 type LinkEditorModalContentProps = Pick<
   LinkEditorModalProps,
-  "linkText" | "linkHref" | "showLinkText" | "linkTypes" | "onSave"
+  "linkText" | "linkHref" | "showLinkText" | "linkTypes" | "onSave" | "onRemove"
 >
 
 const LinkEditorModalContent = ({
@@ -79,6 +81,7 @@ const LinkEditorModalContent = ({
   linkHref,
   showLinkText = true,
   onSave,
+  onRemove,
   linkTypes,
 }: LinkEditorModalContentProps) => {
   const { strippedLinkText, onUploadedFile, buildFinalLinkTextForSave } =
@@ -96,10 +99,16 @@ const LinkEditorModalContent = ({
   } = useZodForm({
     mode: "onChange",
     schema: z.object({
-      linkText: z.string().min(1),
-      // TODO: Refactor to be required
-      // Context: quick hack to ensure error message don't shown for empty linkHref for FileAttachment
-      linkHref: z.string().min(1).optional(),
+      linkText: z.string().min(1, "Link text cannot be empty."),
+      linkHref: z
+        .string()
+        .min(1, "Link destination cannot be empty.")
+        // A cleared External/Email field still lands here as a bare scheme
+        // ("https://"/"mailto:"), not "", so `.min(1)` alone won't catch it.
+        .refine(
+          (href) => !["https://", "mailto:"].includes(href),
+          "Link destination cannot be empty.",
+        ),
     }),
     defaultValues: {
       linkText: strippedLinkText,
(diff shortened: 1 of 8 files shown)


apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -110,12 +119,8 @@ const LinkEditorModalContent = ({
 
   const isEditingLink = !!linkText && !!linkHref
 
-  const onSubmit = handleSubmit(
-    // TODO: Refactor to not have to check for !!linkHref
-    // Context: quick hack to ensure error message don't shown for empty linkHref for FileAttachment
-    ({ linkText, linkHref }) =>
-      !!linkHref &&
-      onSave(buildFinalLinkTextForSave(linkText, linkHref), linkHref),
+  const onSubmit = handleSubmit(({ linkText, linkHref }) =>
```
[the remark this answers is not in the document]
Ssatu: `!fp` — verified against the installed `@tiptap/extension-link@3.29.2` source (not assumed from general docs): its `isAllowedUri` protocol allowlist (`http/https/ftp/ftps/mailto/tel/callto/sms/cid/xmpp`, no `javascript:`/`data:`) is on by default and runs inside `parseHTML()`, exactly the code path `insertContent(htmlString)` goes through. A `javascript:` href fails that check, so no link mark gets created at all — it's inserted as plain unlinked text, not a clickable payload.

For the `linkText` HTML-breakout half: `insertContent` parses via the browser's real `DOMParser` (not string concatenation), and the result is then filtered through ProseMirror's schema (`BASE_EXTENSIONS`/`PROSE_EXTENSIONS` in `useTextEditor/constants.ts`), which has no `script`/`img`/other injectable node types registered — so a literal `<script>` in link text can't become an executable node in this editor.

Library-level defense-in-depth already covers this for the currently pinned version; nothing for this PR to add.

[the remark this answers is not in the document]
Ssatu: that being said, will add a safeguard on our side enforced via zod schema to not purely rely on library's logic

adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

No blocking issues found at medium effort. Verified: all 1764 unit tests pass (including the 11 new `linkEditorSchema` cases), `tsc`/`tsgo --noEmit` is clean, and lint is clean on the touched files. Confirmed the new `hasDisallowedHrefScheme`/`URL_WHITESPACE_REGEX` guard exactly mirrors Tiptap's own `isAllowedUri` whitespace-stripping (verified against `@tiptap/extension-link` source), and Tiptap already blocks `javascript:`/`data:`/`vbscript:` hrefs at the mark-parsing level independently — so this is a genuine defense-in-depth/UX improvement, not a fix for an actual open XSS hole. Also confirmed the real root cause of the "empty link" bug: `LinkHrefEditor.tsx`'s onChange handlers were missing an early `return`, so clearing the External/Email field fell through into `setHref(generateHttpsLink(""))`/`setHref("mailto:")`, silently re-setting a bare scheme — the added `return` statements correctly fix this. Both new theme tokens (`interaction.links.default/hover`, `base.divider.medium`) exist in `theme/generated/colours.ts`.

Two minor, non-blocking observations:

- **Test coverage** — `apps/studio/src/features/editing-experience/components/LinkEditor/LinkHrefEditor.tsx:67,88`: the actual bug fix (early `return` after clearing the External/Email href input) has no direct regression test. The new `LinkEditorModal.test.ts` only exercises `linkEditorSchema.safeParse` with literal `"https://"`/`"mailto:"` strings, so a future regression that removes these `return`s again would not be caught by it.
- **Maintainability** — `apps/studio/src/components/PageEditor/MenuBar/TiptapLinkBubbleMenu.tsx:42` vs `TiptapLinkEditorModal.tsx`: the link-removal chain `editor.chain().focus().extendMarkRange("link").unsetLink().run()` is duplicated verbatim in both places. Worth extracting into a shared helper so the two "Remove link" affordances can't drift apart.

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -44,6 +46,41 @@ import { AttachmentData } from "../AttachmentData"
...
+// interpreted as "javascript:". Same char class Tiptap's own isAllowedUri
+// uses (borrowed from DOMPurify) — must strip it too or this check is a
+// no-op against that bypass.
+const URL_WHITESPACE_REGEX =
+  /[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205F\u3000]/g
+
+const hasDisallowedHrefScheme = (href: string) => {
+  const normalized = href.replace(URL_WHITESPACE_REGEX, "").toLowerCase()
+  return DISALLOWED_HREF_SCHEMES.some((scheme) => normalized.startsWith(scheme))
+}
```
dcshzj: issue: hmm hand-rolling our own security checks is not usually ideal since there can be a lot of edge cases that we are not aware of, so we should offload it to trusted third-party libraries that specialise in this area

In this case, you can consider making use of DOMPurify's `isValidAttribute` which then does the check of the URL, and it covers more than just those that are listed here. So the code would look something like this:

```
import DOMPurify from "isomorphic-dompurify"

export const linkEditorSchema = z.object({
  linkText: z.string().min(1, "Link text cannot be empty."),
  linkHref: z
    .string()
    .min(1, "Link destination cannot be empty.")
    .refine(
      (href) => !["https://", "mailto:"].includes(href),
      "Link destination cannot be empty.",
    )
    .refine(
      (href) => DOMPurify.isValidAttribute("a", "href", href),
      "Link destination is not allowed.",
    ),
})
```

[Reference here](<https://github.com/cure53/DOMPurify/issues/536#issuecomment-1689879538>)

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -222,6 +265,13 @@ export const LinkEditorModal = ({
           onSave(linkText, linkHref)
           onClose()
         }}
+        onRemove={
+          onRemove &&
+          (() => {
+            onRemove()
+            onClose()
+          })
```
dcshzj, line 274: question: why not do it like this:

```
onRemove={() => {
    onRemove?.()
    onClose()
}}
```

This would allow the `onClose()` function to continue working, thus closing the model, if the original `onRemove()` function is not provided but is used by the modal

apps/studio/src/components/PageEditor/MenuBar/TiptapLinkBubbleMenu.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onEdit: () => void
+  isLinkModalOpen: boolean
+}
+
+export const TiptapLinkBubbleMenu = ({
+  editor,
+  onEdit,
+  isLinkModalOpen,
+}: TiptapLinkBubbleMenuProps) => {
+  const options = useMemo(() => ({ placement: "bottom" as const }), [])
```
dcshzj: question: why do we need the `useMemo` here?

apps/studio/src/features/editing-experience/components/form-builder/renderers/TipTapEditor/components.tsx
```diff
@@ -57,6 +62,14 @@ const EditorContentWrapper = ({
...
       onClick={() => editor?.chain().focus().run()}
       cursor="text"
+      sx={{
+        "& a": {
+          color: linkColor,
+          _hover: {
+            color: linkHoverColor,
+          },
+        },
+      }}
```
dcshzj: issue: I think this might not be the right place to update the CSS of the text editor, there is a tiptap.scss file that we store the editor stylings in, under the `.tiptap` class

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
Ssatu: this was quite confusing for me too, please let me know if i make sense!

because LinkEditorModal is shared, onRemove might not be passed in (which represents whether we need the modal to support a remove link button), 

- we want to pass that information down to the layer responsible for rendering the button so that it can choose whether or not to render
- hence if **LinkEditorModal does not receive an onRemove**, we want to also not pass a function object to the button layer which requires the (onRemove &&) step.

otherwise, we will always be passing a function up to the button renderer and it will render "remove link" even for places where the modal should not use it (like in a button's link - BaseLinkControl.tsx)

apps/studio/src/features/editing-experience/components/form-builder/renderers/TipTapEditor/components.tsx
Ssatu: agreed, thanks for catching this

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
Ssatu: agreed!

apps/studio/src/components/PageEditor/MenuBar/TiptapLinkBubbleMenu.tsx
Ssatu: Without it, { placement: "bottom" } is a brand new object every time this component re-renders

1. BubbleMenu watches that options prop, and when it sees a "different" object, it tears down and rebuilds its internal plugin. 
2. That rebuild fires an editor update, and since our editor is configured to re-render on every update (shouldRerenderOnTransaction: true),
3. Re-render creates a new { placement: "bottom" } object again, which loops back to (1)

I hit this as a real crash while testing in the browser ("Maximum update depth exceeded").

useMemo just makes sure we reuse the same object across renders instead of rebuilding it every time!

seaerchin: comment: does placing the object outsid eof the component itself solve this issue? react compares object by ref which is why inside the function body is troublesome but outside the function body maintains a stable reef

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -44,6 +47,25 @@ import { AttachmentData } from "../AttachmentData"
...
+  linkText: z.string().min(1, "Link text cannot be empty."),
+  linkHref: z
+    .string()
+    .min(1, "Link destination cannot be empty.")
+    // A cleared External/Email field still lands here as a bare scheme
+    // ("https://"/"mailto:"), not "", so `.min(1)` alone won't catch it.
+    .refine(
+      (href) => !["https://", "mailto:"].includes(href),
+      "Link destination cannot be empty.",
+    )
```
seaerchin, line 64: question: should we do `href.trim()` or otherwise, stricten validation to use `z.url` if the `linkHref` is not an internal/file link?

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
dcshzj: I see okay, thanks for the explanation!

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
Ssatu: agreed, implemented!

apps/studio/src/components/PageEditor/LinkEditorModal.tsx
```diff
@@ -67,7 +67,19 @@ export const linkEditorSchema = z.object({
...
+    // External/Email are expected to be well-formed URLs. Catches
+    // malformed-but-scheme-safe input (e.g. a stray space in the host)
+    // that DOMPurify doesn't check, since it only validates scheme safety,
+    // not URL structure.
+    .refine((href) => {
+      const type = getLinkHrefType(href)
+      if (type === LINK_TYPES.External || type === LINK_TYPES.Email) {
+        return z.url().safeParse(href).success
+      }
+      return true
```
seaerchin, line 81: comment: maybe we should validate jsut the shape (eg: internal links are :`[resource:siteId:resourceId]` and links are `/uuid/fliename`)

seaerchin (approved):
lgtm but i think we should just buff up the validation for file/internal.

Issue #3056, opened 2026-08-05
Pages cannot be moved directly to the main/root directory


UltraRaptorYT:
**Describe the bug**
Pages cannot be moved directly to the site’s main/root directory from the Move modal.

The root directory is not immediately shown as a selectable destination. To reach it, the user must first navigate into a folder and then click Back to parent until the main directory appears.

This affects pages generally, not only pages that are already nested.

**To Reproduce**
1. Go to a site in Isomer Studio
2. Open the actions menu for any movable page.
3. Click Move.
4. Observe that the main/root directory is not directly available as a destination.
5. Navigate into any folder.
6. Click Back to parent until the main/root directory is shown.
7. The page can then be moved to the root.

**Expected behavior**
The main/root directory should be directly visible or selectable when the Move modal opens.

Users should not need to enter another folder and navigate back upwards before they can choose the root directory.

**Screenshots**
Initial Move modal without a direct root option
<img width="539" height="436" alt="Image" src="https://github.com/user-attachments/assets/01b54e4f-8337-4077-8013-1ffb23217d27" />

GIF explaination
<img width="800" height="500" alt="Image" src="https://github.com/user-attachments/assets/5f2ace9b-0bc2-4f42-89a9-29a6219b3eec" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Google Chrome
 - Version: Version 151.0.7922.71

**Smartphone (please complete the following information):**
 - Device: Not Tested
 - OS: Not Tested
 - Browser: Not Tested
 - Version: Not Tested

**Additional context**
This is a usability and navigation issue rather than a complete inability to move pages to the root.

A direct option such as Home, Site root, or Main directory should be available in the initial destination view.

UltraRaptorYT:
<img width="800" height="500" alt="Image" src="https://github.com/user-attachments/assets/2ba2979f-b573-4706-916a-388137d6d064" />

grMLEqomlkkU5Eeinz4brIrOVCUCkJuN:
This is fixed in #3176 i think, but it is not yet in release

PR #3176, opened 2026-08-18
feat(studio): support selecting the homepage as a link/move destination

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 5 | +88 | -8 |
| 🧪 Test | 2 | +31 | -2 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Closes ISOM-2243

Studio's resource picker (used for both "link to a page" and "move resource") had no way to target the site's homepage — the "Home" row shown at the top of the picker was purely decorative. It wasn't clickable, so there was no way to pick the site root as a link or move destination, and it never reflected whether it was the currently selected destination.

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- The "Home" row in the resource picker (`ResourceSelector`) is now clickable, letting users pick the site root as a link/move destination.
- The Home row now shows the same "selected" highlight styling that a regular folder row gets once it's the chosen destination (`data-selected` + `_selected` styles, mirroring `ResourceItem`).

**Improvements**:

- `isHomeHighlighted` is derived in `useResourceSelector` from `moveDest?.type === RootPage`, avoiding the need to special-case a nullable resource ID just to detect the root selection.

**Bug Fixes**:

- None

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- New Storybook interaction story: `Pages/Site Management/Move Resource Modal` → `Home Selected` — clicks the Home row and asserts it picks up the selected styling.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

This PR allows the shared resource picker to select and highlight the site homepage for link and move workflows.
- Makes the Home row interactive and applies selected-state styling.
- Represents homepage moves as a null parent while preserving the root-page ID for links.
- Encapsulates root-page fetching and homepage stack initialization in the resource-stack hook.
- Adds Storybook coverage for selecting Home and adjusts an existing resource-router test description.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/components/ResourceSelector/ResourceSelector.tsx | Wires the root page into the Home-row selection flow through the resource-stack hook. |
| apps/studio/src/components/ResourceSelector/ResourceSelectorHeader.tsx | Makes the Home header keyboard-accessible and visually reflects its selected state. |
| apps/studio/src/components/ResourceSelector/useResourceSelector.tsx | Tracks homepage highlighting and translates the root page to a null parent only for move operations. |
| apps/studio/src/components/ResourceSelector/useResourceStack.tsx | Encapsulates root-page retrieval and restores homepage selections in the resource ancestry stack. |
| apps/studio/src/stories/Page/MoveResourceModal.stories.tsx | Adds an interaction story asserting that selecting Home applies the selected state. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(studio): resolve root-move destinati..."](https://github.com/opengovsg/isomer/commit/2aed9bdf749c84c83a672130c3c609695d2d8619) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54271676)</sub>

### DIFF
--- a/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
+++ b/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
@@ -4,6 +4,7 @@ import { Box, Flex, Skeleton, Text, VStack } from "@chakra-ui/react"
 import { Button } from "@opengovsg/design-system-react"
 import { Suspense, useMemo } from "react"
 import { useSearchQuery } from "~/hooks/useSearchQuery"
+import { trpc } from "~/utils/trpc"
 import { ResourceType } from "~prisma/generated/generatedEnums"
 
 import {
@@ -71,6 +72,8 @@ const SuspensableResourceSelector = ({
   const isSearchQueryEmpty: boolean = searchQuery.trim().length === 0
   const hasAdditionalLeftPadding: boolean = isSearchQueryEmpty
 
+  const [rootPage] = trpc.page.getRootPage.useSuspenseQuery({ siteId })
+
   const {
     fullPermalink,
     moveDestPermalink,
@@ -105,6 +108,7 @@ const SuspensableResourceSelector = ({
 
   const {
     isResourceIdHighlighted,
+    isHomeHighlighted,
     isResourceItemDisabled,
     hasParentInStack,
     handleClickBackButton,
@@ -139,13 +143,14 @@ const SuspensableResourceSelector = ({
                 title: "Home",
                 permalink: "/",
                 type: ResourceType.RootPage,
-                id: null,
+                id: rootPage.id,
                 parentId: null,
               },
             ])
           }
           searchQuery={searchQuery}
           isLoading={isLoading}
+          isHomeHighlighted={isHomeHighlighted}
         />
       </Suspense>
     )
@@ -157,6 +162,8 @@ const SuspensableResourceSelector = ({
     handleClickResourceItem,
     searchQuery,
     isLoading,
+    isHomeHighlighted,
+    rootPage.id,
   ])
 
   const renderedContent = useMemo(() => {
--- a/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
+++ b/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
@@ -133,6 +133,17 @@ const SuspensableResourceSelector = ({
           hasParentInStack={hasParentInStack}
           handleClickBackButton={handleClickBackButton}
           resourceItemsWithAncestryStack={resourceItemsWithAncestryStack}
+          handleOnClick={() =>
+            handleClickResourceItem([
+              {
+                title: "Home",
+                permalink: "/",
+                type: ResourceType.RootPage,
+                id: null,
+                parentId: null,
+              },
+            ])
+          }
           searchQuery={searchQuery}
           isLoading={isLoading}
         />
@@ -143,6 +154,7 @@ const SuspensableResourceSelector = ({
     hasParentInStack,
     handleClickBackButton,
     resourceItemsWithAncestryStack,
+    handleClickResourceItem,
     searchQuery,
     isLoading,
   ])
--- a/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
+++ b/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
@@ -4,7 +4,6 @@ import { Box, Flex, Skeleton, Text, VStack } from "@chakra-ui/react"
 import { Button } from "@opengovsg/design-system-react"
 import { Suspense, useMemo } from "react"
 import { useSearchQuery } from "~/hooks/useSearchQuery"
-import { trpc } from "~/utils/trpc"
 import { ResourceType } from "~prisma/generated/generatedEnums"
 
 import {
@@ -72,9 +71,8 @@ const SuspensableResourceSelector = ({
   const isSearchQueryEmpty: boolean = searchQuery.trim().length === 0
   const hasAdditionalLeftPadding: boolean = isSearchQueryEmpty
 
-  const [rootPage] = trpc.page.getRootPage.useSuspenseQuery({ siteId })
-
   const {
+    rootPage,
     fullPermalink,
     moveDestPermalink,
     moveDest,
@@ -141,7 +139,7 @@ const SuspensableResourceSelector = ({
             handleClickResourceItem([
               {
                 title: "Home",
-                permalink: "/",
+                permalink: "",
                 type: ResourceType.RootPage,
                 id: rootPage.id,
                 parentId: null,
--- a/apps/studio/src/components/ResourceSelector/ResourceItem.tsx
+++ b/apps/studio/src/components/ResourceSelector/ResourceItem.tsx
@@ -5,7 +5,7 @@ import { dataAttr } from "@chakra-ui/utils"
 import { Button } from "@opengovsg/design-system-react"
 import { getIcon } from "~/utils/resources"
 
-interface ResourceItemProps {
+export interface ResourceItemProps {
   item: ResourceItemContent
   isDisabled?: boolean
   isHighlighted?: boolean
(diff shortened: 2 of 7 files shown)


adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

Note: this PR's base branch (`feature/isom-2503-studio-move-modal-validate-before-moving-not-after`) isn't merged into `main`, so the review scope used was the actual PR diff (base...head), matching what GitHub shows — not a raw `main...HEAD` diff, which would have pulled in unrelated content from the base branch.

- **[Correctness] Reopening an existing homepage link doesn't show "Home" as selected** — `apps/studio/src/components/ResourceSelector/useResourceStack.tsx:16`. When editing a link/block whose href already resolves to the homepage, `LinkEditorModal`/`PageLinkElement` passes the root page's id in as `selectedResourceId`. `useResourceStack` fetches its ancestry via `trpc.resource.getAncestryStack`, but the underlying query (`getBatchAncestryWithSelfQuery` in `resource.service.ts`) explicitly filters out `RootPage`-typed resources in its base case, so it returns an empty stack for a root-page id. That leaves `moveDest` undefined, so `isHomeHighlighted` (`useResourceSelector.tsx:44`, `moveDest?.type === ResourceType.RootPage`) evaluates `false` and the Home row renders unhighlighted on load — even though the link genuinely points at the homepage. This directly contradicts the PR's stated goal ("it never reflected whether it was the currently selected destination") for the one case — reopening an existing homepage selection — that matters most. The new `HomeSelected` Storybook story only exercises the fresh-click path via the Move modal, so it doesn't catch this gap.

- **[Correctness, minor] Selecting "Home" shows "You selected //" (double slash)** — `apps/studio/src/components/ResourceSelector/ResourceSelector.tsx:144`. The synthetic Home item is hardcoded with `permalink: "/"`, but every permalink elsewhere in the codebase is a leading-slash-free segment (see the comments in `resource.service.ts` — "whose permalink is empty" — and `SelectDestinationPageModal.tsx` — "ResourceSelector's permalink has no leading slash"). In `PageLinkElement` (block link editor) and `SelectDestinationPageModal` (redirect destination picker), `showSelectedResourcePreview` is left at its default `true`, so clicking Home makes `fullPermalink` resolve to `"/"`, and the preview renders "You selected /" + "/" = "You selected //". This is display-only — the actual link href is built from `resourceId` alone, and the redirect destination is re-normalized server-side before persisting — but it's directly visible in the exact feature this PR ships. Should be `permalink: ""` to match the rest of the codebase's convention.

- **[Maintainability, low] Shared component now fetches data directly** — `apps/studio/src/components/ResourceSelector/ResourceSelector.tsx:75`. `SuspensableResourceSelector` (under `apps/studio/src/components/`, shared across 3 features) now calls `trpc.page.getRootPage.useSuspenseQuery` itself, which this folder's own `CLAUDE.md` convention discourages ("Don't fetch data in components/... Data fetching lives in feature components or pages"). Not a functional bug, but it couples a shared component to a specific server endpoint/suspense lifecycle rather than receiving the root page via props.

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -133,8 +137,20 @@ const SuspensableResourceSelector = ({
...
+            handleClickResourceItem([
+              {
+                title: "Home",
+                permalink: "/",
+                type: ResourceType.RootPage,
+                id: rootPage.id,
+                parentId: null,
+              },
+            ])
+          }
```
dcshzj, line 148: question: this was flagged out in my AI review so might need you to double check, do we verify that the user is allowed to move items into the root?

**High — Homepage selection bypasses root-move authorization and corrupts hierarchy semantics.** Move mode sends `rootPage.id`, although the backend represents root as `null`. The server consequently treats it as an ordinary parent, bypasses the root permission rule, and stores `parentId = rootPage.id`, potentially making moved content disappear from normal root traversal. Emit `null` in move mode and retain the ID only for link selection. [ResourceSelector.tsx](<https://github.com/opengovsg/isomer/blob/8e1489c26ef03bc5f5ffda26efc7036f5cb19892/apps/studio/src/components/ResourceSelector/ResourceSelector.tsx#L140-L149>), [resource.router.ts](<https://github.com/opengovsg/isomer/blob/8e1489c26ef03bc5f5ffda26efc7036f5cb19892/apps/studio/src/server/modules/resource/resource.router.ts#L453-L475>), [storage invariant](<https://github.com/opengovsg/isomer/blob/8e1489c26ef03bc5f5ffda26efc7036f5cb19892/apps/studio/src/server/modules/resource/resource.service.ts#L885-L890>). Confidence: 1.0.

Ssatu (dismissed):
issue: moving logic as implemented breaks data integrity and resources moved to the homepage no longer appear

https://github.com/user-attachments/assets/1a8902e9-09d9-4c16-bb84-2fd8ee0246b6


similar to the root cause of the auth issue raised above

**Moving a resource to Home doesn't actually make it a top-level resource.**

The move mutation sets parentId based on whether destinationResourceId is truthy (resource.router.ts:524-526):

```
.set({
  parentId: !!destinationResourceId ? String(destinationResourceId) : null,
})
```

Since the new Home click passes destinationResourceId = rootPage.id (a real row id, from the trpc.page.getRootPage fetch in ResourceSelector.tsx), this branch sets the moved resource's parentId to the RootPage's own id — not null.

So a resource "moved to Home" via this new path ends up with a parentId value that doesn't match that invariant anywhere else

The fix has to make the Home selection resolve to destinationResourceId: null by the time it reaches the move mutation (the same value the pre-existing "back out via Back button" path already produces correctly), while still keeping rootPage.id around for the parts of the picker that need a real id (highlighting, the permalink preview).

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
seaerchin: yea this is correct, didn't check after claude added a commit (UI for selected state); it works if we set the `id: null` but not enough cos then the UI would be off (no selected state on hover)

lemme rework it slightly

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
```diff
@@ -71,6 +72,8 @@ const SuspensableResourceSelector = ({
   const isSearchQueryEmpty: boolean = searchQuery.trim().length === 0
   const hasAdditionalLeftPadding: boolean = isSearchQueryEmpty
 
+  const [rootPage] = trpc.page.getRootPage.useSuspenseQuery({ siteId })
```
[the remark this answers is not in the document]
seaerchin: Fixed in f11c026f1 — moved the `trpc.page.getRootPage` fetch out of `ResourceSelector` and into the `useResourceStack` hook, which already owns the other resource-tree fetching in this folder. `ResourceSelector` now just destructures `rootPage` from the hook instead of calling `trpc` directly.

apps/studio/src/components/ResourceSelector/ResourceSelector.tsx
seaerchin: Fixed in f11c026f1. Kept `rootPage.id` as the real internal id for the picker's Home entry (needed for highlighting and for the link flow, which needs a real id to build the reference link) but only translate it to `null` at the point `handleClickResourceItem` calls `onChange` — and only when `interactionType === "move"`. So the move mutation now receives `destinationResourceId: null` for Home, matching the backend's existing convention (same value the Back-button path already produced), while the link flow is unaffected.

seaerchin: Fixed in f11c026f1. `useResourceStack` now fetches `rootPage` and special-cases `selectedResourceId === rootPage.id`, seeding the stack with a synthetic Home entry directly — since `getAncestryStack`'s underlying query always returns an empty stack for the RootPage's own id, this was previously unrecoverable from the ancestry query alone. Reopening a homepage-linked block now correctly shows "Home" as selected.

seaerchin:
Addressed the automated review findings in f11c026f1:

- **Double slash on Home selection** — `apps/studio/src/components/ResourceSelector/ResourceSelector.tsx`: the synthetic Home item's `permalink` is now `""` instead of `"/"`, matching the codebase convention (e.g. `DirectorySidebar.tsx`). Fixes "You selected //".
- **Reopening a homepage link doesn't highlight Home** — replied inline on that thread.
- **Shared component fetching data directly** — replied inline on the Greptile thread; same fix (moved into `useResourceStack`) covers both reports.

dcshzj (approved):

UltraRaptorYT:
Thank you very much. As i think this feature will close the following issue once it is release you may also close the following issue
close #3056

PR #3180, opened 2026-08-18
perf(studio): derive collection thumbnail fallback in SQL for localised sitemap

dcshzj:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 2 | +44 | -47 |
| 🧪 Test | 1 | +153 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Closes ISOM-2322

High CPU load on the app server was traced to slow requests generating more than 4MB of data, virtually all of them from `getLocalisedSitemap` in preview. The query selected each sibling's **entire page body** (`published.content ->> 'content'`), and the immediate-siblings arm has no `LIMIT`, so previewing any page inside a large collection read every sibling's full body. That body was then `JSON.parse`d in Node purely to extract the first image block, which is used only as the collection card thumbnail fallback. The body itself never reaches the client — `IsomerSitemap` has no `content` field — so ~97% of the transferred bytes were pure waste.

## Solution

Project the first image block directly in SQL instead of shipping the whole body to Node. `WITH ORDINALITY` + `ORDER BY ord` keeps document order an explicit guarantee rather than relying on the function scan's emission order, and a `jsonb_typeof(...) = 'array'` guard prevents `jsonb_array_elements` from erroring on blobs whose `content` is not an array.

Measured against local Postgres at 2000 collection articles (~6KB bodies):

| | bytes to Node | PG execution |
| --- | --- | --- |
| Before | 7133 kB | 9.98 ms |
| After | 88 kB | 5.56 ms |

Postgres gets faster too, since the body no longer has to be materialised as text — this is not a database-for-app-server trade. On the Node side, the per-sibling `JSON.parse` and its heap churn are removed entirely, and the outer `UNION` de-duplication no longer hashes a multi-KB text column per row (12.8 ms → 0.84 ms in isolation).

**Breaking Changes**

The tRPC response shape is unchanged. `content` was never part of the returned sitemap; only the internal projection used to derive `firstImage` changed.

**Features**:

- None

**Improvements**:

- `getLocalisedSitemap` no longer reads page bodies for a page's siblings; it selects only the first image block's `src`/`alt`.
- Removed the per-sibling `JSON.parse` from `getSitemapTreeFromArray`.
- Added regression coverage for `firstImage`, which previously had none.

**Bug Fixes**:

- None. Selection of *which* image becomes the thumbnail is now explicitly ordered by document position; previously it depended on `jsonb_array_elements` emission order, which is not a guarantee.

## Before & After Screenshots

No visual change — this is a data-fetching change behind an identical response shape.

## Tests

`pnpm test:unit` in `apps/studio`: 1750 passed, 41 skipped (baseline before this change was 1746 passed). Four new integration tests against real Postgres cover document ordering, no image block, an image block missing `alt`, and non-article layouts. Each was mutation-tested — breaking the SQL predicate or flipping the sort direction makes them fail — so they are not passing vacuously. Typecheck and lint are clean.

- [ ] Open a site with a large collection (e.g. `ura-corp`) in Studio and edit any collection article. Confirm the preview loads and renders as before.
- [ ] In the collection's settings, set "Show thumbnail" fallback to **first image**. Confirm collection items that have no thumbnail of their own display the first image from their page body, and that it is the *first* image for articles containing several.
- [ ] Confirm collection items that do have their own thumbnail still show that thumbnail rather than the first body image.
- [ ] Set the thumbnail fallback to **logo** and confirm items without a thumbnail fall back to the site logo.
- [ ] Open a regular (non-collection) content page containing an image and confirm its preview is unaffected.
- [ ] In DevTools, compare the `site.getLocalisedSitemap` response time for a page in a large collection against production and confirm it is materially faster.

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None

The PR moves collection-thumbnail fallback extraction into PostgreSQL so localized sitemap generation no longer transfers and parses complete sibling page bodies.
- Projects only the first top-level image block’s `src` and `alt` fields.
- Preserves document order with ordinality and safely handles non-array content.
- Updates sitemap tree construction to consume the projected image.
- Adds database-backed regression coverage for ordering, absent images, missing alt text, and non-article layouts.
</details>

The PR appears safe to merge with no actionable correctness or security failures identified.

The SQL projection preserves the former first-top-level-image behavior, safely handles malformed content shapes, and maintains compatible JSONB types across every query branch while eliminating unnecessary page-body transfer and parsing.
</details>

| Filename | Overview |
|----------|----------|
| apps/studio/src/server/modules/resource/resource.service.ts | Replaces full page-body projection with guarded, ordered SQL extraction of first-image metadata across compatible query branches. |
| apps/studio/src/utils/sitemap.ts | Removes per-resource JSON parsing and maps the projected image into the existing sitemap response shape. |
| apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts | Adds integration coverage for first-image ordering, absence, missing alt text, and layout filtering. |

</details>

```mermaid
sequenceDiagram
    participant Preview as Studio preview
    participant Service as getLocalisedSitemap
    participant DB as PostgreSQL
    participant Tree as Sitemap tree builder
    Preview->>Service: Request localized sitemap
    Service->>DB: Query resources and first image metadata
    DB->>DB: Scan top-level blocks in document order
    DB-->>Service: Resource rows with firstImage JSONB
    Service->>Tree: Build IsomerSitemap
    Tree-->>Preview: Sitemap with thumbnail fallback metadata
```
</details>

<sub>Reviews (1): Last reviewed commit: ["perf(studio): derive collection thumbnai..."](https://github.com/opengovsg/isomer/commit/b1e49c6672a55d66f1ba649c8bcc1a0c3f8b9a4e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54284186)</sub>

### DIFF
--- a/apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts
+++ b/apps/studio/src/server/modules/resource/__tests__/resource.service.test.ts
@@ -4,6 +4,7 @@ import {
   setupBlob,
   setupCollection,
   setupCollectionMeta,
+  setupCollectionPage,
   setupFolder,
   setupFolderMeta,
   setupFullSite,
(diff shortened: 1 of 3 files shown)


adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

No issues found at medium effort.

Verification performed: read both modified files in full (`resource.service.ts`, `sitemap.ts`) plus the new test file; confirmed `firstImageSql` is wired consistently across all six union arms (ancestors base/recursive, immediateSiblings, nestedResources base/recursive) the same way `contentSql` was before; traced the `firstImage?.src` truthiness check in `sitemap.ts` against the sole downstream consumer (`getCollectionItems.ts` in `packages/components`), which already treats an empty/missing `src` the same as "no firstImage" — so the new behavior is equivalent to the old JSON.parse-based extraction, not a regression. Ran `pnpm typecheck`, `pnpm lint`, and `pnpm format` (all clean, no new warnings in touched files) and the full `pnpm test:unit` suite (1749 passed, 41 skipped; the 1 failing test is an unrelated pre-existing timeout in `AuditLogExportSection.test.tsx`, not touched by this PR). The new `firstImage` describe block's 4 tests all pass.

seaerchin (approved):
great solution tbh - to make sure i'm understanding this correctly, we project the content of the blob out into an table then attach a cardinality to it. thereafter, we just look through this table and select where the type is `image`?

wondering if we should use a placeholder function for the `image` block since it's entirely possible that other pages/blocks will change in the future but i think this might be abit premature

dcshzj:
> great solution tbh - to make sure i'm understanding this correctly, we project the content of the blob out into an table then attach a cardinality to it. thereafter, we just look through this table and select where the type is `image`?

Yup you are correct, so since it is already ordered, we will pick the first image that it sees, which would then be the first image in the page.

> wondering if we should use a placeholder function for the `image` block since it's entirely possible that other pages/blocks will change in the future but i think this might be abit premature

Possible yeah, if we decide in the future to say consider infopic as a legitimate first image block beyond just image blocks, but I think the change is quite straightforward that it might not require the placeholder function?

PR #3183, opened 2026-08-18
ci: build and publish isomer-components to S3 cache on deploy

seaerchin:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 5 | +346 | -100 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

### 📊 PR diff breakdown

| Bucket | Files | +Lines | \-Lines |
| -- | -- | -- | -- |
| ⚙️ App | 5 | +340 | \-100 |
| 🧪 Test | 0 | +0 | \-0 |
| 📄 Doc | 0 | +0 | \-0 |
| 🚫 Ignore | 0 | +0 | \-0 |

## Problem

Closes [ISOM-2360](https://linear.app/ogp/issue/ISOM-2360/set-up-release-pipeline-to-publish-components-and-reduce-reliance-on)

The `@opengovsg/isomer-components` dist bundle is currently only built as a side effect of whichever per-site CodeBuild publish job happens to run first after a components version bump — `tooling/build/scripts/publisher.sh` builds it and caches it to S3 only if the shared S3 cache doesn't already have an entry for the current commit. That means the build (and the latency of building it) is entirely dependent on site-publish traffic, and there's no way to pre-warm the cache from GitHub itself.

## Solution

Added `.github/workflows/build-components.yml`, a new reusable workflow (`workflow_dispatch` for manual/backfill runs, `workflow_call` for wiring into other workflows) that ports the relevant parts of `publisher.sh`'s dependency and `isomer-components` caching logic:

* Resolves the `opengovsg/isomer` ref to build (an….. explicit input, or the latest release tag) and checks it out.
* Assumes the caller-supplied AWS role and installs pnpm with a project-local store (matching `publisher.sh`'s own setup), so the S3-cached pnpm store tarball stays self-contained.
* Computes the same cache key `publisher.sh` used (the release tag alone for auto-resolved tags, or `<ref>-<short-sha>` for an explicitly-passed ref) so this workflow and the existing CodeBuild site-publish jobs land in the exact same S3 cache locations and don't collide or duplicate work.
* Fetches or installs the pnpm dependency store, then fetches or builds the `isomer-components` dist tarball, publishing it to the shared S3 cache bucket.

Wired this into a new `build_components` job in `deploy_production.yml`, `deploy_staging.yml`, and `deploy_uat.yml`, each pointing at that environment's existing AWS account/role (reusing the same ones the app-deploy job in the same file already uses) and its own `isomer-next-infra-<env>-sites-cache-bucket-private-*` bucket. This job runs independently of (in parallel with, not blocking) the existing app-deploy job in each file, since publishing `isomer-components` to cache has nothing to do with deploying the Studio app itself.

Net effect: a components version bump gets built and cached from GitHub on every deploy, so the next site publish that needs it hits a warm cache instead of building it inline.

**Breaking Changes**

  * Details ...

**Features**:

* New reusable `build-components.yml` workflow that builds and caches `@opengovsg/isomer-components` to the shared S3 cache bucket, callable manually or from another workflow.

**Improvements**:

* `deploy_production.yml`, `deploy_staging.yml`, `deploy_uat.yml` each gain a `build_components` job so every deploy also pre-warms the components cache for that environment, reducing reliance on a site-publish job being the one to build it.

**Bug Fixes**:

* None.

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

* `actionlint` passes clean on all four changed workflow files.
* Recommend a manual `workflow_dispatch` run of `build-components.yml` against staging's role/bucket before relying on the automatic hookup, to confirm the assumed role has the S3 permissions this job needs (`s3:GetObject`/`s3:PutObject` on the cache bucket) — this wasn't verifiable locally since it requires real AWS credentials.
* After merge, confirm the `build_components` job succeeds on the next push to `staging` before trusting it on `production`/`uat`.

**New scripts**:

* None.

**New dependencies**:

* None.

**New dev dependencies**:

* None.

<details><summary>Greptile Summary</summary>

The PR moves production repository preparation into a reusable GitHub workflow and updates CodeBuild publishing to consume the resulting S3 archive.

* Adds environment-specific deploy jobs that populate the shared repository cache.
* Archives an installed checkout under the layout expected by the site publisher.
* Uses an embedded release ref to key downstream components-dist artifacts.

</details>

<details><summary>Confidence Score: 3/5</summary>

The PR is not yet safe to merge because it still leaves the first components build to a site publication and permits older release jobs to overwrite the latest repository cache.

The GitHub workflow never publishes the release-keyed components dist that its consumer requests, and independently grouped release jobs can complete out of order while writing the same mutable S3 object.

**Files Needing Attention:** .github/workflows/build-components.yml, .github/workflows/deploy_production.yml, tooling/build/scripts/publisher.sh

</details>

<details><summary>Important Files Changed</summary>

| Filename | Overview |
| -- | -- |
| .github/workflows/build-components.yml | Adds repository archive publication, but omits components-dist prewarming and writes concurrent releases to one mutable latest key. |
| tooling/build/scripts/publisher.sh | Consumes the repository archive and fixes static dist keying, but still performs the first components build and accepts whichever release occupies the latest key. |
| .github/workflows/deploy_production.yml | Adds the reusable cache job, with per-release concurrency groups that do not serialize writes across release tags. |
| .github/workflows/deploy_staging.yml | Adds the reusable cache job; branch-scoped workflow concurrency prevents overlapping staging pushes. |
| .github/workflows/deploy_uat.yml | Adds the reusable cache job for UAT using the environment-specific role and bucket. |

</details>

<details><summary>Sequence Diagram</summary>

```mermaid
sequenceDiagram
    participant R1 as Older release workflow
    participant R2 as Newer release workflow
    participant S3 as S3 latest cache
    participant Site as Site CodeBuild
    R1->>R1: Install dependencies only
    R2->>R2: Install dependencies only
    R2->>S3: Upload newer isomer.tar.zst
    R1->>S3: Upload older isomer.tar.zst last
    Site->>S3: Download isomer/latest/isomer.tar.zst
    S3-->>Site: Older repository archive
    Site->>S3: Fetch release-keyed components dist
    alt Dist missing
        Site->>Site: Build isomer-components inline
    end
```

</details>

Fix all with Greploop Fix All in Claude Code Fix All in Cursor Fix All in Conductor

<details><summary>Prompt To Fix All With AI</summary>

```markdown
### Issue 1
.github/workflows/build-components.yml:97
**Components dist is never prewarmed**

When the first production site publication runs for an uncached release, this workflow has installed dependencies but has not built or uploaded `isomer-components-dist.tar.zst`, so `publisher.sh` still compiles the components package inline and retains the latency and site-publish dependency this workflow is intended to eliminate.

### Issue 2
.github/workflows/build-components.yml:129-130
**Latest archive permits stale overwrite**

If two production releases overlap and the older job uploads last, both jobs write to the same `isomer/latest/isomer.tar.zst` object without cross-release serialization or release validation, causing subsequent site publications to extract and publish the older repository.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
```

</details>

Reviews (5): Last reviewed commit: ["chore: fix code review findings"](<https://github.com/opengovsg/isomer/commit/cbd65d04948f56aa0911415c267e613a1f386fd0>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=54300185>)

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/.github/workflows/build-components.yml
+++ b/.github/workflows/build-components.yml
@@ -75,14 +75,17 @@ jobs:
           pnpm config set store-dir .pnpm-store --location project
 
       - name: Install pnpm dependencies
-        run: pnpm install --frozen-lockfile
+        run: pnpm install --frozen-lockfile --filter isomer-base-template...
 
       # Shared S3 cache (same bucket/prefix pattern is used by multiple CodeBuild projects).
-      # Rather than caching the pnpm store and a built dist separately, archive the whole
-      # repository (source + node_modules) as a single tarball so downstream per-site
-      # CodeBuild projects can pull it down instead of cloning/installing themselves.
+      # Archived as two separate tarballs so downstream per-site CodeBuild projects can pull
+      # them down instead of cloning/installing themselves:
+      #   1. isomer.tar.zst - the repository itself (source only, no node_modules/.pnpm-store).
+      #   2. isomer-pnpm-store.tar.zst - the project-local pnpm store, cached separately since
+      #      restoring it doesn't touch the repository and it's what publisher.sh's install
+      #      step actually needs to relink node_modules without hitting the registry.
       #
-      # publisher.sh (the consumer) downloads this from isomer/latest/isomer.tar.zst, extracts it with
+      # publisher.sh (the consumer) downloads isomer.tar.zst from isomer/latest/, extracts it with
       # `tar --use-compress-program=zstd -xf`, then `cd isomer/` - mirroring what
       # `git clone .../isomer.git` would have produced. So the archive must be
       # zstd-compressed and rooted at an `isomer/` directory, not the bare working tree.
@@ -102,6 +105,7 @@ jobs:
           }
 
           REPO_TGZ="isomer.tar.zst"
+          STORE_TGZ="isomer-pnpm-store.tar.zst"
           # publisher.sh's production path only ever reads isomer/latest/. This branch
           # exists so a manual/test workflow_dispatch of *this* workflow (passing an
           # explicit ref) writes to its own prefix instead of clobbering isomer/latest/
@@ -111,13 +115,13 @@ jobs:
           # workflow produced for that branch instead of re-cloning from scratch, and
           # only falls back to a fresh git clone if no such archive was published.
           if [ -n "$REQUESTED_REF" ]; then
-            REPO_CACHE_PATH="s3://$S3_CACHE_BUCKET_NAME/isomer/refs/$REQUESTED_REF/$REPO_TGZ"
+            CACHE_PREFIX="s3://$S3_CACHE_BUCKET_NAME/isomer/refs/$REQUESTED_REF"
           else
-            REPO_CACHE_PATH="s3://$S3_CACHE_BUCKET_NAME/isomer/latest/$REPO_TGZ"
+            CACHE_PREFIX="s3://$S3_CACHE_BUCKET_NAME/isomer/latest"
           fi
+          REPO_CACHE_PATH="$CACHE_PREFIX/$REPO_TGZ"
+          STORE_CACHE_PATH="$CACHE_PREFIX/$STORE_TGZ"
 
-          echo "Archiving repository..."
-          start_time=$(date +%s)
           # Record the resolved ref inside the archive so publisher.sh's production
           # path - which has no branch input and no .git to inspect - can key its
           # isomer-components dist cache by the actual release instead of a fixed
@@ -129,12 +133,26 @@ jobs:
           # directory itself and `ln -s` would fail with "File exists".
           ARCHIVE_STAGING=$(mktemp -d)
           ln -s "$(pwd)" "$ARCHIVE_STAGING/isomer"
-          tar --exclude=.git -h --use-compress-program="zstd -6" -cf "$ARCHIVE_STAGING/$REPO_TGZ" -C "$ARCHIVE_STAGING" isomer
+
+          echo "Archiving repository..."
+          start_time=$(date +%s)
+          tar --exclude=.git --exclude=node_modules --exclude=.pnpm-store -h --use-compress-program="zstd -6" -cf "$ARCHIVE_STAGING/$REPO_TGZ" -C "$ARCHIVE_STAGING" isomer
           calculate_duration $start_time
 
           echo "Uploading repository archive to S3..."
           start_time=$(date +%s)
           aws s3 cp --only-show-errors "$ARCHIVE_STAGING/$REPO_TGZ" $REPO_CACHE_PATH
-          rm -rf "$ARCHIVE_STAGING"
           echo "Uploaded repository archive"
+          calculate_duration $start_time
+
+          echo "Archiving pnpm store..."
+          start_time=$(date +%s)
+          tar --use-compress-program="zstd -6" -cf "$ARCHIVE_STAGING/$STORE_TGZ" .pnpm-store
+          calculate_duration $start_time
+
+          echo "Uploading pnpm store archive to S3..."
+          start_time=$(date +%s)
+          aws s3 cp --only-show-errors "$ARCHIVE_STAGING/$STORE_TGZ" $STORE_CACHE_PATH
+          rm -rf "$ARCHIVE_STAGING"
+          echo "Uploaded pnpm store archive"
           calculate_duration $start_time
--- a/.github/workflows/build-components.yml
+++ b/.github/workflows/build-components.yml
@@ -10,25 +10,6 @@ name: Build and publish site
 # repository (source + node_modules) to S3, which downstream per-site
 # CodeBuild projects pull down instead of cloning/installing themselves.
 on:
-  workflow_dispatch:
-    inputs:
-      aws-region:
-        description: "AWS region to use"
-        required: true
-        default: "ap-southeast-1"
-        type: string
-      cicd-role:
-        description: "AWS IAM role to assume by GitHub action runner"
-        required: true
-        type: string
-      s3-cache-bucket-name:
-        description: "S3 bucket used to cache the pnpm store and isomer-components dist between builds (same bucket/prefix pattern as the CodeBuild projects)"
-        required: true
-        type: string
-      isomer-build-repo-branch:
-        description: "Git ref (tag or branch) of opengovsg/isomer to build. Leave empty to use the latest release tag."
-        required: false
-        type: string
   workflow_call:
     inputs:
       aws-region:
--- a/.github/workflows/build-components.yml
+++ b/.github/workflows/build-components.yml
@@ -1,0 +1,181 @@
+name: Build and publish site
+
+# Migrated from tooling/build/scripts/publisher.sh, which used to run as the
+# entrypoint of a per-site AWS CodeBuild project. CodeBuild supplied SITE_NAME /
+# S3_BUCKET_NAME / CLOUDFRONT_DISTRIBUTION_ID / S3_CACHE_BUCKET_NAME etc. as
+# project-level env vars; those become workflow_dispatch inputs here. The
+# git-clone-from-scratch step becomes actions/checkout; the S3 pnpm-store/dist
+# tarball caching is otherwise kept as-is.
+on:
+  workflow_dispatch:
+    inputs:
+      aws-region:
+        description: "AWS region to use"
+        required: true
+        default: "ap-southeast-1"
+        type: string
+      aws-account-id:
+        description: "AWS account ID to use"
+        required: true
+        type: string
+      cicd-role:
+        description: "AWS IAM role to assume by GitHub action runner"
+        required: true
+        type: string
+      s3-cache-bucket-name:
+        description: "S3 bucket used to cache the pnpm store and isomer-components dist between builds (same bucket/prefix pattern as the CodeBuild projects)"
+        required: true
+        type: string
+      isomer-build-repo-branch:
+        description: "Git ref (tag or branch) of opengovsg/isomer to build. Leave empty to use the latest release tag."
+        required: false
+        type: string
+  workflow_call:
+    inputs:
+      aws-region:
+        description: "AWS region to use"
+        required: false
+        default: "ap-southeast-1"
+        type: string
+      aws-account-id:
+        description: "AWS account ID to use"
+        required: true
+        type: string
+      cicd-role:
+        description: "AWS IAM role to assume by GitHub action runner"
+        required: true
+        type: string
+      s3-cache-bucket-name:
+        description: "S3 bucket used to cache the pnpm store and isomer-components dist between builds (same bucket/prefix pattern as the CodeBuild projects)"
+        required: true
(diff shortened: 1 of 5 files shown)


adriangohjw:
## 🤖 Automated code review

**Effort level:** medium

- **[High][Correctness]** `.github/workflows/build-components.yml:123-181` — The S3-cached `.pnpm-store` and `isomer-components` dist tarballs are downloaded but never extracted, so the caching this PR adds is a functional no-op. In "Install pnpm dependencies" (lines 123-150), `aws s3 cp ... $TEMPLATE_DEPS_TGZ || true` fetches the cached store, but no `tar -x` follows before `pnpm install --frozen-lockfile` runs unconditionally (line 141) — the download is discarded and immediately overwritten by an unconditional re-tar+upload (lines 144-149). The same pattern repeats in "Build isomer-components dist" (lines 154-181): the cached dist tarball is downloaded (line 169) but never extracted, and `pnpm --filter @opengovsg/isomer-components run build` (line 172) always does a full rebuild. The original `tooling/build/scripts/publisher.sh` this replaces (lines 70-96, 102-125) wraps the same download in `if [ -f $TGZ ]; then tar -xf ...; else install/build && re-cache; fi` — that branch was dropped in the migration. As written, every run does a full cold install and full rebuild regardless of cache contents, so the cache adds S3/network overhead without ever producing a speedup, undermining the PR's own goal.

- **[Medium][Security]** `.github/workflows/build-components.yml` (inputs at lines 17-28) + `deploy_production.yml:58`, `deploy_staging.yml:63`, `deploy_uat.yml:56` — The new `build_components` job reuses the exact same `cicd-role` ARN as the sibling app-deploy job in each file (verified identical across all three: e.g. `isomer-next-infra-prod-deploy-role`, which `aws_deploy.yml` uses for ECR login and ECS/CodeDeploy), even though `build_components` only needs S3 `GetObject`/`PutObject` on a cache bucket. Combined with `build-components.yml`'s `workflow_dispatch` trigger, which exposes `aws-account-id`/`cicd-role`/`s3-cache-bucket-name` as unrestricted free-text inputs alongside an arbitrary `isomer-build-repo-branch` to check out and build, any repo collaborator able to dispatch workflows could manually run this workflow with the production role ARN and a branch of their choosing — `pnpm install`/`pnpm run build` would then execute that branch's (or its dependencies') scripts while holding credentials scoped for full production ECR/ECS/CodeDeploy access, well beyond what this job needs. Consider a scoped-down role limited to the specific cache bucket/prefix.

- **[Low][Style]** `deploy_production.yml:53`, `deploy_staging.yml:58`, `deploy_uat.yml:51` — Job display name `"Build and publish components "` has a trailing space, duplicated identically in all three files. Cosmetic only.

adriangohjw (commented):
question: instead of publishing this, it better to zip the entire repo and upload it so we can download it from our own s3 in publisher.sh instead of using git clone? this approach does not remove the need to git clone (since it still requires the tooling/template repo), which is the root issue we are addressing of being reliant on github

seaerchin:
> question: instead of publishing this, it better to zip the entire repo and upload it so we can download it from our own s3 in publisher.sh instead of using git clone? this approach does not remove the need to git clone (since it still requires the tooling/template repo), which is the root issue we are addressing of being reliant on github

@adriangohjw wdyt about this approach:

on new release, we upload to 2 locations in s3: `/<version_tag>/isomer` and `/latest/isomer`, where `latest` is always assumed to be the latest release

this allows us to omit the extra step of doing
```
  ISOMER_BUILD_REPO_BRANCH=$(git ls-remote --tags --sort='v:refname' https://github.com/opengovsg/isomer.git | tail -n1 | awk '{print $2}' | sed -E 's|refs/tags/||; s/\^.*$//')
  IS_USING_RELEASE_TAG=true
```

for production builds

adriangohjw:
@seaerchin oh the failure point is less of here but the downstream part https://github.com/opengovsg/isomer/blob/45a0a80e1737ad930bf8fa5ab2e210d0e9e97db7/tooling/build/scripts/publisher.sh#L32

or do you mean `/<version_tag>/isomer` is the repo itself?

seaerchin:
@adriangohjw i meant this: 

> or do you mean /<version_tag>/isomer is the repo itself?

so our entry point is always `/latest/isomer` (assumed latest version tag; our release script handles uploading)

.github/workflows/build-components.yml
```diff
@@ -0,0 +1,156 @@
...
+            end_time=$(date +%s)
+            echo "Time taken: $((end_time - start_time)) seconds"
+          }
+
+          REPO_TGZ="isomer-repo.tar.zst"
+          REPO_CACHE_PATH="s3://$S3_CACHE_BUCKET_NAME/isomer/latest/$REPO_TGZ"
+
+          echo "Archiving repository..."
+          start_time=$(date +%s)
+          tar --exclude=.git --use-compress-program="zstd -6" -cf $REPO_TGZ .
```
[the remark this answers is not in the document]
adriangohjw: @greptileai use ascii diagram to draw the flow on how this issue can occur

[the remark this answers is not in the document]
seaerchin: oops, updating to use `zst` for efficiency

.github/workflows/build-components.yml
```diff
@@ -0,0 +1,154 @@
...
+            end_time=$(date +%s)
+            echo "Time taken: $((end_time - start_time)) seconds"
+          }
+
+          REPO_TGZ="isomer.tar.zst"
+          # publisher.sh's production path only ever reads isomer/latest/. Manual/test
+          # dispatches with an explicit ref get their own prefix so they can never
+          # clobber the artifact a concurrent real deploy is about to consume.
+          if [ -n "$REQUESTED_REF" ]; then
+            REPO_CACHE_PATH="s3://$S3_CACHE_BUCKET_NAME/isomer/refs/$REQUESTED_REF/$REPO_TGZ"
```
dcshzj: question(non-blocking): a bit confused, how come we are doing this when the publisher.sh doesn't reference it, it seems like we are still doing the `git clone` on publisher.sh's line 25

.github/workflows/build-components.yml
```diff
@@ -0,0 +1,154 @@
...
+          # isomer-components dist cache by the actual release instead of a fixed
+          # string, which would otherwise never invalidate across releases.
+          echo "$RESOLVED_REF" > .isomer-release-ref
+          # Stage the rename-symlink in a fresh temp dir rather than the checkout's
+          # parent: actions/checkout's default workspace path is
+          # /home/runner/work/isomer/isomer, so "../isomer" there is the checkout
+          # directory itself and `ln -s` would fail with "File exists".
+          ARCHIVE_STAGING=$(mktemp -d)
+          ln -s "$(pwd)" "$ARCHIVE_STAGING/isomer"
+          tar --exclude=.git -h --use-compress-program="zstd -6" -cf "$ARCHIVE_STAGING/$REPO_TGZ" -C "$ARCHIVE_STAGING" isomer
```
dcshzj: issue: I'm not too familiar with pnpm's internals but I recall they do some sort of symlink, which I think might break when doing the tar. Trouble you to double check this!

.github/workflows/deploy_staging.yml
```diff
@@ -53,3 +53,11 @@ jobs:
       posthog-host: "https://rpfs.isomer.gov.sg"
       posthog-assets-host: "https://us-assets.i.posthog.com"
       posthog-project-token: "phc_uZuYdUk5MvkoNAzxvh9TXdGXXjSL684XAHBjR5r74Ra8"
+
+  build_components:
```
dcshzj, line 57: nit: btw staging and UAT do not use the default branch (they both set `ISOMER_BUILD_REPO_BRANCH`), so the publisher.sh would not retrieve the tar that is being uploaded here.

We either skip building components for staging/UAT, or wire it in some way that publisher.sh treats these the `staging` branch as the main branch on the staging env, and vice versa for the `uat` branch.

seaerchin: yeap that is correct. that is because they do not reference the same commit hash and is the intended design. 

the fallback behaviour is the current behaviour, to clone the repo then get the built deps from s3

.github/workflows/build-components.yml
seaerchin: the flow is as follows: 

```sh
# if not set, assume this is production - fetch from cache 
if [ -z "$ISOMER_BUILD_REPO_BRANCH" ]; then
  aws s3 cp --only-show-errors s3://"$S3_CACHE_BUCKET_NAME"/isomer/latest/isomer.tar.zst .
  tar --use-compress-program=zstd -xf isomer.tar.zst
  cd isomer/
else
  # fallback to existing behaviour to clone the existig repo
  git clone --depth 1 --branch "$ISOMER_BUILD_REPO_BRANCH" https://github.com/opengovsg/isomer.git
```

.github/workflows/build-components.yml
seaerchin: this works because -h follows symlinks but j realised pnpm has its own pnpm pack so will use rhat instead

.github/workflows/build-components.yml
dcshzj: Uh huh, it is only using the `latest` folder, so anything in `refs` are not used

seaerchin: ah gotcha - i can delete this chunk of code; initially was meant to store at `/ref/<ref>/isomer` but kept it simple.

i did this initially because i wanted to have the ability to trigger `workflow_call` (ie manually) without writing to `latest` - lmk if you wanna keep or delete

seaerchin:
tbvh the claude written code on `publisher.sh` is semi complicated, lmk if want a rework but basically:

at deploy time (release etc) 
1. repo is always stored at `/latest`
2. content store is always uploaded to triggering github hash 

at build time (site builds) 
1. check if repo is stored at `/latest` if no build branch set
2. otherwise, fetch from `/build_branch` 
3. (redo the same for others)

seaerchin:
@greptileai thanks

.github/workflows/build-components.yml
```diff
@@ -0,0 +1,180 @@
...
+          pnpm config set store-dir .pnpm-store --location project
+
+      - name: Install pnpm dependencies
+        run: pnpm install --frozen-lockfile --filter isomer-base-template...
+
+      # publisher.sh's per-site builds each need this dist (packages/components/dist is
+      # gitignored, so it isn't in the repository archive below); building it once here
+      # instead of once per site is the whole point of centralizing this in CI.
+      - name: Build isomer-components
+        run: pnpm --filter @opengovsg/isomer-components run build
```
dcshzj, line 87: issue: hmm might need to verify but cos of this filter, only dependencies declared in the components package are captured in the pnpm store, so additional dependencies in `@isomer/db` or in the template may not be captured

dcshzj (approved):

PR #3187, opened 2026-08-18
fix: Preserve array fields when switching combinator variants

adriangohjw:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 11 | +163 | -20 |
| 🧪 Test | 2 | +255 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

## Problem

Toggling InfoCards from with-image to no-image wiped uploaded images from editor state. Toggle back without saving and they were gone.

`keepMatchingArrayFields` was picking only the new schema's keys, so `imageUrl` / `imageAlt` / `imageFit` got dropped on the way to the no-image variant.

We do not need to flush those fields in the editor. Extra fields are valid. Rendering ignores unused ones. S3 does not delete the asset.

closes https://linear.app/ogp/issue/ISOM-2485/template-cards-infocards-flushreset-when-toggling-between-card

## Solution

Keep extras in editor state. Strip them only in the save payload.

On variant switch, `keepMatchingArrayFields` copies array items as they are. Schema defaults still apply for everything else, including `cards: []`, then the kept arrays overwrite that.

On save, every `updatePageBlob` call goes through `serializePageBlob`. That walks each content block, finds the matching `oneOf` / `anyOf` branch, and `pick`s array items down to that branch's keys. `setSavedPageState` / `setPreviewPageState` still get the unstripped editor copy.

An unsaved round trip from with image to no image and back restores images.

Save on no-image, then reload. Images are gone. That's the committed variant.

Save without reload. Extras can still sit in client state, so another toggle can restore them.

This runs for any combinator array field, including InfoCards. Object-level leftovers are not stripped. Hero variant fields are one example, and that's a known gap. A future `updatePageBlob` stringify site that skips `serializePageBlob` would also miss the strip. We moved this off the Zod schema on purpose. Extra fields are already valid on the server.

**Breaking Changes**

**Bug Fixes**:

- Unsaved InfoCards variant toggles keep extra card fields, including images
- Saved page JSON still matches the active variant

## Tests

Unit tests live in `combinatorArrayFields.test.ts` and `page.test.ts`. They cover keep-on-round-trip, pick/strip for the inactive branch, stripped serialize payload vs unstripped editor copy, and schema parse still accepting leftover image fields.

- [ ] Open a page with InfoCards, choose the with-image variant, upload images
- [ ] Switch to no image. Preview should hide the images. Do not save.
- [ ] Switch back to with image. The same images should come back.
- [ ] Switch to no image, save, reload the editor. Images should stay gone.
- [ ] Switch to no image, save, do not reload, switch back to with image. Images can still return from client state.


sehyunidaaa:
Copying over question from Linear: https://linear.app/ogp/issue/ISOM-2485/template-cards-infocards-flushreset-when-toggling-between-card#comment-5b87d97c

adriangohjw:
@sehyunidaaa lemme try. i don't think it's critical to flush them but just worried about edge case where its saved and someone else convert from non-image to image in the future and the old images show up

PR #3315, opened 2026-09-03
fix(publisher): retry CloudFront origin-path update on ETag race

Ssatu:
### 📊 PR diff breakdown

|---|---|---|---|
| ⚙️ App | 1 | +74 | -8 |
| 🧪 Test | 0 | +0 | -0 |
| 📄 Doc | 0 | +0 | -0 |
| 🚫 Ignore | 0 | +0 | -0 |

📄 [Read the full explainer](https://app.notion.com/p/3d077dbba78881588b70f4c14047d4ce) (background, intuition, code walkthrough, quiz)

## Summary
- CodeBuild's `computeBuildChanges` deliberately allows a second build of the same site to start alongside a stale in-progress one, so two builds can concurrently read-modify-write the same CloudFront `DistributionConfig`. Whichever writes second hits a `412 PreconditionFailed` on a stale ETag and the whole build fails, even though the publish content itself built and uploaded fine — a false-positive failure that pages on-call for nothing.
- Fix: on `PreconditionFailed`, re-fetch the distribution and compare the live build number to our own:
  - **Live build >= ours** → a concurrent build legitimately won. Concede, exit 0.
  - **Live build < ours** → we're the one that should be live, just lost the write race. Retry with a fresh ETag (bounded to 5 attempts, exponential backoff + jitter).
  - **Path doesn't parse** (`/$SITE_NAME/<digits>/latest`) → not explained by a normal race (e.g. a manual/IaC change). Fail loudly.
- Credit: the core idea of comparing the live build number to decide whether to concede comes from [PR #2256](https://github.com/opengovsg/isomer/pull/2256), an independent, already-approved fix for the same bug. This PR borrows that comparison and adds retrying specifically when we're the build that should legitimately win — failing there instead would reintroduce the same on-call-paging problem this fix exists to remove.

## Known limitation (not fixed by this PR)
Ordering by `CODEBUILD_BUILD_NUMBER` is not a true content-freshness guarantee: the number is assigned when a build *starts*, not when it fetches its content from the database, so a slower-starting build can legitimately end up with fresher content than a faster one with a higher number. A fully correct fix would need an authoritative content-revision signal threaded from the DB fetch through to the CloudFront promotion step — a structural change spanning `tooling/build/scripts/publishing` and the promotion step, out of scope here. Recommend a follow-up ticket if stronger freshness guarantees are needed. (Both PR #2256 and this PR share this limitation.)

## Test plan
- [ ] `bash -n tooling/build/scripts/publisher.sh` (syntax-checked locally, passes)
- [ ] Trigger two overlapping publishes for the same staging site and confirm the collision is retried (not a hard failure) and the newer build's content ends up live
- [ ] Confirm a genuine (non-ETag) `update-distribution` failure still fails the build immediately with the original AWS exit code

The change handles a CloudFront ETag conflict by re-fetching the live distribution and inspecting its origin path rather than immediately failing. It now:
- Parses the build number only from the exact `/$SITE_NAME/<digits>/latest` format, including site names containing digits.
- Skips the update when the live origin points to an equal or higher-numbered build.
- Fails closed for older or unrecognized live origin paths and fails fast for non-ETag errors.
</details>

The PR appears safe to merge because both previously reported origin-path handling failures are resolved and no blocking failure remains within this follow-up review’s scope.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| tooling/build/scripts/publisher.sh | Adds fail-closed handling for CloudFront ETag conflicts and correctly addresses the two previously reported origin-path parsing and validation defects. |

</details>

```mermaid
sequenceDiagram
  participant Build as Publisher build
  participant CF as CloudFront
  Build->>CF: Get distribution config and ETag
  Build->>CF: Update origin path with ETag
  alt Update succeeds
    CF-->>Build: Success
  else PreconditionFailed
    CF-->>Build: 412 ETag conflict
    Build->>CF: Re-fetch live distribution
    CF-->>Build: Current origin path
    alt "Exact path and live build >= current build"
      Build->>Build: Concede update
    else Older or unrecognized path
      Build->>Build: Exit with failure
    end
  else Other update error
    CF-->>Build: Error
    Build->>Build: Exit immediately
  end
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(publisher): re-add build-number chec..."](https://github.com/opengovsg/isomer/commit/d3695b425274f87b60a523554c3a2e72ce7cd795) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59860174)</sub>

**Context used:**

- Knowledge Base — [Site lifecycle tooling](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/isomer/-/docs/site-lifecycle-tooling.md)

### DIFF
--- a/tooling/build/scripts/publisher.sh
+++ b/tooling/build/scripts/publisher.sh
@@ -217,47 +217,51 @@ calculate_duration $start_time
 #
 # Two builds of the same site can run concurrently (see computeBuildChanges),
 # racing to write the same DistributionConfig; the second writer gets a 412
-# PreconditionFailed on a stale ETag. Retry with a fresh ETag rather than
-# failing the build. This keeps the pre-existing "last write wins" ordering
-# semantics (unchanged from before this fix) instead of trying to order by
-# CODEBUILD_BUILD_NUMBER: that's assigned at build start, not when a build
-# fetches its content, so a slower-starting build can still legitimately
-# fetch fresher content later - ordering by build number would wrongly and
-# permanently lock that fresher content out.
+# PreconditionFailed on a stale ETag. If we lose the race, only concede
+# silently when the live OriginPath proves a build >= ours already won -
+# that's the one state a normal race actually produces. Anything else
+# (an older build won, or the path is unrecognised - e.g. a manual or IaC
+# change) isn't explained by a normal race, so fail loudly instead of
+# guessing and silently overwriting.
 echo "Updating CloudFront origin path..."
 echo "CloudFront distribution ID: $CLOUDFRONT_DISTRIBUTION_ID"
 
-MAX_ATTEMPTS=5
-UPDATED=false
+aws cloudfront get-distribution --id "$CLOUDFRONT_DISTRIBUTION_ID" >distribution.json
+ETag=$(jq -r '.ETag' distribution.json)
 
-for ((attempt = 1; attempt <= MAX_ATTEMPTS; attempt++)); do
-  aws cloudfront get-distribution --id "$CLOUDFRONT_DISTRIBUTION_ID" >distribution.json
-  ETag=$(jq -r '.ETag' distribution.json)
-  echo "Attempt $attempt/$MAX_ATTEMPTS: ETag=$ETag"
+jq '.Distribution.DistributionConfig' distribution.json >distribution-new.json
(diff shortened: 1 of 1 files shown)


adriangohjw:
i got something here https://github.com/opengovsg/isomer/pull/2256 but feel free to close my PR if we are going with yours 👍

Ssatu:
> i got something here #2256 but feel free to close my PR if we are going with yours 👍

@adriangohjw noted on this! will write something short for us to land on this 🤜 🤛

